PR #22, opened 2024-12-07
[Refactor] Phase cleanup part 2

DayKev:
image
image

N/A

Code modernization.

- Many tsdocs added or updated
- Empty constructors removed
- `public`/etc added to all methods
- Return types added to all methods
- `getField()` added to `PartyMemberPokemonPhase` to replace the common pattern of `this.isPlayer ? globalScene.getPlayerField() : globalScene.getEnemyField()`
- Unused parameter removed from `AttemptCapturePhase.failCatch()`
- `readonly` added to all fields it can be applied to
- Some methods and fields renamed for improved clarity/etc
- Unused phase `EnemyPartyMemberPokemonPhase` removed
- Prepended `abstract-` to the filenames of abstract phases
- `HidePartyExpBarPhase` removed and its functionality was extracted to where it was formerly pushed
- `SummonPhase.getTrainerSlot()` added to replace the common pattern of `!(this.fieldIndex % 2) ? TrainerSlot.TRAINER : TrainerSlot.TRAINER_PARTNER`
- Resolved a bunch of leftover `strict-null` `TODO`s

## How to test the changes?
`npm run test`

## Checklist
  - Uhh.... oops


Tempo-anon (commented):
Can you add comments on the class declaration of each phase while you're here? I notice a lot of them aren't commented

MokaStitcher (commented):
only a third of the way through, but since i'm stopping and found a few issues, sharing them right now

Also before I forget, I think we can change the `phases/mystery-encounter-phases` folder to just `phases/mystery-encounter`

MokaStitcher (commented):
i think i'll want to do some amount of playtesting before approving, but code review is done at least

MokaStitcher (commented):
Did some playtesting, found one issue. I confirmed that it doesn't happen on current beta. I have not attempted to figure out where it's coming from though

<details><summary>When starting a new daily run the wrong biome background is used</summary>

 - start daily run. Today's starts in metropolis (for the next 3 hours), but the background and platforms are from Town
 - after clearing a wave, next wave's pokemon comes in with the metropolis platform

This happens regardless of which biome is used on the title screen. But when loading runs things are fine.
I'm actually not sure it's purely related to dailies, but since all other modes start in town and town seems to be used as a default i can't get this issue to happen in other modes

Screenshot_20250101-203958


</details>

MokaStitcher (dismissed):
once that fix gets in it'll be good on my end

<details><summary>much better</summary>

(today's starting biome is Plains so i changed it to something easier to see)

image

</details>

src/battle.ts
```diff
@@ -179,8 +179,8 @@ export default class Battle {
         )
         .map((i) => {
           const ret = i as PokemonHeldItemModifier;
-          //@ts-ignore - this is awful to fix/change
-          ret.pokemonId = null;
+          // this is awful to fix/change
+          ret.pokemonId = null!;
```
torranx, line 183: why is there a bang for `null` here?

src/data/move-attrs/hit-heal-attr.ts
```diff
@@ -60,7 +60,13 @@ export class HitHealAttr extends MoveEffectAttr {
         message = "";
       }
     }
-    globalScene.unshiftPhase(new PokemonHealPhase(user.getBattlerIndex(), healAmount, message, false, true));
+    globalScene.unshiftPhase(
+      new PokemonHealPhase(user.getBattlerIndex(), healAmount, {
+        message: message,
```
torranx:
```suggestion
        message,
```

src/phases/attempt-run-phase.ts
```diff
@@ -63,7 +66,13 @@ export class AttemptRunPhase extends PokemonPhase {
     this.end();
   }
 
-  attemptRunAway(playerField: PlayerPokemon[], enemyField: EnemyPokemon[], escapeChance: NumberHolder) {
+  /**
+   * Calculates the base escape chance based on the current field.
+   * @param playerField - Usually {@linkcode BattleScene.getPlayerField()}
+   * @param enemyField - Usually {@linkcode BattleScene.getEnemyField()}
```
torranx: Why is this termed "usually"? Also I suggest removing the `()`

torranx (commented):
46/153 files reviewed

src/battle.ts
DayKev: The bang replaces the `@ts-ignore`. It achieves a similar effect (telling the compiler to allow the `null`) without disabling *every* error.

src/phases/attempt-run-phase.ts
DayKev: Rewrote the param docs.

src/battle.ts
torranx: Isn't using `@ts-ignore` better than using bang on a `null`? (sorry it's just looks too off to me because we're telling the compiler it's not null but it is actually null)

DayKev: `@ts-ignore` will disable all errors from being displayed, but we only want to disable the error from the `null` assignment. I already ran into this causing a problem once before, an assignment to a `readonly` field was hidden by a `// @ts-ignore` for a different error.

`!` is basically a more limited error suppression command.

`@ts-ignore` should only be used as a last resort imo.

torranx: ok but can we comment the reason why this is `null` and probably a `TODO` as well so that someone who'll look at this won't get confused?

DayKev: I have no idea what the reason is but I can add `TODO` to it.

torranx (approved):

MokaStitcher (approved):
played a few more waves for good measure. Time to send it o7

PR #50, opened 2024-12-12
[Refactor] Integrate a SettingsManager

### LINKED ISSUE #938: [Refactor] Integrate a settings manager

flx-sta:
> [!NOTE]
> - ~~As of right now the automatic reload only works on the language change.~~
>     - ~~_Reason for that is that the whole handling of that is supposed to change anyway so there was no point in integrating it_~~
> - Also no settings-related stuff inside `game-data` was touched as I'm not fully aware of the interaction with the actual game-data object being loaded. **If anyone is better informed about that feel free to help!**

- Reload required is only shown as a `*` now with an explanation of that `*` at the bottom of the settings screen
- Opening the change-language menu doesn't reload if no change happened. It also automatically switches back to the actual language option instead of staying on `Change`
- Move touch controls now has 2 options and automatically updates it label on the fly

All settings were tightly coupled with the battle-scene (as almost everything). This made it impossible to read settings before the battle-scene is ready.

- All settings are now handled by the `SettingsManager` (or rather the intance of it).
- Settings are updated through a `settings.update(<category>, <key>, <value>)` method that also emits events about these changes (in `eventBus`. See `event-bus.ts` file for all the events).
- localstorage keys are now prefixed with `pky/` to make sure that no other localhost development interferes with it.
- added locales: https://github.com/Despair-Games/poketernity-locales/commit/96bf7511264b5c97348463b24a0f3d24e5a1c35f

## Screenshots/Videos

<details>
  <summary> [Videos] </summary>

https://github.com/user-attachments/assets/98b3e329-03bc-40c4-b4b1-aca4e0617c4d

https://github.com/user-attachments/assets/aab3c974-c18b-41b5-a60a-4d7454f5d966

https://github.com/user-attachments/assets/f5d561a2-c8bf-46da-9d37-1b1fcf94f565

https://github.com/user-attachments/assets/528a0b25-2df6-486d-98e6-7bf720700e79

</details>

## How to test the changes?

- Start the game
- Play around with the settings and analyze if they actually take effect!

## Checklist

  - ~~[ ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?~~


flx-sta:
https://github.com/user-attachments/assets/aab3c974-c18b-41b5-a60a-4d7454f5d966

https://github.com/user-attachments/assets/f5d561a2-c8bf-46da-9d37-1b1fcf94f565

https://github.com/user-attachments/assets/528a0b25-2df6-486d-98e6-7bf720700e79

flx-sta:
image
image

innerthunder (dismissed):

MokaStitcher (commented):
again stopping very early on in the review, sorry
I was trying to look at game data, the gamepad/keyboard remapping thing is such a mess wtf, so many different functions to save, called from different places '-'
I'll get back to it properly later, but for now, I think you missed to update the locale storage key here like you did in the gamepad file https://github.com/Despair-Games/poketernity/blob/4dacb5295e1265461b41ca1ec63494824a0e5b40/src/ui/settings/settings-keyboard-ui-handler.ts#L41

MokaStitcher (commented):
Made a first full pass over the code and did some limited testing.

Overall looks good, the main remarks I have so far:
- not sure if you consider the gamepad/keyboard remapping part done (I didn't look into that too closely, following the issues I found in my earlier comment)
- the game-data file needs to be cleaned up to remove everything that is no longer needed if i understand correctly? Probably settings.ts too, so that the deprecated things can be removed?
- tutorials disabled by default, is that intended?
- disabling touchpad no longer asks for confirmation


<details><summary><sub>(this is a note for me or I'll forget, don't mind me)</summary>

not introduced in this PR, but the remapping text + timer are not centered properly (screenshot is italian, but in french it goes out of the window's bounds)

image

</details>

src/@types/Settings.ts
```diff
@@ -0,0 +1,124 @@
...
+  activeIndex: number;
+  enabled: boolean;
+}
+
+export type SettingUiItemOption = {
+  value: number | string | boolean;
+  label: string;
+  needConfirmation?: boolean;
+  confirmationMessage?: string;
+};
```
MokaStitcher: Like we had discussed on Discord a while ago, either these lines need to go, as well as the code using them [here](https://github.com/Despair-Games/poketernity/blob/0f2d8030eef3d66ca9ce1f4184c6673ab70eaf7e/src/ui/settings/abstract-settings-ui-handler.ts#L464)
Or, this https://github.com/Despair-Games/poketernity/blob/0f2d8030eef3d66ca9ce1f4184c6673ab70eaf7e/src/ui/settings/abstract-settings-ui-handler.ts#L588 needs to go, and the [enableTouchControl ui item definition](https://github.com/Despair-Games/poketernity/blob/0f2d8030eef3d66ca9ce1f4184c6673ab70eaf7e/src/ui/settings/settings-ui-items.ts#L129) need to use those `needConfirmation` and `confirmationMessage` field (for reference here's how they were used [before](https://github.com/Despair-Games/poketernity/blob/3c14d91bb7484135dcb7daa6c3ddc0dbcf2cb5d7/src/system/settings/settings.ts#L56))

My opinion hasn't changed that I'd rather go with option 2, because the abstract settings ui handler shouldn't have code for any specific setting. Also because this way the fact that a setting needs confirmation is handled in the same place as the fact that a setting requires reloading, and it's all in one place.

But I won't fight whichever option you go with, just need to pick one or the other and clean up the rest

flx-sta: I totally forgot about that discussion, sorry :D

flx-sta: done

innerthunder (dismissed):
lgtm. I noticed some consts/variables with PokeRogue branding (`PRSFX` specifically), but I suppose that's something for another PR to resolve...

flx-sta:
> lgtm. I noticed some consts/variables with PokeRogue branding (`PRSFX` specifically), but I suppose that's something for another PR to resolve...

correct, that's something for another PR.
Afaik the art team wanted to chip in for that

src/constants.ts
```diff
@@ -17,7 +17,22 @@ export const SAVES_ZIP_PREFIX = "poketernity_";
 export const SAVE_FILE_EXTENSION = "txt";
 
 /** Prefix for local storage keys. */
-export const LS_PREFIX = "pkty";
+export const LS_PREFIX = "pky";
```
Tempo-anon: was there a reason this was changed?

flx-sta: We have a pending vote for the final decision. While that's going on I'll leave it as it is

src/system/game-speed.ts
```diff
@@ -21,7 +19,7 @@ export function initGameSpeed() {
     if (num instanceof FixedNumber) {
       return num.value;
     }
-    return thisArg.gameSpeed === 1 ? num : Math.ceil((num /= thisArg.gameSpeed));
+    return settings.settings.general.gameSpeed === 1 ? num : Math.ceil((num /= settings.settings.general.gameSpeed));
```
MokaStitcher:
```suggestion
    return settings.general.gameSpeed === 1 ? num : Math.ceil((num /= settings.general.gameSpeed));
```

why did that even work

src/system/settings/settings-manager.ts
```diff
@@ -0,0 +1,194 @@
...
+    };
+
+    this.loadFromLocalStorage();
+  }
+
+  /**
+   * Getter for {@linkcode _settings}. No public setter!
+   */
+  get settings() {
+    return this._settings;
```
MokaStitcher: ah, is this why settings.settings works?
Do we need this getter since there's one for each sub category?

src/ui/settings/settings-keyboard-ui-handler.ts
```diff
@@ -37,7 +38,7 @@ export default class SettingsKeyboardUiHandler extends AbstractControlSettingsUi
     this.configs = [cfg_keyboard_qwerty];
     this.commonSettingsCount = 0;
     this.textureOverride = "keyboard";
-    this.localStoragePropertyName = "settingsKeyboard";
+    this.localStoragePropertyName = MAPPING_CONFIG_LS_KEY;
```
MokaStitcher:
```suggestion
```

ah, looks like I forgot to remove [localStoragePropertyName ](https://github.com/Despair-Games/poketernity/blob/7ff8e8d136c1a6e64841e6829ea7d1663260e62b/src/ui/settings/abstract-control-settings-ui-handler.ts#L66) from the parent class. But I'm pretty sure it's not used anymore, so might as well remove it everywhere (will need to remove the MAPPING_CONFIG_LS_KEY import too)

src/ui/settings/settings-ui-items.ts
```diff
@@ -0,0 +1,374 @@
...
+    options: useOnOffOptions(),
+  },
+  {
+    key: "enableTouchControls",
+    label: t("settings:touchControls"),
+    options: useAutoDisabledOptions(
+      {},
+      { requiresConfirmation: true, confirmationMessage: t("settings:confirmDisableTouch") },
+    ),
+    //TODO: needs confirmation for `disabled`
```
MokaStitcher:
```suggestion
```

should be good right?

MokaStitcher (commented):
A few notes. Getting there.
Time to playtest a bunch :D

src/system/settings/default-settings.ts
```diff
@@ -0,0 +1,80 @@
...
+import { PlayerGender } from "#enums/player-gender";
+import { UiWindowType } from "#enums/ui-window-type";
+
+export const defaultGeneralSettings: GeneralSettings = {
+  gameSpeed: 2,
+  hpBarSpeed: HpBarSpeed.DEFAULT,
+  expGainsSpeed: ExpGainsSpeed.DEFAULT,
+  partyExpNotificationMode: ExpNotification.DEFAULT,
+  skipSeenDialogues: false,
+  eggSkipPreference: EggSkipPreference.NEVER,
```
MokaStitcher:
```suggestion
  eggSkipPreference: EggSkipPreference.ASK,
```
Asking was the default [before](https://github.com/Despair-Games/poketernity/blob/0c56fc32b467de4aa12ba6ef982d1b351daa0cb7/src/system/settings/settings.ts#L320)

MokaStitcher (commented):
Testing went well, once the issues in this review and the previous one are fixed i'll be able to approve fast!

edit (x2): nevermind ~~one more~~ couple more issues
- These are used as the locale storage key for custom touch control positions: https://github.com/Despair-Games/poketernity/blob/0b6d618d05d9e10f1d850d1f8ddf2f427d0cb122/src/ui/settings/move-touch-controls-handler.ts#L6-L7 They should probably also use the pky prefix, and if possible not be saved in their own individual entry like they are now? Maybe that second part can be for another time but since this is settings related and other settings like mappings had their keys changed, it would make sense to have it done in this PR

- Another local storage key that didn't get cleaned up here:
https://github.com/Despair-Games/poketernity/blob/0b6d618d05d9e10f1d850d1f8ddf2f427d0cb122/src/system/game-data.ts#L73-L74
Although the settings one is not used, we might as well return the correct key, or add a TODO to update this whole thing


Final thing, we also should merge that locales PR https://github.com/Despair-Games/poketernity-locales/pull/13

src/ui/settings/settings-keyboard-ui-handler.ts
DayKev: Uh, the "suggested" code is identical to the existing code. 🤔

MokaStitcher: hu, right, thanks lol. This suggestion shouldn't be applied on its own anyway it requires touching up 3 files in total I think

src/system/game-data.ts
```diff
@@ -700,106 +671,44 @@ export class GameData {
...
-
     return true; // TODO: is `true` the correct return value?
   }
 
   public saveTutorialFlag(tutorial: Tutorial, flag: boolean): boolean {
     const key = getDataTypeKey(GameDataType.TUTORIALS);
     let tutorials: object = {};
-    if (localStorage.hasOwnProperty(key)) {
-      tutorials = JSON.parse(localStorage.getItem(key)!); // TODO: is this bang correct?
+    const lsItem = localStorage.getItem(key);
```
MokaStitcher, line 705: uh, had forgotten about this change, it's pretty unrelated to the PR and will cause conflicts with #270, but no matter, unless you feel like reverting it

src/ui/settings/move-touch-controls-handler.ts
```diff
@@ -45,41 +42,139 @@ export default class MoveTouchControlsHandler {
...
     this.touchControls = touchControls;
     this.inConfigurationMode = false;
     this.setPositions(this.getSavedPositionsOfCurrentOrientation() ?? []);
-    window.addEventListener("resize", (_event) => {
-      const screenSize = this.getScreenSize();
-      if (screenSize.width > screenSize.height !== this.isLandscapeMode) {
-        this.changeOrientation(screenSize.width > screenSize.height);
-      }
+    this.initListeners();
+    console.log("MoveTouchControlsHandler currentOrientation:", this.currentOrientation);
```
MokaStitcher:
```suggestion
```

src/constants.ts
```diff
@@ -17,10 +17,25 @@ export const SAVES_ZIP_PREFIX = "poketernity_";
...
 
 /** Value used for byapssing login values */
 export const bypassLogin = import.meta.env.VITE_BYPASS_LOGIN === "1";
+
+/** Key for the local storage item storing the settings. */
+export const SETTINGS_LS_KEY = `${LS_PREFIX}/settings`;
+
+/** Key prefix for the local storage item storing the users data. */
+export const DATA_LS_KEY_PREFIX = `${LS_PREFIX}/data`;
+
```
MokaStitcher:
```suggestion
```

What is this for? It doesn't seem to be used, better leave it for a later PR that will handle updating all local storage keys

src/constants.ts
```diff
@@ -17,10 +17,25 @@ export const SAVES_ZIP_PREFIX = "poketernity_";
 export const SAVE_FILE_EXTENSION = "txt";
 
 /** Prefix for local storage keys. */
-export const LS_PREFIX = "pkty";
+export const LS_PREFIX = "pky";
```
MokaStitcher:
```suggestion
export const LS_PREFIX = "pkty";
```
adding it since i made other suggestions so you don't need a separate commit

src/ui/settings/move-touch-controls-handler.ts
```diff
@@ -109,52 +204,37 @@ export default class MoveTouchControlsHandler {
...
       this.saveCurrentPositions();
       this.disableConfigurationMode();
+      eventBus.emit("touchControls/move/save");
+      eventBus.emit("touchControls/move/end");
     });
-    resetButton.addEventListener("click", () => {
+    this.resetBtn.addEventListener("click", () => {
       this.resetPositions();
+      eventBus.emit("touchControls/move/reset");
+      eventBus.emit("touchControls/move/end");
```
MokaStitcher:
```suggestion
```

I don't think the reset option should emit the end event, since it doesn't close the view

src/ui/settings/settings-ui-items.ts
```diff
@@ -0,0 +1,373 @@
...
+      },
+      {
+        value: 1,
+        label: t("settings:configure"),
+      },
+    ],
+    touchscreenOnly: true,
+  },
+  {
+    key: "enableVibration",
```
MokaStitcher: if I could have one final nit, it would be to put the enableVibration setting above the enable touch control one, as it was before

MokaStitcher (commented):
few final nits. Looks good!
I merged the remaining 2 locale PRs so you can go ahead and update the submodule in this PR (if you use command line for it: `git submodule update --remote --recursive --force` )

src/ui/settings/move-touch-controls-handler.ts
flx-sta: good point

src/system/game-data.ts
flx-sta: I think it's fine. Whoever comes first will easily resolve it

MokaStitcher (dismissed):
🚀🚀🚀

src/ui-inputs.ts
```diff
@@ -216,23 +216,13 @@ export class UiInputs {
...
 
   buttonSpeedChange(up = true): void {
-    const settingGameSpeed = settingIndex(SettingKeys.Game_Speed);
-    if (up && globalScene.gameSpeed < 5) {
-      globalScene.gameData.saveSetting(
-        SettingKeys.Game_Speed,
-        Setting[settingGameSpeed].options.findIndex((item) => item.label === `${globalScene.gameSpeed}x`) + 1,
-      );
+    if (up && settings.general.gameSpeed < 5) {
+      settings.update("general", "gameSpeed", settings.general.gameSpeed + 1);
```
PigeonBar: This approach causes edge cases when pressing the "speed up" or "speed down" keys while at a fractional game speed.

https://github.com/user-attachments/assets/23f82016-008a-49b5-a8a1-0b17bee23774

(Most likely, we'll have to replace these with separate functions to search up the previous or next speed option.)

flx-sta: Good catch

flx-sta: Fixed: 3a55ba0

innerthunder (dismissed):

src/ui-inputs.ts
```diff
@@ -216,23 +217,17 @@ export class UiInputs {
   }
 
   buttonSpeedChange(up = true): void {
```
MokaStitcher, line 219: Proposed rewrite of the function (untested). I don't like that this is calling the uiHandler's `show` function and would rather just **NOT** allow changing speed with the shortcuts while in the settings. Also this would only update the setting if there is a change

```ts
      buttonSpeedChange(up = true): void {
        if (globalScene.ui?.getMode() === Mode.SETTINGS) {
          return;
        }
        const index = gameSpeeds.findIndex((n) => n === settings.general.gameSpeed) ?? 0;
        let newIndex = up ? Math.min(index + 1, gameSpeeds.length - 1) : Math.max(index - 1, 0);
        if (newIndex !== index) {
          settings.update("general", "gameSpeed", gameSpeeds[newIndex]);
        }
      }
```

flx-sta: ~~@MokaStitcher doesn’t this eliminate the whole purpose of the “quick access” of those buttons ? 
I’d rather keep them as they are. But I’m open for reasons to apply your change~~

mb, I missed the not

MokaStitcher: The quick buttons are meant to be used during gameplay, not while in the settings, that makes absolutely no sense to me. Why would you want to change game speed with a shortcut no one knows about while being in the audio settings menu for example?
<sub>I'd also argue that no one uses them either way.

I don't know that show is basically re-rendering the whole settings I don't see a world where it wouldn't cause issues to do that while you're navigating said settings. But I guess I should just try it and see if that's the case

MokaStitcher: if you don't want to change that, fine but can you at least move the duplicate 
```
settings.update("general", "gameSpeed", gameSpeeds[nextIndex]);
      if (globalScene.ui?.getMode() === Mode.SETTINGS) {
        (globalScene.ui.getHandler() as SettingsUiHandler).show([]);
      }
```
outside of the if/else, and only execute it if nextIndex is not the same as the current one? (also I just noticed it was `if (up) ... else if (!up)` instead of just `if (up) ... else ...` lol)

MokaStitcher: Here's an argument against allowing the shortcuts' use in the settings:
if you press the shortcut button because you want to map it to another action, it also calls the function and updates the speed
image

I didn't find other issues with calling show itself, aside from the fact that if you're in the General tab it scrolls you back to the top of the tab, which is a bit weird but fine

flx-sta: mb, I missed the `not`

innerthunder (approved):

MokaStitcher (approved):

PR #120, opened 2024-12-26
[Move] Partially Reimplement Future Sight and Doom Desire

innerthunder:

This makes some improvements to [Future Sight](https://bulbapedia.bulbagarden.net/wiki/Future_Sight_(move)) (and Doom Desire):
- The move can now be used consecutively against different targets in a double battle, and will fail when used consecutively against the same target.
- If the original target fainted or was otherwise removed from the field in a double battle on the turn a delayed attack would activate, the delayed attack is now redirected to the target's ally (if available).

However, this does not fully implement the move, with the known differences in the moves being:
- The moves apply the user's abilities and held items regardless of the user's party position. In mainline, these are only applied when the user is on the field.
- When these moves are called by other moves like Metronome, they attack immediately, ignoring the delayed effect.

Untangling some spaghetti in an old move implementation.

All uses of Future Sight and Doom Desire are now tracked by a single `DelayedAttackTag` arena tag and under a single `DELAYED_ATTACK` tag type. The tag contains a queue of delayed attacks, with each element holding the following data:
```ts
interface DelayedAttack {
  source: Pokemon | null;
  move: Moves;
  targetIndex: BattlerIndex;
  turnCount: number;
}
```
Whenever Future Sight or Doom Desire is used, it adds this tag to the arena (if the tag doesn't already exist) and then adds an attack to the tag's queue. When the tag lapses at the end of the turn, all attacks in the tag's queue have their `turnCount` decremented. When an attack's `turnCount` reaches 0, a new `MoveEffectPhase` is queued invoking the `move` as if it were used by the `source` Pokemon against the given `targetIndex`. If the targeted index doesn't have an active Pokemon in a double battle, the attack will try to redirect to the original target's ally index.

Other tasks related to this implementation:

- [x] Adjusted params to arena tag functions -- The old Future Sight tag was the only one to use the `targetIndex` param.
- [x] Removed temporary fixes to game-breaking Future Sight bugs in `MovePhase` and `MoveEffectPhase`
- [x] Added tests to `future_sight.test.ts` to cover edge cases.
- [x] Fix flyout display issues caused by the new implementation -- Might require locale changes
  - Future Sight and Doom Desire are now excluded from the arena flyout; the arena tag's new format doesn't lend itself to providing useful information within the flyout. 

## Screenshots/Videos

WIP

## How to test the changes?

`npm run test future_sight`

## Checklist


Snailman11:
> [Remove delayed attacks from arena flyout](https://github.com/Despair-Games/poketernity/pull/120/commits/35c8f54f2ecb61bc84c96094d190154cd59ee43d)

I understand that you might be removing the arena flyout from Future sight and Doom Desire. Would that mean you'd also need to remove the Locales for those arena tags in the flyout as well?
<https://github.com/Despair-Games/poketernity-locales/blob/e760388a00f371eb3264102238c2bf3f9853468e/en/arena-flyout.json#L27>

innerthunder:
> I understand that you might be removing the arena flyout from Future sight and Doom Desire. Would that mean you'd also need to remove the Locales for those arena tags in the flyout as well? https://github.com/Despair-Games/poketernity-locales/blob/e760388a00f371eb3264102238c2bf3f9853468e/en/arena-flyout.json#L27

I can remove them, but leaving them in should be harmless and we might want to save them if the arena flyout is improved or reworked

MokaStitcher (approved):
lgtm

flx-sta (approved):
lgtm: 

https://github.com/user-attachments/assets/50106ced-b06b-4bc2-9c59-729cf4955f60

<sub>I've changed the turns to `2` for testing (from `3`)</sub>

PR #159, opened 2025-01-01
[Test] Add new helpers for using moves and getting pokemon

DayKev:
N/A

There's 2 reasons (for the move helper changes):
- To remove the need to use the moveset overrides entirely in tests (less stuff test writers need to do).
- So that you can reliably check a pokemon's moveset as part of a test.

If you don't need to check the pokemon's moveset as part of the test you use `use()` or `forceEnemyMove()`, otherwise you set the pokemon's moveset with `changeMoveset()` and use `select()` or `selectEnemyMove()`.

The reason for this is that the way the moveset override works is that it constantly overwrites the pokemon's moveset (partially in the case of the player and fully in the case of the AI) by modifying `Pokemon.getMoveset()` rather than just overwriting it once (these overrides were designed to be used in manual in-game testing, not automated tests), meaning any test that needs to check changed properties of a pokemon's moveset could run into hard-to-debug issues if they use the moveset overrides (or the test could falsely pass with the test writer not knowing this).

New methods added to the `MoveHelper` class:
- Added `useMove()` which will overwrite the player pokemon's moveset and then call `select()`. Removes the need for modifying the global moveset override.
- Added `selectEnemyMove()` which is copied from `GameManager.forceEnemyMove()` and the old method is marked as `@deprecated`.
- Added `forceEnemyMove()` which is the enemy pokemon version of `useMove()`.

A new helper class `PokemonHelper` is added that contains two methods `getPlayerPokemon()` and `getEnemyPokemon()` which are passthroughs for `globalScene.get*Pokemon()` that tell the compiler to ignore the possibility of an `undefined` return value. This removes the need to add `!` to every `game.scene.get*Pokemon()` usage in tests.

## How to test the changes?
The tests for Whirlwind, Wide Guard and Will-O-Wisp were updated with the new methods as a proof of concept.

## Checklist


Tempo-anon (dismissed):

Tempo-anon (commented):
Oh the rename ballooned the changes huh oops.

Still a nit with variable naming but should be fine

torranx (dismissed):

test/testUtils/helpers/pokemonHelper.ts
```diff
@@ -0,0 +1,36 @@
...
+   * Note: `undefined` can still be returned if there is no valid pokemon, but the compiler will
+   * no longer warn of that possibility or require `null` checks in order to simplify test writing.
+   * @param includeSwitching Whether a pokemon that is currently switching out is valid, default `true`
+   * @returns The first {@linkcode PlayerPokemon} that is {@linkcode globalScene.getPlayerField on the field}
+   * and {@linkcode PlayerPokemon.isActive is active}
+   * (aka {@linkcode PlayerPokemon.isAllowedInBattle is allowed in battle}).
+   */
+  public getPlayerPokemon(includeSwitching: boolean = true): PlayerPokemon {
+    return this.game.scene.getPlayerPokemon(includeSwitching)!;
+  }
```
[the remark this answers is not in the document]
innerthunder: I think keeping the `expect` is better. The failure is more immediate and points directly to `getPlayerPokemon`.

test/testUtils/helpers/pokemonHelper.ts
```diff
@@ -0,0 +1,41 @@
...
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { type globalScene } from "#app/global-scene";
+
+/** Helper to manage pokemon */
+export class PokemonHelper extends GameManagerHelper {
+  /**
+   * Passthrough for {@linkcode globalScene.getPlayerPokemon} that ignores the `undefined` return value.
+   *
+   * Note: `undefined` can still be returned if there is no valid pokemon, but the compiler will
+   * no longer warn of that possibility or require `null` checks in order to simplify test writing.
```
MokaStitcher: is that doc still correct? Since there's the `expect` now the execution will just stop if it's undefined, right?

This doc and the getEnemyPokemon one probably need to be updated to clarify that

DayKev: Oh yeah. Changed the docs.

MokaStitcher (approved):

innerthunder (approved):

PR #164, opened 2025-01-03
[UI][Temporary] Moved addWindow and addTextObject into class functions of SceneBase

frutescens:

None.

This will make breaking globalScene up much easier. 

addWindow and addTextObject, as they are now, are coded with the default assumption that they are adding a Phaser object to global scene. We could extend them by having the current scene be a parameter; however, I feel like that leads to more confusion. By making them class functions of SceneBase, any scene can use these methods + it logically makes sense that this.addWindow() / this.addTextObject() is adding a Phaser object to this (the specific scene).

I personally do not want this to be a permanent change and ideally, these functions would go into a scene that governs UI specifically. But, this should make the process of breaking down globalScene much easier. This is pretty much adding a little bit of spaghetti to clean up a larger chunk of spaghetti however.

## How to test the changes?
Game should function as normal. 

## Checklist

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -50,7 +50,7 @@ import type { Phase } from "#app/phase";
 import { initGameSpeed } from "#app/system/game-speed";
 import { Arena, ArenaBase } from "#app/field/arena";
 import { GameData } from "#app/system/game-data";
-import { addTextObject, getTextColor, TextStyle } from "#app/ui/text";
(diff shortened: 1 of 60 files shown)


DayKev (commented):
It seems like you didn't remove the original functions in `ui-theme.ts`?

flx-sta:
@frutescens do you still actively work on this?

frutescens:
I'm taking a break from UI stuff. Feel free to close it and I'll probably learn the lessons of this

PR #165, opened 2025-01-03
[Refactor] Damage Reduction abilities return their multiplier value instead of processing damage after calculation. 

frutescens:

None (ideally).

Problem for testing if abilities are working properly. I think it also makes logical sense to move these multipliers up too. 

DamageBoostAbAttr, ReceivedDamageAbAttr, and AlliedFieldDamageReductionAbAttr now return their multiplier values, instead of changing damage itself, and these values are used when calculating the final damage product (before tokens are applied). 

## How to test the changes?
`npm run test`

## Checklist


frutescens:
> Seems like Ice Face and Disguise have to be decoupled from `ReceivedMoveDamageMultiplierAbAttr`. The best way to do so is probably to apply them in a similar way to Substitute in `MoveEffectPhase#applyMove()`. That way, they can properly modify damage after it has already been calculated.

I was thinking of doing that in another PR to keep this PR focused. What are your thoughts?

innerthunder (approved):
Assuming the changes to Disguise and Ice Face will be fleshed out in another PR

src/field/pokemon.ts
```diff
@@ -3393,11 +3380,43 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
           cancelled,
           simulated,
-          damage,
+          alliedFieldDamageMultiplier,
         );
       }
     }
 
+    /** If damage is nullified by a form-ability (Eiscue's Ice Face, Mimikyu's Disguise), then damage is set to 0 */
+    if (receivedDamageMultiplier.value > 0) {
```
DayKev, line 3389: If rDM is 0 then the damage will be 0 because it will be multiplied by 0, this `if`/`else` doesn't seem necessary.

DayKev: Wait I remember now, `toDmgValue` forces min 1... hm.

frutescens: It's because toDamage turns any 0s into 1s.

DayKev (approved):
TODO: Ice Face / Disguise

PR #185, opened 2025-01-06
[Ability][Checkup] Cleaned up Early Bird's implementation and tests

### LINKED ISSUE #114: [Ability][Checkup] Early Bird
## Ability Requirements
- [x] Documentation
- [x] Tests
- [x] Early Bird halves the sleep duration. When the Pokémon is put to sleep, the number of turns it will remain asleep is preset, between 1 and 5 (the range depends on the generation). 
- [x] This number of turns is halved for a Pokémon with Early Bird, rounded down if it is odd. So if only 1 turn is preset, it is rounded down to 0, causing the Pokémon to wake up the next time it moves.
- [x] This also applies to the self-induced sleep from Rest, reducing its length from 2 turns to 1 turn asleep (not counting the turn in which Rest is used). 

## Related PRs
- https://github.com/Despair-Games/poketernity/pull/253
- https://github.com/pagefaultgames/pokerogue/pull/4632

## Related Bugs
- None as of 01-06-2024

## References
- [Bulbapedia - Early Bird](https://bulbapedia.bulbagarden.net/wiki/Early_Bird_(Ability))

frutescens:

None.

Ability checkups

Renamed Early Bird's attribute to ReduceSleepDurationAbAttr and updated its file name to reflect that.
Brought its implementation closer to mainline where the turn duration is reduced when the # of turns of asleep is generated/provided. 
Cleaned up the tests. As a result, this significantly reduced the # of turns needed to pass the requirements. 
Updated trySetStatus() to consider turnsRemaining as a parameter before attempting to generate a value for it when setting sleep. 

## How to test the changes?

`npm run test early_bird`

## Checklist


DayKev:
The previous functionality (an extra -1 turn counter per turn) is the mainline/showdown behavior.

frutescens:
Changed implementation:
image

Beta:
image

200 tiny milliseconds

DayKev:
If we don't care about matching mainline then Overgrow et al shouldn't have been changed (which increased complexity too), otherwise this shouldn't be changed away from mainline behavior.

PR #186, opened 2025-01-06
[Refactor] Consolidate anims into a single class

torranx:
none

anims are all over the place, with some functions duplicated

- created a new class `Animation` that holds all animations other than battle-anims
- - I didn't use the `globalScene` here and opted for a parameter accounting for an eventual scene breakdown (hopefully)
- - as discussed with the team, the animations are now attach to `scene`
- moved all anims from `src/phases/abstract-form-change-base-phase.ts` to the new class
- moved all anims from `src/data/mystery-encounters/utils/encounter-transformation-sequence.ts` to the new class

## How to test the changes?
play test or `npm run test`

## Checklist


src/anims.ts
```diff
@@ -0,0 +1,594 @@
...
+   */
+  public doSpiralUpward(
+    transformationBaseBg: Phaser.GameObjects.Image,
+    transformationContainer: Phaser.GameObjects.Container,
+    xOffset: number = 0,
+    yOffset: number = 0,
+  ): void {
+    let f = 0;
+
+    this.scene.tweens.addCounter({
```
MokaStitcher: this is the only place in the class where `this.scene` is used. Everywhere else is the globalScene

I'm a bit wary of replacing all instances of globalScene but at the same time if we don't the `scene` field here is not really needed and these functions could just exist outside of a class, as utils

But I'm assuming the point of the class is to make it independent of the globalScene for possible future splitting of the  BattleScene?

torranx: > But I'm assuming the point of the class is to make it independent of the globalScene for possible future splitting of the BattleScene?

yes

innerthunder (dismissed):

src/anims.ts
```diff
@@ -0,0 +1,594 @@
+import { globalScene } from "#app/global-scene";
+import type { SceneBase } from "#app/scene-base";
+import { getFrameMs, randGauss, randInt } from "#app/utils";
+import { PokeballType } from "#enums/pokeball";
+import type { Variant } from "./data/variant";
+// tsdoc import
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { BattleAnim } from "./data/battle-anims";
```
DayKev: Btw I found that it's better to put these above the normal imports so that the IDE's import management doesn't mix them together.
```suggestion
```

src/anims.ts
MokaStitcher: then we should probably yeet globalScene everywhere and replace it with this.scene? ^^

src/anims.ts
```diff
@@ -0,0 +1,594 @@
+import { globalScene } from "#app/global-scene";
```
DayKev:
```suggestion
// -- start tsdoc imports --
// eslint-disable-next-line @typescript-eslint/no-unused-vars
import { BattleAnim } from "#app/data/battle-anims";
// -- end tsdoc imports --

import { globalScene } from "#app/global-scene";
```

src/anims.ts
torranx: > then we should probably yeet globalScene everywhere and replace it with this.scene? ^^

oh thanks for catching that, I forgot to replace them lol

src/data/mystery-encounters/encounters/fun-and-games-encounter.ts
```diff
@@ -338,6 +338,8 @@ async function showWobbuffetHealthBar() {
 
 function summonPlayerPokemonAnimation(pokemon: PlayerPokemon): Promise<void> {
   return new Promise<void>((resolve) => {
+    const anims = new Animation();
```
DayKev: Doing `const anims = new Animation();` in every file seems redundant, couldn't it just be done once globally?

src/anims.ts
MokaStitcher: I'm wondering though... Because this means we need to instanciate a new Animation object in so many places, if we shouldn't create a single one and use it everywhere
Maybe the scene should hold onto its own animation object?

src/phases/egg-hatch-phase.ts
```diff
@@ -1,21 +1,21 @@
...
-import type { EggLapsePhase } from "#app/phases/egg-lapse-phase";
 import { achvs } from "#app/system/achv";
 import EggCounterContainer from "#app/ui/egg-counter-container";
 import type EggHatchSceneHandler from "#app/ui/egg-hatch-scene-handler";
 import PokemonInfoContainer from "#app/ui/pokemon-info-container";
 import { Mode } from "#app/ui/ui";
-import { fixedInt, getFrameMs, randInt } from "#app/utils";
 import i18next from "i18next";
 import SoundFade from "phaser3-rex-plugins/plugins/soundfade";
+import { fixedInt, getFrameMs, randInt } from "#app/utils";
```
DayKev: That's weird, why did the imports get mixed up like this?

src/anims.ts
torranx: Hmm makes sense, will update the pr 👍

src/phases/egg-hatch-phase.ts
torranx: merge conflicts

src/anims.ts
flx-sta, line 1: Can we write the full word for the file?
`animations.ts`

torranx: I made it `anims` to be consistent with `battle-anims` and the now-deleted-file `src/field/anims.ts`

src/data/mystery-encounters/encounters/fun-and-games-encounter.ts
flx-sta: Or make the methods static. WHich would basically make it kind of like a util but it works (?)

DayKev: That could work too.

torranx: but that would make it stick to a single `scene`, this needs to be instantiated if we want to have a dynamic `scene`, unless we wanna pass `scene` to every method

pom-eranian: but is that not already handled by the [AnimationManager](https://docs.phaser.io/api-documentation/class/animations-animationmanager#create)?

torranx: the animations on this PR is not the same with the animations that you create when using phaser's AnimationManager

Screenshot 2025-01-08 at 2 43 03 AM

src/animations.ts
```diff
@@ -1,9 +1,10 @@
+// -- start tsdoc imports --
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { BattleAnim } from "./data/battle-anims";
+// -- end tsdoc imports --
```
flx-sta: I didn't follow all the discussion: What the goal here of this whole mess? Simply include the file?

src/phases/evolution-phase.ts
```diff
@@ -101,7 +101,7 @@ export class EvolutionPhase extends FormChangeBasePhase {
                   this.bgVideo.play();
                 });
                 globalScene.playSound("se/charge");
-                this.anims.doSpiralUpward(this.baseBgImg, this.container);
+                globalScene.animations.doSpiralUpward(this.baseBgImg, this.container);
```
flx-sta: NIT: consider destrcuturing the globalScene 

```ts
const { animations } = globalScene;

File: src/animations.ts (hunk 6 above)
Comment (torranx) [reply src/animations.ts]: yes, for proper typing on tsdocs

File: src/animations.ts [hunk 8]
```diff
@@ -0,0 +1,594 @@
...
+import { getFrameMs, randGauss, randInt } from "#app/utils";
+import { PokeballType } from "#enums/pokeball";
+import type { Variant } from "./data/variant";
+import type BattleScene from "./battle-scene";
+
+/**
+ * Class for handling general animations such as particle effects.
+ * For battle animations, see {@linkcode BattleAnim}.
```
flx-sta, line 12:
```suggestion
 * For battle animations, see {@linkcode import("./data/battle-anims").BattleAnim}.
```
Maybe use that instead of the 
```ts
// -- start tsdoc imports --
// eslint-disable-next-line @typescript-eslint/no-unused-vars
import { BattleAnim } from "./data/battle-anims";
// -- end tsdoc imports --
```
?

src/animations.ts
flx-sta: Oh okay. I've added a follow up comment for that

src/animations.ts
torranx: I've tried this before and it does not work
Screenshot 2025-01-09 at 5 10 29 AM

flx-sta: That's dumb

MokaStitcher: I've been wondering about these imports for doc purposes in the way they've been added in a few recent PRs. Could these be impacting execution order and cause more of some of the issues we've had with that? We've been trying to cut circular dependencies too, how do these work in regards to that?
Are they just fully ignored when building because they are not used?

flx-sta: @MokaStitcher yes the will indeed influence the build since the file itself will be imported!
That's why I'm so eager to eliminate it. 
BUT if I'm not mistaken using a `type` import `import type { [...]` might eliminate this issue since type imports are for development/compiler only

----

> When you import a type (using `import { type SomeType } from 'module'`), it has **no** impact on the runtime because TypeScript removes type information during compilation

MokaStitcher: Interesting
Yeah if they can influence the build I'm not sure I'd want them added like that either

torranx: > BUT if I'm not mistaken using a type import import type { [...] might eliminate this issue since type imports are for development/compiler only

yes type import eliminates the issue (I didnt notice I wasnt using one, nice catch)

torranx: updated

flx-sta (dismissed):
Enough said! 
_Gotta merge them all!_

PigeonBar (dismissed):

PR #149, opened 2024-12-30
[Refactor][Move] Make `args` type-inferrable for move attributes

innerthunder:
## What are the changes the user will see?

N/A

## What are the changes from a developer perspective?

`MoveAttr#apply`'s signature has been changed from

```ts
apply(_user: Pokemon | null, _target: Pokemon | null, _move: Move, _args: any[]): boolean { }
```

to

```ts
apply(_user: Pokemon | null, _target: Pokemon | null, _move: Move, ..._args: unknown[]): boolean { }
```

This allows for type inference for any argument passed to subclasses that override this function. Most move attributes, as well as `applyMoveAttrs` and related functions, have been adjusted to reflect this change.

## Why am I making these changes?

The added type inference helps with documentation, readability, and overall implementation stability for new and existing move attributes. This pattern was done previously for modifiers and arena tags' apply functions, and should be applied to abilities after this PR.

## Screenshots/Videos

Results of `npm run test`:

image

## How to test the changes?

`npm run test:silent`

## Checklist

- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test`)
  - [x] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?
- [n/a] Have I provided screenshots/videos of the changes (if applicable)?
  - [n/a] Have I made sure that any UI change works for both UI themes (default and legacy)?

PR #197, opened 2025-01-07
[Refactor] Make `args` type-inferrable for Ability attributes

innerthunder:

N/A

Cleaner code is nice. This was also done for move attributes in #149 

`AbAttr#apply`'s signature has been changed from

```ts
apply(
  _pokemon: Pokemon,
  _passive: boolean,
  _simulated: boolean,
  _cancelled: BooleanHolder | null,
  _args: any[],
): boolean {
```

to

```ts
apply(_pokemon: Pokemon, _passive: boolean, _simulated: boolean, ..._args: unknown[]): boolean {
```

allowing for overrides of the function to have named parameters for any additional data instead of being forced to use the pattern

```ts
const arg = args[0] as X
// do stuff with arg
```

## How to test the changes?

`npm run test`

## Checklist

### DIFF
--- a/src/data/ab-attrs/copy-fainted-ally-ability-ab-attr.ts
+++ b/src/data/ab-attrs/copy-fainted-ally-ability-ab-attr.ts
@@ -7,13 +7,7 @@ import { PostKnockOutAbAttr } from "./post-knock-out-ab-attr";
 import { UncopiableAbilityAbAttr } from "./uncopiable-ability-ab-attr";
 
 export class CopyFaintedAllyAbilityAbAttr extends PostKnockOutAbAttr {
-  override applyPostKnockOut(
-    pokemon: Pokemon,
-    _passive: boolean,
-    simulated: boolean,
-    knockedOut: Pokemon,
-    _args: any[],
-  ): boolean {
+  override applyPostKnockOut(pokemon: Pokemon, _passive: boolean, simulated: boolean, knockedOut: Pokemon): boolean {
     if (pokemon.isPlayer() === knockedOut.isPlayer() && !knockedOut.getAbility().hasAttr(UncopiableAbilityAbAttr)) {
       if (!simulated) {
         pokemon.summonData.ability = knockedOut.getAbility().id;
--- a/src/data/ab-attrs/copy-fainted-ally-ability-ab-attr.ts
+++ b/src/data/ab-attrs/copy-fainted-ally-ability-ab-attr.ts
@@ -7,14 +7,22 @@ import { PostKnockOutAbAttr } from "./post-knock-out-ab-attr";
 import { UncopiableAbilityAbAttr } from "./uncopiable-ability-ab-attr";
 
 export class CopyFaintedAllyAbilityAbAttr extends PostKnockOutAbAttr {
-  override applyPostKnockOut(pokemon: Pokemon, _passive: boolean, simulated: boolean, knockedOut: Pokemon): boolean {
-    if (pokemon.isPlayer() === knockedOut.isPlayer() && !knockedOut.getAbility().hasAttr(UncopiableAbilityAbAttr)) {
+  override applyPostKnockOut(
+    pokemon: Pokemon,
+    _passive: boolean,
+    simulated: boolean,
+    knockedOutPokemon: Pokemon,
+  ): boolean {
+    if (
+      pokemon.isPlayer() === knockedOutPokemon.isPlayer()
+      && !knockedOutPokemon.getAbility().hasAttr(UncopiableAbilityAbAttr)
+    ) {
       if (!simulated) {
-        pokemon.summonData.ability = knockedOut.getAbility().id;
+        pokemon.summonData.ability = knockedOutPokemon.getAbility().id;
         globalScene.queueMessage(
           i18next.t("abilityTriggers:copyFaintedAllyAbility", {
-            pokemonNameWithAffix: getPokemonNameWithAffix(knockedOut),
-            abilityName: allAbilities[knockedOut.getAbility().id].name,
+            pokemonNameWithAffix: getPokemonNameWithAffix(knockedOutPokemon),
+            abilityName: allAbilities[knockedOutPokemon.getAbility().id].name,
           }),
         );
       }
--- a/src/data/ab-attrs/post-damage-ab-attr.ts
+++ b/src/data/ab-attrs/post-damage-ab-attr.ts
@@ -11,8 +11,8 @@ export class PostDamageAbAttr extends AbAttr {
     _damage: number,
     _passive: boolean,
     _simulated: boolean,
-    _args: any[],
     _source?: Pokemon,
+    ..._args: any[]
   ): boolean {
     return false;
   }
--- a/src/data/ab-attrs/post-damage-ab-attr.ts
+++ b/src/data/ab-attrs/post-damage-ab-attr.ts
@@ -12,7 +12,7 @@ export class PostDamageAbAttr extends AbAttr {
     _passive: boolean,
     _simulated: boolean,
     _source?: Pokemon,
-    ..._args: any[]
+    ..._args: unknown[]
   ): boolean {
     return false;
   }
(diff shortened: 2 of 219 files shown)


Tempo-anon (dismissed):
Tests pass

src/data/ab-attrs/post-damage-ab-attr.ts
```diff
@@ -11,8 +11,8 @@ export class PostDamageAbAttr extends AbAttr {
     _damage: number,
     _passive: boolean,
     _simulated: boolean,
-    _args: any[],
     _source?: Pokemon,
+    ..._args: any[]
```
MokaStitcher: should this be unknown like the rest of the changes or is it not needed since nothing extends it?
```suggestion
    ..._args: unknown[]
```

src/data/move-conditions.ts
```diff
@@ -108,7 +108,7 @@ export const failIfDampCondition: MoveConditionFunc = (user, _target, move) => {
   globalScene
     .getField(true)
     .map((p) =>
-      applyAbAttrs(FieldPreventExplosionLikeAbAttr, p, cancelled, undefined, getPokemonNameWithAffix(user), move.name),
+      applyAbAttrs(FieldPreventExplosionLikeAbAttr, p, undefined, cancelled, getPokemonNameWithAffix(user), move.name),
```
MokaStitcher, line 111: why undefined and not false?

MokaStitcher (commented):
code review only, 2 small comments

src/data/move-conditions.ts
flx-sta: If I had to guess: Basically a mirror of the original?

src/data/ab-attrs/post-damage-ab-attr.ts
flx-sta: +1

src/data/ab-attrs/post-knock-out-stat-stage-change-ab-attr.ts
```diff
@@ -15,13 +15,7 @@ export class PostKnockOutStatStageChangeAbAttr extends PostKnockOutAbAttr {
...
   }
 
-  override applyPostKnockOut(
-    pokemon: Pokemon,
-    _passive: boolean,
-    simulated: boolean,
-    _knockedOut: Pokemon,
-    _args: any[],
-  ): boolean {
+  override applyPostKnockOut(pokemon: Pokemon, _passive: boolean, simulated: boolean, _knockedOut: Pokemon): boolean {
```
flx-sta: NIT:
```suggestion
  override applyPostKnockOut(pokemon: Pokemon, _passive: boolean, simulated: boolean, _knockedOutPokemon: Pokemon): boolean {
```

The name is bugging me. If I were to see that variable name I'd expected a boolean or so

src/data/ab-attrs/stat-multiplier-ab-attr.ts
```diff
@@ -36,28 +36,26 @@ import { AbAttr } from "./ab-attr";
...
  */
-export class StatMultiplierAbAttr extends AbAttr {
-  public stat: BattleStat;
+export class StatMultiplierAbAttr extends StatStageAbAttr {
   private readonly multiplier: number;
   private readonly condition?: PokemonAttackCondition;
 
   constructor(stat: BattleStat, multiplier: number, condition?: PokemonAttackCondition) {
-    super(false);
+    super(stat);
```
flx-sta, line 44: Q: Was that a bug?

src/data/ab-attrs/terrain-event-type-change-ab-attr.ts
```diff
@@ -72,7 +65,7 @@ export class TerrainEventTypeChangeAbAttr extends PostSummonAbAttr {
   override applyPostSummon(pokemon: Pokemon, passive: boolean, simulated: boolean, _args: any[]): boolean {
```
flx-sta: Does the `any[]` here stay?

src/data/ab-attrs/terrain-event-type-change-ab-attr.ts
```diff
@@ -72,7 +65,7 @@ export class TerrainEventTypeChangeAbAttr extends PostSummonAbAttr {
...
     if (globalScene.arena.getTerrainType() !== TerrainType.NONE) {
       // TODO: `apply()` probably shouldn't be used this way
-      return this.apply(pokemon, passive, simulated, new BooleanHolder(false), []);
+      return this.apply(pokemon, passive, simulated);
```
flx-sta, line 68: NIT: What happened to the boolean-holder?

src/data/ab-attrs/post-damage-ab-attr.ts
innerthunder: I left in `args` fields for all of the different ability attribute types, even if they aren't currently used. I must've just missed this one.

Side note: I'm gonna make a follow-up PR to clean up the `applyXAbAttrs` functions (I'll probably end up merging them into one function using a method similar to #199 ). That should also help with any extra/unused `args`.

flx-sta (commented):
I'm curious about all those `simulated` and `cancelled` swaps. Was that actually causing side-effects or this simply necessary because of this change?

src/data/move-conditions.ts
innerthunder: Yeah, the original used `undefined` to refer to the default setting for `simulated`. It's probably meant to be replaced when move condition checks have simulated support.

src/data/ab-attrs/stat-multiplier-ab-attr.ts
MokaStitcher: I don't think so, a new parent class was added that does the super(false)

src/data/ab-attrs/terrain-event-type-change-ab-attr.ts
innerthunder: Nope, just missed it

src/data/ab-attrs/terrain-event-type-change-ab-attr.ts
innerthunder: the removed params were fillers for when `cancelled` and `args` weren't optional

innerthunder:
> I'm curious about all those `simulated` and `cancelled` swaps. Was that actually causing side-effects or this simply necessary because of this change?

It's just a necessary change because `cancelled` is now implemented as an optional argument in `args`.

src/data/ability.ts
```diff
@@ -682,14 +682,14 @@ export function applyPostAttackAbAttrs(
 export function applyPostKnockOutAbAttrs(
   attrType: Constructor<PostKnockOutAbAttr>,
   pokemon: Pokemon,
-  knockedOut: Pokemon,
+  knockedOutPokemon: Pokemon,
   simulated: boolean = false,
   ...args: any[]
```
flx-sta: any[]?

innerthunder: Now switched to `unknown` for every external apply function (except for `applyFieldStatMultiplierAbAttrs` because there isn't a parent class to allow for `args` to be typecast).

flx-sta (approved):

src/data/ab-attrs/copy-fainted-ally-ability-ab-attr.ts
```diff
@@ -7,14 +7,22 @@ import { PostKnockOutAbAttr } from "./post-knock-out-ab-attr";
...
+    simulated: boolean,
+    knockedOutPokemon: Pokemon,
+  ): boolean {
+    if (
+      pokemon.isPlayer() === knockedOutPokemon.isPlayer()
+      && !knockedOutPokemon.getAbility().hasAttr(UncopiableAbilityAbAttr)
+    ) {
       if (!simulated) {
-        pokemon.summonData.ability = knockedOut.getAbility().id;
+        pokemon.summonData.ability = knockedOutPokemon.getAbility().id;
```
MokaStitcher, line 21: Absolutely unrelated but this made me discover that receiver is a thing. Shouldn't that ability consider passives in some way?
Like if receiver is a mon's passive and not ability (at the moment I assume none have this), it should probably replace the passive in summonData, not the normal ability 🤔 
Anyways, this is probably for the ability checkups

MokaStitcher (approved):

PR #202, opened 2025-01-07
[UI][P3] Temporarily use luxury ball to show epic tier items in shop

MokaStitcher:

the pokeball animation in the shop is no longer using master ball for epic tier items nor open great ball for master tier

it looks messy, caused by removing the rogue ball

- Add luxury ball in the PokeballType enum. It appears below the Master ball to avoid it being considered usable by the player
- Fix up the ball ui handler so that it takes the proper size even with the unused ball in the enum
- In modifier select ui handler, add a temporary function that maps a ModifierTier to a PokeballType, instead of expecting both types indices to match

## Screenshots/Videos

<details><summary>Reward screen for items of each tier</summary>

Before:

https://github.com/user-attachments/assets/89b57c7c-2c55-449a-9451-3642f7533f11

After:

https://github.com/user-attachments/assets/fb1e3458-2498-4b6b-b56c-0ca6a683b6dd

</details>

<details><summary>In battle ball menu didn't change</summary>

Before:

https://github.com/user-attachments/assets/c964db00-a0f7-4711-9bcb-9e3f3215bd6e

After:

https://github.com/user-attachments/assets/abbec3f7-4ebc-4a37-8a49-440f0dfd4c09

</details>

## How to test the changes?

Overrides and reroll items in shop. Can also give yourself pokeballs to make sure catching still works as before
```
  STARTING_MODIFIER_OVERRIDE: [{name: "GOLDEN_POKEBALL", count: 5}, {name: "LOCK_CAPSULE", count: 1}],
  WAIVE_ROLL_FEE_OVERRIDE: true,
  POKEBALL_OVERRIDE: {
    active: true,
    pokeballs: {
      [PokeballType.POKEBALL]: 5,
      [PokeballType.GREAT_BALL]: 5,
      [PokeballType.ULTRA_BALL]: 5,
      [PokeballType.MASTER_BALL]: 5,
    },
  }
```

## Checklist

- [X] Are all unit tests still passing? (`npm run test`)
  - [X] Have I made sure that any UI change works for both UI themes (default and legacy)?

~~Are there any localization additions or changes? If so:~~


src/data/pokeball.ts
```diff
@@ -31,6 +33,9 @@ export function getPokeballName(type: PokeballType): string {
     case PokeballType.ULTRA_BALL:
       ret = i18next.t("pokeball:ultraBall");
       break;
+    case PokeballType.LUXURY_BALL:
+      ret = i18next.t("pokeball:luxuryBall");
+      break;
```
flx-sta: NOTE: We should really get this function `simplified` in the near future.
we can simply set `ret` to the i18n key and at the end call `i18next.t(ret)`.

it's a minimal change and won't do much good or bad But it just bothers me :D

src/data/pokeball.ts
```diff
@@ -59,6 +66,8 @@ export function getPokeballTintColor(type: PokeballType): number {
       return 0x94b4de;
     case PokeballType.ULTRA_BALL:
       return 0xe6cd31;
+    case PokeballType.LUXURY_BALL:
+      return 0xe74c18; // TODO find proper color
```
flx-sta: Is this TODO still relevant/open?

src/enums/pokeball.ts
```diff
@@ -3,5 +3,7 @@ export enum PokeballType {
   GREAT_BALL,
   ULTRA_BALL,
   MASTER_BALL,
-  // LUXURY_BALL,
+  // Temporarily used to represent epic tier items in the shop, not available to the player
+  // For that, it needs to be placed after the master ball
```
flx-sta: NIT: Should the comment display on `inspect`? 
Then it should use 
```ts
/** */
``` 
Or the multi-line version of it

src/data/pokeball.ts
MokaStitcher: I just added it so yes
I put the color that was used for the rogue ball before I think, not sure anymore. Not entirely sure what this color actually achieves too

src/data/pokeball.ts
MokaStitcher: True. Or we could just return the value directly, like the other functions in the file. will update

flx-sta (dismissed):
lgtm

https://github.com/user-attachments/assets/14cf40ca-dc2f-4899-8b43-85c2166967ec

src/data/pokeball.ts
```diff
@@ -59,6 +59,8 @@ export function getPokeballTintColor(type: PokeballType): number {
...
```
PigeonBar: For reference, before #2, the tint color for Rogue Ball was `0xd52929`, and the tint color for Luxury Ball was `0xffde6a`

image

PigeonBar (dismissed):

src/data/pokeball.ts
MokaStitcher: huh, you're right. Not sure why I didn't check that. Thank you

PR #205, opened 2025-01-08
[Bug][Move] Refactor moves that call a random move

DayKev:
> [!NOTE]
> Text copied from PR3380 

This includes the moves: Metronome, Assist, Sleep Talk, Copycat, Mirror Move

## What are the changes?
Properly implements Assist. 
Refactors `RandomMovesetMoveAttr` to extend `RandomMoveAttr`. 
Creates arrays of invalid moves for each move: Assist, Sleep Talk, Metronome, since they each have different sets of uncallable moves.

## Why am I doing these changes?
Assist did not call ally moves, had useless code, and was given no distinction in logic from Sleep Talk. 
Metronome calls random moves in a similar manner to the two other moves, so it makes sense to have Assist/Sleep Talk reuse the code for calling moves.
None of Assist/Sleep Talk/Metronome had properly implemented lists of uncallable moves.

## What did change?
- Assist now uses the entire team's party
- Assist/Sleep Talk's attribute now extends `RandomMoveAttr` and uses its callMoves function (which has been modified slightly to accommodate this) to call the move
- Sleep Talk targeting is changed to `NEAR_ENEMY`
- Copycat's last move used is now unaffected by moves that fail, see Copycat's [bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Copycat_(move)#:~:text=The%20last%20move%20used%20is%20now%20unaffected%20if%20a%20Pok%C3%A9mon%20tries%20to%20use%20a%20move%20but%20is%20prevented%20by%20a%20status%20condition%20or%20effect.%20In%20other%20words%2C%20the%20last%2Dused%20move%20remains%20what%20it%20was%20before%20the%20Pok%C3%A9mon%27s%20turn.)
- Mirror Move can now properly copy status moves
- Mirror Move now targets a pokemon and copies that pokemon's last used move
- Assist/Sleep Talk/Metronome/Copycat now have individual arrays listing uncallable moves. These arrays were created referencing Bulbapedia's list of uncallable moves.
- Uncallable moves for [Assist](https://bulbapedia.bulbagarden.net/wiki/Assist_(move)#:~:text=in%20a%20battle.-,Uncallable%20moves,-This%20move%20effect)
- Uncallable moves for [Sleep Talk](https://bulbapedia.bulbagarden.net/wiki/Sleep_Talk_(move)#:~:text=Z%2DPower%20effect.-,Unselectable%20moves,-This%20section%20is)
- Uncallable moves for [Metronome](https://bulbapedia.bulbagarden.net/wiki/Metronome_(move)#:~:text=Z%2DPower%20effect.-,Unselectable%20moves,-This%20move%20effect)
- Uncallable moves for [Copycat](https://bulbapedia.bulbagarden.net/wiki/Copycat_(move)#:~:text=or%20failing%20entirely.-,Uncallable%20moves,-This%20move%20effect)

Note: Sleep Talk specifies that it targets the [opposite opponent](https://bulbapedia.bulbagarden.net/wiki/Sleep_Talk_(move)#:~:text=In%20Double%20and%20Triple%20Battles%2C%20because%20of%20the%20inability%20to%20select%20a%20target%2C%20the%20chosen%20move%20will%20always%20hit%20the%20opponent%20directly%20across%20from%20the%20user%2C%20if%20the%20move%20called%20normally%20hits%20only%20one%20Pok%C3%A9mon.). Since there is no `MoveTarget` specifically for opposite opponents, I chose to make it `NEAR_ENEMY` for now.

### Screenshots/Videos
<details>
  <summary>Metronome</summary>

https://github.com/user-attachments/assets/c8861801-acd0-4ae0-b817-181cffa549d7

</details>
<details>
  <summary>Assist</summary>

https://github.com/user-attachments/assets/9146fc45-9294-4d59-a726-ffcdf31b5777

</details>
<details>
  <summary>Sleep Talk</summary>

https://github.com/user-attachments/assets/afd243b0-b2e9-4d84-9ba9-13a924c4b439

</details>

## How to test the changes?
Use src/overrides.ts to give yourself one of the three moves. If using Metronome, observe it will never call certain moves. If using Assist, observe it calling your party pokemon's moves. If using Sleep Talk, observe it only calling its own moveset.
Note that each move will no longer call certain moves: attempt Assist/Sleep Talk with 2 turn moves to see them fail. 
List of moves that won't be callable are given above by links to bulbapedia.

## Checklist
- [x] Have I considered writing automated tests for the issue?
- [x] Have I tested the changes (manually)?
- [x] Are the changes visual?
  - [x] Have I provided screenshots/videos of the changes?


src/data/all-moves.ts
```diff
@@ -245,6 +245,282 @@ import { ArenaTagRelativeSide } from "#enums/arena-tag-relative-side";
 
 export const allMoves: Move[] = [];
 
+const invalidMetronomeMoves: Moves[] = [
```
innerthunder: Maybe these arrays should be put in `move-utils.ts`?

src/data/move-attrs/call-move-attr.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Attribute used to call a move.
+ * Used by other move attributes: {@linkcode RandomMoveAttr}, {@linkcode RandomMovesetMoveAttr}, {@linkcode CopyMoveAttr}
+ * @see {@linkcode apply} for move call
+ * @extends OverrideMoveEffectAttr
+ */
+export abstract class CallMoveAttr extends OverrideMoveEffectAttr {
+  protected invalidMoves: Moves[];
+  protected hasTarget: boolean;
+  override apply(user: Pokemon, target: Pokemon, move: Move, _overridden: BooleanHolder, _virtual: boolean): boolean {
```
innerthunder: This should set `overridden.value` to `true` at the end to skip the rest of the `MoveEffectPhase` this is applied in.

src/data/move-attrs/copy-move-attr.ts
```diff
@@ -1,41 +1,40 @@
...
 
 /**
- * Attribute to copy the last move used in battle and invoke it against random target(s).
- * Used for {@link https://bulbapedia.bulbagarden.net/wiki/Mirror_Move_(move) | Mirror Move}
- * and {@link https://bulbapedia.bulbagarden.net/wiki/Copycat_(move) | Copycat}.
- * @extends OverrideMoveEffectAttr
+ * Attribute used to copy a previously-used move.
+ * Used for {@linkcode Moves.COPYCAT} and {@linkcode Moves.MIRROR_MOVE}
+ * @see {@linkcode apply} for move selection and move call
+ * @extends CallMoveAttr to call a selected move
```
innerthunder:
```suggestion
 * @extends CallMoveAttr
```
Current comment doesn't gel well with VSCode's doc formatter...

src/data/move-attrs/call-move-attr.ts
innerthunder: You might also be able to omit `virtual` from this and subclasses' signatures

src/data/move-attrs/random-moveset-move-attr.ts
```diff
@@ -1,67 +1,63 @@
...
-    const moves = moveset.filter((m) => !m.getMove().hasFlag(MoveFlags.IGNORE_VIRTUAL));
-    if (moves.length) {
-      const move = moves[user.randSeedInt(moves.length)];
-      const moveIndex = moveset.findIndex((m) => m.moveId === move?.moveId);
-      const moveTargets = getMoveTargets(user, move.moveId);
-      if (!moveTargets.targets.length) {
-        return false;
+  override apply(user: Pokemon, target: Pokemon, _move: Move, overriden: BooleanHolder, virtual: boolean): boolean {
+    return super.apply(user, target, allMoves[this.moveId], overriden, virtual);
+  }
```
innerthunder:
```suggestion
  override apply(user: Pokemon, target: Pokemon, _move: Move, overridden: BooleanHolder, virtual: boolean): boolean {
    return super.apply(user, target, allMoves[this.moveId], overridden, virtual);
  }
```

src/data/move-attrs/recoil-attr.ts
```diff
@@ -42,7 +42,10 @@ export class RecoilAttr extends MoveEffectAttr {
     }
 
     // Chloroblast and Struggle should not deal recoil damage if the move was not successful
-    if (this.useHp && [MoveResult.FAIL, MoveResult.MISS].includes(user.getLastXMoves(1)[0]?.result)) {
+    if (
+      this.useHp
+      && [MoveResult.FAIL, MoveResult.MISS].includes(user.getLastXMoves(1)[0]?.result ?? MoveResult.FAIL)
+    ) {
```
innerthunder: I don't think this condition is necessary anymore with the MEPhase rewrite

innerthunder (dismissed):

src/data/all-moves.ts
DayKev: Would they need to be exported for use elsewhere?

src/data/move-attrs/call-move-attr.ts
DayKev: Setting `overridden.value = true` breaks Sketch.

src/data/move-attrs/call-move-attr.ts
```diff
@@ -0,0 +1,47 @@
...
+              ? target.getBattlerIndex()
+              : moveTargets.targets[user.randSeedInt(moveTargets.targets.length)],
+          ]; // account for Mirror Move having a target already
+
+    user.getMoveQueue().push({ move: move.id, targets: targets, virtual: true, ignorePP: true });
+    globalScene.unshiftPhase(new MovePhase(user, targets, new PokemonMove(move.id, 0, 0, true), true, true));
+
+    // Promise.resolve(initMoveAnim(user.scene, move.id).then(() => {
+    //   loadMoveAnimAssets(user.scene, [ move.id ], true);
+    // }));
```
DayKev: Unlike in PKR this isn't an async function anymore, what to do about this?

innerthunder: Use `LoadMoveAnimPhase` (`load-move-anim-phase.ts`):
```suggestion
    globalScene.unshiftPhase(new LoadMoveAnimPhase(move.id));
```

src/data/all-moves.ts
innerthunder: I wouldn't think so, but this really bloats the `all-moves` file...

src/data/move-attrs/call-move-attr.ts
innerthunder: ...how? Sketch isn't even a subclass of this??

DayKev: image

innerthunder: Ah, seems like `overridden` stops the move from being pushed to the user's move history?

src/data/all-moves.ts
DayKev: Maybe we should remove all the Z-Moves and Max moves from the file (I just found out a bunch of Z-Moves weren't marked as `.unimplemented()` and so Metronome could call them :pikastare:).

Tempo-anon: I think logically to me these lists should be in the attr of the move itself, but I worry that might create a circular dependency. I think if that's a concern then they can just live here

src/data/move-attrs/copy-move-attr.ts
```diff
@@ -1,41 +1,41 @@
...
+import type { BooleanHolder } from "#app/utils";
+import type { Moves } from "#enums/moves";
 
 /**
- * Attribute to copy the last move used in battle and invoke it against random target(s).
- * Used for {@link https://bulbapedia.bulbagarden.net/wiki/Mirror_Move_(move) | Mirror Move}
- * and {@link https://bulbapedia.bulbagarden.net/wiki/Copycat_(move) | Copycat}.
- * @extends OverrideMoveEffectAttr
+ * Attribute used to copy a previously-used move.
+ * Used for {@linkcode Moves.COPYCAT} and {@linkcode Moves.MIRROR_MOVE}
```
Tempo-anon: Should add some doc on how mirror move is different than copycat

Tempo-anon (dismissed):

src/data/all-moves.ts
DayKev: Moving some of them would be a bit awkward as well. For Copycat and Metronome it's simple, but for Assist and Sleep Talk it would require adding some sort of `if (move === Moves.SLEEP_TALK) / if (move === Moves.ASSIST)` to the move attr and I'm pretty sure we don't want to do that.

Tempo-anon: Yeah, having a bespoke move check like that is not ideal. Could always just make AssistAttr and SleepTalkAttr if that would be clearer

DayKev:
Waiting for #221 to hopefully make fixing the circular import issue easier (or at least to prevent merge conflicts).

DayKev:
Turns out the issue is only in the tests, the game plays just fine otherwise. 🫠

src/data/all-moves.ts
```diff
@@ -2750,39 +3245,34 @@ export function initMoves() {
...
-      .ignoresVirtual(),
+      .attr(ClearTerrainAttr)
+      .makesContact(false),
     new AttackMove(Moves.CLANGOROUS_SOULBLAZE, Type.DRAGON, MoveCategory.SPECIAL, 185, -1, 1, 100, 0, 7)
+      .unimplemented()
       .attr(StatStageChangeAttr, [Stat.ATK, Stat.DEF, Stat.SPATK, Stat.SPDEF, Stat.SPD], 1, true, {
         firstTargetOnly: true,
       })
       .soundMove()
       .unimplemented()
```
PigeonBar:
```suggestion
```
(Duplicate)

src/data/all-moves.ts
```diff
@@ -247,6 +247,282 @@ import { StatStageChangeAllOppsAttr } from "./move-attrs/stat-stage-change-all-o
...
+  Moves.BESTOW,
+  Moves.CELEBRATE,
+  Moves.CHATTER,
+  Moves.CIRCLE_THROW,
+  Moves.COPYCAT,
+  Moves.COUNTER,
+  Moves.COVET,
+  Moves.DESTINY_BOND,
+  Moves.DETECT,
+  Moves.DRAGON_TAIL,
```
PigeonBar:
```suggestion
  Moves.DRAGON_TAIL,
  Moves.DYNAMAX_CANNON,
```

src/data/move-attrs/call-move-attr.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @see {@linkcode apply} for move call
+ * @extends OverrideMoveEffectAttr
+ */
+export abstract class CallMoveAttr extends OverrideMoveEffectAttr {
+  protected invalidMoves: Moves[];
+  protected hasTarget: boolean;
+  public readonly callsOtherMoves: boolean = true;
+
+  override apply(user: Pokemon, target: Pokemon, move: Move, overridden: BooleanHolder): boolean {
+    const replaceMoveTarget = move.moveTarget === MoveTarget.NEAR_OTHER ? MoveTarget.NEAR_ENEMY : undefined;
```
PigeonBar:
```suggestion
    const replaceMoveTarget = [MoveTarget.NEAR_OTHER, MoveTarget.DRAGON_DARTS].includes(move.moveTarget) ? MoveTarget.NEAR_ENEMY : undefined;
```

src/data/all-moves.ts
```diff
@@ -247,6 +247,282 @@ import { StatStageChangeAllOppsAttr } from "./move-attrs/stat-stage-change-all-o
...
+  Moves.ME_FIRST,
+  Moves.METEOR_ASSAULT,
+  Moves.METRONOME,
+  Moves.MIMIC,
+  Moves.MIND_BLOWN,
+  Moves.MIRROR_COAT,
+  Moves.MIRROR_MOVE,
+  Moves.MOONGEIST_BEAM,
+  Moves.NATURE_POWER,
+  Moves.NATURES_MADNESS,
```
PigeonBar:
```suggestion
  Moves.NATURES_MADNESS,
  Moves.NONE,
```

src/data/all-moves.ts
```diff
@@ -247,6 +247,282 @@ import { StatStageChangeAllOppsAttr } from "./move-attrs/stat-stage-change-all-o
...
+  Moves.HELPING_HAND,
+  Moves.HOLD_HANDS,
+  Moves.KINGS_SHIELD,
+  Moves.MAT_BLOCK,
+  Moves.ME_FIRST,
+  Moves.METRONOME,
+  Moves.MIMIC,
+  Moves.MIRROR_COAT,
+  Moves.MIRROR_MOVE,
+  Moves.NATURE_POWER,
```
PigeonBar:
```suggestion
  Moves.NATURE_POWER,
  Moves.NONE,
```

src/data/all-moves.ts
```diff
@@ -247,6 +247,282 @@ import { StatStageChangeAllOppsAttr } from "./move-attrs/stat-stage-change-all-o
...
+  Moves.DYNAMAX_CANNON,
+  Moves.FREEZE_SHOCK,
+  Moves.FLY,
+  Moves.FOCUS_PUNCH,
+  Moves.GEOMANCY,
+  Moves.ICE_BURN,
+  Moves.ME_FIRST,
+  Moves.METRONOME,
+  Moves.MIRROR_MOVE,
+  Moves.MIMIC,
```
PigeonBar:
```suggestion
  Moves.MIMIC,
  Moves.NONE,
```

src/data/all-moves.ts
```diff
@@ -247,6 +247,282 @@ import { StatStageChangeAllOppsAttr } from "./move-attrs/stat-stage-change-all-o
...
+  Moves.FOLLOW_ME,
+  Moves.HELPING_HAND,
+  Moves.HOLD_HANDS,
+  Moves.KINGS_SHIELD,
+  Moves.MAT_BLOCK,
+  Moves.ME_FIRST,
+  Moves.METRONOME,
+  Moves.MIMIC,
+  Moves.MIRROR_COAT,
+  Moves.MIRROR_MOVE,
```
PigeonBar:
```suggestion
  Moves.MIRROR_MOVE,
  Moves.NONE,
```

src/data/move-attrs/random-move-attr.ts
```diff
@@ -1,46 +1,48 @@
...
+   * Invalid moves are indicated by what is passed in to invalidMoves: {@linkcode invalidMetronomeMoves}
+   * @param user Pokemon that used the move and will call a random move
+   * @param target Pokemon that will be targeted by the random move (if single target)
+   * @param move Move being used
+   * @param args Unused
+   */
+  override apply(user: Pokemon, target: Pokemon, _move: Move, overridden: BooleanHolder): boolean {
+    const moveIds = getEnumValues(Moves).map((m) =>
+      !this.invalidMoves.includes(m) && !allMoves[m].name.endsWith(" (N)") ? m : Moves.NONE,
     );
```
PigeonBar:
```suggestion
    const moveIds = getEnumValues(Moves).filter((m) =>
      !this.invalidMoves.includes(m) && !allMoves[m].name.endsWith(" (N)")
    );
```

src/data/move-attrs/random-move-attr.ts
```diff
@@ -1,46 +1,48 @@
...
-          ? [target.getBattlerIndex()]
-          : [moveTargets.targets[user.randSeedInt(moveTargets.targets.length)]];
-    user.getMoveQueue().push({ move: moveId, targets: targets, ignorePP: true });
-    globalScene.unshiftPhase(new LoadMoveAnimPhase(moveId));
-    globalScene.unshiftPhase(new MovePhase(user, targets, new PokemonMove(moveId, 0, 0, true), true));
-    return true;
+    let moveId: Moves = Moves.NONE;
+    do {
+      moveId = this.getMoveOverride() ?? moveIds[user.randSeedInt(moveIds.length)];
+    } while (moveId === Moves.NONE);
```
PigeonBar:
```suggestion
    const moveId = this.getMoveOverride() ?? moveIds[user.randSeedInt(moveIds.length)];
```

src/data/move-attrs/random-moveset-move-attr.ts
```diff
@@ -1,67 +1,64 @@
...
+        return false;
       }
-      const targets = selectTargets;
-      user.getMoveQueue().push({ move: move.moveId, targets: targets, ignorePP: true });
-      globalScene.unshiftPhase(new MovePhase(user, targets, moveset[moveIndex], true));
-      return true;
-    }
 
-    return false;
+      this.moveId = moves[user.randSeedInt(moves.length)].moveId;
```
PigeonBar, line 61:
```suggestion
      moves.sort((a, b) => a.moveId - b.moveId);
      this.moveId = moves[user.randSeedInt(moves.length)].moveId;
```
Sorting the moves before picking a random move prevents the player from RNG manipulating the selected move by swapping moveset orders in the Pokemon summary UI.

src/data/all-moves.ts
```diff
@@ -570,11 +839,10 @@ export function initMoves() {
       { failOnOverlap: true },
     ),
     new AttackMove(Moves.BIDE, Type.NORMAL, MoveCategory.PHYSICAL, -1, -1, 10, -1, 1, 1)
-      .ignoresVirtual()
       .target(MoveTarget.USER)
       .unimplemented(),
-    new SelfStatusMove(Moves.METRONOME, Type.NORMAL, -1, 10, -1, 0, 1).attr(RandomMoveAttr).ignoresVirtual(),
-    new StatusMove(Moves.MIRROR_MOVE, Type.FLYING, -1, 20, -1, 0, 1).attr(CopyMoveAttr).ignoresVirtual(),
+    new SelfStatusMove(Moves.METRONOME, Type.NORMAL, -1, 10, -1, 0, 1).attr(RandomMoveAttr, invalidMetronomeMoves),
+    new StatusMove(Moves.MIRROR_MOVE, Type.FLYING, -1, 20, -1, 0, 1).attr(CopyMoveAttr, true),
```
PigeonBar:
```suggestion
    new StatusMove(Moves.MIRROR_MOVE, Type.FLYING, -1, 20, -1, 0, 1)
      .attr(CopyMoveAttr, true)
      .edgeCase(), // Need to find out which moves are not copyable by Mirror Move
```

src/data/move-attrs/copy-move-attr.ts
```diff
@@ -1,41 +1,43 @@
...
 
-    return true;
+  override apply(user: Pokemon, target: Pokemon, _move: Move, overridden: BooleanHolder): boolean {
+    this.hasTarget = this.mirrorMove;
+    const lastMove = this.mirrorMove ? target.getLastXMoves()[0].move : globalScene.currentBattle.lastMove;
+    return super.apply(user, target, allMoves[lastMove], overridden);
   }
 
   override getCondition(): MoveConditionFunc {
-    return lastMoveCopiableCondition;
```
PigeonBar, line 37: Now that this condition is unused, it can also be deleted in `move-conditions.ts`.

test/moves/assist.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    game.move.changeMoveset(feebas, [Moves.ASSIST, Moves.SKETCH, Moves.PROTECT, Moves.DRAGON_TAIL]);
+    game.move.changeMoveset(shuckle, [Moves.ASSIST, Moves.SKETCH, Moves.PROTECT, Moves.DRAGON_TAIL]);
+
+    game.move.select(Moves.ASSIST, 0);
+    game.move.select(Moves.SKETCH, 1);
+    await game.setTurnOrder([BattlerIndex.ENEMY, BattlerIndex.ENEMY_2, BattlerIndex.PLAYER_2, BattlerIndex.PLAYER]);
+    // Player_2 uses Sketch, copies Swords Dance, Player_1 uses Assist, uses Player_2's Sketched Swords Dance
+    await game.toNextTurn();
+
+    expect(game.scene.getPlayerPokemon()!.getStatStage(Stat.ATK)).toBe(2); // Stat raised from Assist -> Swords Dance
```
PigeonBar:
```suggestion
    expect(feebas.getStatStage(Stat.ATK)).toBe(2); // Stat raised from Assist -> Swords Dance
```

test/moves/assist.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  it("should fail if there are no allies", async () => {
+    await game.classicMode.startBattle([Species.FEEBAS]);
+
+    const feebas = game.scene.getPlayerPokemon()!;
+    game.move.changeMoveset(feebas, [Moves.ASSIST, Moves.SKETCH, Moves.PROTECT, Moves.DRAGON_TAIL]);
+
+    game.move.select(Moves.ASSIST, 0);
+    await game.toNextTurn();
+    expect(game.scene.getPlayerPokemon()!.getLastXMoves()[0].result).toBe(MoveResult.FAIL);
```
PigeonBar:
```suggestion
    expect(feebas.getLastXMoves()[0].result).toBe(MoveResult.FAIL);
```

test/moves/assist.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    game.move.select(Moves.PROTECT, 1);
+    await game.setTurnOrder([BattlerIndex.ENEMY, BattlerIndex.ENEMY_2, BattlerIndex.PLAYER, BattlerIndex.PLAYER_2]);
+    // Player uses Sketch to copy Swords Dance, Player_2 stalls a turn. Player will attempt Assist and should have no usable moves
+    await game.toNextTurn();
+    game.move.select(Moves.ASSIST, 0);
+    await game.phaseInterceptor.to(CommandPhase);
+    game.move.select(Moves.PROTECT, 1);
+    await game.toNextTurn();
+
+    expect(game.scene.getPlayerPokemon()!.getLastXMoves()[0].result).toBe(MoveResult.FAIL);
```
PigeonBar:
```suggestion
    expect(feebas.getLastXMoves()[0].result).toBe(MoveResult.FAIL);
```

test/moves/assist.test.ts
```diff
@@ -0,0 +1,105 @@
...
+  });
+
+  it("should apply secondary effects of a move", async () => {
+    game.override.moveset([Moves.ASSIST, Moves.WOOD_HAMMER, Moves.WOOD_HAMMER, Moves.WOOD_HAMMER]);
+    await game.classicMode.startBattle([Species.FEEBAS, Species.SHUCKLE]);
+
+    const [feebas, shuckle] = game.scene.getPlayerField();
+    game.move.changeMoveset(feebas, [Moves.ASSIST, Moves.SKETCH, Moves.PROTECT, Moves.DRAGON_TAIL]);
+    game.move.changeMoveset(shuckle, [Moves.ASSIST, Moves.SKETCH, Moves.PROTECT, Moves.DRAGON_TAIL]);
```
PigeonBar:
```suggestion
```

test/moves/assist.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    game.move.select(Moves.ASSIST, 0);
+    await game.phaseInterceptor.to(CommandPhase);
+    game.move.select(Moves.ASSIST, 1);
+    await game.toNextTurn();
+
+    expect(game.scene.getPlayerPokemon()!.isFullHp()).toBeFalsy(); // should receive recoil damage from Wood Hammer
```
PigeonBar:
```suggestion
    expect(feebas.isFullHp()).toBeFalsy(); // should receive recoil damage from Wood Hammer
```

src/data/all-moves.ts
DayKev: Moved the arrays into the move attr files and then imported them into `all-moves.ts`. Feels a bit awkward but it works I guess.

PigeonBar:
The Bulbapedia article for [Sleep Talk](https://bulbapedia.bulbagarden.net/wiki/Sleep_Talk_(move)#Effect) was recently edited to say that it targets a random opponent, not the opponent directly across from the user.

src/data/all-moves.ts
```diff
@@ -878,10 +873,9 @@ export function initMoves() {
       .condition((user, target, _move) => user.isOppositeGender(target)),
     new SelfStatusMove(Moves.SLEEP_TALK, Type.NORMAL, -1, 10, -1, 0, 2)
       .attr(BypassSleepAttr)
-      .attr(RandomMovesetMoveAttr)
+      .attr(RandomMovesetMoveAttr, invalidSleepTalkMoves)
       .condition(userSleptOrComatoseCondition)
-      .target(MoveTarget.ALL_ENEMIES)
-      .ignoresVirtual(),
+      .target(MoveTarget.ALL_ENEMIES),
```
PigeonBar:
```suggestion
```
(Sleep Talk is technically self-targeting.)

src/phases/move-phase.ts
```diff
@@ -347,6 +342,14 @@ export class MovePhase extends BattlePhase {
 
     const success = passesConditions && !failedDueToWeather && !failedDueToTerrain;
 
+    // Update the battle's "last move" pointer, unless we're currently mimicking a move.
+    if (!allMoves[this.move.moveId].hasAttr(CopyMoveAttr)) {
+      // The last move used is unaffected by moves that fail
+      if (success) {
+        globalScene.currentBattle.lastMove = this.move.moveId;
```
PigeonBar: This line also needs to be copied to the `chargeMove()` function, so that Copycat can copy charging moves during their charging turn.

src/data/move-attrs/copy-move-attr.ts
```diff
@@ -1,41 +1,91 @@
...
+ *
+ * Used for {@linkcode Moves.COPYCAT} and {@linkcode Moves.MIRROR_MOVE}
+ * @see {@linkcode apply} for move selection and move call
+ * @extends CallMoveAttr
  */
-export class CopyMoveAttr extends OverrideMoveEffectAttr {
-  override apply(user: Pokemon, target: Pokemon, _move: Move): boolean {
-    const lastMove = globalScene.currentBattle.lastMove;
+export class CopyMoveAttr extends CallMoveAttr {
+  private mirrorMove: boolean;
```
innerthunder:
```suggestion
  private readonly mirrorMove: boolean;
```

src/data/move-attrs/nature-power-attr.ts
```diff
@@ -15,6 +15,8 @@ import { OverrideMoveEffectAttr } from "#app/data/move-attrs/override-move-effec
  * @extends OverrideMoveEffectAttr
  */
 export class NaturePowerAttr extends OverrideMoveEffectAttr {
```
innerthunder: Should this extend `CallMoveAttr`?

src/data/move-attrs/call-move-attr.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @extends OverrideMoveEffectAttr
+ */
+export abstract class CallMoveAttr extends OverrideMoveEffectAttr {
+  protected invalidMoves: Moves[];
+  protected hasTarget: boolean;
+  public readonly callsOtherMoves: boolean = true;
+
+  override apply(user: Pokemon, target: Pokemon, move: Move, overridden: BooleanHolder): boolean {
+    const replaceMoveTarget = [MoveTarget.NEAR_OTHER, MoveTarget.DRAGON_DARTS].includes(move.moveTarget) ? MoveTarget.NEAR_ENEMY : undefined;
+    const moveTargets = getMoveTargets(user, move.id, replaceMoveTarget);
```
innerthunder: I feel like this can be done without pushing `replaceMoveTarget` through `getMoveTargets` as a param, e.g.

```ts
const replaceMoveTarget = [MoveTarget.NEAR_OTHER, MoveTarget.DRAGON_DARTS].includes(move.moveTarget);
const moveTargets = getMoveTargets(user, move.id);
const oppTargets = user.getOpponents().map((opp) => opp.getBattlerIndex());

if (replaceMoveTarget) {
  moveTargets.targets = moveTargets.targets.filter((t) => oppTargets.includes(t));
}
// ...
```

src/data/move-attrs/copy-move-attr.ts
Tempo-anon: Do you think it's worthy to have a MirrorMoveAttr instead of a variable here? Want to think what the team thinks

src/enums/move-flags.ts
Tempo-anon, line 1: Does the all_moves.json need to be updated because the flags are now all off by 1?

PigeonBar: No, there is no sequential correspondence between pkty's flags and the all_moves.json flags, so they just get matched manually in all_moves.test.ts:
https://github.com/Despair-Games/poketernity/blob/95e25a4e5b08bc607aa6c0623bcf1c4b9d3a8179/test/data/all_moves.test.ts#L30-L32

src/data/move-attrs/copy-move-attr.ts
DayKev: The amount of overlap between the two moves within `CopyMoveAttr` was small enough that I split it into two `Attr`s.

src/data/move-attrs/nature-power-attr.ts
DayKev: Seems like it works, changed.

src/data/move-attrs/call-move-attr.ts
DayKev: You mean this?
```ts
const replaceMoveTarget = [MoveTarget.NEAR_OTHER, MoveTarget.DRAGON_DARTS].includes(move.moveTarget) ? true : false;
const moveTargets = getMoveTargets(user, move.id);
const oppTargets = user.getOpponents().map((opp) => opp.getBattlerIndex());

if (replaceMoveTarget) {
  moveTargets.targets = moveTargets.targets.filter((t) => oppTargets.includes(t));
}
```
You left `replaceMoveTarget` as a parameter to `getMoveTargets()`.

src/data/ab-attrs/pokemon-type-change-ab-attr.ts
```diff
@@ -25,13 +23,7 @@ export class PokemonTypeChangeAbAttr extends PreAttackAbAttr {
...
        * @see {@link https://bulbapedia.bulbagarden.net/wiki/Category:Moves_that_call_other_moves}
        */
-      && !move.findAttr(
-        (attr) =>
-          attr instanceof RandomMovesetMoveAttr
-          || attr instanceof RandomMoveAttr
-          || attr instanceof NaturePowerAttr
-          || attr instanceof CopyMoveAttr,
-      )
+      && !move.findAttr((attr: CallMoveAttr | NaturePowerAttr) => attr.callsOtherMoves)
```
innerthunder:
```suggestion
      && !move.findAttr((attr: CallMoveAttr) => attr.callsOtherMoves)
```

src/data/move-attrs/call-move-attr.ts
innerthunder: updated my original comment

DayKev: It seems like it would be simpler to use the already existing function parameter though.

src/data/move-attrs/random-moveset-move-attr.ts
```diff
@@ -1,67 +1,154 @@
...
+    return super.apply(user, target, allMoves[this.moveId], overridden);
+  }
+
+  override getCondition(): MoveConditionFunc {
+    return (user, _target, _move) => {
+      // includeParty will be true for Assist, false for Sleep Talk
+      let allies: Pokemon[];
+      if (this.includeParty) {
+        allies = user.isPlayer()
+          ? globalScene.getPlayerParty().filter((p) => p !== user)
```
MokaStitcher:
```suggestion
          ? globalScene.getPlayerParty().filter((p) => p !== user && p.isAllowedInChallenge())
```
I think we need this otherwise it will use moves from invalid Pokemon in challenges and we probably don't want that.
(we can't call `globalScene.getPokemonAllowedInBattle()` instead of `getPlayerParty()` because it would exclude fainted mons. But maybe a `getPlayerValidParty()` function would be useful to have)

<sub>We should probably make a pass over all calls to `getPlayerParty` calls in the code at some point to make sure there are no similar issues

src/data/all-moves.ts
```diff
@@ -2962,45 +3189,57 @@ export function initMoves() {
...
-      .makesContact(false)
       .unimplemented()
-      .edgeCase() // I assume it's because the user needs spirit shackle and decidueye
-      .ignoresVirtual(),
+      .makesContact(false)
+      .edgeCase(), // I assume it's because the user needs spirit shackle and decidueye
     new AttackMove(MoveId.MALICIOUS_MOONSAULT, ElementalType.DARK, MoveCategory.PHYSICAL, 180, -1, 1, -1, 0, 7)
+      .unimplemented()
       .attr(AlwaysHitMinimizeAttr)
       .unimplemented()
```
MokaStitcher:
```suggestion
```
duplicate unimplemented

src/data/all-moves.ts
```diff
@@ -2962,45 +3189,57 @@ export function initMoves() {
...
-      .edgeCase() // I assume it's because it needs thunderbolt and Alola Raichu
-      .ignoresVirtual(),
+      .edgeCase(), // I assume it's because it needs thunderbolt and Alola Raichu
     new AttackMove(MoveId.PULVERIZING_PANCAKE, ElementalType.NORMAL, MoveCategory.PHYSICAL, 210, -1, 1, -1, 0, 7)
-      .edgeCase() // I assume it's because it needs giga impact and snorlax
       .unimplemented()
-      .ignoresVirtual(),
+      .edgeCase(), // I assume it's because it needs giga impact and snorlax
     new SelfStatusMove(MoveId.EXTREME_EVOBOOST, ElementalType.NORMAL, -1, 1, -1, 0, 7)
       .unimplemented()
```
MokaStitcher:
```suggestion
```

src/data/all-moves.ts
```diff
@@ -2962,45 +3189,57 @@ export function initMoves() {
...
       .attr(StatStageChangeAttr, [Stat.ATK, Stat.DEF, Stat.SPATK, Stat.SPDEF, Stat.SPD], 2, true)
-      .ignoresVirtual(),
+      .unimplemented(),
     new AttackMove(MoveId.GENESIS_SUPERNOVA, ElementalType.PSYCHIC, MoveCategory.SPECIAL, 185, -1, 1, 100, 0, 7)
       .unimplemented()
```
MokaStitcher:
```suggestion
```

src/data/all-moves.ts
```diff
@@ -3187,8 +3419,8 @@ export function initMoves() {
     /* Unused */
     new SelfStatusMove(MoveId.MAX_GUARD, ElementalType.NORMAL, -1, 10, -1, 4, 8)
       .attr(ProtectAttr)
-      .condition(failIfLastCondition)
-      .ignoresVirtual(),
+      .unimplemented()
```
MokaStitcher: hum, why is it unimplemented now?

src/data/move-attrs/nature-power-attr.ts
```diff
@@ -142,16 +109,10 @@ export class NaturePowerAttr extends OverrideMoveEffectAttr {
         moveId = MoveId.PSYCHIC;
         break;
       default:
-        // Just in case there's no match
         moveId = MoveId.TRI_ATTACK;
```
MokaStitcher, line 112: huh, i would've done the same thing for default and case TerrainType.NONE: use the biome moveId
doesn't really matter but

src/data/move-attrs/random-moveset-move-attr.ts
```diff
@@ -1,67 +1,154 @@
...
-   * against a random legal target
-   * @param user the {@linkcode Pokemon} using the move
-   * @param target the {@linkcode Pokemon} targeted by the move
-   * @param _move the {@linkcode Move} being used
-   * @returns `true` if a move is successfully invoked
+   * User calls a random moveId selected in {@linkcode getCondition}
+   * @param user Pokemon that used the move and will call a random move
+   * @param target Pokemon that will be targeted by the random move (if single target)
+   * @param move Move being used
+   * @param args Unused
```
MokaStitcher: there no args?
instead there's `overriden: boolean` and it's not unused

src/phases/command-phase.ts
```diff
@@ -149,12 +150,7 @@ export class CommandPhase extends FieldPhase {
    * @returns `true` if the command was successful
    * @overload
    */
-  public handleCommand(
-    command: BattleCommand.FIGHT,
-    cursor: number,
-    ignorePp?: boolean,
-    targets?: MoveTargetSet,
-  ): boolean;
+  public handleCommand(command: BattleCommand.FIGHT, cursor: number, ignorePp?: boolean, turnMove?: TurnMove): boolean;
```
MokaStitcher, line 160: could you add the doc for `turnMove`?

src/phases/command-phase.ts
```diff
@@ -183,22 +179,32 @@ export class CommandPhase extends FieldPhase {
     switch (command) {
       case BattleCommand.FIGHT:
         const ignorePp = args[0] as boolean | undefined;
-        const targets = args[1] as MoveTargetSet | undefined;
+        //const targets = args[1] as MoveTargetSet | undefined;
```
MokaStitcher:
```suggestion
```

MokaStitcher (commented):
just code review

Does this PR need to update the locales? If not please revert it / update it to the current commit on beta ([3da1f38](https://github.com/Despair-Games/poketernity-locales/tree/3da1f38cffae39a12b69dfd0e795015a497e4d76)) so that it doesn't create conflicts with other PRs

<sub> conflicts with 426 are going to be a lot :c

DayKev:
Ugh how did the locales get out of sync...

src/data/all-moves.ts
DayKev: I thought I removed this... ugh, must have been a merge conflict.

src/data/move-attrs/random-moveset-move-attr.ts
DayKev: Hm, I guess the tsdocs got out of sync over time. Will fix them up.

src/phases/command-phase.ts
DayKev: Okay no I removed this too... ugh. :despair:

MokaStitcher: ~~Did something happen with the force push from yesterday maybe?~~
Nevermind it was just one line lol

DayKev: Nah it would have been from one of the merges, the force push was just a single line change to remove an import I left in.

src/data/all-moves.ts
```diff
@@ -2962,45 +3189,57 @@ export function initMoves() {
...
       .attr(HitsTagAttr, BattlerTagType.MINIMIZED, true)
-      .edgeCase() // I assume it's because it needs darkest lariat and incineroar
-      .ignoresVirtual(),
+      .edgeCase(), // I assume it's because it needs darkest lariat and incineroar
     new AttackMove(MoveId.OCEANIC_OPERETTA, ElementalType.WATER, MoveCategory.SPECIAL, 195, -1, 1, -1, 0, 7)
-      .edgeCase() // I assume it's because it needs sparkling aria and primarina
-      .unimplemented()
-      .ignoresVirtual(),
-    new AttackMove(MoveId.GUARDIAN_OF_ALOLA, ElementalType.FAIRY, MoveCategory.SPECIAL, -1, -1, 1, -1, 0, 7)
       .unimplemented()
```
flx-sta:
```suggestion
```

src/data/all-moves.ts
```diff
@@ -3099,39 +3337,33 @@ export function initMoves() {
...
-      .ignoresVirtual(),
+      .ignoresAbilities(),
     new AttackMove(MoveId.MENACING_MOONRAZE_MAELSTROM, ElementalType.GHOST, MoveCategory.SPECIAL, 200, -1, 1, -1, 0, 7)
-      .ignoresAbilities()
       .unimplemented()
-      .ignoresVirtual(),
+      .ignoresAbilities(),
     new AttackMove(MoveId.LETS_SNUGGLE_FOREVER, ElementalType.FAIRY, MoveCategory.PHYSICAL, 190, -1, 1, -1, 0, 7)
-      .edgeCase() // I assume it needs play rough and mimikyu
       .unimplemented()
```
flx-sta:
```suggestion
      .unimplemented()
```

src/data/move-attrs/nature-power-attr.ts
```diff
@@ -1,133 +1,100 @@
...
+          return MoveId.DRACO_METEOR;
+        case Biome.FACTORY:
+          return MoveId.FLASH_CANNON;
+        case Biome.LABORATORY:
+          return MoveId.ZAP_CANNON;
+        case Biome.POWER_PLANT:
+          return MoveId.CHARGE_BEAM;
+        case Biome.END:
+          return MoveId.ETERNABEAM;
+      }
```
flx-sta, line 92: NIT: This switch should probably be replaced with a an object/dict
e.g.
```ts
 const biomeMoveMap = {
    [Biome.TOWN]: MoveId.ROUND,
    [Biome.METROPOLIS]: MoveId.TRI_ATTACK,
    [Biome.SLUM]: MoveId.SLUDGE_BOMB,
    [Biome.PLAINS]: MoveId.SILVER_WIND,
    [Biome.GRASS]: MoveId.GRASS_KNOT,
    [Biome.TALL_GRASS]: MoveId.POLLEN_PUFF,
    [Biome.MEADOW]: MoveId.GIGA_DRAIN,
    [Biome.FOREST]: MoveId.BUG_BUZZ,
    [Biome.JUNGLE]: MoveId.LEAF_STORM,
    [Biome.SEA]: MoveId.HYDRO_PUMP,
    [Biome.SWAMP]: MoveId.MUD_BOMB,
    [Biome.BEACH]: MoveId.SCALD,
    [Biome.LAKE]: MoveId.BUBBLE_BEAM,
    [Biome.SEABED]: MoveId.BRINE,
    [Biome.ISLAND]: MoveId.LEAF_TORNADO,
    [Biome.MOUNTAIN]: MoveId.AIR_SLASH,
    [Biome.BADLANDS]: MoveId.EARTH_POWER,
    [Biome.DESERT]: MoveId.SCORCHING_SANDS,
    [Biome.WASTELAND]: MoveId.DRAGON_PULSE,
    [Biome.CONSTRUCTION_SITE]: MoveId.STEEL_BEAM,
    [Biome.CAVE]: MoveId.POWER_GEM,
    [Biome.ICE_CAVE]: MoveId.ICE_BEAM,
    [Biome.SNOWY_FOREST]: MoveId.FROST_BREATH,
    [Biome.VOLCANO]: MoveId.LAVA_PLUME,
    [Biome.GRAVEYARD]: MoveId.SHADOW_BALL,
    [Biome.RUINS]: MoveId.ANCIENT_POWER,
    [Biome.TEMPLE]: MoveId.EXTRASENSORY,
    [Biome.DOJO]: MoveId.FOCUS_BLAST,
    [Biome.FAIRY_CAVE]: MoveId.ALLURING_VOICE,
    [Biome.ABYSS]: MoveId.OMINOUS_WIND,
    [Biome.SPACE]: MoveId.DRACO_METEOR,
    [Biome.FACTORY]: MoveId.FLASH_CANNON,
    [Biome.LABORATORY]: MoveId.ZAP_CANNON,
    [Biome.POWER_PLANT]: MoveId.CHARGE_BEAM,
    [Biome.END]: MoveId.ETERNABEAM
};
```
<sup> DISCLAIMER: Untested code! </sup>

flx-sta: Also: What's the default? If there is no biome. I mean it shouldn't be the case but TECHNICALLY it can occur

src/data/move-attrs/random-moveset-move-attr.ts
```diff
@@ -1,67 +1,154 @@
...
+ * Used for {@linkcode MoveId.ASSIST} and {@linkcode MoveId.SLEEP_TALK}
+ *
+ * Fails if the user has no callable moves.
+ * @extends CallMoveAttr
+ * @see {@linkcode getCondition} for move selection
  */
-export class RandomMovesetMoveAttr extends OverrideMoveEffectAttr {
-  private enemyMoveset: boolean | null;
+export class RandomMovesetMoveAttr extends CallMoveAttr {
+  private includeParty: boolean;
```
flx-sta: Q: Does this ever change after int? Otherwise => readonly

```suggestion
  private readonly includeParty: boolean;
```

src/data/move-attrs/call-move-attr.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Attribute used to call a different move.
+ *
+ * Used by other move attributes: {@linkcode MetronomeAttr}, {@linkcode RandomMovesetMoveAttr}, {@linkcode CopycatAttr}
+ * and {@linkcode NaturePowerAttr}
+ * @see {@linkcode apply} for move call
+ * @extends OverrideMoveEffectAttr
+ */
+export abstract class CallMoveAttr extends OverrideMoveEffectAttr {
+  protected invalidMoves: MoveId[];
+  protected hasTarget: boolean;
```
flx-sta, line 30: Q: Do these ever change? Otherwise => `readonly`

test/moves/sketch.test.ts
```diff
@@ -82,8 +82,8 @@ describe("Moves - Sketch", () => {
 
   it("should sketch moves that call other moves", async () => {
     const randomMoveAttr = allMoves[MoveId.METRONOME].findAttr(
-      (attr) => attr instanceof RandomMoveAttr,
-    ) as RandomMoveAttr;
+      (attr) => attr instanceof MetronomeAttr,
+    ) as MetronomeAttr;
```
flx-sta, line 86: Oh boy... I can already smell the new dep. violations

src/phases/move-phase.ts
```diff
@@ -421,6 +420,16 @@ export class MovePhase extends BattlePhase {
     }
   }
 
+  protected updateLastMoveId(success: boolean): void {
+    // Update the battle's "last move" pointer, unless we're currently mimicking a move.
+    if (!allMoves[this.move.moveId].hasAttr(CopycatAttr)) {
```
flx-sta: NIT: Shouldn't this comment be above the methods?

src/phases/move-effect-phase.ts
```diff
@@ -93,6 +93,13 @@ export class MoveEffectPhase extends HitCheckPhase {
     applyMoveAttrs(OverrideMoveEffectAttr, user, targets[0], move, overridden, this.move.virtual);
     // If other effects were overridden, stop this phase before they can be applied
     if (overridden.value) {
+      this.moveHistoryEntry = {
+        moveId: this.move.moveId,
+        targets: this.adjustedTargets ?? this.targets,
+        result: MoveResult.SUCCESS,
+        virtual: this.move.virtual,
+      };
+      user.pushMoveHistory(this.moveHistoryEntry);
```
flx-sta: NIT: Consider deconstructing `this.move`:
```suggestion
      const { moveId, virtual } = this.move;
      this.moveHistoryEntry = {
        moveId,
        targets: this.adjustedTargets ?? this.targets,
        result: MoveResult.SUCCESS,
        virtual,
      };
      user.pushMoveHistory(this.moveHistoryEntry);
```

src/data/move.ts
```diff
@@ -1105,17 +1095,20 @@ export type MoveTargetSet = {
   multiple: boolean;
```
flx-sta, line 968: Q: What's the use of `multiple` ? I mean `targets` is already plural, so why multiple? Or is it related to something else?

src/data/move.ts
```diff
@@ -1105,17 +1095,20 @@ export type MoveTargetSet = {
...
   user.getOpponents().forEach((p) => applyMoveAttrs(VariableTargetAttr, user, p, allMoves[moveId], variableTarget));
 
-  const moveTarget = allMoves[moveId].hasAttr(VariableTargetAttr)
-    ? variableTarget.value
-    : moveId
-      ? allMoves[moveId].moveTarget
-      : moveId === undefined
-        ? MoveTarget.NEAR_ENEMY
-        : [];
+  let moveTarget: MoveTarget | undefined;
```
flx-sta, line 975: If I'm not mistaken any `let` that isn't defined in the beginning automatically has `undefined` as type added.
I might be wrong.

MokaStitcher:
lots of conflicts to solve
I promise we'll get this out eventually 😭

src/data/init-moves.ts
```diff
@@ -3092,39 +2977,32 @@ export function initMoves() {
     new AttackMove(MoveId.PHOTON_GEYSER, ElementalType.PSYCHIC, MoveCategory.SPECIAL, 100, 100, 5, -1, 0, 7)
       .attr(UseHigherAttackingStatAttr)
       .ignoresAbilities(),
-    /* Unused */
+    // #region USUM Z-Moves (unused)
     new AttackMove(MoveId.LIGHT_THAT_BURNS_THE_SKY, ElementalType.PSYCHIC, MoveCategory.SPECIAL, 200, -1, 1, -1, 0, 7)
       .attr(UseHigherAttackingStatAttr)
```
MokaStitcher, line 2963: was the unimplemented removed on purpose?

```suggestion
      .unimplemented()
      .attr(UseHigherAttackingStatAttr)
```

src/data/move-attrs/mirror-move-attr.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+export class MirrorMoveAttr extends CallMoveAttr {
+  constructor() {
+    super();
+    this.invalidMoves = invalidMirrorMoveMoves;
+    this.hasTarget = true;
+  }
+
+  override apply(user: Pokemon, target: Pokemon, _move: Move, overridden: BooleanHolder): boolean {
+    const lastMove = target.getLastXMoves()[0].move;
```
MokaStitcher, line 24: according to [bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Mirror_Move_(move)), mirror move should use the last move that targetted the user from a mon currently on the field. 
> Mirror Move uses the last move targeted at the user by a Pokémon still on the field

This seems to imply that 
- the move to copy doesn't need to have been used by the target, but could be a move used by its double partner, or even the user's partner.
- the move to copy wouldn't necessarily be its last used move either, but the last move that targetted the user.
- as a result it shouldn't fail even if the target itself hasn't done any action

I'm not sure how much of that is worth marking the move as partial. This is definitely only relevant in double battles, but it seems like a big enough difference to this implementation to be noted.


edit: i have attempted to playtest it in a double battle, and the opponents end up in an infinite loop of using mirror move over and over :')
This may be related to one of them attempting to target the other because double battles are scuffed? Not sure

MokaStitcher (commented):
I have done some limited testing so far, but mirror move is fully broken in double battles
```
  ENEMY_LEVEL_OVERRIDE: 12, //  Not sure if this is needed to reproduce consistently, but I had this in my testing, so enemies are faster than my mons. turn 1 the moves fail, turn 2 they end up in an infinite loop.
  ENEMY_MOVESET_OVERRIDE: MoveId.MIRROR_MOVE,
```
(this doesn't happen on the turn command branch, probably not on beta either but i haven't tested it to make sure.
Although there's no infinite loop there, the 2nd opponent ends up hitting its friend with the move so, things are weird too, just not game breaking like here)

(I haven't playtested the other moves at all yet)

src/data/init-moves.ts
DayKev: No, the Z-Moves should still be marked unimplemented. Will fix.

Snailman11:
Should Fix #550 ([Bug] Future Sight hits instantly when used indirectly)

src/data/move-attrs/nature-power-attr.ts
DayKev: Using a switch statement provides type safety (for example if I remove the `case Biome.TOWN:` section it'll throw this error:
image

That's also why there's no `default:` case, so there will be an error if a Biome is missing.

src/data/move-attrs/call-move-attr.ts
DayKev: The subclasses re-assign them.
image

src/phases/move-effect-phase.ts
DayKev: Will look into using destructuring for `MoveEffectPhase` separately.

src/data/move.ts
DayKev: I think `multiple` means "this move *can* target multiple pokemon", while the `targets` array contains the *actual* targets of the move, and it might only contain one pokemon because there's only one valid target on the field.

src/data/move.ts
DayKev: Adding `| undefined` removes the following error:
image

Note that this is preserving the existing behavior, just rewriting it to be more readable.

src/data/move-attrs/mirror-move-attr.ts
DayKev: Showdown has it implemented as based on the last move used by the target of Mirror Move:
```ts
  mirrormove: {
    num: 119,
    accuracy: true,
    basePower: 0,
    category: "Status",
    isNonstandard: "Past",
    name: "Mirror Move",
    pp: 20,
    priority: 0,
    flags: {failencore: 1, nosleeptalk: 1, noassist: 1, failcopycat: 1, failmimic: 1, failinstruct: 1},
    onTryHit(target, pokemon) {
      const move = target.lastMove;
      if (!move?.flags['mirror'] || move.isZ || move.isMax) {
        return false;
      }
      this.actions.useMove(move.id, pokemon, {target});
      return null;
    },
    callsMove: true,
    secondary: null,
    target: "normal",
    type: "Flying",
    zMove: {boost: {atk: 2}},
    contestType: "Clever",
  },
```
This is also supported by the in-game description of the move in gens 5-7 (from the Bulbapedia link):
> The user counters the target by mimicking the target's last move.

It might need some testing on a gen7 cart to be sure.

test/moves/sketch.test.ts
DayKev: Well unlike the ability attrs you didn't add the same flag system for move attrs, though it doesn't seem to matter atm.

src/data/move-attrs/mirror-move-attr.ts
DayKev: Testing by Snailman confirms Showdown's implementation: https://cdn.discordapp.com/attachments/1229525892624154684/1341680621461835806/PXL_20250219_075640199.mp4?ex=67b6e0f3&is=67b58f73&hm=76cac61a2abdffbb178a803e1e686c8cc7084e8f1ff92cb2666ac688e7d477e5&

Also, it seems like Mirror Move's effect might have changed at some point between Gen1 and Gen7, based on a comment on Bulbapedia's discussion page for the move (which explains why Bulbapedia's page is inaccurate): https://bulbapedia.bulbagarden.net/wiki/Talk:Mirror_Move_(move)#Effect_out_of_date_for_later_generations

DayKev: Oh yeah re: Mirror Move causing an infinite loop, it's because I forgot to add the list of invalid moves to Mirror Move, doing that now.

DayKev: List added.

src/data/move.ts
innerthunder: - `targets` = all legal targets on the field
- `multiple` = if `true`, this move hits every target in `targets` rather than just 1 selected target

src/battle.ts
```diff
@@ -55,11 +54,12 @@ import {
   TUTORIAL_BATTLE_WAVE,
 } from "./data/special-waves";
 import type { Move } from "#app/data/move";
+import type { TurnMove } from "#app/@types/TurnMove";
 
 export interface TurnCommand {
   command: BattleCommand;
   cursor?: number;
-  move?: QueuedMove;
+  move?: TurnMove;
```
innerthunder: Should this be renamed? The references to `TurnCommand.move.move` seem weird.

DayKev: I'll rename it to `turnMove`. 👍

innerthunder (dismissed):

MokaStitcher:
just dropping a few issues that may be kind of related. Does this PR fix them?
#593 
#550

src/data/move-attrs/mirror-move-attr.ts
MokaStitcher: ok that actually makes things much easier :')
I'll try and get back to review/playtesting this weekend so we can finally get this out

DayKev:
> just dropping a few issues that may be kind of related. Does this PR fix them?

550 - No. Taking a look at the `DelayedAttackAttr`, it was intentionally set up to be this way (so in that sense it's not a "bug"). I wonder why it was done this way. 🤔
593 - Probably not? Dancer's move attr isn't touched.

innerthunder (dismissed):

src/data/init-moves.ts
```diff
@@ -642,11 +640,10 @@ export function initMoves() {
       { failOnOverlap: true },
     ),
     new AttackMove(MoveId.BIDE, ElementalType.NORMAL, MoveCategory.PHYSICAL, -1, -1, 10, -1, 1, 1)
-      .ignoresVirtual()
       .target(MoveTarget.USER)
       .unimplemented(),
-    new SelfStatusMove(MoveId.METRONOME, ElementalType.NORMAL, -1, 10, -1, 0, 1).attr(RandomMoveAttr).ignoresVirtual(),
-    new StatusMove(MoveId.MIRROR_MOVE, ElementalType.FLYING, -1, 20, -1, 0, 1).attr(CopyMoveAttr).ignoresVirtual(),
+    new SelfStatusMove(MoveId.METRONOME, ElementalType.NORMAL, -1, 10, -1, 0, 1).attr(MetronomeAttr),
+    new StatusMove(MoveId.MIRROR_MOVE, ElementalType.FLYING, -1, 20, -1, 0, 1).attr(MirrorMoveAttr).edgeCase(), // Need to find out which moves are not copyable by Mirror Move
```
MokaStitcher: is this comment solved now?

src/field/pokemon.ts
```diff
@@ -4573,20 +4572,22 @@ export class EnemyPokemon extends Pokemon {
     if (movePool.length) {
       // If there's only 1 move in the move pool, use it.
       if (movePool.length === 1) {
-        return { moveId: movePool[0].moveId, targets: this.getNextTargets(movePool[0].moveId) };
+        const move = movePool[0].getMove();
+        return { move, targets: this.getNextTargets(movePool[0].moveId), type: this.getMoveType(move) };
```
```suggestion
        return { move, targets: this.getNextTargets(move.id), type: this.getMoveType(move) };
```
it's the same, but more consistent style wise imo

src/field/pokemon.ts
```diff
@@ -4573,20 +4572,22 @@ export class EnemyPokemon extends Pokemon {
...
       }
       // If a move is forced because of Encore, use it.
       const encoreTag = this.getTag<EncoreTag>(BattlerTagType.ENCORE);
       if (encoreTag) {
         const encoreMove = movePool.find((m) => m.moveId === encoreTag.moveId);
         if (encoreMove) {
-          return { moveId: encoreMove.moveId, targets: this.getNextTargets(encoreMove.moveId) };
+          const move = encoreMove.getMove();
+          return { move, targets: this.getNextTargets(encoreMove.moveId), type: this.getMoveType(move) };
```
```suggestion
          return { move, targets: this.getNextTargets(move.id), type: this.getMoveType(move) };
```

test/moves/future_sight.test.ts
```diff
@@ -253,5 +255,17 @@ describe("Moves - Future Sight", () => {
 
   it.todo("should not apply the user's held items when dealing damage if the user is inactive");
 
-  it.todo("should invoke the move's first phase when called by Metronome");
+  it.todo("should invoke the move's first phase when called by Metronome", async () => {
```
MokaStitcher, line 258: is it intended to still be "todo"?

src/data/move-attrs/recoil-attr.ts
```diff
@@ -38,11 +37,6 @@ export class RecoilAttr extends MoveEffectAttr {
       return false;
     }
 
-    // Chloroblast and Struggle should not deal recoil damage if the move was not successful
-    if (this.useHp && [MoveResult.FAIL, MoveResult.MISS].includes(user.getLastXMoves(1)[0]?.result)) {
```
MokaStitcher, line 42: can you explain where this change is coming from?

public/locales
MokaStitcher, line 1: can you update the locales please (feel free to update to latest since the fusion removal PR was merged)

MokaStitcher (dismissed):
Haven't been able to playtest more after all
Feel free to consider the PR as approved again by me once the locale repo is updated + small nits addressed

src/data/init-moves.ts
DayKev: Oops yeah, will remove.

test/moves/future_sight.test.ts
DayKev: Yes, Future Sight's move attr wasn't changed yet.

src/data/move-attrs/recoil-attr.ts
DayKev: Iirc IT said it shouldn't be necessary anymore due to changes to `MoveEffectPhase` and/or `MoveEffectAttr`

DayKev: Just tested and confirmed it works as it's supposed to without this `if` check.

PR #218, opened 2025-01-08
[Refactor] Improvements on FixedInt

torranx:
none

- `IntegerHolder` is deprecated
- `FixedInt` was mistakenly tagged as deprecated
- type `int` is being scrapped

- changed `FixedInt` and `fixedInt` to `FixedNumber` and `fixedNumber` respectively
- removed `IntegerHolder` and changed the base class of `FixedNumber` to `NumberHolder`
- removed deprecated tag on `FixedNumber`
- added docs on [`FixedNumber`, `fixedNumber`](https://github.com/Despair-Games/poketernity/pull/218/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351), and [`transformValue`](https://github.com/Despair-Games/poketernity/pull/218/files#diff-edd9ef70f3f68aa9cf56787d0e2f43b0a8c09f7bf238026cb068c9a598732ec7R20)

## How to test the changes?
regular playtest, tweak game speed, there shouldn't be errors since I did `rename symbol`

## Checklist

### DIFF
--- a/src/system/game-speed.ts
+++ b/src/system/game-speed.ts
@@ -13,15 +13,15 @@ export function initGameSpeed() {
 
   /**
    * Adjusts the given value based on the game speed or returns it as is.
-   * @param value - The value to be evaluated. Can be either a `number` or an instance of {@linkcode FixedNumber}.
+   * @param num - The value to be evaluated. Can be either a `number` or an instance of {@linkcode FixedNumber}.
    * @returns The original numeric value if `value` is an instance of {@linkcode FixedNumber} or if {@linkcode BattleScene.gameSpeed | gameSpeed} is `1`;
    * otherwise, the value adjusted for the game speed.
    */
-  const transformValue = (value: number | FixedNumber): number => {
-    if (value instanceof FixedNumber) {
-      return (value as FixedNumber).value;
+  const transformValue = (num: number | FixedNumber): number => {
+    if (num instanceof FixedNumber) {
+      return num.value;
     }
-    return thisArg.gameSpeed === 1 ? value : Math.ceil((value /= thisArg.gameSpeed));
+    return thisArg.gameSpeed === 1 ? num : Math.ceil((num /= thisArg.gameSpeed));
   };
 
   const originalAddEvent = this.time.addEvent;
--- a/src/system/game-speed.ts
+++ b/src/system/game-speed.ts
@@ -3,7 +3,7 @@ import type FadeIn from "phaser3-rex-plugins/plugins/audio/fade/FadeIn";
 import type FadeOut from "phaser3-rex-plugins/plugins/audio/fade/FadeOut";
 import type BattleScene from "#app/battle-scene";
 import { globalScene } from "#app/global-scene";
-import { FixedInt } from "#app/utils";
+import { FixedNumber } from "#app/utils";
 
 type FadeIn = typeof FadeIn;
 type FadeOut = typeof FadeOut;
@@ -11,9 +11,15 @@ type FadeOut = typeof FadeOut;
 export function initGameSpeed() {
   const thisArg = this as BattleScene;
 
-  const transformValue = (value: number | FixedInt): number => {
-    if (value instanceof FixedInt) {
-      return (value as FixedInt).value;
+  /**
+   * Adjusts the given value based on the game speed or returns it as is.
+   * @param value - The value to be evaluated. Can be either a `number` or an instance of {@linkcode FixedNumber}.
+   * @returns The original numeric value if `value` is an instance of {@linkcode FixedNumber} or if {@linkcode BattleScene.gameSpeed | gameSpeed} is `1`;
+   * otherwise, the value adjusted for the game speed.
+   */
+  const transformValue = (value: number | FixedNumber): number => {
+    if (value instanceof FixedNumber) {
+      return (value as FixedNumber).value;
     }
     return thisArg.gameSpeed === 1 ? value : Math.ceil((value /= thisArg.gameSpeed));
   };
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,3 +1,7 @@
+// -- start tsdoc imports --
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import type { initGameSpeed } from "#app/system/game-speed";
+// -- end tsdoc imports --
 import { MoneyFormat } from "#enums/money-format";
 import { Moves } from "#enums/moves";
 import i18next from "i18next";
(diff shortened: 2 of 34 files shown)


Tempo-anon (dismissed):

flx-sta (changes requested):
-

src/utils.ts
```diff
@@ -353,23 +357,23 @@ export class NumberHolder {
...
-}
-
-/** @deprecated Use {@linkcode NumberHolder}*/
-export class FixedInt extends IntegerHolder {
+/**
+ * Holds a fixed number value, this is solely used to differentiate between a regular number and a fixed number.
+ * @see `transformValue` in {@linkcode initGameSpeed}
+ * @extends NumberHolder
+ */
+export class FixedNumber extends NumberHolder {
```
flx-sta:
```suggestion
export class FixedNumber extends NumberHolder {

  public override readonly value: number;
  
```

flx-sta (changes requested):
Since `fixed` stands for _"the value being restricted to changes"_ ~~the class should override the `value` property~~ to be `readonly`

see @DayKev s final suggestion:
https://github.com/Despair-Games/poketernity/pull/218#discussion_r1908025347

src/system/game-speed.ts
```diff
@@ -3,17 +3,23 @@ import type FadeIn from "phaser3-rex-plugins/plugins/audio/fade/FadeIn";
...
-      return (value as FixedInt).value;
+  /**
+   * Adjusts the given value based on the game speed or returns it as is.
+   * @param value - The value to be evaluated. Can be either a `number` or an instance of {@linkcode FixedNumber}.
+   * @returns The original numeric value if `value` is an instance of {@linkcode FixedNumber} or if {@linkcode BattleScene.gameSpeed | gameSpeed} is `1`;
+   * otherwise, the value adjusted for the game speed.
+   */
+  const transformValue = (value: number | FixedNumber): number => {
+    if (value instanceof FixedNumber) {
+      return (value as FixedNumber).value;
```
DayKev:
```suggestion
      return value.value;
```
Also might want to rename the `value` parameter so we don't have `value.value` lol.

src/utils.ts
flx-sta: Begs the question at this point if it should even extend NumberHolder, lol.
changing the one only thing it has

src/utils.ts
```diff
@@ -353,23 +357,23 @@ export class NumberHolder {
...
-export class IntegerHolder extends NumberHolder {
-  constructor(value: number) {
-    super(value);
-  }
-}
-
-/** @deprecated Use {@linkcode NumberHolder}*/
-export class FixedInt extends IntegerHolder {
+/**
+ * Holds a fixed number value, this is solely used to differentiate between a regular number and a fixed number.
```
DayKev:
```suggestion
 * If passed as the delay for {@linkcode Phaser.Time.Clock.delayedCall} or similar
 * (normally called via {@linkcode BattleScene.time} or {@linkcode BattleScene.tweens}),
 * will cause the current game speed to not modify the delay.
*/
```

src/utils.ts
```diff
@@ -1,3 +1,7 @@
+// -- start tsdoc imports --
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import type { initGameSpeed } from "#app/system/game-speed";
```
DayKev, line 3:
```suggestion
/* eslint-disable @typescript-eslint/no-unused-vars */
import type BattleScene from "#app/battle-scene";
import type { initGameSpeed } from "#app/system/game-speed";
/* eslint-enable @typescript-eslint/no-unused-vars */
```

src/utils.ts
```diff
@@ -353,23 +357,23 @@ export class NumberHolder {
...
   constructor(value: number) {
     super(value);
   }
 }
 
-/** @deprecated */
-export function fixedInt(value: number): number {
-  return new FixedInt(value) as unknown as number;
+/**
+ * Helper method to create a FixedNumber
```
DayKev:
```suggestion
 * Helper method to create a {@linkcode FixedNumber}
```

src/utils.ts
DayKev: Yeah it probably doesn't need to extend `NumberHolder`, can skip the `override` too then.

src/utils.ts
```diff
@@ -353,23 +357,23 @@ export class NumberHolder {
...
+/**
+ * Holds a fixed number value, this is solely used to differentiate between a regular number and a fixed number.
+ * @see `transformValue` in {@linkcode initGameSpeed}
+ * @extends NumberHolder
+ */
+export class FixedNumber extends NumberHolder {
   constructor(value: number) {
     super(value);
   }
 }
```
DayKev:
```suggestion
export class FixedNumber {
  public readonly value: number;

  constructor(value: number) {
    this.value = value;
  }
}
```
Based on flx's suggestion.

src/utils.ts
flx-sta: Closing this in favor of @DayKev s suggestion

src/system/game-speed.ts
flx-sta: `num`. You may include this in your suggestion

src/system/game-speed.ts
```diff
@@ -3,17 +3,23 @@ import type FadeIn from "phaser3-rex-plugins/plugins/audio/fade/FadeIn";
...
     }
     return thisArg.gameSpeed === 1 ? value : Math.ceil((value /= thisArg.gameSpeed));
   };
```
DayKev:
```suggestion
  const transformValue = (num: number | FixedNumber): number => {
    if (num instanceof FixedNumber) {
      return num.value;
    }
    return thisArg.gameSpeed === 1 ? num : Math.ceil((num /= thisArg.gameSpeed));
  };
```

src/utils.ts
torranx: I feel like this suggestion does not communicate well what the intention of the class is or what is it for, since it is passed to different indexes. I opted for `this is solely used to differentiate between a regular number and a fixed number` and linked the `initGameSpeed` to provide context.

what do you think? @DayKev @flx-sta

DayKev: This is what the class is used for, it's only used in context of game speed and delayedCall/etc. Your tsdoc doesn't explain what it does or what it is, someone reading it isn't going to know what a "fixed number" is. Following the link to `initGameSpeed` should be optional for someone who wants/needs to dig into it further.

flx-sta: I feel like @DayKev s suggestion is missing a piece of the sentence:

>  If passed as the delay for {@linkcode Phaser.Time.Clock.delayedCall} or (normally called via {@linkcode BattleScene.time} or {@linkcode BattleScene.tweens}), will cause the current game speed to not modify the delay.

`or (normally [...]), will cause the [...]` ? huh?

torranx: what I don't like is how it emphasizes the `delay` (`FixedNumber` is also used for `duration` and `hold`). another option is:

```
/**
 * Holds a fixed number value, this is solely used to differentiate between a regular number
 * and a constant or "fixed" number.
 * This is currently used in the game speed system to differentiate between a fixed game speed and a dynamic one.
 * @see `transformValue` in {@linkcode initGameSpeed}
 * @extends NumberHolder
 */
```

flx-sta: This solution seems to combine both so seems fine by me.

<sup>I honestly don't care enough to discuss further.<sup>

DayKev: > `or (normally [...]), will cause the [...]` ? huh?

It's `or similar (normally ...)`.

> what I don't like is how it emphasizes the `delay` (`FixedNumber` is also used for `duration` and `hold`). another option is:
> 
> [example]

That might be better.

DayKev (approved):

flx-sta (approved):

PR #219, opened 2025-01-08
[P1 Bug] Fix Sticky Hold and Unaware bugs from AbAttr refactor

PigeonBar:
## What are the changes the user will see?

Sticky Hold and Unaware work properly.

## Why am I making these changes?

Fixing some bugs introduced by #197.

<sup>Hopefully I didn't miss any other bugs...</sup>

## What are the changes from a developer perspective?

1. Uses of Sticky Hold and Unaware now use correct parameters.
2. Added some basic tests for Sticky Hold and Unaware.

## Screenshots/Videos

<details>
<summary> Before the changes: Sticky Hold crashes the game </summary>

https://github.com/user-attachments/assets/ade1bcb1-18f8-422d-8ed0-3602b5e40914

</details>
<details>
<summary> After the changes: Sticky Hold works correctly </summary>

https://github.com/user-attachments/assets/9fc447b7-af65-4bdd-b49e-08b7fa1ea638

</details>
<details>
<summary> Before the changes: Unaware Dondozo takes extra damage after using Swagger </summary>

https://github.com/user-attachments/assets/f1dd33c7-6a38-41f2-880e-cb96c117aa6c

</details>
<details>
<summary> After the changes: Unaware Dondozo no longer takes extra damage after using Swagger </summary>

https://github.com/user-attachments/assets/6bf426e2-9c17-4dcc-8fe7-9255be49e3d3

</details>

## How to test the changes?

`npm run test [sticky_hold | unaware]`

## Checklist

- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test`)
  - [x] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?
- [x] Have I provided screenshots/videos of the changes (if applicable)?

PR #199, opened 2025-01-07
[Refactor] Improved type enforcement for `applyMoveAttrs`

innerthunder:
## What are the changes the user will see?

N/A

## Why am I making these changes?

I made this as a proof of concept for consolidating the `applyXAbAttr` wrapper functions.

## What are the changes from a developer perspective?

`applyMoveAttrs` and related functions have new signatures.

Old:

```ts
export function applyMoveAttrs(
  attrType: Constructor<MoveAttr>,
  user: Pokemon | null,
  target: Pokemon | null,
  move: Move,
  ...args: unknown[]
): void;
```

New:

```ts
export function applyMoveAttrs<TAttr extends MoveAttr>(
  attrType: Constructor<TAttr>,
  ...params: Parameters<TAttr["apply"]>
): void;
```

The new `params` field pulls function parameters directly from the templated attribute's `apply` function, meaning that extended arguments (`args`) have stricter enforcement on their types

## Screenshots/Videos

> [!NOTE]
> All screenshots are pulled from VSCode. Other platforms may not support the changes as well.

### TSDocs for existing apply calls

Old:

image

New:

image

### Suggestions when writing apply calls

Old:

image

New:

image

## How to test the changes?

`npm run test`. You can also experiment with the new functions in your IDE.

## Checklist

- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test`)
  - [x] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?
- [n/a] Have I provided screenshots/videos of the changes (if applicable)?
  - [n/a] Have I made sure that any UI change works for both UI themes (default and legacy)?

PR #221, opened 2025-01-09
[Refactor][Ability] Consolidate Ability application functions

innerthunder:

Some bugs related to ability application parameters should be fixed (overlaps with #219 ).

29 different functions to apply abilities (following the pattern `applyXAbAttrs`) are now consolidated into a single function:

```ts
/**
 * Applies a Pokemon's ability attributes of matching type
 * @param attrType The type of attribute to apply
 * @param params The parameters for the given attribute's `apply` function. This should include:
 * - `pokemon`: The {@linkcode Pokemon} with the ability
 * - `simulated`: If `true`, suppresses changes to game state when applying.
 * - Any additional necessary arguments for the specific attribute type
 * @returns The message(s) displayed when the ability applies
 * @see {@linkcode AbAttr}
 */
export function applyAbAttrs<TAttr extends AbAttr>(
  attrType: Constructor<TAttr>,
): string[];
```

With this, the following changes have been made to ability attributes and their apply functions:
- All attributes now override `apply` with named additional arguments instead of using their own specialized apply functions.
- The `passive` parameter is now removed. `AbAttr` now has a field `source: Ability` to support some attributes that relied on `passive` to look up ability names.
- `AbAttr` also now has a `showAbilityInstant` field to replace logic from the old `applyXAbAttrs` functions

These changes come with several benefits compared to the old version, including:
- The parameters passed through `applyAbAttrs` now have stricter type enforcement (see #199 for more details)
- Assuming #199 is also added, the new `applyAbAttrs` API is now much closer to how move attributes are applied.
- One-off attribute types no longer need a parent class with an empty apply function just to add named parameters.
- Open-ended `args` params are no longer needed for subclasses (mostly because of #197 , but this finishes the job), and parameters to `apply` can generally be implemented in a much cleaner way.

## How to test the changes?

`npm run test`

## Checklist


src/@types/PokemonAttackCondition.ts
```diff
@@ -1,4 +1,4 @@
 import type { Move } from "#app/data/move";
 import type { Pokemon } from "#app/field/pokemon";
 
-export type PokemonAttackCondition = (user: Pokemon | null, target: Pokemon | null, move: Move) => boolean;
+export type PokemonAttackCondition = (user: Pokemon | null, target: Pokemon | null, move?: Move) => boolean;
```
DayKev: Instead of `| null` these should probably all be optional (previously this wasn't possible because `move` wasn't optional and was the third param, for some reason...).

src/data/ab-attrs/ab-attr.ts
```diff
@@ -1,25 +1,28 @@
 import type { AbAttrCondition } from "#app/@types/AbAttrCondition";
 import { type Pokemon } from "#app/field/pokemon";
+import type { Ability } from "../ability";
```
DayKev:
```suggestion
import type { Ability } from "#app/data/ability";
```

src/data/ab-attrs/add-second-strike-ab-attr.ts
```diff
@@ -22,19 +23,17 @@ export class AddSecondStrikeAbAttr extends PreAttackAbAttr {
...
+  override apply(
     pokemon: Pokemon,
-    _passive: boolean,
     _simulated: boolean,
     _defender: Pokemon,
     move: Move,
-    hitCount: NumberHolder,
-    multiplier: NumberHolder,
+    hitCount: NumberHolder | null,
+    multiplier: NumberHolder | null,
```
DayKev: Shouldn't these be optional instead?

DayKev (commented):
image

src/data/ab-attrs/add-second-strike-ab-attr.ts
flx-sta: It's opinionated but I'd agree with generally considering `undefined` over `null` in the future. Takes less spaces, lol.

```ts
hitCount?: NumberHolder,
multiplier?: Numberholder,
```

1 thing to watch out for is that after an optional arg there can't be a non-optional. That's why you sometimes see `var: MyType | undefined` instead of `var?: MyType`

DayKev: > 1 thing to watch out for is that after an optional arg there can't be a non-optional. That's why you sometimes see `var: MyType | undefined` instead of `var?: MyType`

Yeah we should try to avoid that, it's a bad pattern.

flx-sta: The existing ones are of historical nature afaik so we are good

frutescens (dismissed):

src/data/ab-attrs/ab-attr.ts
```diff
@@ -1,25 +1,28 @@
...
+import type { Ability } from "#app/data/ability";
 
 export abstract class AbAttr {
+  public source: Ability;
   public showAbility: boolean;
+  public showAbilityInstant: boolean;
   private extraCondition: AbAttrCondition;
 
-  constructor(showAbility: boolean = true) {
+  constructor(showAbility: boolean = true, showAbilityInstant: boolean = false) {
```
flx-sta, line 11: This might be quite the refactor but couldn't we replace `showAbility` with an enum/flag instead?

Something like:
```ts
  export enum ShowAbilityFlag {
    NORMAL,
    INSTANT,
  }
```

src/data/ab-attrs/ab-attr.ts
```diff
@@ -1,25 +1,28 @@
...
 
   /**
    * Applies the effects of this attribute
-   * @param _pokemon The {@linkcode Pokemon} with the ability
-   * @param _passive `true` if the source ability is a Pokemon's passive
-   * @param _simulated `true` if attribute effects should be resolved without changing game state
-   * @param _args Any additional parameters or data to modify
+   * @param pokemon The {@linkcode Pokemon} with the ability
+   * @param simulated `true` if attribute effects should be resolved without changing game state
+   * @param args Any additional parameters or data to modify
```
flx-sta, line 20:
```suggestion
   * @param _pokemon The {@linkcode Pokemon} with the ability
   * @param _simulated `true` if attribute effects should be resolved without changing game state
   * @param _args Any additional parameters or data to modify
```

They have to match the names, don't they?

src/data/ab-attrs/arena-trap-ab-attr.ts
```diff
@@ -24,13 +24,7 @@ export class ArenaTrapAbAttr extends CheckTrappedAbAttr {
...
    * @returns `true` if enemy Pokemon is trapped
    */
-  override applyCheckTrapped(
-    pokemon: Pokemon,
-    _passive: boolean,
-    _simulated: boolean,
-    trapped: BooleanHolder,
-    otherPokemon: Pokemon,
-  ): boolean {
+  override apply(pokemon: Pokemon, _simulated: boolean, trapped: BooleanHolder, otherPokemon: Pokemon): boolean {
```
flx-sta: NIT:
```suggestion
  override apply(pokemon: Pokemon, _simulated: boolean, trapped: BooleanHolder, affectedPokemon: Pokemon): boolean {
```

src/data/ab-attrs/bypass-speed-chance-ab-attr.ts
```diff
@@ -27,11 +27,11 @@ export class BypassSpeedChanceAbAttr extends AbAttr {
   /**
    * bypass move order in their priority bracket when pokemon choose damaging move
    * @param pokemon {@linkcode Pokemon} applying this ability
-   * @param _passive N/A
-   * @param args [0] {@linkcode BooleanHolder} set to true when the ability activated
+   * @param passive N/A
```
flx-sta:
```suggestion
   * @param simulated N/A
```

src/data/ab-attrs/change-move-priority-ab-attr.ts
```diff
@@ -9,27 +9,19 @@ import { AbAttr } from "./ab-attr";
...
+  /** The condition moves must follow for the priority change to apply */
   private readonly moveFunc: (pokemon: Pokemon, move: Move) => boolean;
+  /** The amount of priority added or subtracted */
   private readonly changeAmount: number;
 
-  /**
-   * @param moveFunc applies priority-change to moves within a provided category
-   * @param changeAmount the amount of priority added or subtracted
-   */
   constructor(moveFunc: (pokemon: Pokemon, move: Move) => boolean, changeAmount: number) {
```
flx-sta: NIT:
```ts
(pokemon: Pokemon, move: Move) => boolean;
```

Should be defined as a type as it's used twice

src/data/ab-attrs/check-trapped-ab-attr.ts
```diff
@@ -18,14 +18,16 @@ export class CheckTrappedAbAttr extends AbAttr {
...
+  /**
+   * Applies an effect when another Pokemon attempts to leave the field.
+   * @param pokemon The {@linkcode Pokemon} with this ability
+   * @param simulated If `true`, suppresses changes to game state
+   * @param trapped A {@linkcode BooleanHolder} which, if `true`, prevents
+   * Pokemon from leaving the field.
+   * @param otherPokemon The {@linkcode Pokemon} attempting to leave the field
+   * @returns
+   */
+  override apply(_pokemon: Pokemon, _simulated: boolean, _trapped: BooleanHolder, _otherPokemon: Pokemon): boolean {
```
flx-sta:
```suggestion
  /**
   * Applies an effect when another Pokemon attempts to leave the field.
   * @param _pokemon The {@linkcode Pokemon} with this ability
   * @param _simulated If `true`, suppresses changes to game state
   * @param _trapped A {@linkcode BooleanHolder} which, if `true`, prevents
   * Pokemon from leaving the field.
   * @param _trappedPokemon The {@linkcode Pokemon} attempting to leave the field
   * @returns
   */
  override apply(_pokemon: Pokemon, _simulated: boolean, _trapped: BooleanHolder, _trappedPokemon: Pokemon): boolean {
```

src/data/ab-attrs/confusion-on-status-effect-ab-attr.ts
```diff
@@ -18,26 +17,24 @@ export class ConfusionOnStatusEffectAbAttr extends PostAttackAbAttr {
...
-    super((_user, _target, _move) => true);
+    super(false);
     this.effects = effects;
   }
+
   /**
    * Applies confusion to the target pokemon.
    * @param pokemon {@link Pokemon} attacking
-   * @param _passive N/A
+   * @param passive N/A
```
flx-sta:
```suggestion
   * @param simluated N/A
```

src/data/ab-attrs/damage-boost-ab-attr.ts
```diff
@@ -21,17 +21,18 @@ export class DamageBoostAbAttr extends PreAttackAbAttr {
...
-   * @param pokemon the attacker pokemon
-   * @param _passive N/A
-   * @param defender the target pokemon
-   * @param move the move used by the attacker pokemon
-   * @param args Utils.NumberHolder as damage
-   * @returns true if the function succeeds
+   * Multiplies a move's damage by {@linkcode damageMultiplier}
+   * if the attribute's {@linkcode condition} is met.
+   * @param pokemon The {@linkcode Pokemon} with this ability
+   * @param simulated If `true`, suppresses changes to game state
```
flx-sta, line 32:
```suggestion
   * @param _simulated If `true`, suppresses changes to game state
```

src/data/ab-attrs/field-move-power-boost-ab-attr.ts
```diff
@@ -22,11 +22,10 @@ export class FieldMovePowerBoostAbAttr extends PreAttackAbAttr {
...
   }
 
-  override applyPreAttack(
-    pokemon: Pokemon | null,
-    _passive: boolean | null,
+  override apply(
+    pokemon: Pokemon,
     _simulated: boolean,
-    defender: Pokemon | null,
+    defender: Pokemon,
```
flx-sta: Q: Where did `| null` go? I'd expect to see `defender?: Pokemon`

Or is that outdated?

src/data/ab-attrs/field-multiply-stat-ab-attr.ts
```diff
@@ -25,24 +23,21 @@ export class FieldMultiplyStatAbAttr extends AbAttr {
 
   /**
    * Tries to multiply a Pokemon's Stat
-   * @param _pokemon N/A
-   * @param _passive N/A
+   * @param pokemon The {@linkcode Pokemon} with this ability
+   * @param simulated If `true`, suppresses changes to game state
```
flx-sta, line 27:
```suggestion
   * @param _pokemon The {@linkcode Pokemon} with this ability
   * @param _simulated If `true`, suppresses changes to game state
```

src/data/ab-attrs/form-block-damage-ab-attr.ts
```diff
@@ -37,20 +37,18 @@ export class FormBlockDamageAbAttr extends ReceivedMoveDamageMultiplierAbAttr {
    * Removes the appropriate `BattlerTagType` when hit by an attack and is in its defense form.
    *
    * @param pokemon The Pokémon with the ability.
-   * @param _passive n/a
+   * @param passive n/a
    * @param attacker The attacking Pokémon.
    * @param move The move being used.
-   * @param _cancelled n/a
+   * @param cancelled n/a
```
flx-sta:
```suggestion
```

Not there anymore

src/data/ab-attrs/form-block-damage-ab-attr.ts
```diff
@@ -71,7 +69,7 @@ export class FormBlockDamageAbAttr extends ReceivedMoveDamageMultiplierAbAttr {
    * Gets the message triggered when the Pokémon avoids damage using the form-changing ability.
    * @param pokemon The Pokémon with the ability.
    * @param abilityName The name of the ability.
-   * @param _args n/a
+   * @param args n/a
```
flx-sta, line 70:
```suggestion
   * @param _args n/a
```

src/data/ab-attrs/full-hp-resist-type-ab-attr.ts
```diff
@@ -15,21 +15,17 @@ export class FullHpResistTypeAbAttr extends PreDefendAbAttr {
   /**
    * Reduces a type multiplier to 0.5 if the source is at full HP.
    * @param pokemon {@linkcode Pokemon} the Pokemon with this ability
-   * @param _passive n/a
-   * @param _simulated n/a (this doesn't change game state)
-   * @param _attacker n/a
+   * @param simulated n/a (this doesn't change game state)
+   * @param attacker n/a
```
flx-sta, line 19:
```suggestion
   * @param _attacker n/a
```

src/data/ab-attrs/ignore-opponent-stat-stages-ab-attr.ts
```diff
@@ -18,20 +18,14 @@ export class IgnoreOpponentStatStagesAbAttr extends AbAttr {
...
   /**
    * Modifies a BooleanHolder and returns the result to see if a stat is ignored or not
-   * @param _pokemon n/a
-   * @param _passive n/a
-   * @param _simulated n/a
-   * @param stat The {@linkcode EffectiveStat} to be ignored by this ability
+   * @param pokemon n/a
+   * @param passive n/a
+   * @param simulated n/a
+   * @param stat The {@linkcode BattleStat} to be ignored by this ability
```
flx-sta:
```suggestion
   * @param _pokemon n/a
   * @param _simulated n/a
   * @param stat The {@linkcode BattleStat} to be ignored by this ability
```

src/data/ab-attrs/infiltrator-ab-attr.ts
```diff
@@ -9,13 +9,13 @@ import { AbAttr } from "./ab-attr";
 export class InfiltratorAbAttr extends AbAttr {
   /**
    * Sets a flag to bypass screens, Substitute, Safeguard, and Mist
-   * @param _pokemon n/a
-   * @param _passive n/a
-   * @param _simulated n/a
+   * @param pokemon n/a
+   * @param passive n/a
+   * @param simulated n/a
    * @param bypassed a {@linkcode BooleanHolder} containing the flag
```
flx-sta:
```suggestion
   * @param pokemon n/a
   * @param simulated n/a
   * @param bypassed a {@linkcode BooleanHolder} containing the flag
```

src/data/ab-attrs/post-attack-ab-attr.ts
```diff
@@ -1,56 +1,50 @@
...
 import type { Move } from "#app/data/move";
 import { MoveCategory } from "#enums/move-category";
 import type { Pokemon } from "#app/field/pokemon";
-import type { HitResult } from "#enums/hit-result";
 import { AbAttr } from "./ab-attr";
 
 export class PostAttackAbAttr extends AbAttr {
-  private readonly attackCondition: PokemonAttackCondition;
+  /** Does this effect only apply to attack moves? */
+  private readonly attackOnly: boolean;
```
flx-sta: Is this a question from you or the "explanation"?

```suggestion
  /** Whether it only applies to attack moves. */
  private readonly attackMovesOnly: boolean;
```

src/data/ab-attrs/post-attack-ab-attr.ts
```diff
@@ -1,56 +1,50 @@
...
 
-  /** The default attackCondition requires that the selected move is a damaging move */
-  constructor(
-    attackCondition: PokemonAttackCondition = (_user, _target, move) => move.category !== MoveCategory.STATUS,
-    showAbility: boolean = true,
-  ) {
+  constructor(attackOnly: boolean = true, showAbility: boolean = true) {
```
flx-sta:
```suggestion
  constructor(attackMovesOnly: boolean = true, showAbility: boolean = true) {
```

src/data/ab-attrs/post-attack-ab-attr.ts
```diff
@@ -1,56 +1,50 @@
...
     super(showAbility);
 
-    this.attackCondition = attackCondition;
+    this.attackOnly = attackOnly;
```
flx-sta:
```suggestion
    this.attackMovesOnly = attackMovesOnly;
```

src/data/ab-attrs/post-attack-ab-attr.ts
```diff
@@ -1,56 +1,50 @@
...
   /**
-   * This method is only called after {@link applyPostAttack} has already been applied. Use this for handling checks specific to the ability in question.
+   * Called by {@linkcode apply} after a move category check satisfies
+   * the attribute's {@linkcode attackOnly} condition to apply effects.
+   * @param pokemon The {@linkcode Pokemon} with this ability
+   * @param simulated If `true`, suppresses changes to game state
+   * @param defender The {@linkcode Pokemon} attacked by the source
+   * @param move The {@linkcode Move} being used
+   * @param args Additional arguments for subclasses
+   * @returns `true` if effects apply successfully
```
flx-sta, line 41:
```suggestion
   * the attribute's {@linkcode attackOnly} condition to apply effects.
   * @param _pokemon The {@linkcode Pokemon} with this ability
   * @param _simulated If `true`, suppresses changes to game state
   * @param _defender The {@linkcode Pokemon} attacked by the source
   * @param _move The {@linkcode Move} being used
   * @param _args Additional arguments for subclasses
   * @returns `true` if effects apply successfully
```

flx-sta (commented):
Partial Review:
image

src/data/ab-attrs/ab-attr.ts
innerthunder: Would also need a `NONE` option, but yeah, it's possible.

Having the two flags separate is nice for how they're used in `applyAbAttrs`, though.

src/data/ab-attrs/ab-attr.ts
innerthunder: It's not a strict rule in the sense that the params are still highlighted and the links are still functional

Removing the underscores makes the docs a bit neater for subclasses as their `apply` overrides are being made, which is the main point of this and similar comment blocks.

src/data/ab-attrs/damage-boost-ab-attr.ts
```diff
@@ -6,7 +6,7 @@ import { PreAttackAbAttr } from "./pre-attack-ab-attr";
 
 /**
  * Class for abilities that boost the damage of moves
- * For abilities that boost the base power of moves, see VariableMovePowerAbAttr
+ * For abilities that boost the base power of moves, see {@linkcode VariableMovePowerAbAttr}
```
DayKev, line 14: Adding the `@linkcode` will require a tsdoc import. Example:
https://github.com/Despair-Games/poketernity/blob/7539c9f4946fb32387136bd65c0954f17bd6efdd/src/phases/form-change-phase.ts#L1-L5

src/data/ab-attrs/money-ab-attr.ts
```diff
@@ -9,13 +9,7 @@ import { PostBattleAbAttr } from "./post-battle-ab-attr";
  * @see {@linkcode applyPostBattle}
```
DayKev:
```suggestion
```

src/data/ab-attrs/moody-ab-attr.ts
```diff
@@ -5,22 +5,15 @@ import { EFFECTIVE_STATS } from "#enums/stat";
 import { PostTurnAbAttr } from "./post-turn-ab-attr";
 
 /**
- * Attribute used for Moody
+ * Attribute to randomly increase one stat stage by 2 and decrease a different
+ * stat stage by 1. Any stat stage at +6 or -6 is excluded from being increased
+ * or decreased, respectively.
+ * Used for {@link https://bulbapedia.bulbagarden.net/wiki/Moody_(Ability) | Moody}.
  * @extends PostTurnAbAttr
  * @see {@linkcode applyPostTurn}
```
DayKev:
```suggestion
```

src/data/ab-attrs/field-move-power-boost-ab-attr.ts
innerthunder: It's outdated, as `defender` is openly available from `move#calculateBattlePower`, where the subclass `AllyMoveCategoryPowerBoostAbAttr` is applied

src/data/ab-attrs/move-type-change-ab-attr.ts
```diff
@@ -15,14 +15,13 @@ export class MoveTypeChangeAbAttr extends PreAttackAbAttr {
...
 
   // TODO: Decouple this into two attributes (type change / power boost)
-  override applyPreAttack(
+  override apply(
     pokemon: Pokemon,
-    _passive: boolean,
     _simulated: boolean,
-    defender: Pokemon,
+    defender: Pokemon | undefined,
     move: Move,
```
DayKev: Can these be swapped so that `defender` can be made optional?

src/data/ab-attrs/pokemon-type-change-ab-attr.ts
```diff
@@ -17,13 +17,7 @@ import { PreAttackAbAttr } from "./pre-attack-ab-attr";
...
   private moveType: Type;
 
-  override applyPreAttack(
-    pokemon: Pokemon,
-    _passive: boolean,
-    simulated: boolean,
-    _defender: Pokemon,
-    move: Move,
-  ): boolean {
+  override apply(pokemon: Pokemon, simulated: boolean, _defender: Pokemon | undefined, move: Move): boolean {
```
DayKev: Same here, swap `defender` and `move` to make `defender` optional.

src/data/ab-attrs/check-trapped-ab-attr.ts
DayKev, line 1: I wonder if there's even a point to this `AbAttr` when it's only extended by exactly one `AbAttr` (`ArenaTrapAbAttr`)...

src/data/ab-attrs/post-faint-clear-weather-ab-attr.ts
```diff
@@ -13,16 +13,14 @@ import { PostFaintAbAttr } from "./post-faint-ab-attr";
 export class PostFaintClearWeatherAbAttr extends PostFaintAbAttr {
   /**
    * @param pokemon The {@linkcode Pokemon} with the ability
```
DayKev:
```suggestion
```

src/data/ab-attrs/post-faint-clear-weather-ab-attr.ts
```diff
@@ -13,16 +13,14 @@ import { PostFaintAbAttr } from "./post-faint-ab-attr";
...
-   * @param _attacker N/A
-   * @param _move N/A
-   * @param _hitResult N/A
-   * @param _args N/A
+   * @param passive N/A
+   * @param attacker N/A
+   * @param move N/A
+   * @param hitResult N/A
    * @returns Returns true if the weather clears, otherwise false.
    */
```
DayKev:
```suggestion
```

src/data/ab-attrs/post-faint-unsuppressed-weather-form-change-ab-attr.ts
```diff
@@ -14,16 +14,15 @@ export class PostFaintUnsuppressedWeatherFormChangeAbAttr extends PostFaintAbAtt
   /**
    * Triggers {@linkcode Arena.triggerWeatherBasedFormChanges | triggerWeatherBasedFormChanges}
    * when the user of the ability faints
```
DayKev:
```suggestion
```

src/data/ab-attrs/post-faint-unsuppressed-weather-form-change-ab-attr.ts
```diff
@@ -14,16 +14,15 @@ export class PostFaintUnsuppressedWeatherFormChangeAbAttr extends PostFaintAbAtt
...
-   * @param _attacker n/a
-   * @param _move n/a
-   * @param _hitResult n/a
+   * @param pokemon the fainted {@linkcode Pokemon}
+   * @param passive n/a
+   * @param attacker n/a
+   * @param move n/a
+   * @param hitResult n/a
    * @returns whether the form change was triggered
    */
```
DayKev:
```suggestion
```

src/data/ab-attrs/post-summon-form-change-by-weather-ab-attr.ts
```diff
@@ -28,7 +27,7 @@ export class PostSummonFormChangeByWeatherAbAttr extends PostSummonAbAttr {
    * @param passive n/a
    * @returns whether the form change was triggered
    */
```
DayKev:
```suggestion
```
This tsdoc should be removed.

src/data/ab-attrs/post-summon-stat-stage-change-on-arena-ab-attr.ts
```diff
@@ -35,9 +35,9 @@ export class PostSummonStatStageChangeOnArenaAbAttr extends PostSummonStatStageC
    * @param passive Whether the effect is passive
    * @returns Returns `true` if the stat change was applied, otherwise `false`
    */
```
DayKev:
```suggestion
```
rmdoc

src/data/ab-attrs/post-summon-user-field-remove-status-effect-ab-attr.ts
```diff
@@ -25,10 +25,10 @@ export class PostSummonUserFieldRemoveStatusEffectAbAttr extends PostSummonAbAtt
    * Removes supplied status effect from the user's field when user of the ability is summoned.
    *
    * @param pokemon - The Pokémon that triggered the ability.
-   * @param _passive - n/a
+   * @param passive - n/a
    * @returns A boolean or a promise that resolves to a boolean indicating the result of the ability application.
    */
```
DayKev:
```suggestion
```
rmdoc (it still references promises lol)

src/data/ab-attrs/post-summon-weather-suppressed-form-change-ab-attr.ts
```diff
@@ -11,12 +11,11 @@ import { PostSummonAbAttr } from "./post-summon-ab-attr";
 export class PostSummonWeatherSuppressedFormChangeAbAttr extends PostSummonAbAttr {
   /**
    * Triggers {@linkcode globalScene.arena.triggerWeatherBasedFormChangesToNormal | triggerWeatherBasedFormChangesToNormal}
```
DayKev:
```suggestion
```

src/data/ab-attrs/post-summon-weather-suppressed-form-change-ab-attr.ts
```diff
@@ -11,12 +11,11 @@ import { PostSummonAbAttr } from "./post-summon-ab-attr";
...
-   * @param _pokemon the {@linkcode Pokemon} with this ability
-   * @param _passive n/a
-   * @param _args n/a
-   * @returns whether a Pokemon was reverted to its normal form
+   * @param pokemon the {@linkcode Pokemon} with this ability
+   * @param simulated if `true`, suppresses changes to game state
+   * @returns `true` if a Pokemon was reverted to its normal form
    */
```
DayKev:
```suggestion
```

src/data/ab-attrs/pre-attack-ab-attr.ts
```diff
@@ -3,11 +3,19 @@ import type { Pokemon } from "#app/field/pokemon";
...
+   * @param args Any additional parameters for this effect
+   * @returns `true` if effects from this attribute apply successfully
+   */
+  override apply(
     _pokemon: Pokemon,
-    _passive: boolean,
     _simulated: boolean,
-    _defender: Pokemon | null,
+    _defender: Pokemon | undefined,
     _move: Move,
```
DayKev: These should be swapped so `defender` can be optional.

src/data/ab-attrs/pre-defend-ab-attr.ts
```diff
@@ -1,16 +1,22 @@
...
+   * @param move The {@linkcode Move} being used
+   * @param args Additional arguments required for the specific effect
+   * @returns `true` if effects from this attribute successfully apply
+   */
+  override apply(
     _pokemon: Pokemon,
-    _passive: boolean,
     _simulated: boolean,
     _attacker: Pokemon,
     _move: Move | null,
```
DayKev:
```suggestion
    _move?: Move,
```

src/data/ab-attrs/pre-weather-effect-ab-attr.ts
DayKev, line 1: This web of `AbAttr`s is a bit weird...

src/data/ab-attrs/post-set-status-ab-attr.ts
DayKev, line 1: I assume this was removed because only a single `AbAttr` extended it? This isn't the only "base" `AbAttr` that's like this though. Also, [Magic Bounce](https://bulbapedia.bulbagarden.net/wiki/Magic_Bounce_(Ability)) might want to extend this in the future.

DayKev (commented):
image

src/data/ab-attrs/pre-weather-effect-ab-attr.ts
innerthunder: Seems like this attribute was just used to enforce the apply arguments? This web is never applied in full in the same way as other webs, and `SuppressWeatherEffectAbAttr` doesn't have much in common with the other attributes.

src/data/ab-attrs/post-set-status-ab-attr.ts
innerthunder: I considered this, but Magic Bounce and Synchronize together would probably have to apply in a very specific order that would make things hard to generalize under this class.

DayKev: Oh that's true, it would have to apply first... hm.

src/data/ab-attrs/post-damage-force-switch-out-ab-attr.ts
```diff
@@ -0,0 +1,98 @@
...
+   * Checks various conditions based on the moves used by the Pokémon, the opponents' moves, and
+   * the Pokémon's health after damage to determine whether the switch-out should occur.
+   *
+   * @param pokemon The Pokémon that took damage.
+   * @param simulated Whether the ability is being simulated.
+   * @param damage The amount of damage taken by the Pokémon.
+   * @param passive N/A
+   * @param args N/A
+   * @param source The Pokemon that dealt damage
+   * @returns `true` if the switch-out logic was successfully applied
```
flx-sta:
```suggestion
   * @param pokemon The Pokémon that took damage.
   * @param simulated Whether the ability is being simulated.
   * @param damage The amount of damage taken by the Pokémon.
   * @param source The Pokemon that dealt damage
   * @returns `true` if the switch-out logic was successfully applied
```

src/data/ab-attrs/post-damage-force-switch-out-ab-attr.ts
```diff
@@ -0,0 +1,98 @@
...
+        return this.helper.switchOutLogic(pokemon);
+      } else {
+        return false;
+      }
+    } else {
+      return false;
+    }
+  }
+  public getFailedText(_user: Pokemon, target: Pokemon, _move: Move, _cancelled: BooleanHolder): string | null {
+    return this.helper.getFailedText(target);
```
flx-sta: NIT: missing new line

```suggestion
  }
  
  public getFailedText(_user: Pokemon, target: Pokemon, _move: Move, _cancelled: BooleanHolder): string | null {
    return this.helper.getFailedText(target);
```

src/data/ab-attrs/post-damage-force-switch-out-ab-attr.ts
```diff
@@ -0,0 +1,98 @@
...
+          if (!this.helper.getSwitchOutCondition(pokemon, opponent)) {
+            return false;
+          }
+        }
+        return this.helper.switchOutLogic(pokemon);
+      } else {
+        return false;
+      }
+    } else {
+      return false;
```
flx-sta: NIT: typos

```suggestion
  public override apply(pokemon: Pokemon, _simulated: boolean, damage: number, source?: Pokemon): boolean {
    const moveHistory = pokemon.getMoveHistory();
    // Will not activate when the Pokémon's HP is lowered by cutting its own HP
    const forbiddenAttackingMoves = [Moves.BELLY_DRUM, Moves.SUBSTITUTE, Moves.CURSE, Moves.PAIN_SPLIT];
    if (moveHistory.length > 0) {
      const lastMoveUsed = moveHistory[moveHistory.length - 1];
      if (forbiddenAttackingMoves.includes(lastMoveUsed.move)) {
        return false;
      }
    }

    // Dragon Tail and Circle Throw switch out Pokémon before the Ability activates.
    const forbiddenDefendingMoves = [Moves.DRAGON_TAIL, Moves.CIRCLE_THROW];
    if (source) {
      const enemyMoveHistory = source.getMoveHistory();
      if (enemyMoveHistory.length > 0) {
        const enemyLastMoveUsed = enemyMoveHistory[enemyMoveHistory.length - 1];
        // Will not activate if the Pokémon's HP falls below half while it is in the air during Sky Drop.
        if (forbiddenDefendingMoves.includes(enemyLastMoveUsed.move) || pokemon.getTag(BattlerTagType.SKY_DROP)) {
          return false;
          // Will not activate if the Pokémon's HP falls below half by a move affected by Sheer Force.
        } else if (allMoves[enemyLastMoveUsed.move].chance >= 0 && source.hasAbility(Abilities.SHEER_FORCE)) {
          return false;
          // Activate only after the last hit of multistrike moves
        } else if (source.turnData.hitsLeft > 1) {
          return false;
        }
        if (source.turnData.hitCount > 1) {
          damage = pokemon.turnData.damageTaken;
        }
      }
    }

    if (pokemon.hp + damage >= pokemon.getMaxHp() * this.hpRatio) {
      // Activates if it falls below half and recovers back above half from a Shell Bell
      const shellBellHeal = calculateShellBellRecovery(pokemon);
      if (pokemon.hp - shellBellHeal < pokemon.getMaxHp() * this.hpRatio) {
        for (const opponent of pokemon.getOpponents()) {
          if (!this.helper.getSwitchOutCondition(pokemon, opponent)) {
            return false;
          }
        }
        return this.helper.switchOutLogic(pokemon);
      } else {
        return false;
      }
    } else {
      return false;
```

src/data/ab-attrs/post-summon-ab-attr.ts
```diff
@@ -8,12 +8,12 @@ import { AbAttr } from "./ab-attr";
 export class PostSummonAbAttr extends AbAttr {
   /**
    * Applies ability post summon (after switching in)
-   * @param _pokemon {@linkcode Pokemon} with this ability
-   * @param _passive Whether this ability is a passive
-   * @param _args Set of unique arguments needed by this attribute
+   * @param pokemon {@linkcode Pokemon} with this ability
+   * @param passive Whether this ability is a passive
+   * @param args Set of unique arguments needed by this attribute
```
flx-sta:
```suggestion
   * @param pokemon {@linkcode Pokemon} with this ability
   * @param simulated If `true`, suppresses changes to game state
   * @param args Set of unique arguments needed by this attribute
```

src/data/ab-attrs/post-turn-hurt-if-sleeping-ab-attr.ts
```diff
@@ -10,19 +10,12 @@ import { BlockNonDirectDamageAbAttr } from "./block-non-direct-damage-ab-attr";
 import { PostTurnAbAttr } from "./post-turn-ab-attr";
 
 /**
- * Attribute used for abilities (Bad Dreams) that damages the opponents for being asleep
+ * Attribute to damage all sleeping opponents by 1/8 of their max hp at the end of turn.
+ * Used for {@linkcode https://bulbapedia.bulbagarden.net/wiki/Bad_Dreams_(Ability) | Bad Dreams}.
```
flx-sta:
```suggestion
 * Used for {@link https://bulbapedia.bulbagarden.net/wiki/Bad_Dreams_(Ability) | Bad Dreams}.
```

src/data/ab-attrs/post-weather-lapse-ab-attr.ts
```diff
@@ -14,13 +14,15 @@ export class PostWeatherLapseAbAttr extends AbAttr {
...
-    _weather: Weather | null,
-    ..._args: unknown[]
-  ): boolean {
+  /**
+   * Applies an effect after the weather on the field lapses.
+   * @param pokemon The {@linkcode Pokemon} with this ability
+   * @param simulated If `true`, suppresses changes to game state
+   * @param weather The {@linkcode Weather} on the field
+   * @param args
+   * @returns
```
flx-sta: Q: Not sure if I still might misunderstand this
```suggestion
   * @param weather The {@linkcode Weather} on the field
   * @returns
```

src/data/ab-attrs/reduce-burn-damage-ab-attr.ts
```diff
@@ -13,12 +13,12 @@ export class ReduceBurnDamageAbAttr extends AbAttr {
 
   /**
    * Applies the damage reduction
-   * @param _pokemon N/A
-   * @param _passive N/A
+   * @param pokemon N/A
+   * @param passive N/A
```
flx-sta:
```suggestion
   * @param pokemon N/A
   * @param simulated N/A
```

src/data/ab-attrs/reverse-drain-ab-attr.ts
```diff
@@ -18,21 +17,13 @@ export class ReverseDrainAbAttr extends PostDefendAbAttr {
    * Examples include: Absorb, Draining Kiss, Bitter Blade, etc.
    * Also displays a message to show this ability was activated.
    * @param pokemon {@linkcode Pokemon} with this ability
-   * @param _passive N/A
+   * @param passive N/A
```
flx-sta:
```suggestion
   * @param simulated N/A
```

src/data/ab-attrs/stat-multiplier-ab-attr.ts
```diff
@@ -37,27 +37,38 @@ import { StatStageAbAttr } from "./stat-stage-ab-attr";
 +-----------------------+-------+--------+----------------------------------+
 ```
  */
-export class StatMultiplierAbAttr extends StatStageAbAttr {
-  private readonly multiplier: number;
-  private readonly condition?: PokemonAttackCondition;
+export class StatMultiplierAbAttr extends AbAttr {
+  protected stat: BattleStat;
```
flx-sta, line 41: Q: Why not readonly? It doesn't seem like it's changed after the constructor either

src/data/ab-attrs/type-immunity-ab-attr.ts
```diff
@@ -26,15 +26,14 @@ export class TypeImmunityAbAttr extends PreDefendAbAttr {
   /**
    * Applies immunity if this ability grants immunity to the type of the given move.
    * @param pokemon - The defending {@linkcode Pokemon}
-   * @param _passive - N/A
+   * @param passive - N/A
```
flx-sta:
```suggestion
   * @param simulated - N/A
```

src/data/all-abilities.ts
```diff
@@ -668,7 +675,7 @@ export function initAbilities() {
     new Ability(Abilities.TINTED_LENS, 4).attr(
       DamageBoostAbAttr,
       2,
-      (user, target, move) => (target?.getMoveEffectiveness(user!, move) ?? 1) <= 0.5,
+      (user, target, move) => (target?.getMoveEffectiveness(user!, move!) ?? 1) <= 0.5, // TODO: Fix these bangs
```
flx-sta:
```suggestion
      (user, target, move) => {
        if (!user || !target || !move) {
          return false;
        }
        return (target.getMoveEffectiveness(user, move) ?? 1) <= 0.5;
      },
```

src/data/all-abilities.ts
```diff
@@ -1255,7 +1265,7 @@ export function initAbilities() {
     ),
     new Ability(Abilities.NEUROFORCE, 7).attr(
       MovePowerBoostAbAttr,
-      (user, target, move) => (target?.getMoveEffectiveness(user!, move) ?? 1) >= 2,
+      (user, target, move) => (target?.getMoveEffectiveness(user!, move!) ?? 1) >= 2, // TODO: Fix these bangs
```
flx-sta:
```suggestion
      (user, target, move) => {
      if (!user || !target || !move) {
        return false; // If any are missing, return false
      }
      return (target.getMoveEffectiveness(user, move) ?? 1) >= 2;
    },
```

src/data/move-attrs/psycho-shift-effect-attr.ts
flx-sta, line 1: consider adding the 
```ts
const statusToApply: StatusEffect | undefined =
      user.status?.effect ?? (user.hasAbility(Abilities.COMATOSE) ? StatusEffect.SLEEP : undefined);
```

as a helper
e.g.
```ts
private getStatusToApply(user: Pokemon): StatusEffect | undefined {
  return user.status?.effect ?? (user.hasAbility(Abilities.COMATOSE) ? StatusEffect.SLEEP : undefined);
}
```

src/data/move-conditions.ts
```diff
@@ -106,11 +106,10 @@ export const failIfSingleBattle: MoveConditionFunc = (_user, _target, _move) =>
 
 export const failIfDampCondition: MoveConditionFunc = (user, _target, move) => {
   const cancelled = new BooleanHolder(false);
-  globalScene
-    .getField(true)
-    .map((p) =>
-      applyAbAttrs(FieldPreventExplosionLikeAbAttr, p, undefined, cancelled, getPokemonNameWithAffix(user), move.name),
-    );
+  globalScene.getField(true).map((p) =>
+    // TODO: Add simulated support for this effect
```
flx-sta: Q: Which "effect"? The `FieldPreventExplosionLikeAbAttr` or the whole `failIfDampCondition` function?

src/phases/move-effect-phase.ts
```diff
@@ -127,7 +122,7 @@ export class MoveEffectPhase extends HitCheckPhase {
       // Assume single target for multi hit
       applyMoveAttrs(MultiHitAttr, user, this.getFirstTarget(), move, hitCount);
       // If Parental Bond is applicable, add another hit
-      applyPreAttackAbAttrs(AddSecondStrikeAbAttr, user, null, move, false, hitCount, null);
+      applyAbAttrs(AddSecondStrikeAbAttr, user, false, move, targets[0], hitCount);
```
flx-sta, line 125: NIT: use deconstructor to get the first target

```ts
const [firstTarget] = this.getTargets();
// [...]
applyAbAttrs(AddSecondStrikeAbAttr, user, false, move, firstTarget, hitCount);
```

src/data/move-conditions.ts
innerthunder: The whole condition (and move conditions in general) need simulated support since the AI uses them for move scoring

src/phases/move-effect-phase.ts
innerthunder: The full `targets` array is used later on (the loop starting on L210)

src/data/ab-attrs/stat-multiplier-ab-attr.ts
innerthunder: This is modified in `src/data/ab-attrs/low-hp-move-type-attack-multiplier-ab-attr.ts`

src/data/move-conditions.ts
flx-sta: Might wanna move it to the top of the function then as a `@todo`

flx-sta (dismissed):

flx-sta (dismissed):

flx-sta (dismissed):

src/data/ab-attrs/add-second-strike-ab-attr.ts
```diff
@@ -21,20 +22,20 @@ export class AddSecondStrikeAbAttr extends PreAttackAbAttr {
...
    * @param move the {@linkcode Move} used by the ability source
-   * @param args Additional arguments:
-   * - `[0]` the number of strikes this move currently has ({@linkcode NumberHolder})
-   * - `[1]` the damage multiplier for the current strike ({@linkcode NumberHolder})
-   * @returns
+   * @param defender n/a
+   * @param hitCount a {@linkcode NumberHolder} containing the number of strikes this move currently has
+   * @param multiplier a {@linkcode NumberHolder} containing the damage multiplier for the current strike
+   * @returns `true` if the given move is modified by this effect
    */
```
DayKev, line 31:
```suggestion
```
Another override that needs removed docs.

src/data/ab-attrs/allied-field-damage-reduction-ab-attr.ts
```diff
@@ -18,16 +17,13 @@ export class AlliedFieldDamageReductionAbAttr extends PreDefendAbAttr {
 
   /**
    * Handles the damage reduction
-   * @param args
-   * - `[0]` {@linkcode NumberHolder} - The damage being dealt
+   * @param multiplier {@linkcode NumberHolder} - The damage being dealt
    */
```
DayKev:
```suggestion
```
rmdoc

src/data/ab-attrs/move-type-change-ab-attr.ts
```diff
@@ -15,14 +15,13 @@ export class MoveTypeChangeAbAttr extends PreAttackAbAttr {
...
-  override applyPreAttack(
+  override apply(
     pokemon: Pokemon,
-    _passive: boolean,
     _simulated: boolean,
-    defender: Pokemon,
     move: Move,
-    moveType: NumberHolder,
-    power: NumberHolder,
+    defender: Pokemon | undefined,
```
DayKev:
```suggestion
    defender?: Pokemon,
```

src/data/ability.ts
```diff
@@ -167,9 +129,7 @@ export class Ability implements Localizable {
   }
 }
 
-type AbAttrApplyFunc<TAttr extends AbAttr> = (attr: TAttr, passive: boolean) => boolean;
-
-class ForceSwitchOutHelper {
+export class ForceSwitchOutHelper {
```
DayKev, line 131: I assume circular imports prevented this from being removed too?

src/data/ability.ts
```diff
@@ -320,7 +280,7 @@ class ForceSwitchOutHelper {
  * @param pokemon - The Pokémon whose Shell Bell recovery is being calculated.
  * @returns The amount of health recovered by Shell Bell.
  */
-function calculateShellBellRecovery(pokemon: Pokemon): number {
+export function calculateShellBellRecovery(pokemon: Pokemon): number {
```
DayKev: Also here?

src/data/all-abilities.ts
```diff
@@ -597,7 +598,7 @@ export function initAbilities() {
     new Ability(Abilities.DOWNLOAD, 4).attr(DownloadAbAttr),
     new Ability(Abilities.IRON_FIST, 4).attr(
       MovePowerBoostAbAttr,
-      (_user, _target, move) => move.hasFlag(MoveFlags.PUNCHING_MOVE),
+      (_user, _target, move) => !!move && move.hasFlag(MoveFlags.PUNCHING_MOVE),
```
DayKev:
```suggestion
      (_user, _target, move) => !!move?.hasFlag(MoveFlags.PUNCHING_MOVE),
```

src/data/all-abilities.ts
```diff
@@ -665,11 +672,12 @@ export function initAbilities() {
...
-    new Ability(Abilities.TINTED_LENS, 4).attr(
-      DamageBoostAbAttr,
-      2,
-      (user, target, move) => (target?.getMoveEffectiveness(user!, move) ?? 1) <= 0.5,
-    ),
+    new Ability(Abilities.TINTED_LENS, 4).attr(DamageBoostAbAttr, 2, (user, target, move) => {
+      if (!user || !target || !move) {
+        return false;
+      }
+      return (target.getMoveEffectiveness(user, move) ?? 1) <= 0.5;
```
DayKev:
```suggestion
      return target.getMoveEffectiveness(user, move) <= 0.5;
```
This `??` is not be necessary anymore due to the `if (!...)` that was added.

src/data/all-abilities.ts
```diff
@@ -702,7 +710,7 @@ export function initAbilities() {
     new Ability(Abilities.FRISK, 4).attr(FriskAbAttr),
     new Ability(Abilities.RECKLESS, 4).attr(
       MovePowerBoostAbAttr,
-      (_user, _target, move) => move.hasFlag(MoveFlags.RECKLESS_MOVE),
+      (_user, _target, move) => !!move && move.hasFlag(MoveFlags.RECKLESS_MOVE),
```
DayKev:
```suggestion
      (_user, _target, move) => !!move?.hasFlag(MoveFlags.RECKLESS_MOVE),
```

src/data/all-abilities.ts
```diff
@@ -915,14 +923,14 @@ export function initAbilities() {
       .edgeCase(), // Should not boost stats if switching into court changed sticky web
     new Ability(Abilities.STRONG_JAW, 6).attr(
       MovePowerBoostAbAttr,
-      (_user, _target, move) => move.hasFlag(MoveFlags.BITING_MOVE),
+      (_user, _target, move) => !!move && move.hasFlag(MoveFlags.BITING_MOVE),
```
DayKev:
```suggestion
      (_user, _target, move) => !!move?.hasFlag(MoveFlags.BITING_MOVE),
```

src/data/all-abilities.ts
```diff
@@ -941,7 +949,7 @@ export function initAbilities() {
     ),
     new Ability(Abilities.MEGA_LAUNCHER, 6).attr(
       MovePowerBoostAbAttr,
-      (_user, _target, move) => move.hasFlag(MoveFlags.PULSE_MOVE),
+      (_user, _target, move) => !!move && move.hasFlag(MoveFlags.PULSE_MOVE),
```
DayKev:
```suggestion
      (_user, _target, move) => !!move?.hasFlag(MoveFlags.PULSE_MOVE),
```

src/data/all-abilities.ts
```diff
@@ -950,14 +958,14 @@ export function initAbilities() {
     new Ability(Abilities.SYMBIOSIS, 6).unimplemented(),
     new Ability(Abilities.TOUGH_CLAWS, 6).attr(
       MovePowerBoostAbAttr,
-      (_user, _target, move) => move.hasFlag(MoveFlags.MAKES_CONTACT),
+      (user, target, move) => !!user && !!move && move.checkFlag(MoveFlags.MAKES_CONTACT, user, target ?? null),
```
DayKev:
```suggestion
      (user, target, move) => !!user && !!move?.checkFlag(MoveFlags.MAKES_CONTACT, user, target ?? null),
```

src/data/all-abilities.ts
```diff
@@ -1075,8 +1083,11 @@ export function initAbilities() {
       .condition(getSheerForceHitDisableAbCondition()),
     new Ability(Abilities.SLUSH_RUSH, 7).attr(WeatherBasedSpeedDoublerAbAttr, [WeatherType.HAIL, WeatherType.SNOW]),
     new Ability(Abilities.LONG_REACH, 7).attr(IgnoreContactAbAttr),
-    new Ability(Abilities.LIQUID_VOICE, 7).attr(MoveTypeChangeAbAttr, Type.WATER, 1, (_user, _target, move) =>
-      move.hasFlag(MoveFlags.SOUND_BASED),
+    new Ability(Abilities.LIQUID_VOICE, 7).attr(
+      MoveTypeChangeAbAttr,
+      Type.WATER,
+      1,
+      (_user, _target, move) => !!move && move.hasFlag(MoveFlags.SOUND_BASED),
```
DayKev:
```suggestion
      (_user, _target, move) => !!move?.hasFlag(MoveFlags.SOUND_BASED),
```

src/data/all-abilities.ts
```diff
@@ -1250,12 +1261,22 @@ export function initAbilities() {
     ),
     new Ability(Abilities.PRISM_ARMOR, 7).attr(
       ReceivedMoveDamageMultiplierAbAttr,
-      (target, user, move) => target.getMoveEffectiveness(user, move) >= 2,
+      (user, target, move) => {
+        if (!user || !target || !move) {
+          return false; // If any are missing, return false
+        }
+        return (target.getMoveEffectiveness(user, move) ?? 1) >= 2;
```
DayKev:
```suggestion
        return target.getMoveEffectiveness(user, move) >= 2;
```

src/data/all-abilities.ts
```diff
@@ -1250,12 +1261,22 @@ export function initAbilities() {
...
       0.75,
     ),
     new Ability(Abilities.NEUROFORCE, 7).attr(
       MovePowerBoostAbAttr,
-      (user, target, move) => (target?.getMoveEffectiveness(user!, move) ?? 1) >= 2,
+      (user, target, move) => {
+        if (!user || !target || !move) {
+          return false; // If any are missing, return false
+        }
+        return (target.getMoveEffectiveness(user, move) ?? 1) >= 2;
```
DayKev:
```suggestion
        return target.getMoveEffectiveness(user, move) >= 2;
```

src/data/all-abilities.ts
```diff
@@ -1250,12 +1261,22 @@ export function initAbilities() {
...
+      },
       0.75,
     ),
     new Ability(Abilities.NEUROFORCE, 7).attr(
       MovePowerBoostAbAttr,
-      (user, target, move) => (target?.getMoveEffectiveness(user!, move) ?? 1) >= 2,
+      (user, target, move) => {
+        if (!user || !target || !move) {
+          return false; // If any are missing, return false
```
DayKev:
```suggestion
          return false;
```

src/data/all-abilities.ts
```diff
@@ -1250,12 +1261,22 @@ export function initAbilities() {
     ),
     new Ability(Abilities.PRISM_ARMOR, 7).attr(
       ReceivedMoveDamageMultiplierAbAttr,
-      (target, user, move) => target.getMoveEffectiveness(user, move) >= 2,
+      (user, target, move) => {
+        if (!user || !target || !move) {
+          return false; // If any are missing, return false
```
DayKev:
```suggestion
          return false;
```

src/data/all-abilities.ts
```diff
@@ -1301,7 +1322,7 @@ export function initAbilities() {
       6,
     ),
     new Ability(Abilities.PUNK_ROCK, 8)
-      .attr(MovePowerBoostAbAttr, (_user, _target, move) => move.hasFlag(MoveFlags.SOUND_BASED), 1.3)
+      .attr(MovePowerBoostAbAttr, (_user, _target, move) => !!move && move.hasFlag(MoveFlags.SOUND_BASED), 1.3)
```
DayKev:
```suggestion
      .attr(MovePowerBoostAbAttr, (_user, _target, move) => !!move?.hasFlag(MoveFlags.SOUND_BASED), 1.3)
```

src/data/all-abilities.ts
```diff
@@ -1583,7 +1604,7 @@ export function initAbilities() {
     new Ability(Abilities.CUD_CHEW, 9).unimplemented(),
     new Ability(Abilities.SHARPNESS, 9).attr(
       MovePowerBoostAbAttr,
-      (_user, _target, move) => move.hasFlag(MoveFlags.SLICING_MOVE),
+      (_user, _target, move) => !!move && move.hasFlag(MoveFlags.SLICING_MOVE),
```
DayKev:
```suggestion
      (_user, _target, move) => !!move?.hasFlag(MoveFlags.SLICING_MOVE),
```

src/data/battler-tags.ts
```diff
@@ -2921,12 +2920,8 @@ export class MysteryEncounterPostSummonTag extends BattlerTag {
     const ret = super.lapse(pokemon, lapseType);
 
     if (lapseType === BattlerTagLapseType.CUSTOM) {
-      const cancelled = new BooleanHolder(false);
-      applyAbAttrs(ProtectStatAbAttr, pokemon, false, cancelled);
```
DayKev, line 2925: Is this handled elsewhere now?

src/data/move-attrs/force-switch-out-attr.ts
```diff
@@ -15,7 +15,8 @@ import { SwitchSummonPhase } from "#app/phases/switch-summon-phase";
 import { BooleanHolder } from "#app/utils";
 import i18next from "i18next";
 import { ForceSwitchOutImmunityAbAttr } from "#app/data/ab-attrs/force-switch-out-immunity-ab-attr";
-import { PostDamageForceSwitchAbAttr, applyAbAttrs } from "#app/data/ability";
+import { applyAbAttrs } from "#app/data/ability";
+import { PostDamageForceSwitchAbAttr } from "../ab-attrs/post-damage-force-switch-out-ab-attr";
```
DayKev:
```suggestion
import { PostDamageForceSwitchAbAttr } from "#app/data/ab-attrs/post-damage-force-switch-out-ab-attr";
```

src/data/move-attrs/flinch-attr.ts
```diff
@@ -5,7 +5,7 @@ import { ArenaTagType } from "#enums/arena-tag-type";
 import { BattlerTagType } from "#enums/battler-tag-type";
 import { IgnoreMoveEffectsAbAttr } from "../ab-attrs/ignore-move-effect-ab-attr";
 import { MoveEffectChanceMultiplierAbAttr } from "../ab-attrs/move-effect-chance-multiplier-ab-attr";
-import { applyAbAttrs, applyPreDefendAbAttrs } from "../ability";
+import { applyAbAttrs } from "../ability";
 import type { Move } from "../move";
```
DayKev:
```suggestion
import { applyAbAttrs } from "#app/data/ability";
import type { Move } from "#app/data/move";
```

src/data/move-attrs/flinch-attr.ts
```diff
@@ -5,7 +5,7 @@ import { ArenaTagType } from "#enums/arena-tag-type";
 import { BattlerTagType } from "#enums/battler-tag-type";
 import { IgnoreMoveEffectsAbAttr } from "../ab-attrs/ignore-move-effect-ab-attr";
 import { MoveEffectChanceMultiplierAbAttr } from "../ab-attrs/move-effect-chance-multiplier-ab-attr";
```
DayKev:
```suggestion
import { IgnoreMoveEffectsAbAttr } from "#app/data/ab-attrs/ignore-move-effect-ab-attr";
import { MoveEffectChanceMultiplierAbAttr } from "#app/data/ab-attrs/move-effect-chance-multiplier-ab-attr";
```

src/data/move-attrs/remove-held-item-attr.ts
```diff
@@ -32,7 +32,8 @@ export class RemoveHeldItemAttr extends MoveEffectAttr {
     }
 
     const cancelled = new BooleanHolder(false);
-    applyAbAttrs(BlockItemTheftAbAttr, target, false, cancelled); // Check for abilities that block item theft
+    // Check for abilities that block item theft
```
DayKev:
```suggestion
```

src/data/move-attrs/secret-power-attr.ts
```diff
@@ -13,7 +13,7 @@ import { StatusEffectAttr } from "#app/data/move-attrs/status-effect-attr";
 import { NumberHolder } from "#app/utils";
 import { IgnoreMoveEffectsAbAttr } from "../ab-attrs/ignore-move-effect-ab-attr";
 import { MoveEffectChanceMultiplierAbAttr } from "../ab-attrs/move-effect-chance-multiplier-ab-attr";
-import { applyAbAttrs, applyPreDefendAbAttrs } from "../ability";
+import { applyAbAttrs } from "../ability";
```
DayKev:
```suggestion
import { applyAbAttrs } from "#app/data/ability";
```

src/data/move-attrs/secret-power-attr.ts
```diff
@@ -13,7 +13,7 @@ import { StatusEffectAttr } from "#app/data/move-attrs/status-effect-attr";
 import { NumberHolder } from "#app/utils";
 import { IgnoreMoveEffectsAbAttr } from "../ab-attrs/ignore-move-effect-ab-attr";
 import { MoveEffectChanceMultiplierAbAttr } from "../ab-attrs/move-effect-chance-multiplier-ab-attr";
```
DayKev:
```suggestion
import { IgnoreMoveEffectsAbAttr } from "#app/data/ab-attrs/ignore-move-effect-ab-attr";
import { MoveEffectChanceMultiplierAbAttr } from "#app/data/ab-attrs/move-effect-chance-multiplier-ab-attr";
```

DayKev (commented):
Went through all the files finally.

src/data/ab-attrs/add-second-strike-ab-attr.ts
innerthunder: This one has different args (`hitCount` and `multiplier`) from `PreAttackAbAttr` on its own, so I feel like this should stay

src/data/ability.ts
innerthunder: Yeah, probably because this is pushing phases

src/data/ab-attrs/add-second-strike-ab-attr.ts
DayKev: Oh I missed that.

src/data/battler-tags.ts
innerthunder: No, but I'm pretty sure this is irrelevant. This tag seems to only be used for self-buffs, and even if any are self-debuffs they wouldn't activate Clear Body, etc.

flx-sta (approved):
<sub>Can't wait to see the after-fix PR from pigeon raywheeze</sub>

flx-sta (approved):

Tempo-anon (approved):
227 files changed all tests pass safetycat

PR #230, opened 2025-01-09
[Misc] Replace `!!` with `> 0` check in `MovePhase.canMove`

DayKev:
N/A

Because checking that something is nonzero should be done with an actual numerical comparator and not `!!`.

`!!this.targets.length` replaced with `this.targets.length > 0`.

## Checklist


innerthunder (approved):

Tempo-anon (approved):
There's a lot of places in the codebase where !variable or !!variable is used when it would make more sense to use `> 0`

torranx (approved):

PR #238, opened 2025-01-09
[UI] Refactor Option Select and Confirm UI handlers

MokaStitcher:

Most of it is refactoring, but some UI issues got fixed thanks to it / as a proof of concept of the changes
- The up and down arrows in the menus have been replaced by a scroll bar
- Some UI elements that needed it now get placed properly based on other element's sizes
   - Text Window in main menu
   - Community and Manage Data sub menus
   - Pokemon Info Container slides to proper distance when offering to add a caught Pokémon to the team
- Save & Quit option no longer appears in the menu outside of battle

<details><summary>Context: What's OptionSelectUiHandler and ConfirmUiHandler</summary>

- `OptionSelectUiHandler` takes care of all the little windows with several options to choose from. It displays the options, resizes automatically based on their size, handles input processing, wrap around, scrolling, input delay, etc.
image
- `ConfirmUiHandler` extends `OptionSelectUiHandler` but only ever shows Yes and No options
</details>

Many UIs have a similar menu but don't actually make use of this handler (examples: main menu, pokeball menu, egg gacha, learn move menu, transfer item menu, ...). This results in a lot of duplicated code and some of these UIs not having all the benefits of fixes or updates to the OptionSelectUiHandler.

<details><summary>Current issues with the handler itself:</summary>

- A few things would prevent some of these UIs from being able to use the handler:
   - It currently does not allow options to use different text colors, which the move learning menu and transfer menu need
   - It does not extend MessageUiHandler, preventing handlers that extend it from showing text boxes easily
- The handler itself has a few bugs and its code is quite inefficient. For example it computes its expected size unnecessarily every time the cursor gets moved. This is causing all UIs that copy and pasted its code to be very slow when there are a lot of elements (ie, transfer menu in late endless)
- `ConfirmUiHandler` extends `OptionSelectUiHandler` but unlike it it doesn't use a custom interface to take its parameters, making using it very cumbersome.
- `ConfirmUiHandler` also has specific code for the Summary/Yes/No menu when catching a Pokemon with a full team which is messy

</details>

This PR aims to address those issues so that in the future the UIs who should use it can do it more easily.
It does *not* update those UIs except the main menu as a proof of concept

#### The `AbstractOptionSelectUIHandler` has pretty much been fully rewritten, but here are the key changes:
- Move its configuration interface `OptionSelectModeConfig` to its own file
- Internally, each `OptionSelectItem` from the configuration is mapped to a `UIOptionSelectItem` that will keep track of any information needed for it to be displayed later, to help with optimization (more details below)
- Make the class generic, to work with any subclass of `OptionSelectItem`. This is not needed currently, but will be very useful to write children classes that do a bit more complex things like the transfer item menu. 
<details><summary>I made a prototype of the transfer menu with this (not included in this PR)</summary>

It uses its own subclass of `OptionSelectItem` to keep track of the current and max stack of items for each element and thanks to the generic class it can access both these attributes and the `UIOptionSelectItem` specific attributes from the `AbstractOptionSelectUIHandler`. Which allows it to easily show the max stack items in red and to update the labels when left and right are pressed.

https://github.com/user-attachments/assets/dd276de7-68ba-45b6-bc71-81716a07aaf3

</details>

- Make the handler extends `MessageUiHandler`, allowing children classes to display messages outside of battle if needed
- Make the TextObject used to display the menu items a BBCodeTextObject
- Remove the ↑ and ↓ arrows and replace them with a `ScrollBar` (the scrollbar look will be updated at some point don't worry)
- The `OptionSelectUiHandler` is no longer placed 48pixels up from the bottom by default. All configs were updated accordingly. This default offset was kept for the `ConfirmUiHandler` however.

<details><summary>OptionSelectUIHandler optimization details:</summary>

- The handler keeps a rolling array of Sprite object for the icons, to avoid sprites getting constantly destroyed and recreated when scrolling through a menu with icons
- At the menu creation only the first 15 elements menu are considered when computing the window's size. Before it was all options, which could make things quite heavy.
- How computing the elements size works:
  - Internally, each element from the configuration is mapped to a `UIOptionSelectItem` that will keep track of any information needed for it to be displayed later
  - If the element has an icon, make a temporary sprite object and measure its size. Since the OptionSelectUiHandler only uses a single TextObject to display all items, separated by new lines, the space needed for the icon needs to be taken into account for each individual option. For this, the element's label is padded with as many spaces as needed at the start to fit the icon. The updated label and needed space for the icon is kept in the `UIOptionSelectItem.displayLabel`, so that this only needs to be done once and the original provided label is left untouched.
  - If the element asks for a specific color, the `displayLabel` is updated with the BBCode for this color
  - That text is put in a temporary text object and its width measured
  - Finally, the it is marked as `initialized` and won't need be checked like this again
  - The menu's size is adapted if the maximum width of all measured options is bigger than its current width
- Whenever the menu is scrolled, it checks for any non initialized elements, and adapts the size for them as needed.

</details>

#### Changes in `OptionSelectModeConfig`
- Added an attribute `canBypassInputDelay` to compliment the `inputDelay`. With this, it is now possible to say whether one can cancel out of a menu while it is blocked due to a provided `inputDelay`
   - `canBypassInputDelay: true` has been added for example to the erase save confirmation, so that if one clicks the wrong save they don't need to wait for the delay to expire before exiting out
   - This also helps to improve the Egg Skip prompt. In this case we don't want people to cancel out of the delay on accident. Before, to prevent this, the `blockCancelButton` flag had to be set, which would also prevent using the cancel button to say "No" altogether, even after the delay had passed. Now, by default it is not possible to cancel the input delay, but once the delay is done it is possible to press the cancel button to select the "No" option.
- Added a `color` attribute that allows an element to use a different TextStyle from the default one
- Added an `onResize` callback. It gets called whenever the menu resizes itself, allowing other elements in the UI to adapt to the proper size. So far this is being used by:
   - The `MenuUiHandler`, to properly size its message box and place its sub menus
   - The `AttemptCapturePhase` (and duplicate code in ME utils), to offset the Pokemon Info Container by the right amount when offering to add a new Pokemon to the team

#### ConfirmUIHandler:
- Remove the summary-specific code from the handler. That code now uses the normal Option Select mode
- Define an interface for the mode's configuration
- Update all instances of `setMode(Mode.CONFIRM, ..)` in the game to use the interface

#### MenuUiHandler:
- As a test of the changes the `MenuUiHandler` has been rewritten to extend `OptionSelectUiHandler`
- It now place its elements properly based on the menu's size
- It will now be able to accommodate more than 10 options if needed
- Cleaned up some duplicate code relative to excluded options
- Added `Save & Quit` to the excluded options when not in game
- Partially removed the `Test Dialogue` specific code. I did not remove entirely just yet so that we can make sure that the AutoCompleteUiHandler still works as before

#### Others:
- **ui/text.ts**: Updated a few BBCode related helper functions, and removed uiTheme as an argument from them, as it should always use the theme from the settings by default
- **Tests**: Added mocking for the BBCodeTextObject and InputText objects similar to the one for the TextObject, to prevent the tests from failing due to the TextObject > BBCodeTextObject change

## Screenshots/Videos

<details><summary>Menu with scrolling:</summary>

Before:

https://github.com/user-attachments/assets/26eebcad-8ae9-419f-b750-ad1a7dc505ac

After:

https://github.com/user-attachments/assets/2287b46a-54ea-418d-8563-15880b6fded3

</details>

<details><summary>Menu element placement:</summary>

Before:
image
image
image

After:
image
image
image

</details>

<details><summary>Summary/Yes/No menu placement:</summary>

Before:
image
image

After:
image
image

</details>

<details><summary>Couple legacy mode pics:</summary>

image
image

</details>

## How to test the changes?

You'll have to run the game and try to see as different option menus as possible, and make sure their placement and behavior is still correct.

## Checklist

- [ ] The PR is self-contained and cannot be split into smaller PRs? The MenuUiHandler changes could be moved to another PR if needed
- [X] Are all unit tests still passing? (`npm run test`)
  - [X] Have I made sure that any UI change works for both UI themes (default and legacy)?

~~Are there any localization additions or changes? If so:~~


src/data/mystery-encounters/utils/encounter-phase-utils.ts
```diff
@@ -547,6 +547,7 @@ export function selectPokemonForOption(
                 const fullOptions = secondaryOptions
                   .map((option) => {
                     // Update handler to resolve promise
+                    // TODO: don't update the handler like this
```
flx-sta, line 550: Q: Was this supposed to be resolved or is this supposed to stay after your PR?

src/system/game-data.ts
```diff
@@ -1366,53 +1367,51 @@ export class GameData {
             );
           }
 
+          // TODO: move this outside of game data
```
flx-sta, line 1370: Q: TODO still relevant?

src/ui/abstract-option-select-ui-handler.ts
```diff
@@ -0,0 +1,465 @@
...
+ *  - Creating the window of the appropriate size to hold all elements.
+ *  - Handle scrolling through the menu items, resizing as needed.
+ *  - Handle selecting the menu items, or cancelling out of the menu.
+ *
+ * At initialization the size of the first {@linkcode NUM_PRE_COMPUTED_OPTIONS} is measured.
+ * Then the window's size is updated as needed when a non initialized option needs to be displayed.
+ *
+ * @template T the specifc type of {@linkcode OptionSelectItem} that this handler displays
+ */
+export default abstract class AsbtractOptionSelectUiHandler<T extends OptionSelectItem> extends MessageUiHandler {
```
flx-sta: Typo
```suggestion
export default abstract class AbstractOptionSelectUiHandler<T extends OptionSelectItem> extends MessageUiHandler {
```
This might affect other files too

src/ui/abstract-option-select-ui-handler.ts
```diff
@@ -0,0 +1,465 @@
...
+  private config: OptionSelectModeConfig<T> | null;
+  private options: (UIOptionSelectItem & T)[];
+  private maxOptions: number;
+  protected fullyInitialized: boolean;
+
+  protected optionSelectContainer: Phaser.GameObjects.Container;
+  protected optionSelectBg: Phaser.GameObjects.NineSlice;
+  protected optionSelectText: BBCodeText;
+  protected optionSelectIcons: Phaser.GameObjects.Sprite[];
```
flx-sta, line 44: NIT: Consider initialization with empty array `[]`

src/ui/abstract-option-select-ui-handler.ts
flx-sta, line 1: We should consider adding a `Scrollable` interface that can be added as `implements` to identify ui-windows/component  that can be scrolled

```ts
interface Scrollable {
  scrollCursor: number;
  scrollBar: ScrollBar | null;
  // maybe other stuff
}
```

src/ui/autocomplete-ui-handler.ts
```diff
@@ -1,36 +1,22 @@
...
-
-  getWindowWidth(): number {
-    return 64;
-  }
+export default class AutoCompleteUiHandler extends OptionSelectUiHandler {
+  private modalContainer: Phaser.GameObjects.Container;
 
   override show(args: any[]): boolean {
-    if (args[0].modalContainer) {
+    if (args[0].hasOwnProperty("modalContainer")) {
```
flx-sta: NIT: please try to avoid using `hasOwnProperty()`.
TS compiler doesn't like it

```suggestion
    if (args[0]?.modalContainer) {
```

src/ui/battle-message-ui-handler.ts
```diff
@@ -11,6 +11,8 @@ import { PERMANENT_STATS, getStatKey } from "#enums/stat";
 import { settings } from "#app/system/settings/settings-manager";
 
 export default class BattleMessageUiHandler extends MessageUiHandler {
+  private readonly wordWrapWidth: number = 1780;
```
flx-sta, line 14: NIT: can be static, doesn't have to be initialized for every instance of the UiHandler (even though it's probably just one)

src/ui/confirm-ui-handler.ts
```diff
@@ -1,123 +1,46 @@
...
-
-      super.show([config]);
-
-      this.switchCheck = args.length >= 5 && args[4] !== null && (args[4] as boolean);
-
-      const xOffset = args.length >= 6 && args[5] !== null ? (args[5] as number) : 0;
-      const yOffset = args.length >= 7 && args[6] !== null ? (args[6] as number) : 0;
-
-      this.optionSelectContainer.setPosition(globalScene.game.canvas.width / 6 - 1 + xOffset, -48 + yOffset);
+    if (!args[0] || !args[0].hasOwnProperty("yesHandler") || !args[0].hasOwnProperty("noHandler")) {
```
flx-sta: Please avoid using `hasOwnProperty()` the ts compiler can't handle it well
```suggestion
    if (!args[0] || !args[0].yesHandler || !args[0].noHandler) {
```

src/ui/interfaces/confirm-menu-config.ts
```diff
@@ -0,0 +1,15 @@
+// -- start tsdoc imports --
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import type { Mode } from "#app/ui/ui";
+// -- end tsdoc imports --
+import type { OptionMenuSettings } from "#app/ui/interfaces/option-select-config";
+
+/**
+ * Customizations options for UI's {@linkcode Mode.CONFIRM}
```
flx-sta, line 8: Q: Does this link even work? for me they never did

src/ui/interfaces/option-select-config.ts
```diff
@@ -0,0 +1,83 @@
...
+  readonly yOffset?: number;
+  /**
+   * Set to `true` to prevent closing the menu with the cancel button.
+   * Otherwise using the cancel button will act as if the last option was selected.
+   */
+  readonly noCancel?: boolean;
+  /** Optional delay (in ms) before the player is allowed to make a selection. */
+  readonly inputDelay?: number;
+  /** Set to `true` to allow bypassing the inputDelay with the cancel button. */
+  readonly canCancelDelay?: boolean;
```
flx-sta: NIT: 
```suggestion
  readonly canBypassInputDelay?: boolean;
```

src/ui/interfaces/option-select-config.ts
```diff
@@ -0,0 +1,83 @@
...
+  readonly maxOptions?: number;
+  /** Horizontal offset for the window compared to the default (right of screen) */
+  readonly xOffset?: number;
+  /** Vertical offset for the window compared to the default (bottom of screen) */
+  readonly yOffset?: number;
+  /**
+   * Set to `true` to prevent closing the menu with the cancel button.
+   * Otherwise using the cancel button will act as if the last option was selected.
+   */
+  readonly noCancel?: boolean;
```
flx-sta: This one is a bit confusing to me, even with the description. SO will it basically not present a `cancel` option? or does the cancel option simply not close the window? huh!?

src/ui/interfaces/option-select-config.ts
```diff
@@ -0,0 +1,83 @@
...
+   * Handler called when that option is selected.
+   * @returns `true` to play the "success" sfx, `false` for the "error" sfx
+   */
+  handler: () => boolean;
+  /** Optional handler for when the cursor is moved to that option. */
+  readonly onHover?: () => void;
+  /** Set to `true` to keep the menu open after this option was selected. */
+  readonly keepOpen?: boolean;
+  /** Set to `true` to prevent the default menu sound effects from playing. */
+  readonly overrideSound?: boolean;
```
flx-sta: huh? A boolean? What is it overwritten with then?

src/ui/login-form-ui-handler.ts
```diff
@@ -233,6 +233,8 @@ export default class LoginFormUiHandler extends FormModalUiHandler {
         globalScene.ui.setOverlayMode(Mode.OPTION_SELECT, {
           options: options,
           delay: 1000,
+          xOffset: globalScene.scaledCanvas.width,
+          yOffset: globalScene.scaledCanvas.height - this.usernameInfoImage.displayHeight - 16 * dataKeys.length - 22,
```
flx-sta: NIT: might be worth to destructure `globalScene`

```suggestion
        const { ui, scaledCanvas: { width, height } } = globalScene;
       ui.setOverlayMode(Mode.OPTION_SELECT, {
          options: options,
          delay: 1000,
          xOffset: width,
          yOffset: height - this.usernameInfoImage.displayHeight - 16 * dataKeys.length - 22,
```

src/ui/menu-ui-handler.ts
```diff
@@ -35,172 +35,154 @@ enum MenuOptions {
...
     ui.bgmBar = this.bgmBar;
 
-    this.menuContainer = globalScene.add.container(1, -(globalScene.game.canvas.height / 6) + 1);
-    this.menuContainer.setName("menu");
-    this.menuContainer.setInteractive(
-      new Phaser.Geom.Rectangle(0, 0, globalScene.game.canvas.width / 6, globalScene.game.canvas.height / 6),
-      Phaser.Geom.Rectangle.Contains,
-    );
+    const width = globalScene.scaledCanvas.width;
+    const height = globalScene.scaledCanvas.height;
```
flx-sta: NIT: 
```suggestion
    const { width, height } = globalScene.scaledCanvas;
```

src/ui/menu-ui-handler.ts
```diff
@@ -322,6 +310,8 @@ export default class MenuUiHandler extends MessageUiHandler {
         keepOpen: true,
       },
     );
+
+    // TODO: fully remove test dialogue option and related handlers
```
flx-sta, line 313: Q: Still relevant TODO?

flx-sta (commented):
image

----

When the Money is close to the purchase price I can't purchase the pokemon. Idk if that has to do with this ticket
<details>
   <summary> [Screenshot 🇩🇪] </summary>

image
</details>

MokaStitcher:
> When the Money is close to the purchase price I can't purchase the pokemon. Idk if that has to do with this ticket

Nah, that's because that ME is not supposed to spawn when we don't have a certain amount of money for it (something like 2-3 times the wave's money). And the same check is used to disable the option, instead of checking that we have the required amount to make the actual purchase. Doesn't come up in actual gameplay, that's fully because of the override forcing the ME even without fulfilling the requirements lol

src/ui/abstract-option-select-ui-handler.ts
MokaStitcher: Yeah. There is a component to handle scrollable grids already that could be used for this too with some updated/improvements. Lots of the UI code should be rewritten in regards to scrolling indeed, but not for this PR

src/ui/interfaces/option-select-config.ts
MokaStitcher: Nothing, that's how it was named before, does `noSoundEffects` seem fine? Feel free to suggest a better name

src/ui/interfaces/option-select-config.ts
MokaStitcher: Default behavior in the option select menu: pressing the cancel *button* acts as if you had moved the cursor to the last option in the menu, and pressed the action button.
-> For example in a Yes/No menu, pressing the cancel button selects "No" without having to press down + action
-> For the mode selection menu, pressing the cancel button selects "Cancel"
But this only works if the menu has a cancel option at the end. If it doesn't, for example in the main menu, we need to block that behavior or pressing the cancel button would select Save & Quit or whatever. Even for some Yes/No menus, it doesn't make sense for that behavior to happen, so we need a way to prevent it.

I attempted to make things clearer in the comment + name, let me know

src/ui/menu-ui-handler.ts
MokaStitcher: yes, this is not for this PR

src/ui/interfaces/confirm-menu-config.ts
MokaStitcher: Yes, thanks to the import at the top of the file in VSCode I can click on it and it brings me to it 
image
image

src/ui/battle-message-ui-handler.ts
MokaStitcher: Fair. Each ui handler is only ever instanciated once though, so for this PR I'll leave it at that.
The whole message system needs to be heavily overhauled

src/ui/abstract-option-select-ui-handler.ts
MokaStitcher: omg thank you... That handler is cursed, before it was the filename which had a typo

src/system/game-data.ts
MokaStitcher: yes, this is still UI stuff happenning in game data which I don't like. Not related to this PR though

src/data/mystery-encounters/utils/encounter-phase-utils.ts
MokaStitcher: It's just something I noticed and didn't like but not related to this PR

src/ui/abstract-option-select-ui-handler.ts
MokaStitcher: moved it to the constructor instead of setup(), even if in the case of UiHandlers both are called at the same time I'm pretty sure. Is that what you meant or do you want it done on this line specifically?

flx-sta: No, that's fine

src/ui/interfaces/option-select-config.ts
flx-sta: Oooooh... So `cancel` is in regards to the button (e.g. `B` on XBOX cursors). aaaah

src/ui/interfaces/option-select-config.ts
flx-sta: yeah sounds good!
- `noSoundEffects`

Alternative suggestions 
- `disableSoundEffects`
- `disableSound`
- `suppressSound`
- `suppressSoundEffects`

All works

src/ui/interfaces/option-select-config.ts
```diff
@@ -0,0 +1,80 @@
...
+import type { Mode } from "#app/ui/ui";
+// -- end tsdoc imports --
+import type { TextStyle } from "#app/ui/text";
+
+/**
+ * Customizations options for UI's {@linkcode Mode.OPTION_SELECT}
+ *
+ * @template T the specifc type of {@linkcode OptionSelectItem} contained by this config
+ */
+export interface OptionSelectModeConfig<T extends OptionSelectItem> extends OptionMenuSettings {
```
flx-sta: NIT: Considering that in 90% of the cases it's being used like this:
```ts
const config: OptionSelectModeConfig<OptionSelectItem> = // [...]
```

It might be worth to set this as default/omitting it in that case:

```suggestion
export interface OptionSelectModeConfig<T extends OptionSelectItem = OptionSelectItem> extends OptionMenuSettings {
```

---->

```ts
const config: OptionSelectModeConfig = // [...]
```

MokaStitcher: ooh, didn't know you could do that, thanks! Having to add it every time was annoying and I almost considered defining a type just to avoid it, this is much better. Done!

frutescens (approved):

flx-sta (approved):

PR #242, opened 2025-01-10
[Feature] Add starter code for badges

Tempo-anon:

None

Adding starting code for badges

* Created a `Badge` class
Please let me know if you think there are any other variables we should or should not be adding in there
* Created `all-badges.ts`
Should this go in the balance folder? The balance folder was mostly for the balance team to make it easier on them but there's always been some confusion to me as to what goes in there and what doesn't
* Created `badgeType` enum
These are aligned so each enum lines up with the gym leader that gives them out

## Screenshots/Videos

n/a

## How to test the changes?

`n/a`

## Checklist

- [ ] Are all unit tests still passing? (`npm run test`)

### DIFF
--- a/src/data/all-badges.ts
+++ b/src/data/all-badges.ts
@@ -1,0 +1,7 @@
+import type { Badge } from "./badge";
+
+export const allBadges: Badge[] = [];
+
+export function initBadges() {
+  // TODO: initialize all badges here
+}
--- a/src/data/all-badges.ts
+++ b/src/data/all-badges.ts
@@ -1,7 +1,10 @@
-import type { Badge } from "./badge";
+import { BadgeType } from "#enums/badge-type";
+import { Type } from "#enums/type";
+import { Badge } from "./badge";
 
 export const allBadges: Badge[] = [];
 
 export function initBadges() {
   // TODO: initialize all badges here
+  allBadges.push(new Badge(BadgeType.BOULDER, "BoulderBadge", "TBD", 1, Type.ROCK));
 }
--- a/src/data/badge.ts
+++ b/src/data/badge.ts
@@ -1,0 +1,28 @@
+import type { Localizable } from "#app/interfaces/locales";
+import type { BadgeType } from "#enums/badge-type";
+import type { Type } from "#enums/type";
+
+/**
+ * Badges are key items that are given to the player and affect the run
+ *
+ * @param id the corresponding BadgeType enum
+ * @param name the name of the badge
+ * @param effect the description of what the badge does
+ * @param generation which generation the badge originated
+ * @param type the pokemon type associated with the badge
+ * @param powerLevel a way to tweak how effective a badge is
+ */
+export abstract class Badge implements Localizable {
+  public id: BadgeType;
+  public name: string;
+  public effect: string;
+  public generation: number;
+  public type: Type;
+  public powerLevel: number;
+
+  localize(): void {
+    // TODO: FILL IN
+    this.name = "";
+    this.effect = "";
+  }
+}
--- a/src/data/badge.ts
+++ b/src/data/badge.ts
@@ -1,28 +1,39 @@
-import type { Localizable } from "#app/interfaces/locales";
+import type { BadgeGrade } from "#enums/badge-grades";
 import type { BadgeType } from "#enums/badge-type";
 import type { Type } from "#enums/type";
+import i18next from "i18next";
 
 /**
  * Badges are key items that are given to the player and affect the run
  *
  * @param id the corresponding BadgeType enum
- * @param name the name of the badge
- * @param effect the description of what the badge does
+ * @param localeKey the locale key of the badge
+ * @param iconImage the string representation of the image
  * @param generation which generation the badge originated
  * @param type the pokemon type associated with the badge
- * @param powerLevel a way to tweak how effective a badge is
+ * @param grade a way to tweak how effective a badge is
  */
-export abstract class Badge implements Localizable {
+export class Badge {
   public id: BadgeType;
-  public name: string;
-  public effect: string;
+  public localeKey: string;
+  public iconImage: string;
   public generation: number;
   public type: Type;
-  public powerLevel: number;
+  public grade: BadgeGrade;
 
-  localize(): void {
-    // TODO: FILL IN
-    this.name = "";
-    this.effect = "";
+  constructor(id: BadgeType, localeKey: string, iconImage: string, generation: number, type: Type) {
+    this.id = id;
+    this.localeKey = localeKey;
+    this.iconImage = iconImage;
+    this.generation = generation;
+    this.type = type;
+  }
+
+  getName(): string {
+    return i18next.t(`${this.localeKey}.name` as any);
+  }
+
+  getDescription(): string {
+    return i18next.t(`${this.localeKey}.description` as any);
   }
 }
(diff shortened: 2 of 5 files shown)


gerolau:
> Should this go in the balance folder?

Do you want to _keep_ the balance folder?

src/enums/badge-type.ts
```diff
@@ -0,0 +1,93 @@
...
+  FAIRY_OPAL,
+  FAIRY_BEDE,
+  ROCK,
+  ICE,
+  DARK_PIERS,
+  DARK_MARNIE,
+  DRAGON,
+
+  // Paldea
+  CONTONDO,
```
innerthunder:
```suggestion
  CORTONDO,
```

src/data/badge.ts
```diff
@@ -0,0 +1,28 @@
...
+ * @param type the pokemon type associated with the badge
+ * @param powerLevel a way to tweak how effective a badge is
+ */
+export abstract class Badge implements Localizable {
+  public id: BadgeType;
+  public name: string;
+  public effect: string;
+  public generation: number;
+  public type: Type;
+  public powerLevel: number;
```
innerthunder: Maybe make this a `grade` enum?

innerthunder:
The Galar badge names :pikastare:... Maybe we should use custom names for some badges

gerolau:
> The Galar badge names :pikastare:... Maybe we should use custom names for some badges

if it adds some variety, we can probably pull some from the town names as for at least a few of the gyms the town is also a pun off of their gym
Milo - Turf Badge
Nessa - Bow Badge (swimmer pun)
Kabu - Engine Badge
Stow-on-Side is called Lateral Town in japanese, I can see interpretations of this reading for Bea or Allister, so i think making a conditional badge named the same thing for each of them is fine here (so like LATERAL_BEA and LATERAL_ALLISTER like the trio badge)
Opal/Bede - Arabesque Badge (admittedly i did have too google what this one meant, but its a decorative style used on pottery and other art forms; i do believe its canonical that Ballonlea is very artistry inspired)
Piers/Marnie - Spike Badge (a bit questionable, but if were taking it the _evil_ type route spikes could be considered an unfair battling advantage. also, punk city.)

The main issue with this concept comes with melony and gordie, because circhester is Circus town and besides a slight nudging over mr. rimes existance this makes minimal sense for melony and no sense at all for gordie. Additionally, Hammerlockes name in japanese is Knuckle City, which not only doesnt fit Raihans team but the knuckle badge already exists. These would need custom names if we followed this

or, we can leave them the lame canon names, it is the crowds digression pepezen

src/data/badge.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Badges are key items that are given to the player and affect the run
+ *
+ * @param id the corresponding BadgeType enum
+ * @param name the name of the badge
+ * @param effect the description of what the badge does
+ * @param generation which generation the badge originated
+ * @param type the pokemon type associated with the badge
+ * @param powerLevel a way to tweak how effective a badge is
+ */
+export abstract class Badge implements Localizable {
```
flx-sta: Please don't implement `Localizable` anymore. It's an anti pattern!

```suggestion
export abstract class Badge {
```

flx-sta (changes requested):

src/data/badge.ts
flx-sta: I'm not entirely sure how this would be replaced for the existing ones though.

innerthunder (commented):
If we want to have parity between this and moves/abilities, `Badge` shouldn't be abstract.

src/data/badge.ts
```diff
@@ -0,0 +1,28 @@
+import type { Localizable } from "#app/interfaces/locales";
+import type { BadgeType } from "#enums/badge-type";
+import type { Type } from "#enums/type";
+
+/**
+ * Badges are key items that are given to the player and affect the run
+ *
+ * @param id the corresponding BadgeType enum
+ * @param name the name of the badge
+ * @param effect the description of what the badge does
```
CodeTappert:
```suggestion
 * @param effectDescription the description of what the badge does
```

src/data/badge.ts
```diff
@@ -0,0 +1,28 @@
...
+ * @param generation which generation the badge originated
+ * @param type the pokemon type associated with the badge
+ * @param powerLevel a way to tweak how effective a badge is
+ */
+export abstract class Badge implements Localizable {
+  public id: BadgeType;
+  public name: string;
+  public effect: string;
```
CodeTappert:
```suggestion
  public effectDescription: string;
```

innerthunder:
Related discussion on badge names: #246

src/data/all-badges.ts
```diff
@@ -1,7 +1,10 @@
...
+import { BadgeType } from "#enums/badge-type";
+import { Type } from "#enums/type";
+import { Badge } from "./badge";
 
 export const allBadges: Badge[] = [];
 
 export function initBadges() {
   // TODO: initialize all badges here
+  allBadges.push(new Badge(BadgeType.BOULDER, "BoulderBadge", "TBD", 1, Type.ROCK));
 }
```
flx-sta, line 10: DO we really want to follow this `init` pattern again?
Wouldn't it make more sense to initialize them as soon as they are needed?
Piece by Piece?

innerthunder: Yeah, I think it'd make more sense to load in badges individually and make a subclass for each badge type

(maybe `Badge` should be abstract after all lol)

src/data/badge.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  constructor(id: BadgeType, localeKey: string, iconImage: string, generation: number, type: Type) {
+    this.id = id;
+    this.localeKey = localeKey;
+    this.iconImage = iconImage;
+    this.generation = generation;
+    this.type = type;
+  }
+
+  getName(): string {
```
innerthunder, line 32:
```suggestion
  public get name(): string {
```

i.e. you can use `Badge.name` to get a localized name for the badge.

src/data/badge.ts
```diff
@@ -0,0 +1,39 @@
...
+    return i18next.t(`${this.localeKey}.name` as any);
+  }
+
+  getDescription(): string {
```
innerthunder, line 36:
```suggestion
  public get description(): string {
```
We might also want to consider `effectDescription` for clarity

innerthunder (commented):
Should there be a `trainer` field linking the badge to its respective Gym Leader?

src/data/all-badges.ts
Tempo-anon: well this is why we're having these discussions :pikasmile:

Do we want an abstract badge class and then create new classes for each badge? Or a concrete Badge class that can be instantiated with the parameters we need?

src/data/badge.ts
Tempo-anon: the `getThing()` and `get thing()` discrepancy is a thing in all of the codebase. We should decide on which pattern to use. 

re: `trainer` linking, yeah we can definitely have that on the Badge class. Not like extra information will hurt

src/data/all-badges.ts
innerthunder: I think it being abstract enforces better implementation practices, and it might make it easier to make type guards instead of doing `instanceof` checks everywhere (less circular dependency hell)

src/data/badge.ts
innerthunder: I find `get thing()` to be cleaner personally, but `getThing()` would still be needed in cases where parameters are needed. `get thing(foo)` doesn't compile afaik.

Tempo-anon:
Closing in favor of #443

PR #247, opened 2025-01-10
[Test] Enable `--no-isolate` for unit tests

PigeonBar:
## What are the changes for developers?

- Unit tests are actually fast now.
- (Introduces a somewhat higher risk of tests not working properly if stuff isn't sanitized properly between tests.)

I found some [Vitest docs](https://vitest.dev/guide/improving-performance) that recommend enabling `--no-isolate` to improve test performance. (My understanding is that once a thread is done running a test file, `--no-isolate` allows the thread to be reused for the next test file, so Vitest no longer has to spend as much time on creating new threads. If anyone else has a deeper understanding of how this stuff works, please let me know.)

I tried organizing each change into an individual commit to hopefully make the changes easier to follow.

1. Enabled the `--no-isolate` flag for the commands `npm run test`, `npm run test:silent`, and GitHub test workflows.
2. Moved some initialization code into a function `initTestFile()` that runs at the beginning of each test file (via the `beforeAll()` placed in `vitest.setup.ts`).
3. Changed `MockConsoleLog` to an actual class. (This prevents `mockConsoleLog` from crashing due to `this` somehow being undefined.)
4. The `vitest-canvas-mock` library seems to have issues when running tests with `--no-isolate`, so I replaced it with the stub objects `mockContext` and `mockCanvas`.
5. ~~A few overrides are now sanitized in `gameManager.ts` for each test.~~ (Outdated by change 15 below)
6. Added `listenersManager.ts` to help with removing unneeded `MessageListener`s. (Vitest opens a new listener for every test file but does not remove them, so the manager is needed to prevent an error occurring from having too many listeners active.)
7. Fixed how `select-modifier-phase.ts` handles phases.
8. Removed an `i18next.init()` call from `status_effect.test.ts` to prevent it from interfering with other tests that use `i18next`.
9. The visual update function `Pokemon.updateInfo()` is now mocked to complete instantly. This function is tied to `DamageAnimPhase` ending, so this change fixed some issues with `DamageAnimPhase` not ending properly in one test, then breaking a completely different test by deciding to end during that test.
10. Added a function to initialize a server for each API test file (due to previously used servers not being sanitized).
11. `battlerTags\substitute.test.ts` no longer creates new BattleScenes, since creating a new BattleScene messes with `globalScene`.
12. Configured Vitest`s test file sorter to place specific test files earlier, for the following reasons:
  - `battle_scene.test.ts` must be run first so that it cannot reuse an already existing `globalScene`. This ensures that `battle_scene.test.ts` creates a new BattleScene, which is required for the test to function.
  - The gamepad tests in `input.test.ts` randomly fail if the test file is not run early. I have tried looking into this, but I have not been able to figure anything out.
13. Fixed `fusion_flare_bolt.test.ts` trying to access certain moves before they are initialized.
14. Small updates to GitHub docs.
15. (New) Added a function to the overridesHelper to sanitize all overrides, instead of only sanitizing a select few overrides in `gameManager.ts`.

## How to test the changes?

`npm run test:silent`.

To make sure that tests still fail when they should fail: Introduce a bug somewhere in the code, then check that the corresponding unit test fails.

## Screenshots/Videos

<details>
<summary> After the changes: Unit tests take under 2 minutes to run on my device </summary>

20250110 No Isolate Success

</details>
<details>
<summary> After the changes: Unit tests fail if I change Freeze-Dry to be super-effective against Fire-types instead of Water-types </summary>

20250110 No Isolate Failure

</details>

## Checklist


torranx (commented):
as far as I understand, enabling `--no-isolate` risks side-effects between tests since we're making the tests run in the same shared environment. unless we have a comprehensive way to reset states in between tests, I'm worried of the potential flaky tests this may bring

Tempo-anon (commented):
Tests are not really my forte, but I have to say that the formatting of this branch with every significant update in its own well documented commit truly goes above and beyond.

The performance improvement is definitely very impressive but I do see that there may be concern for tests not sanitizing properly. Perhaps we can just go ahead with this change and then either roll it back if it's causing too many flaky tests or we can start stamping out the tests that have not been reset properly

test/testUtils/gameManager.ts
```diff
@@ -57,6 +57,7 @@ import fs from "fs";
 import { expect, vi } from "vitest";
 import { globalScene } from "#app/global-scene";
 import type StarterSelectUiHandler from "#app/ui/starter-select-ui-handler";
+import { MockFetch } from "./mocks/mockFetch";
```
MokaStitcher:
```suggestion
import { MockFetch } from "#test/testUtils/mocks/mockFetch";
```

MokaStitcher (dismissed):
Considering the globalScene is already being reused between tests I don't see the potential introduced flakiness to be enough reason to refuse this change.
If anything, I expect the globalScene one to potentially cause more trouble in the future than this one, especially if we want to break the game between multiple scenes.

This is of course knowing that this doesn't impact the game itself and can be reverted as needed

<sub> For reference, from 40+ minutes on my computer before globalScene reuse, to less than 3 minutes with this PR. Big yes

Tempo-anon (dismissed):

src/overrides.ts
```diff
@@ -243,6 +243,11 @@ class DefaultOverrides {
    * Note that, for all items in the array, `count` is not used.
    */
   readonly ITEM_REWARD_OVERRIDE: ModifierOverride[] = [];
+
+  /**
+   * If `true`, disable all non-scripted opponent trainer encounters.
+   */
+  readonly DISABLE_TRAINERS_OVERRIDE: boolean = false;
```
MokaStitcher: Should this be named DISABLE_RANDOM_TRAINERS_OVERRIDE if it doesn't include fixed battles?

torranx (dismissed):
since others are ok with the risks involved + it's a separate thing from the game itself, lgtm

(also that's a huge improvement on test time)

test/testUtils/listenersManager.ts
```diff
@@ -0,0 +1,41 @@
...
+ * The list of listeners that were present during the first time this manager is run.
+ * These initial listeners are needed throughout the entire test suite, so we never remove them.
+ */
+const initialListeners: unknown[] = [];
+
+/**
+ * The current listener that is only needed for the current test file.
+ * We plan to delete it during the next test file, when it is no longer needed.
+ */
+let currentListener;
```
DayKev:
```suggestion
let currentListener: any;
```
Should this be `any` or `unknown`?

DayKev:
The tests run so fast my CPU literally can't keep up (previously running the test suite would pin my CPU near 100% usage, now it bounces from 10-20% usage).

image

Here I set max threads to 32 (normally it defaults to 16 since I have an 8C/16T CPU) and that actually shaved off 15s. Crazy.

image

test/testUtils/mocks/mockContext.ts
```diff
@@ -0,0 +1,15 @@
+import { mockCanvas } from "#test/testUtils/mocks/mockCanvas";
+
+/**
+ * A minimal stub object to mock CanvasRenderingContext2D
+ */
+export const mockContext: any = {
+  font: "",
+  //@ts-ignore
+  measureText: () => new TextMetrics(""),
```
DayKev: What does passing `""` here do?
image

MokaStitcher:
for reference I got a failure of the inputs test locally.
But it's not the first time it's happened to me even before this, usually it's because my computer lags during test. It was the first test run for the day so it took a while to start, which may explain the issues.

test/testUtils/mocks/mockContext.ts
PigeonBar: Not sure what's going on here: Inside the library, the constructor for `TextMetrics` has 1 param, but the compiler doesn't recognize it for some reason.

image

I have replaced it with an empty object as a workaround.

MokaStitcher (approved):

test/testUtils/mocks/mockContext.ts
DayKev: Weird.

DayKev (approved):

PR #64, opened 2024-12-17
[Move] Fully Implement Sky Drop

innerthunder:
## What are the changes the user will see?

[Sky Drop](https://bulbapedia.bulbagarden.net/wiki/Sky_Drop_(move)) will now carry the target up to the air and immobilize them.

## Why am I making these changes?

I just made Dragon Darts and wanted to enter a new level of hell

## What are the changes from a developer perspective?

- Sky Drop implemented with
  - new move attributes `SkyDropAttr` and `NoDamageAgainstFlyingAttr`. `SkyDropAttr` is implemented as a charge attribute
  - new `hitCheckOnCharge` flag for charge moves to force Sky Drop to calculate a hit check during the move's charge phase.
  - new battler tag `SkyDropTag` representing the airborne status from Sky Drop.
- New abstract phase `HitCheckPhase` contains the hit check calculation function as well as some related helper functions. This is now inherited by both `MoveChargePhase` and `MoveEffectPhase`, so both phases can make use of the function.
- Adjusted some code for Wimp Out, Commander, and Gravity to account for edge cases with Sky Drop.

## Screenshots/Videos

<details>
  <summary>Single Battle: Sky Drop executed over 2 turns </summary>

https://github.com/user-attachments/assets/a9d85bc3-d8e0-49aa-9964-27348c292ce0

</details>

<details>
<summary>Double Battle</summary>

https://github.com/user-attachments/assets/5c3e9b50-3eb9-4ea8-9e9e-6f310e843916

</details>

## How to test the changes?

`npm run test sky_drop` - new

`npm run test wimp_out` - 1 new test

image

## Checklist

- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test`)
  - [x] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?
- [x] Have I provided screenshots/videos of the changes (if applicable)?
  - [n/a] Have I made sure that any UI change works for both UI themes (default and legacy)?

PR #263, opened 2025-01-12
[P1 Bug] Several Sky Drop updates

PigeonBar:

1. A Pokemon immobilized by Sky Drop can no longer switch out or flee.
2. Both the user and target of Sky Drop are considered ungrounded (confirmed correct behaviour by [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Grounded)).
3. Sky Drop gets cancelled properly if the user or target faints from indirect damage.
4. Sky Drop now fails against a Commanding Tatsugiri.

(Edit) NOT FIXED: Catching a wild Pokemon in the middle of Sky Drop still causes issues. (As discussed in #64, this bug is planned to be addressed in a separate PR that handles catching Pokemon during charging moves in general.)

<sup> I need to learn to review code before it gets merged, not after </sup>

1. `Pokemon.isTrapped()` and `Pokemon.isGrounded()` now also check if the Pokemon has a `SkyDropTag`.
2. (Edited) Moved the function `clearSkyDropAffects()` from `commander-ab-attr.ts` to ~~`battle-scene.ts`~~ `SkyDropTag`. This function, along with some other code to clean up a fainting Pokemon, is now also called during `FaintPhase`.
3. Added a condition in `SkyDropAttr` to make the move fail against a Commanding Tatsugiri.
4. Added some tests.

## Screenshots/Videos

<details>
<summary> Before the changes: A Pokemon immobilized by Sky Drop can switch out </summary>

https://github.com/user-attachments/assets/b6f05975-5b31-41d3-b1a2-6ae56e96f257

</details>
<details>
<summary> After the changes: A Pokemon immobilized by Sky Drop can no longer switch out or flee </summary>

https://github.com/user-attachments/assets/c81361ed-8fe3-4ed6-b766-7d7f7b8360b5

</details>
<details>
<summary> Before the changes: Grassy Terrain tries to heal a Pokemon immobilized by Sky Drop </summary>

https://github.com/user-attachments/assets/3c884d23-ec26-4aba-a0c0-b0308df6ca96

</details>
<details>
<summary> After the changes: Grassy Terrain no longer tries to heal a Pokemon immobilized by Sky Drop </summary>

https://github.com/user-attachments/assets/7695ef4a-1116-4f16-9771-da7f08144194

</details>
<details>
<summary> Before the changes: Sky Drop is not cleared when the user faints </summary>

https://github.com/user-attachments/assets/a1721bc3-f17d-44f1-bf2f-bf4e2d27fd53

</details>
<details>
<summary> After the changes: Sky Drop is cleared when the user faints </summary>

https://github.com/user-attachments/assets/9b10fdf4-7f91-4fe2-b5d7-e37e17ecdd8a

</details>
<details>
<summary> Before the changes: Sky Drop succeeds against Commanding Tatsugiri </summary>

https://github.com/user-attachments/assets/5ba0ca76-84f8-4ffe-9f22-73aeba6d46bc

</details>
<details>
<summary> After the changes: Sky Drop fails against Commanding Tatsugiri </summary>

https://github.com/user-attachments/assets/463545b4-ee3e-4815-a8a6-d72cc496fec4

</details>

## How to test the changes?

`npm run test sky_drop`

## Checklist


src/field/pokemon.ts
```diff
@@ -1803,7 +1804,8 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
       || (!this.isOfType(Type.FLYING, true, true)
         && !this.hasAbility(Abilities.LEVITATE)
         && !this.getTag(BattlerTagType.FLOATING)
-        && !this.getTag(SemiInvulnerableTag))
+        && !this.getTag(SemiInvulnerableTag)
+        && !this.getTag(SkyDropTag))
```
Tempo-anon, line 1784: Should the GroundedTag.onAdd also be updated to have the SkyDropTag check? Should probably also add a comment here explaining why there's some copypasted code there

PigeonBar: According to Showdown, if a Pokemon is affected by Sky Drop and then gets hit by Smack Down, the game shouldn't display "[Pokemon] fell straight down", and also shouldn't inflict the Smack Down status on the target

image

Tempo-anon: Huh, all right. That's so weird though. Might be worth leaving a comment in the GroundedTag then so someone in the future doesn't take a look and also get confused

src/phases/command-phase.ts
```diff
@@ -334,7 +334,7 @@ export class CommandPhase extends FieldPhase {
             true,
           );
         } else {
-          const trapTag = playerPokemon.getTag(TrappedTag);
+          const trapTag = playerPokemon.getTag(TrappedTag) ?? playerPokemon.getTag(SkyDropTag);
```
innerthunder, line 343: Should Sky Drop be using the same message as binding tags and other trapping tags?

If not, omitting this should force the "can't escape/flee" message to be more generic (L340/343)

src/phases/faint-phase.ts
```diff
@@ -84,6 +84,14 @@ export class FaintPhase extends PokemonPhase {
...
     this.grudgeTag = grudgeTag;
     this.source = source;
+
+    /**
+     * Clear Sky Drop effects immediately when we know this Pokemon is fainting.
+     * Otherwise, the Pokemon's summon data may get reset before this phase starts,
+     * which would prevent the function from properly clearing Sky Drop effects.
+     */
+    const faintPokemon = this.getPokemon();
+    globalScene.clearSkyDropEffects(faintPokemon);
```
innerthunder: Should this be done with other things in `clearEffectsOnFaint`?

src/phases/command-phase.ts
PigeonBar: Someone would need to test this on cart, I think Showdown just displays a generic "You are trapped and cannot switch!" regardless of the trapping effect.

Screenshot 2025-01-11 101106

src/phases/faint-phase.ts
PigeonBar: My intent was for this to replace clearing Sky Drop effects in `clearEffectsOnFaint` in MoveEffectPhase, since `clearEffectsOnFaint` only runs if a Pokemon faints directly from an attack. (I can add it back to `clearEffectsOnFaint` if needed.)

flx-sta:
For the sake of cooperation I've changed the title to `update`. It was known that the first PR will not cover everything

<sub>Meta comment</sub>

src/phases/faint-phase.ts
innerthunder: Oh sorry, I meant should the other things in `clearEffectsOnFaint` be moved here?

Tempo-anon (dismissed):

src/battle-scene.ts
```diff
@@ -919,6 +919,28 @@ export default class BattleScene extends SceneBase {
...
+            globalScene.tryRemovePhase((phase) => phase instanceof MovePhase && phase.pokemon.id === p.id);
+            p.getMoveQueue().shift();
+            p.removeTag(BattlerTagType.CHARGING);
+          }
+          p.removeTag(BattlerTagType.SKY_DROP);
+        }
+      });
+    }
+  }
+
```
flx-sta: NIT: Could also be put as static method inside `SkyDropTag`

e.g.
```ts
export class SkyDropTag extends BattlerTag {
  //...
  
  /**
   * Remove's Sky Drops effects from the given Pokemon
   * and whoever else is involved (when applicable).
   * @param pokemon The {@linkcode Pokemon} with this attribute
   */
  public static clearEffect(pokemon: Pokemon) {
   const skyDropTagId = pokemon.getTag(BattlerTagType.SKY_DROP)?.sourceId;
    if (skyDropTagId) {
      globalScene.getField(true).forEach((p) => {
        if (p && p.getTag(BattlerTagType.SKY_DROP)?.sourceId === skyDropTagId) {
          // Cancel the Sky Drop user's next use of Sky Drop
          if (skyDropTagId === p.id) {
            globalScene.tryRemovePhase((phase) => phase instanceof MovePhase && phase.pokemon.id === p.id);
            p.getMoveQueue().shift();
            p.removeTag(BattlerTagType.CHARGING);
          }
          p.removeTag(BattlerTagType.SKY_DROP);
        }
      });
    }
  }
  
  //...
}
```

and then call `SkyDropTag.clearEffect(...)` instead of bloating up the `globalScene` yet more

flx-sta (commented):
Looking at all these `expect(enemyPokemon.getTag([...])).toBeUndefined();` makes me wonder if we should add a `hasTag()` helper method that returns a boolean instead

flx-sta (dismissed):
<sub>Meant to approve with comments</sub>

flx-sta (commented):
@PigeonBar did you mean to **NOT** make it a static method?

PigeonBar:
> @PigeonBar did you mean to **NOT** make it a static method?

Yes (mainly so it could access `this.sourceId`)

flx-sta (approved):

flx-sta:
> > @PigeonBar did you mean to **NOT** make it a static method?
> 
> Yes (mainly so it could access `this.sourceId`)

Didn't notice that initially. Noice

Tempo-anon (approved):

PR #249, opened 2025-01-11
[Refactor] Remove string assignments from Tutorial and Tag Type enums

innerthunder:
## What are the changes the user will see?

N/A

## Why am I making these changes?

String enums aren't supported by the language(s) planned for server code, so this should make server development a bit smoother.

## What are the changes from a developer perspective?

3 server-facing enums have been made numeric:
- `ArenaTagType`
- `BattlerTagType`
- `Tutorial`

This means that the only remaining server-facing enum in the codebase is `SpeciesFormKey`, which should be tackled when pokemon species are refactored in the future.

## Screenshots/Videos

N/A

## How to test the changes?

`npm run test`

## Checklist

- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test`)
  - [x] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?
- [?] Have I provided screenshots/videos of the changes (if applicable)?
  - [n/a] Have I made sure that any UI change works for both UI themes (default and legacy)?

PR #270, opened 2025-01-13
[Beta][P2] Fix tutorial seen flags not being set/read properly

MokaStitcher:

No more seeing the tutorials on repeat if tutorials are enabled

Got broken in #249 when removing the strings from the Tutorial enum

- change the tutorial local storage key to use the pk(t)y prefix
- change the stored data to a simple boolean array instead of the `string, flag` object that was used before
- add a `wasTutorialSeen` function in game data so that the rest of the code doesn't need to know how it's stored
- add try/catch statements around tutorial data loading/saving operations

## How to test the changes?

- Open the game in a private windows (or enable tutorials in the settings)
- The intro tutorial should play
- Open the menu. The tutorial should play
- Close the menu then reopen it. Neither the intro or tutorial menu should play again
- Start a run. The "press M to open the menu" tutorial should not play since seeing the menu tutorial also disables this one

## Checklist

~~ Are there any localization additions or changes? If so: ~~


src/system/game-data.ts
```diff
@@ -795,44 +791,50 @@ export class GameData {
...
+      console.warn("Failed to saved tutorial data in local storage", err);
+      return false;
     }
+  }
 
-    return ret;
+  /**
+   * Checks if the given tutorial is marked as seen in local storage
+   * @param tutorial the {@linkcode Tutorial} to get the flag for
+   * @returns `true` if the tutorial has already been seen, false otherwise
```
torranx:
```suggestion
   * @returns `true` if the tutorial has already been seen, `false` otherwise
```

src/system/game-data.ts
```diff
@@ -795,44 +791,50 @@ export class GameData {
...
+    }
+    return tutorials;
   }
 
-  public getTutorialFlags(): TutorialFlags {
+  /**
+   * Registers the given tutorial's seen status in local storage
+   * @param tutorial the {@linkcode Tutorial} to update the flag for
+   * @param flag whether the tutorial should be marked as seen or not. Default: `true`
+   * @returns `true` if saving was successful
```
torranx:
```suggestion
   * @returns `true` if saving was successful, `false` otherwise
```

torranx (commented):
minor nits, otherwise lgtm

torranx (dismissed):

src/system/game-data.ts
```diff
@@ -795,44 +791,50 @@ export class GameData {
...
   }
 
-  public getTutorialFlags(): TutorialFlags {
+  /**
+   * Registers the given tutorial's seen status in local storage
+   * @param tutorial the {@linkcode Tutorial} to update the flag for
+   * @param flag whether the tutorial should be marked as seen or not. Default: `true`
+   * @returns `true` if saving was successful, `false` otherwise
+   */
+  public saveTutorialFlag(tutorial: Tutorial, flag: boolean = true): boolean {
```
Tempo-anon: Is there ever a time when `flag` should be false? It looks like both uses of this in tutorial.ts call it with true, and having the option implies the name of this function should be `setTutorialFlag`

Tempo-anon (dismissed):

src/system/game-data.ts
MokaStitcher: Got lazy on that one and left it "just in case". You make a good point about the name though, updated

src/system/game-data.ts
```diff
@@ -795,44 +791,49 @@ export class GameData {
...
-    Object.keys(Tutorial).forEach((key) => {
-      if (key === Tutorial[tutorial]) {
-        tutorials[key] = flag;
-      } else {
-        tutorials[key] ??= false;
+    if (localStorage.hasOwnProperty(key)) {
+      try {
+        tutorials = JSON.parse(localStorage.getItem(key)!);
+      } catch (err) {
+        console.warn("Failed to parse tutorial data from local storage", err);
```
flx-sta: Please don't use `hasOwnProperty`. The ts compiler can't make the connection that way.
Instead simply use `getItem` and check the content:

```suggestion
    const lsItem = localStorage.getItem(key);
    if (lsItem) {
      try {
        tutorials = JSON.parse(lsItem);
      } catch (err) {
        console.warn("Failed to parse tutorial data from local storage", err);
```

src/system/game-data.ts
```diff
@@ -795,44 +791,50 @@ export class GameData {
...
     }
+  }
 
-    return ret;
+  /**
+   * Checks if the given tutorial is marked as seen in local storage
+   * @param tutorial the {@linkcode Tutorial} to get the flag for
+   * @returns `true` if the tutorial has already been seen, `false` otherwise
+   */
+  public wasTutorialSeen(tutorial: Tutorial): boolean {
```
flx-sta: NIT: using “was” is an uncommon appearance for me

```suggestion
  public isSeenTutorial(tutorial: Tutorial): boolean {
```

flx-sta (commented):
It is worth considering to either clarify this lskey to have the word “seen” in it or use a sub-key in the object that indicates `seen`.
Basically `tutorials.seen: […]`

cause as of right now it’s not clear what this storage is exactly doing. It just states “tutorial”

src/system/game-data.ts
MokaStitcher: LS key updated to seenTutorials to match the seenDialogues one

src/system/game-data.ts
```diff
@@ -833,7 +833,7 @@ export class GameData {
    * @param tutorial the {@linkcode Tutorial} to get the flag for
    * @returns `true` if the tutorial has already been seen, `false` otherwise
    */
-  public wasTutorialSeen(tutorial: Tutorial): boolean {
+  public isTutorialSeen(tutorial: Tutorial): boolean {
```
flx-sta: NIT: swap seen & tutorial here
```suggestion
  public isSeenTutorial(tutorial: Tutorial): boolean {
```

flx-sta (dismissed):

frutescens (approved):

flx-sta (commented):
The way the flags are stored feels kinda weird to me
```json
[true,null,null,true]
```

I know this means that I've seen the tutorials 0 and 3 but I expected to basically see the numbers seen 
e.g. 
```json
[0, 3]
```

MokaStitcher:
> The way the flags are stored feels kinda weird to me

That's js for you... I can put back the key in there I guess...

flx-sta:
@MokaStitcher Have a look at #286 and decide if you want to use the `boolean` approach or the `Set` approach

Tempo-anon (approved):

PR #268, opened 2025-01-13
[Checkup][Moves] Move Validation

frutescens:
## What are the changes the user will see?

- Sheer Cold's listed accuracy was corrected from 20 to 30. Merely a cosmetic change since SheerColdAccuracyAttr does not use the value.
- Life Dew and Jungle Healing now correctly listed as triage moves.
- Added `makesContact()` flag to Aqua Step, Bitter Blade, and Headlong Rush
- Spit Up's accuracy is set to 100

## Why am I making these changes?

Making move checkups easier. 

## What are the changes from a developer perspective?
- Test created using data from PokeAPI that checks for the following characteristics:
     - Move Category
     - Power
     - Accuracy
     - Secondary Effect Chance
     - Type
     - Certain flags (Sound, Punch, Ballistics, etc.)
- Z Moves are marked as unimplemented. 

## Checklist

- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test:silent`)
  - [x] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?

PR #282, opened 2025-01-14
[Bug] Fix Misty Explosion making contact

PigeonBar:

Misty Explosion no longer makes contact.

Bug discovered while reviewing #268.

The root cause was that if `setFlag(flag, false)` is called when the move already doesn't have the flag, the function XOR's the flag, which incorrectly makes the move obtain the flag. To fix this, the function now only XOR's the flag if the move has the flag.

## Screenshots/Videos

<details>
<summary> Before the changes: Misty Explosion makes contact and is boosted by Tough Claws </summary>

https://github.com/user-attachments/assets/e2dcc46a-5525-49ca-a1ea-f4802f46fc7b

</details>
<details>
<summary> After the changes: Misty Explosion no longer makes contact and is no longer boosted by Tough Claws </summary>

https://github.com/user-attachments/assets/7604f367-f6cf-47ab-a78f-210327ee48b5

</details>

## How to test the changes?

Try interactions between Misty Explosion and battle effects that interact with contact moves.

## Checklist

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -359,13 +359,13 @@ export abstract class Move implements Localizable {
   /**
    * Sets the flags of the move
    * @param flag {@linkcode MoveFlags}
-   * @param on a boolean, if True, then "ORs" the flag onto existing ones, if False then "XORs" the flag onto existing ones
+   * @param on If `true`, sets the move to have the flag; if `false`, sets the move to NOT have the flag.
    */
   private setFlag(flag: MoveFlags, on: boolean): void {
     // bitwise OR and bitwise XOR respectively
     if (on) {
       this.flags |= flag;
-    } else {
+    } else if (this.hasFlag(flag)) {
       this.flags ^= flag;
     }
   }


src/data/move.ts
```diff
@@ -359,13 +359,13 @@ export abstract class Move implements Localizable {
...
-   * @param on a boolean, if True, then "ORs" the flag onto existing ones, if False then "XORs" the flag onto existing ones
+   * @param on If `true`, sets the move to have the flag; if `false`, sets the move to NOT have the flag.
    */
   private setFlag(flag: MoveFlags, on: boolean): void {
     // bitwise OR and bitwise XOR respectively
     if (on) {
       this.flags |= flag;
-    } else {
+    } else if (this.hasFlag(flag)) {
       this.flags ^= flag;
```
Tempo-anon, line 369: ~~If the move already has the flag and then we `setFlag(false)` wouldn't that still keep the flag as true since we'll be xor'ing a 1 with a 0?~~

wait nvm the flag itself is just a mask like 0000100000

Tempo-anon (approved):
We may also want to consider how moveflags are handled as data in the first place since right now it's just one giant `number` that's a binary representation of everything.

torranx (approved):

PR #283, opened 2025-01-14
[Move] Implement G-Max effects

Tempo-anon:

None

There was some interest expressed in implementing G-Max moves. 

* Refactored the `StealthRockTag` to `TypeHazardTag` to account for different typed entry hazards
* Created a `GMaxArenaTag` for the ArenaTags that damage Pokemon every turn if they aren't of a specific type
* Created a `ChiStrikeCritBoostTag` for stackable instances of critboost and it checked in `pokemon.ts`
* Created a `add-battler-tag-all-opps-attr.ts` for applying battlerTags to all opponents
* Created a `multi-status-all-opps-effect-attr.ts` for applying status to all opponents
* Created a `stat-stage-change-all-opps.ts` for applying stat changes to all opponents
* Added basic effects to G-Max moves

## Screenshots/Videos

n/a

## How to test the changes?

Override with the moves

## Checklist


src/data/arena-tag.ts
```diff
@@ -925,53 +925,39 @@ export class DelayedAttackTag extends ArenaTag {
...
-      );
+      if (this.damagingType === Type.ROCK) {
+        globalScene.queueMessage(
+          i18next.t("arenaTag:stealthRockOnAdd", { opponentDesc: source.getOpponentDescriptor() }),
+        );
+      } else {
+        globalScene.queueMessage(
+          i18next.t("arenaTag:sharpSteelOnAdd", { opponentDesc: source.getOpponentDescriptor() }),
+        );
+      }
```
flx-sta, line 954: You probably want to avoid hard-coding those messages and instead provide the message as parameter.

src/data/arena-tag.ts
```diff
@@ -925,53 +925,39 @@ export class DelayedAttackTag extends ArenaTag {
...
-        damageHpRatio = 0.25;
-        break;
-      case 4:
-        damageHpRatio = 0.5;
-        break;
-    }
-
-    return damageHpRatio;
+    const effectiveness = pokemon.getAttackTypeEffectiveness(this.damagingType, undefined, true);
+    return 0.125 * effectiveness;
```
flx-sta, line 960: What if "effectiveness" becomes negative? 
Is that something to be aware of?

src/data/arena-tag.ts
```diff
@@ -989,9 +975,15 @@ class StealthRockTag extends ArenaTrapTag {
...
-      );
+      if (this.damagingType === Type.ROCK) {
+        globalScene.queueMessage(
+          i18next.t("arenaTag:stealthRockActivateTrap", { pokemonNameWithAffix: getPokemonNameWithAffix(pokemon) }),
+        );
+      } else {
+        globalScene.queueMessage(
+          i18next.t("arenaTag:sharpSteelActivateTrap", { pokemonNameWithAffix: getPokemonNameWithAffix(pokemon) }),
+        );
+      }
```
flx-sta, line 986: You probably want to avoid hard-coding those messages and instead provide the message as parameter.

src/data/arena-tag.ts
```diff
@@ -1008,6 +1000,62 @@ class StealthRockTag extends ArenaTrapTag {
   }
 }
 
+/**
+ * Class to describe the field effect of G-Max moves that damage all Pokemon not of the type to take 1/6th damage each turn
+ */
+class GMaxArenaTag extends ArenaTag {
+  public immuneType: Type;
+  constructor(tagType, sourceMove: Moves, sourceId: number, immuneType: Type) {
```
flx-sta, line 1008: NIT: missing new line

src/data/arena-tag.ts
```diff
@@ -1008,6 +1000,62 @@ class StealthRockTag extends ArenaTrapTag {
...
+    super(tagType, 4, sourceMove, sourceId, ArenaTagSide.ENEMY);
```
flx-sta, line 1009: Can only enemy take damage from it?

src/data/arena-tag.ts
```diff
@@ -1008,6 +1000,62 @@ class StealthRockTag extends ArenaTrapTag {
...
+      case Type.GRASS:
+        return CommonAnim.WRAP;
+      case Type.FIRE:
+        return CommonAnim.FIRE_SPIN;
+      case Type.WATER:
+        return CommonAnim.WHIRLPOOL;
+      case Type.ROCK:
+        return CommonAnim.SAND_TOMB;
+      default:
+        return CommonAnim.WRAP;
```
flx-sta, line 1024: NIT: `Type.GRASS` can be fallthrough or simply removed since it's the same as `default`

src/data/arena-tag.ts
```diff
@@ -1008,6 +1000,62 @@ class StealthRockTag extends ArenaTrapTag {
...
+    }
+  }
+
+  override onAdd(_arena: Arena): void {
+    globalScene.queueMessage(
+      i18next.t(
+        `arenaTag:GMaxArenaOnAdd${this.side === ArenaTagSide.PLAYER ? "Player" : this.side === ArenaTagSide.ENEMY ? "Enemy" : ""}${Type[this.immuneType]}`,
```
flx-sta, line 1031: Please no more multi-ternary!!!!

src/data/battler-tags.ts
```diff
@@ -2109,6 +2114,48 @@ export class CritBoostTag extends BattlerTag {
   }
 }
 
+/**
+ * Used for an instance of a stackable crit boost
+ * Granted by G-Max Chi Strike
+ */
+export class ChiStrikeCritBoostTag extends BattlerTag {
+  public stackCount: number = 0;
```
flx-sta, line 2122: why public?

src/data/battler-tags.ts
```diff
@@ -2109,6 +2114,48 @@ export class CritBoostTag extends BattlerTag {
...
+    super.onRemove(pokemon);
+
+    globalScene.queueMessage(
+      i18next.t("battlerTags:chiStrikeCritBoostOnRemove", { pokemonNameWithAffix: getPokemonNameWithAffix(pokemon) }),
+    );
+  }
+
+  override loadTag(source: BattlerTag | any): void {
+    super.loadTag(source);
+    this.stackCount = source.stackCount || 0;
```
flx-sta, line 2155: Always `??` over `||` unless you are 100% sure you want `||`

```suggestion
    this.stackCount = source.stackCount ?? 0;
```

src/data/move-attrs/add-battler-tag-all-opps-attr.ts
```diff
@@ -0,0 +1,180 @@
...
+ * Confusion - G-Max Smite
+ * Infatuation - G-Max Cuddle
+ * Trapped - G-Max Terror
+ * Drowsy - G-Max Snooze
+ * Torment - G-Max Meltdown
+ * @extends MoveEffectAttr
+ * @see {@linkcode BattlerTag}
+ */
+export class AddBattlerTagAllOppsAttr extends MoveEffectAttr {
+  public tagType: BattlerTagType;
```
flx-sta, line 31: Why public?
pOtentially `readonly`

src/data/all-moves.ts
```diff
@@ -3254,107 +3257,148 @@ export function initMoves() {
...
-      .target(MoveTarget.ALL_NEAR_ENEMIES)
-      .unimplemented(),
-    new AttackMove(Moves.G_MAX_MALODOR, Type.POISON, MoveCategory.PHYSICAL, 10, -1, 10, -1, 0, 8)
-      .target(MoveTarget.ALL_NEAR_ENEMIES)
-      .unimplemented(),
-    new AttackMove(Moves.G_MAX_STONESURGE, Type.WATER, MoveCategory.PHYSICAL, 10, -1, 10, -1, 0, 8)
-      .target(MoveTarget.ALL_NEAR_ENEMIES)
-      .unimplemented(),
+    new AttackMove(Moves.G_MAX_WILDFIRE, Type.FIRE, MoveCategory.PHYSICAL, 10, -1, 10, -1, 0, 8).attr(
+      AddArenaTrapTagHitAttr,
```
innerthunder, line 3261: Use `AddArenaTagAttr` instead. This attr is mainly for enforcing Sheer Force's interaction with Stone Axe / Ceaseless Edge, and otherwise should be deprecated.

src/data/move-attrs/add-battler-tag-all-opps-attr.ts
```diff
@@ -0,0 +1,180 @@
...
+      case BattlerTagType.IGNORE_ACCURACY:
+      case BattlerTagType.AQUA_RING:
+        return 3;
+      case BattlerTagType.PROTECTED:
+      case BattlerTagType.FLYING:
+      case BattlerTagType.CRIT_BOOST:
+      case BattlerTagType.ALWAYS_CRIT:
+        return 5;
+      default:
+        console.warn(`BattlerTag ${BattlerTagType[this.tagType]} is missing a score!`);
```
flx-sta, line 168: NIT: 
```suggestion
        console.warn(`BattlerTag ${BattlerTagType[this.tagType]} has no defined benefit score!`);
```

flx-sta (commented):
~~Nothing obvious as far as I can tell except~~ 
- https://github.com/Despair-Games/poketernity/pull/283#discussion_r1917282842
- https://github.com/Despair-Games/poketernity/pull/283#discussion_r1917296277

src/data/all-moves.ts
```diff
@@ -3254,107 +3257,148 @@ export function initMoves() {
...
+      MultiStatusAllOppsEffectAttr,
+      [StatusEffect.PARALYSIS],
+      true,
+    ),
+    new AttackMove(Moves.G_MAX_GOLD_RUSH, Type.NORMAL, MoveCategory.PHYSICAL, 10, -1, 10, -1, 0, 8).attr(MoneyAttr), // TODO: Should give 100x user level money
+    new AttackMove(Moves.G_MAX_CHI_STRIKE, Type.FIGHTING, MoveCategory.PHYSICAL, 10, -1, 10, -1, 0, 8).attr(
+      AddBattlerTagAttr,
+      BattlerTagType.CHI_STRIKE_CRIT_BOOST,
+      true,
+      { failOnOverlap: false },
```
innerthunder, line 3279:
```suggestion
```
`false` is the default

src/data/arena-tag.ts
Tempo-anon: It shouldn't ever be negative, no. It should only ever be a positive power of 2

src/data/arena-tag.ts
Tempo-anon: Oh this is a good point. If an enemy uses this it should be ArenaTagSide.PLAYER. I got "enemy" confused with opponent in this instance

src/data/arena-tag.ts
Tempo-anon: Oh I agree but my reason for ordering it like this is that
* It makes sense due to dex order (grass, fire, water, rock)
* Helps future proof if we ever need to add types after rock

src/data/battler-tags.ts
Tempo-anon: I stole that code from stockpile, I can add that over there too then

src/data/arena-tag.ts
flx-sta: sure. Maybe add a comment so we will know in the future

src/data/all-moves.ts
```diff
@@ -3254,107 +3257,148 @@ export function initMoves() {
...
+      { turnCount: 5 },
+    ),
+    new AttackMove(Moves.G_MAX_VOLCALITH, Type.ROCK, MoveCategory.PHYSICAL, 10, -1, 10, -1, 0, 8).attr(
+      AddArenaTrapTagHitAttr,
+      ArenaTagType.G_MAX_VOLCALITH,
+    ),
+    new AttackMove(Moves.G_MAX_SANDBLAST, Type.GROUND, MoveCategory.PHYSICAL, 10, -1, 10, -1, 0, 8).attr(
+      TrapAttr,
+      BattlerTagType.SAND_TOMB,
+    ),
```
innerthunder, line 3337: This effect should be slightly different from Sand Tomb in that it shouldn't disappear if the user leaves the field

src/data/all-moves.ts
```diff
@@ -3254,107 +3257,148 @@ export function initMoves() {
...
+    new AttackMove(Moves.G_MAX_FOAM_BURST, Type.WATER, MoveCategory.PHYSICAL, 10, -1, 10, -1, 0, 8).attr(
+      StatStageChangeAllOppsAttr,
+      [Stat.SPD],
+      -2,
+    ),
+    new AttackMove(Moves.G_MAX_CENTIFERNO, Type.FIRE, MoveCategory.PHYSICAL, 10, -1, 10, -1, 0, 8).attr(
+      TrapAttr,
+      BattlerTagType.FIRE_SPIN,
+    ),
```
innerthunder, line 3372: Same here

src/data/arena-tag.ts
```diff
@@ -925,53 +925,39 @@ export class DelayedAttackTag extends ArenaTag {
...
- * who is summoned into the trap, based on the Rock type's type effectiveness.
+ * Arena Tag class for {@link https://bulbapedia.bulbagarden.net/wiki/Stealth_Rock_(move) Stealth Rock}
+ * and {@link https://bulbapedia.bulbagarden.net/wiki/G-Max_Steelsurge_(move) G-Max Steelsurge}.
+ * Applies up to 1 layer of hazards, dealing percentage-based damage to any Pokemon summoned into the trap
+ * based on the type effectiveness
  */
-class StealthRockTag extends ArenaTrapTag {
-  constructor(sourceId: number, side: ArenaTagSide) {
-    super(ArenaTagType.STEALTH_ROCK, Moves.STEALTH_ROCK, sourceId, side, 1);
+class TypeHazardTag extends ArenaTrapTag {
```
innerthunder, line 933: I think this would be cleaner if:
1. This class were made abstract
2. Each of Stealth Rock and Steelspike (sharp steel?) were given a subclass
3. This class had abstract `getOnAddMessage` and `getActivationMessage` helper functions that were overridden by the new subclasses

src/data/arena-tag.ts
```diff
@@ -1008,6 +1000,62 @@ class StealthRockTag extends ArenaTrapTag {
   }
 }
 
+/**
+ * Class to describe the field effect of G-Max moves that damage all Pokemon not of the type to take 1/6th damage each turn
+ */
+class GMaxArenaTag extends ArenaTag {
```
innerthunder, line 1006: This could probably use a more descriptive name referring to the DOT effect, e.g. `NonTypeDamageTag`

src/data/battler-tags.ts
```diff
@@ -2109,6 +2114,48 @@ export class CritBoostTag extends BattlerTag {
...
+ */
+export class ChiStrikeCritBoostTag extends BattlerTag {
+  public stackCount: number = 0;
+
+  constructor(tagType: BattlerTagType, sourceMove: Moves) {
+    super(tagType, BattlerTagLapseType.TURN_END, 1, sourceMove, undefined, true);
+  }
+
+  override onAdd(pokemon: Pokemon): void {
+    super.onAdd(pokemon);
```
innerthunder, line 2129:
```suggestion
```

`BattlerTag.onAdd` doesn't do anything

src/data/move-attrs/add-battler-tag-all-opps-attr.ts
```diff
@@ -0,0 +1,180 @@
...
+import i18next from "i18next";
+import type { Move } from "../move";
+import type { MoveConditionFunc } from "../move-conditions";
+import { type MoveEffectAttrOptions, MoveEffectAttr } from "./move-effect-attr";
+import { getPokemonNameWithAffix } from "#app/messages";
+import { Type } from "#enums/type";
+
+interface AddBattlerTagAllOppsAttrOptions extends MoveEffectAttrOptions {
+  /** Should the move fail if the target already has a tag of the same type? */
+  failOnOverlap?: boolean;
```
innerthunder, line 13: I think this can be removed? Every G-Max move is an attack, so they shouldn't fail completely if the target(s) have a matching tag

src/data/move-attrs/add-battler-tag-all-opps-attr.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  /**
+   * If `true`, causes the move to fail if the target already
+   * has a tag of the same type.
+   * @default false
+   */
+  public get failOnOverlap() {
+    return this.options?.failOnOverlap ?? false;
+  }
```
innerthunder, line 48:
```suggestion
```

src/data/move-attrs/add-battler-tag-all-opps-attr.ts
```diff
@@ -0,0 +1,180 @@
...
+   */
+  public get turnCountMax() {
+    return this.options?.turnCountMax ?? this.turnCountMin;
+  }
+
+  override apply(user: Pokemon, target: Pokemon, move: Move): boolean {
+    let result = false;
+    let applyTag = false;
+    const moveChance = this.getMoveChance(user, target, move, this.selfTarget, true);
+    if (moveChance < 0 || moveChance === 100 || user.randSeedInt(100) < moveChance) {
```
innerthunder, line 70: Is this intended to be a single roll for both opponents? Also, does any max move actually use chance?

src/data/move-attrs/add-battler-tag-all-opps-attr.ts
```diff
@@ -0,0 +1,180 @@
...
+              i18next.t("moveTriggers:safeguard", { targetName: getPokemonNameWithAffix(target) }),
+            );
+          } else {
+            // Only make drowsy if the target has no status
+            applyTag = !opp.status;
+          }
+        }
+        if (this.tagType === BattlerTagType.TRAPPED && !opp.isOfType(Type.GHOST)) {
+          applyTag = true;
+        }
```
innerthunder, line 105: These should probably be handled in their respective tags' `canAdd` methods.

If that's not possible, I feel like this would be cleaner as a `switch` on tag type

src/data/move-attrs/add-battler-tag-all-opps-attr.ts
```diff
@@ -0,0 +1,180 @@
...
+      });
+      return result;
+    }
+
+    return false;
+  }
+
+  override getCondition(): MoveConditionFunc | null {
+    return this.failOnOverlap ? (user, target, _move) => !(this.selfTarget ? user : target).getTag(this.tagType) : null;
+  }
```
innerthunder, line 121:
```suggestion
```

src/data/move-attrs/add-battler-tag-all-opps-attr.ts
```diff
@@ -0,0 +1,180 @@
...
+      case BattlerTagType.PROTECTED:
+      case BattlerTagType.FLYING:
+      case BattlerTagType.CRIT_BOOST:
+      case BattlerTagType.ALWAYS_CRIT:
+        return 5;
+      default:
+        console.warn(`BattlerTag ${BattlerTagType[this.tagType]} is missing a score!`);
+        return 0;
+    }
+  }
```
innerthunder, line 171: Should be evaluated for each opponent of the user, not for the target. Not a big deal though since I want to remove these scoring functions anyway 😅

src/data/move-attrs/multi-status-all-opps-effect-attr.ts
```diff
@@ -0,0 +1,62 @@
...
+    effects: StatusEffect[],
+    sameStatus: boolean = false,
+    selfTarget?: boolean,
+    turnsRemaining?: number,
+    overrideStatus?: boolean,
+  ) {
+    super(effects[0], selfTarget, turnsRemaining, overrideStatus);
+    this.effects = effects;
+    this.sameStatus = sameStatus;
+  }
```
innerthunder, line 32: `selfTarget` doesn't make much sense as a param here. I think it should just be set to `true`, since the effect bypasses substitutes (need to verify, should be `false` otherwise).

`turnsRemaining` and `overrideStatus` may also not apply here; they're only used for Rest afaik

src/data/move-attrs/add-battler-tag-all-opps-attr.ts
Tempo-anon: There's only one gmax move that does (G-max snooze) but it's also single target so I could probably use the other attr and just make this a guaranteed check, though this is mainly for futureproofing if we want to do things that mess with chance procs

src/data/battler-tags.ts
flx-sta: I also wonder, if this is safe to be a `number` or must be a `NumberHolder`. I'm unfortunately not able to tell

Tempo-anon:
PR too large, will break this down into smaller PRs

PR #287, opened 2025-01-14
[Beta][Feature] Use `Set` for seen tutorials

flx-sta:

They may have all tutorials displayed again as there is no mechanic to "adapt" from the `boolean[]` pattern to the `Set<Tutorial>` pattern

## Why am I making these changes

I proposed/suggested them in #270  via #286 but then #270 was merged in the meantime so now I'm proposing it as a standalone PR

> [!NOTE] 
> This is a pending decision to be taken by @MokaStitcher

Replaces the current
```json
[true, null, null, true]
```
way of storing seen tutorials with
```json
[0, 3]
```

To avoid redundancy of numbers I've used a `Set<Tutorial>` (which is effectively a `Set<number>`).

## How to test the changes?

1. Start the game
2. Let the initial tutorial "rain on you"
3. Press "New Game"
4. Inspect the localstorage item `pkty/seenTutorials`
    - It should say something like `[0, 3]` instead of ~~`[true, null, null, true]`

## Checklist 

    - Reference to #270  


torranx (commented):
I initially thought this needs data migration but realized we're doing a fresh start :pikasquint:

src/system/game-data.ts
```diff
@@ -720,10 +730,10 @@ export class GameData {
    */
   public saveTutorialAsSeen(tutorial: Tutorial): boolean {
     const key = getDataTypeKey(GameDataType.TUTORIALS);
-    const tutorials = this.getTutorialFlags();
-    tutorials[tutorial] = true;
+    const tutorials = this.getSeenTutorialsSet();
+    tutorials.add(tutorial);
```
Tempo-anon, line 734: modifying a const?

flx-sta: I'm not modifying the `const`. It's always referencing the same `Set` created initially.
I'm merely adding a value to the set, which isn't a modification of the const itself

Tempo-anon: weird but okay

Tempo-anon (approved):

torranx (approved):

PR #292, opened 2025-01-15
[Misc] Update CREDITS.md

DayKev:
N/A

To keep the credits file updated.


f-fsantos (approved):

torranx (approved):

CREDITS.md
```diff
@@ -1,665 +1,5 @@
-# 🎵 Music
+# Pokérogue Contributors
+For a list of Pokérogue contributors, see: https://github.com/pagefaultgames/pokerogue/blob/62e4a7ec54e4fdee78503adfd55cd44c00e3b399/CREDITS.md
```
Tempo-anon, line 2: Whys is this pointing to a blob instead of beta/main? https://github.com/pagefaultgames/pokerogue?tab=readme-ov-file#-credits

Can also just link the credits https://github.com/pagefaultgames/pokerogue/blob/beta/CREDITS.md

Tempo-anon (changes requested):

CREDITS.md
flx-sta: Cause it's a permanent link thus avoiding linking to a _wrongfully_ updated version of it or basically linking to the point the fork was created

flx-sta (approved):

CREDITS.md
f-fsantos: Ideally we link to the point of the fork since that's when they contributed to the current codebase, anything afterwards is/should be irrelevant to this repository

Tempo-anon: That's fair. I'm good so long as the link points to when the fork happened

Tempo-anon (approved):
Make sure the link is to when the fork happened

CREDITS.md
DayKev: Yeah, I made it a permalink to the current version of the file as it exists right now (it hasn't been updated to include anyone that contributed to PR after the fork happened yet). Future changes won't be reflected in the link.

PR #294, opened 2025-01-15
[Move] Add gmax arena damage over time effects

Tempo-anon:

None

* There was interest in adding G-Max moves
* Breaking up #283 into smaller chunks

* Uncommented out G-Max moves and enums
* Created `arena-tag.TypeImmuneDamageOverTimeTag`

## Screenshots/Videos

n/a

## How to test the changes?

```ts
const overrides = {
  MOVESET_OVERRIDE: [Moves.COURT_CHANGE, Moves.G_MAX_VINE_LASH],
  OPP_MOVESET_OVERRIDE: [Moves.G_MAX_VINE_LASH]
}
```

## Checklist

  - [x] If so, please leave a link to it here: https://github.com/Despair-Games/poketernity-locales/pull/17


Tempo-anon:
Tests are failing because the json for all moves is missing gmax moves

test/moves/all_moves.test.ts
```diff
@@ -49,11 +49,11 @@ describe("All Moves", async () => {
...
 
-  const filename = resolve("./test/moves/all_moves.json");
-  const file = readFileSync(filename, { encoding: "utf-8" });
-  const moveData: MoveData[] = JSON.parse(file);
+  // const filename = resolve("./test/moves/all_moves.json");
+  // const file = readFileSync(filename, { encoding: "utf-8" });
+  // const moveData: MoveData[] = JSON.parse(file);
 
-  it.each(moveData)("$identifier, if implemented, should have correct move data", async (move: MoveData) => {
+  it.todo("$identifier, if implemented, should have correct move data", async (move: MoveData) => {
```
PigeonBar: Even though we don't have a consensus on the desired ordering of the `Moves` enum, I think that moving all G-Max moves to the end of both `Moves` and `allMoves` would be a better temporary solution than disabling the all_moves test.

Tempo-anon: Correct me if I'm wrong, but I believe the orderings of the `Moves` enum is the internal id of the move, ref: https://bulbapedia.bulbagarden.net/wiki/List_of_moves

Looks like while max moves and z-moves appear in that list, G-Max moves are in their own list which is :pikastare: I guess for now they can go at the very beginning or very end of the `Moves` enum and the `all_moves.test` but we also now have two copies of the same data (and we would need to generate the json for it)

test/moves/all_moves.json
```diff
@@ -16470,6 +16470,468 @@
       "effect_chance": 50,
       "flags": ""
     },
+    {
+      "id": 8001,
+      "identifier": "g-max-wildfire",
+      "generation_id": 8,
+      "type_id": 9,
```
PigeonBar: A bunch of these types are off by 1 because PokeAPI starts its type ID's from 1 instead of 0:

https://github.com/Despair-Games/poketernity/blob/33e955a900051cb7cfd9b08d038e17a8bef6624c/test/moves/all_moves.test.ts#L59

test/moves/all_moves.json
```diff
@@ -16470,6 +16470,468 @@
...
+      "effect_chance": -1
+    },
+    {
+      "id": "8014",
+      "identifier": "g-max-finale",
+      "generation_id": "8",
+      "type_id": "17",
+      "power": "10",
+      "pp": "10",
+      "accuracy": "-1",
```
PigeonBar: The json values being strings instead of numbers can also cause `all_moves.test.ts` to fail

src/data/arena-tag.ts
```diff
@@ -1389,6 +1389,74 @@ class GrassWaterPledgeTag extends ArenaTag {
...
+  private getAnimationForType() {
+    switch (this.immuneType) {
+      case Type.GRASS:
+        return CommonAnim.WRAP;
+      case Type.FIRE:
+        return CommonAnim.FIRE_SPIN;
+      case Type.WATER:
+        return CommonAnim.WHIRLPOOL;
+      case Type.ROCK:
+        return CommonAnim.SAND_TOMB;
```
PigeonBar:
```suggestion
        return CommonAnim.SALT_CURE;
```
This animation might work better

PigeonBar (dismissed):

src/data/arena-tag.ts
```diff
@@ -1389,6 +1389,74 @@ class GrassWaterPledgeTag extends ArenaTag {
...
+    globalScene.queueMessage(i18next.t(localeKey));
+  }
+
+  override lapse(arena: Arena): boolean {
+    const field: Pokemon[] =
+      this.side === ArenaTagSide.PLAYER ? globalScene.getPlayerField() : globalScene.getEnemyField();
+
+    field
+      .filter((pokemon) => !pokemon.isOfType(this.immuneType) && !pokemon.switchOutStatus)
+      .forEach((pokemon) => {
```
PigeonBar, line 1443:
```suggestion
      .forEach((pokemon) => {
        const cancelled = new BooleanHolder(false);
        applyAbAttrs(BlockNonDirectDamageAbAttr, pokemon, false, cancelled);
        if (cancelled.value) {
          return;
        }
```

PigeonBar: Since this code seems to be copied from FireGrassPledgeTag, it means that FireGrassPledgeTag likely also has the same bug

Tempo-anon: Excellent catch! I'll apply the magic guard check here and we can put the firegrass fix in a different PR

test/moves/gmax_dot_arena_test.ts
```diff
@@ -0,0 +1,77 @@
+import { BattlerIndex } from "#enums/battler-index";
+import { Abilities } from "#enums/abilities";
+import { Moves } from "#enums/moves";
+import { Species } from "#enums/species";
+import { GameManager } from "#test/testUtils/gameManager";
+import Phaser from "phaser";
+import { afterEach, beforeAll, beforeEach, describe, it, expect } from "vitest";
+
+describe("Moves - Pledge Moves", () => {
```
PigeonBar:
```suggestion
describe("Moves - G-Max Damage-over-time Effects", () => {
```

test/moves/gmax_dot_arena_test.ts
PigeonBar, line 1: File extension needs to be `.test.ts`

test/moves/gmax_dot_arena.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    game.move.select(Moves.G_MAX_WILDFIRE, 0, BattlerIndex.ENEMY);
+    game.move.select(Moves.SPLASH, 1);
+
+    await game.setTurnOrder([BattlerIndex.PLAYER, BattlerIndex.PLAYER_2, BattlerIndex.ENEMY, BattlerIndex.ENEMY_2]);
+    await game.phaseInterceptor.to("MoveEndPhase", false);
+
+    game.scene.getEnemyField().forEach((pokemon) => {
+      expect(pokemon.hp).toBeLessThanOrEqual((pokemon.getMaxHp() * 5) / 6);
+    });
```
PigeonBar:
```suggestion

    const enemyParty = game.scene.getEnemyParty();
    const enemyStartingHp = enemyParty.map(p => p.hp);

    await game.toNextTurn();

    const enemyEndingHp = enemyParty.map(p => p.hp);

    enemyParty.forEach((pokemon, index) => {
      const damage = enemyStartingHp[index] - enemyEndingHp[index];
      expect(damage).toBe(toDmgValue(pokemon.getMaxHp() / 6));
    });
```
Checks the damage dealt by the arena tag more precisely. You can also do something similar for the other tests.

(Also, the arena tag's damage doesn't activate until `await game.toNextTurn()` is run.)

PigeonBar (approved):

innerthunder (approved):

PR #295, opened 2025-01-16
[Refactor] Add `AbstractConstructor` util + Convert parent move/ability attributes to `abstract`

innerthunder:

N/A

Trying to make implementations less error-prone.

- `utils`: Added a new util `AbstractConstructor`, which is basically an abstract version of `Constructor`. Using this instead of `Constructor` for type checks (i.e. `instanceof`) allows for comparing against abstract types.
- Some parent classes for arena tags, move attributes, and ability attributes are now `abstract`.

## How to test the changes?

`npm run test`

## Checklist

### DIFF
--- a/src/data/move-attrs/variable-def-attr.ts
+++ b/src/data/move-attrs/variable-def-attr.ts
@@ -8,7 +8,6 @@ import type { NumberHolder } from "#app/utils";
  * @extends MoveAttr
  */
 export abstract class VariableDefAttr extends MoveAttr {
-
   /**
    * Changes the defensive stat used in the current attack's damage calculation
    * @param _user the {@linkcode Pokemon} using the move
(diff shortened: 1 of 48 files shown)


src/data/move-attrs/variable-def-attr.ts
```diff
@@ -7,7 +7,7 @@ import type { NumberHolder } from "#app/utils";
  * Attribute to change the defensive stat to be used in a move's damage calculations.
  * @extends MoveAttr
  */
-export class VariableDefAttr extends MoveAttr {
+export abstract class VariableDefAttr extends MoveAttr {
   constructor() {
     super();
   }
```
DayKev:
```suggestion
```

src/data/move-attrs/variable-def-attr.ts
```diff
@@ -7,10 +7,7 @@ import type { NumberHolder } from "#app/utils";
  * Attribute to change the defensive stat to be used in a move's damage calculations.
  * @extends MoveAttr
  */
-export class VariableDefAttr extends MoveAttr {
-  constructor() {
-    super();
-  }
+export abstract class VariableDefAttr extends MoveAttr {
```
DayKev:
```suggestion
```
Apparently there was an extra newline hidden by github.

DayKev (dismissed):

DayKev (approved):

torranx (approved):

PR #275, opened 2025-01-14
[MISC] Vite 6.0.7  | vitest 2.1.8

flx-sta:
## Why am I making these changes?

Stay up to date

## What are the changes from a developer perspective?

> [!IMPORTANT]
> Major update for vite
> From ~~`5.4.8`~~ -> `6.0.7`

- update vitest from ~~`2.0.4`~~ -> `2.1.8`
    - <sub>update `@vitest/coverage-istanbul` from ~~`2.0.4`~~ -> `2.1.8`</sub>
    

## How to test the changes?

- Run a build
- Run the dev script
- Run all scripts and see if they still work

## Checklist

- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?

- [x] Are all unit tests still passing? (`npm run test:silent`)

PR #304, opened 2025-01-17
[Test][Security] Upgrade to vitest `3.0.2`, update vitest configuration

flx-sta:

Vitest 3 was finally released today and since it fully supports vite 6 we should adopt this too.

- upgrade `vitest` to `3.0.2`
- upgrade @vitest/coverage-istanbul to `3.0.2`
- fix stockpiling.test.ts*
    - [*]: spys no longer get reset since they are being re-used as long as they stay in the same scope 
- fix `sequence.sequencer` in vitest config
- remove vitest workspaces/projects
    - That system is not really in use anymore. We just never really removed it properly.
- Reduce testing shards down to `5` (from ~~`10`~~) 
   - Some trial & error has shown that there is no difference in total time. We can scale it up again whenever necessary 

> [!NOTE]
> Closes https://github.com/Despair-Games/poketernity/security/dependabot/3

## How to test the changes?

```sh
npm run test:silent
```

## Checklist


test/battlerTags/stockpiling.test.ts
```diff
@@ -149,12 +149,8 @@ describe("BattlerTag - StockpilingTag", () => {
       subject.onOverlap(mockPokemon);
       expect(subject.stockpiledCount).toBe(3);
 
-      vi.spyOn(game.scene, "unshiftPhase").mockImplementationOnce((_phase) => {
-        throw new Error("Should not be called a fourth time");
-      });
-
       // fourth stack should not be applied
-      subject.onOverlap(mockPokemon);
+      // subject.onOverlap(mockPokemon);
```
PigeonBar:
```suggestion
      subject.onOverlap(mockPokemon);
```
(The test is testing that calling `onOverlap()` a fourth time does not apply a fourth layer of the Stockpiling tag)

flx-sta: Oh whops, this must have slipped through while debugging. Thanks

PigeonBar:
When running the tests, I am getting the same "Failed to load source map for node_modules/i18next-korean-postposition-processor/index.js" error that we got while trying vitest 2.1.6 in https://github.com/Despair-Games/poketernity/pull/275#issuecomment-2588924217. The tests still run perfectly fine though.

One hacky solution I've found is in `i18n.ts`, to replace the `import ... from "i18next-korean-postposition-processor"` statement with an `await import(...)` statement inside `initI18n()`:

image

This may require someone who is fluent in Korean to verify that Korean text still gets translated properly.

innerthunder (dismissed):

package.json
```diff
@@ -12,16 +12,16 @@
...
-    "test:cov": "vitest run --project main --coverage",
+    "test:cov": "vitest run --coverage",
     "posttest:cov": "git rev-parse --short HEAD",
 
-    "test:watch": "vitest watch --project main --coverage",
+    "test:watch": "vitest watch --coverage",
     "posttest:watch": "git rev-parse --short HEAD",
 
-    "test:silent": "vitest run --project main --silent --no-isolate",
+    "test:silent": "vitest run --silent --no-isolate",
```
DayKev, line 24: I'd like to replace all the `:` with `-` in these commands.

package.json
```diff
@@ -31,14 +31,10 @@
     "eslint-ci": "eslint .",
 
     "prettier": "prettier . --write",
-
```
DayKev: We shouldn't have inconsistent spacing, either keep the spacing or remove it all (as the one that added the spacing for distinct commands I'd argue to keep it but I don't feel too strongly about it).

torranx (dismissed):
blocking this pr with regards to @PigeonBar 's comment

package.json
flx-sta: It should be removed in total.
it's a hassle to re-add it every time I ran 
```shell
npm i <module>@<version>
```
But i wanted to keep the PR scope intact so for now, no changes

package.json
flx-sta: you mean `test-silent`?
I mean. I couldn't care less but that's a very common pattern in package.json scripts

flx-sta:
> When running the tests, I am getting the same "Failed to load source map for node_modules/i18next-korean-postposition-processor/index.js" error that we got while trying vitest 2.1.6 in [#275 (comment)](https://github.com/Despair-Games/poketernity/pull/275#issuecomment-2588924217). The tests still run perfectly fine though.
> 
> One hacky solution I've found is in `i18n.ts`, to replace the `import ... from "i18next-korean-postposition-processor"` statement with an `await import(...)` statement inside `initI18n()`:
> 
> ![image]
> 
> This may require someone who is fluent in Korean to verify that Korean text still gets translated properly.

Another way is to exclude it from dependency optimization:
```ts
export const defaultConfig: UserConfig = {
  //...
  
  optimizeDeps: {
    exclude: ["i18next-korean-postposition-processor"],
  }
  
  //...
}
```

package.json
DayKev: I mentioned it because there were some changes (such as at the line I commented on) but only some of the whitespace was removed.

package.json
DayKev: Interesting. The `:` is annoying because occasionally I type too fast and hold shift for a fraction of a second too long which causes the command to become something like `npm run start:Dev` which doesn't work because they're all case sensitive.

PigeonBar:
> Another way is to exclude it from dependency optimization:
> 
> ```ts
> export const defaultConfig: UserConfig = {
>   //...
>   
>   optimizeDeps: {
>     exclude: ["i18next-korean-postposition-processor"],
>   }
>   
>   //...
> }
> ```

I tried out this solution, it unfortunately did not work on my device.

flx-sta:
> > Another way is to exclude it from dependency optimization:
> > ```ts
> > export const defaultConfig: UserConfig = {
> >   //...
> >   
> >   optimizeDeps: {
> >     exclude: ["i18next-korean-postposition-processor"],
> >   }
> >   
> >   //...
> > }
> > ```
> 
> I tried out this solution, it unfortunately did not work on my device.

Neither for me anymore, lol..
Guess I didn't check enough, sorry

package.json
flx-sta: You should make a classic poll

package.json
flx-sta: Probably worth a poll

package.json
flx-sta: done

flx-sta:
> When running the tests, I am getting the same "Failed to load source map for node_modules/i18next-korean-postposition-processor/index.js" error that we got while trying vitest 2.1.6 in [#275 (comment)](https://github.com/Despair-Games/poketernity/pull/275#issuecomment-2588924217). The tests still run perfectly fine though.
> 
> One hacky solution I've found is in `i18n.ts`, to replace the `import ... from "i18next-korean-postposition-processor"` statement with an `await import(...)` statement inside `initI18n()`:
> 
> image
> 
> This may require someone who is fluent in Korean to verify that Korean text still gets translated properly.

As far as I am concerned it seems to work like a charm. 
I also don't see why It shouldn't, as it's simply a lazy import, instead of eager
https://github.com/Despair-Games/poketernity/pull/304/commits/1b9e45818ada93b303003e13ad8e19865c57cd66

DayKev:
Huh, something weird happened the first time I ran the tests and a few of them failed for some reason (and the post-command didn't get run either).
image
Worked the second time though.
image

I ran `npm ci` first, then both test runs were done with no other commands between them. It's probably not an issue? Just documenting it in case it's relevant later.

DayKev (dismissed):

PigeonBar:
> Huh, something weird happened the first time I ran the tests and a few of them failed for some reason (and the post-command didn't get run either).

After some investigation, it turns out that this error is consistently reproducible by adding an `await new Promise<void>(resolve => setTimeout(() => resolve(), 1000));` inside `initI18n()`. So, it's possible that adding lazy imports to `initI18n()` can make the same error occur some of the time.

My best guess is that this occurs due to the scattered `initI18n()` calls that do not properly `await` it. For example:

https://github.com/Despair-Games/poketernity/blob/f212ecbdda4c2049a9f66b655c9c344a4e4f3dc2/src/data/trainer-config.ts#L323-L334

This seems to not cause problems for regular gameplay because when booting up the game normally (not during tests), there is a proper `await initI18n()` during `startGame()`:

https://github.com/Despair-Games/poketernity/blob/f212ecbdda4c2049a9f66b655c9c344a4e4f3dc2/src/main.ts#L37-L40

PigeonBar:
Here's a potential solution, which calls `await initI18n()` as soon as it is being imported:

image

```
vi.mock("#app/plugins/i18n", async (importOriginal) => {
  const importedStuff: any = await importOriginal();
  const { initI18n } = importedStuff;
  await initI18n();
  return importedStuff;
});
```

flx-sta:
> Here's a potential solution, which calls `await initI18n()` as soon as it is being imported:
> 
> image
> 
> ```
> vi.mock("#app/plugins/i18n", async (importOriginal) => {
>   const importedStuff: any = await importOriginal();
>   const { initI18n } = importedStuff;
>   await initI18n();
>   return importedStuff;
> });
> ```

Thanks. I’ll have a look as soon as I can. I wish we didn’t even have those scattered init i18n calls as they make no sense whatsoever

DayKev (approved):

PigeonBar (approved):

PR #323, opened 2025-01-19
[Refactor] Various improvements to achievements

DayKev:
N/A

Cleaning up some 🍝 

- Unused `name` and `description` parameters removed from `Achv` class and subclass constructors.
- `readonly` added to some class fields.
- `args?: any[]` replaced with `...args: unknown[]` pattern.
- Add `public` marker to classes.
- Refactor description code (Moka).
- Create Mono-Gen and Mono-Type achievement classes (Moka).
- Misc minor improvements.

## Screenshots/Videos
<details>
  <summary>Various screenshots</summary>

image
image
image
image
image
image
image
image

</details>

## How to test the changes?
`npm run test achievement`
Go through the in-game achievements menu, attempt to gain an achievement, etc.

## Checklist

### DIFF
--- a/src/system/achv.ts
+++ b/src/system/achv.ts
@@ -23,7 +23,7 @@ export class Achv {
   protected readonly localizationKey: string;
   protected descriptionKey: string;
   protected descriptionLocArgs: Record<string, unknown>;
-  protected readonly iconImage: string;
+  protected readonly _iconImage: string;
   public readonly score: number;
   public readonly id: string;
 
@@ -36,7 +36,7 @@ export class Achv {
   constructor(localizationKey: string, iconImage: string, score: number, conditionFunc?: ConditionFn) {
     this.localizationKey = localizationKey;
     this.descriptionKey = localizationKey;
-    this.iconImage = iconImage;
+    this._iconImage = iconImage;
     this.score = score;
     this.conditionFunc = conditionFunc;
   }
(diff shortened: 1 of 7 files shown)


MokaStitcher:
I had a look at what I had done back then with achievements and made a PR with some of the things that are not in this one. Most of the changes revolve around getting rid of the terrible `getAchievementDescription` function #325 
Feel free to do whatever with this PR, make any changes directly or outright reject it.
At least it's no longer on the back of my mind now :)
(I trusted past me and didn't playtest everything again, but i will do it properly tomorrow)

MokaStitcher (dismissed):

src/system/achv.ts
```diff
@@ -95,679 +95,213 @@ export class Achv {
...
-    "MONO_GEN_ONE",
-    "",
-    "MONO_GEN_ONE.description",
-    "ribbon_gen1",
-    100,
-    (c) =>
-      c instanceof SingleGenerationChallenge
-      && c.value === 1
-      && !globalScene.gameMode.challenges.some((c) => c.id === Challenges.INVERSE_BATTLE && c.value > 0),
+    () => globalScene.gameData.gameStats.sessionsWon === 0,
```
PigeonBar: Should we get rid of this check for the "Classic Victory" achv? There is already code elsewhere to prevent the player from repeatedly unlocking the same achv, and this check could lock a player out of obtaining this achv if they somehow win a run without obtaining this achv.

MokaStitcher: That makes sense actually. I'm pretty sure this was done because of past issues of the voucher getting awarded multiple times. It shouldn't be needed for us since we are starting from scratch so we don't have legacy accounts in weird states to deal with

DayKev: Sure, done.

PigeonBar (approved):

Tempo-anon (approved):
We should pick which achievements we want to keep (if any) and which ones to scrap before launch

PR #325, opened 2025-01-19
[Refactor] Cleanup achievement description code

MokaStitcher:
An addition to PR #323 based on things I had done months ago

Mostly:
- The achievement class handles its description instead of the old `getAchievementDescription` function
- The player gender is no longer passed as an argument to `getName` and `getDescription`, as it is now directly available from the settings manager
- Achievement subclasses take care of setting the correct localization key for the description as well as any argument that needs to be passed to i18n. For this purpose, subclasses have been created for MonoGen and monoType achievements (which also greatly reduces the duplicate code)
- Tests and UI classes updated to accommodate the changes

### DIFF
--- a/src/system/achv.ts
+++ b/src/system/achv.ts
@@ -17,14 +17,15 @@ import { Challenges } from "#enums/challenges";
 import { globalScene } from "#app/global-scene";
 import { settings } from "./settings/settings-manager";
 import { AchvTier } from "#enums/achv-tier";
+import { Type } from "#enums/type";
 
 export class Achv {
-  public readonly localizationKey: string;
-  public readonly id: string;
-  public name: string = "";
-  public description: string = "";
+  protected readonly localizationKey: string;
+  protected descriptionKey: string;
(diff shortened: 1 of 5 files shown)


src/system/achv.ts
```diff
@@ -34,22 +35,34 @@ export class Achv {
...
     this.conditionFunc = conditionFunc;
   }
 
   /**
    * Get the name of the achievement based on the gender of the player
-   * @param playerGender - the gender of the player (default: {@linkcode PlayerGender.UNSET})
    * @returns the name of the achievement localized for the player gender
    */
-  public getName(playerGender: PlayerGender = PlayerGender.UNSET): string {
+  public getName(): string {
```
Tempo-anon, line 48: If there isn't an input variable should we be moving to `get name()`?

DayKev: Sure. Converted them over in the main PR.

MokaStitcher: ah I like that, done

Issue #184, opened 2025-01-06
[Move][P2] Body Press and Foul Play do not calculate Attack modifiers correctly


frutescens:
## Foul Play
> Foul Play inflicts damage. It uses the target's Attack stat to calculate damage instead of the user's.
The target's Attack stat stage-modifiers are applied (rather than the user's), but otherwise the user's Attack modifiers are used (including held item, Ability, and burn). For example, the user's Huge Power or held Choice Band will increase the stat used in calculation, whereas the target's will not. 

## Body Press
> Body Press inflicts damage, but it uses the user's Defense stat instead of its Attack stat to calculate damage.
The user's Defense stat stage-modifiers are applied (rather than Attack stage-modifiers), but otherwise its Attack modifiers are used (including held item, Ability, and burn). For example, the user's Huge Power or a held Choice Band will increase the stat used in calculation, whereas Fur Coat or a held Eviolite will not. In contrast, the user's Slow Start will decrease the stat used in calculation. 

Foul Play's attribute `TargetAtkUserAtkAttr` and Body Press' attribute `DefAtkAttr` does not consider the user's Huge Power/attack modifiers correctly in getEffectiveStat() 

## Past Bug Reports
https://github.com/pagefaultgames/pokerogue/issues/5036

PR #254, opened 2025-01-11
[Balance] Removed Tokens from the game

frutescens:
## What are the changes the user will see?

The game will not give enemy tokens at the end of every X50 wave in endless. 

## Why am I making these changes?

Part of an larger effort to scrap and overhaul modifiers. 

## What are the changes from a developer perspective?

EnemyPersistentModifier has been removed from the game.
All modifier applications have also been removed. 
In addition, add-enemy-buff phase has been removed from the game and will not be called. 

## Screenshots/Videos

https://github.com/user-attachments/assets/506f99d1-2342-4a01-901c-d9d8039acc45

## How to test the changes?

Try defeating a Pokemon at endless stage 50. No modifier should be added. 

## Checklist

- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test`)
  - [x] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?
- [x] Have I provided screenshots/videos of the changes (if applicable)?

PR #329, opened 2025-01-20
[Refactor][Ability] Add `REVEALED` and `IGNORE` modes for ability application

innerthunder:

- All abilities on a Pokemon are now tracked individually for the purpose of "revealing" them to type hints and the enemy AI. Revealing a Pokemon's ability in this way should no longer also reveal its passive.
- Field-affecting abilities (e.g. Friend Guard) are now properly ignored by the enemy AI if not already revealed.
- Abilities are now completely ignored during Shell Side Arm, Photon Geyser, Tera Blast, and Tera Starstorm's move category resolution. **This fully implements Tera Blast and Tera Starstorm**.
- Download now also ignores the target's abilities when determining its offensive stat to boost.
- Abilities given by Trace, Mummy, and other ability-changing abilities, along with Frisk, now immediately reveal the affected Pokemon's abilities to type hints and the AI.
- This PR has a side effect where Body Press and Foul Play now ignore the effects of the target's Unaware ability. This should be fixed alongside #184  in a future PR.

Cleaning up simulated/revealing abilities before the AI rework.

Several AI-facing functions in `Pokemon` for damage calculation currently have parameters to restrict ability application, e.g.

```ts
getAttackDamage(
  source: Pokemon,
  move: Move,
  ignoreAbility: boolean = false,
  ignoreSourceAbility: boolean = false,
  isCritical: boolean = false,
  simulated: boolean = true,
  effectiveness?: TypeDamageMultiplier,
): DamageCalculationResult;
```

This PR replaces these parameters with a single `AbilityApplyMode` enum that defines how abilities are applied:

```ts
getAttackDamage(
  source: Pokemon,
  move: Move,
  abilityApplyMode: AbilityApplyMode = AbilityApplyMode.DEFAULT,
  isCritical: boolean = false,
  simulated: boolean = true,
  effectiveness?: TypeDamageMultiplier,
): DamageCalculationResult;
```

The enum currently supports 3 modes for applying abilities:
- `DEFAULT`: Abilities are applied without any restrictions (as long as they aren't suppressed/ignored and meet their conditions)
- `REVEALED`: Abilities are only applied if they have previously been applied or otherwise revealed in the current wave. This is mainly used for type hints and the Enemy AI during move selection.
- `IGNORE`: Abilities are not applied entirely. This is mainly used for Shell Side Arm and similar moves during their forecasting to evaluate move category.

Also, the `abilityRevealed` flag in `Pokemon.battleData` has been replaced with an `abilitiesRevealed` array to track abilities individually as they are revealed. This array is updated in almost the same way as `abilitiesApplied`, except that it's also updated when an ability such as Frisk or Trace replaces a Pokemon's ability.

## Screenshots/Videos

The following screenshots are from sessions with the overrides below. Both Pokemon are Bronzongs with Pressure and Levitate and with the moves Bulldoze and Covet.

```ts
const overrides = {
  STARTER_SPECIES_OVERRIDE: Species.BRONZONG,
  ABILITY_OVERRIDE: Abilities.PRESSURE,
  PASSIVE_ABILITY_OVERRIDE: Abilities.LEVITATE,
  ENEMY_SPECIES_OVERRIDE: Species.BRONZONG,
  ENEMY_ABILITY_OVERRIDE: Abilities.PRESSURE,
  ENEMY_PASSIVE_ABILITY_OVERRIDE: Abilities.LEVITATE,
  MOVESET_OVERRIDE: [Moves.BULLDOZE, Moves.COVET],
  ENEMY_MOVESET_OVERRIDE: [Moves.BULLDOZE, Moves.COVET]
```

### Type Hints

**Before**: Pressure reveals itself *and* Levitate, modifying Bulldoze's type hint.

image

**After**: Pressure reveals itself, but Bulldoze still shows as 2x effective.

image

### Enemy AI

**Before**: Because Levitate was "revealed" by Pressure's activation, the AI avoids using Bulldoze.

image

(Bulldoze is scored -20 points; Covet is scored 5)

**After**: The AI no longer detects Levitate, giving Bulldoze a score of 32 points since it's seen as super effective.

image

## How to test the changes?

- `npm run test shell_side_arm` (1 new test)
- `npm run test photon_geyser`

## Checklist


src/data/all-moves.ts
Tempo-anon, line 1: Should add some comments in body press and foul play about ignoring unaware now

src/data/ability-filter-options.ts
```diff
@@ -0,0 +1,5 @@
+export interface AbilityFilterOptions {
+  ignoreOverride?: boolean;
+  revealedOnly?: boolean;
+  canApplyOnly?: boolean;
```
Tempo-anon: There should be documentation here just like how there is in `enums/ability-apply-mode`

I also think the `ignoreOverride` name should be renamed to something like `baseAbilityOnly` because overrides make me think of debug overrides

src/field/pokemon.ts
```diff
@@ -1586,6 +1586,37 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+  }
+
+  /**
+   * Obtains the Pokemon's abilities.
+   * @param options Optional flags to filter the output:
+   * - `ignoreOverride`: If `true`, obtains the Pokemon's base ability instead
+   * of overriding abilities (e.g. obtains Trace instead of whatever Trace copies)
+   * - `canApplyOnly`: If `true`, filters out abilities that are suppressed or ignored
+   * - `revealedOnly`: If `true`, filters out abilities that haven't been revealed yet
+   * @returns
```
Tempo-anon: nit: @linkcode the AbilityData or move it to be before this function

src/field/pokemon.ts
```diff
@@ -5924,12 +5924,21 @@ export class PokemonSummonData {
   public addedType: Type | null = null;
 }
 
+/** Container for Pokemon-specific data that resets at the end of each wave. */
 export class PokemonBattleData {
+  /** How many hits the Pokemon has taken */
   public hitCount: number = 0;
-  public endured: boolean = false;
```
Tempo-anon, line 5937: Was `endured` an unused variable?

innerthunder: It was previously only used by Endure Tokens, which were removed with #254

Tempo-anon (dismissed):

Tempo-anon (approved):

flx-sta (approved):

PR #337, opened 2025-01-21
[Move][Checkup] Present RNG Fixes

### LINKED ISSUE #313: [Move][Checkup] Present
## Details
- [x] Passes all_moves.test.ts
- [x] Present can either deal damage against the target or heal it for 1/4 of its maximum HP. If Present deals damage, then the base power of Present is chosen at random from 40, 80, or 120.
- [x] Present's effect is chosen randomly as shown below.
     Damage with 40 base power 	       40%
     Damage with 80 base power 	        30%
     Damage with 120 base power 	        10%
     Heal 1/4 of the target's max HP 	20% 
- [x] If a Pokémon with Parental Bond uses Present and heals the target, it will only hit once. However, if it damages the target, it will hit twice. 
- [x] If Present is used on an ally with Telepathy, Present will always be blocked, even if it would have healed the target. 

## Development Notes
- [x] Does it have a test? Does it need a test? https://github.com/Despair-Games/poketernity/pull/375
- [x] Are the move attributes it uses documented?
- [x] Does it match the basic details above?

## Related PRs
- https://github.com/Despair-Games/poketernity/pull/337
- https://github.com/Despair-Games/poketernity/pull/375

## Related Bugs
- Uses global RNG instead of seeded RNG
- If healing is selected, the user is damaged for 1 HP and then healed.

## References
- [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Present_(move))

frutescens:

Present will no longer use global RNG when it is used. 
Present will also no longer damage the target for 1 HP before healing the target.
Present's outcomes were shifted by 1 before. This PR fixes that so that the chances of each of the 4 outcomes are accurate to mainline.

Move Checkups.
#313 

`firstHit` variable renamed to `isFirstHit`
Present-related test removed from Triage since its unnecessary.
powerSeed generated with user.randSeedInt()
Fixed off-by-1 Present RNG

## How to test the changes?

Using Present should not affect the events of the next wave.

https://github.com/user-attachments/assets/70c72698-7ee6-43c7-b76e-38ab3759aa06

## Checklist


src/data/move-attrs/present-power-attr.ts
```diff
@@ -19,16 +19,16 @@ export class PresentPowerAttr extends VariablePowerAttr {
      * If this move is multi-hit, and this attribute is applied to any hit
      * other than the first, this move cannot result in a heal.
      */
-    const firstHit = user.turnData.hitCount === user.turnData.hitsLeft;
+    const isFirstHit = user.turnData.hitCount === user.turnData.hitsLeft;
```
Tempo-anon: nit: `const isFirstHit = (user.turnData.hitCount === user.turnData.hitsLeft);`

src/data/move-attrs/present-power-attr.ts
```diff
@@ -19,16 +19,16 @@ export class PresentPowerAttr extends VariablePowerAttr {
...
-    const powerSeed = randSeedInt(firstHit ? 100 : 80);
+    const powerSeed = user.randSeedInt(isFirstHit ? 100 : 80);
     if (powerSeed <= 40) {
       power.value = 40;
     } else if (40 < powerSeed && powerSeed <= 70) {
       power.value = 80;
     } else if (70 < powerSeed && powerSeed <= 80) {
       power.value = 120;
-    } else if (80 < powerSeed && powerSeed <= 100) {
+    } else if (80 < powerSeed) {
```
Tempo-anon: nit: this can just be an `else`

src/data/move-attrs/present-power-attr.ts
```diff
@@ -19,16 +19,16 @@ export class PresentPowerAttr extends VariablePowerAttr {
      * If this move is multi-hit, and this attribute is applied to any hit
      * other than the first, this move cannot result in a heal.
```
Tempo-anon, line 23: Can you clarify this? Reading the bulbapedia article seems to imply that if the first hit is a heal then the rest of the hits don't happen

Tempo-anon (dismissed):

src/data/move-attrs/present-power-attr.ts
frutescens: Yes. That's pretty much it + it is elaborated upon in the comment 2-3 lines below

src/data/move-attrs/present-power-attr.ts
frutescens: True, but I think it does a good way of laying out all the possibilities.

Tempo-anon (dismissed):

torranx (dismissed):

src/data/move-attrs/present-power-attr.ts
```diff
@@ -19,16 +19,16 @@ export class PresentPowerAttr extends VariablePowerAttr {
...
-    const firstHit = user.turnData.hitCount === user.turnData.hitsLeft;
+    const isFirstHit = user.turnData.hitCount === user.turnData.hitsLeft;
 
-    const powerSeed = randSeedInt(firstHit ? 100 : 80);
+    const powerSeed = user.randSeedInt(isFirstHit ? 100 : 80);
     if (powerSeed <= 40) {
       power.value = 40;
     } else if (40 < powerSeed && powerSeed <= 70) {
       power.value = 80;
     } else if (70 < powerSeed && powerSeed <= 80) {
```
PigeonBar:
```suggestion
    if (powerSeed < 40) {
      power.value = 40;
    } else if (powerSeed < 70) {
      power.value = 80;
    } else if (powerSeed < 80) {
```
(`powerSeed` ranges from 0 to 99.)

PigeonBar:
I think there's just one small bug remaining where if Present heals the target, it still also does 1 damage:

https://github.com/user-attachments/assets/5e5c8ffc-e472-4f83-951e-8b58a01a0511

Tempo-anon:
Probably caused because of `toDmgValue` making all attacks do a min of 1 damage. Does the 1 tick of damage happen before or after the healing applies?

innerthunder:
> Probably caused because of `toDmgValue` making all attacks do a min of 1 damage. Does the 1 tick of damage happen before or after the healing applies?

The video seems to show it happens after (which makes sense -- the heal phase is pushed when move power is resolved and before damage is finalized/rounded up)

PigeonBar:
I think the damage actually happens before the healing, since if Present heals a Pokemon that starts at full HP, the Pokemon also ends at full HP:

https://github.com/user-attachments/assets/b085c7bc-0993-444c-a83a-0622e14bee10

PigeonBar:
I also tried a potential fix where the healing functionality instead gets handled by a `PresentHealAttr` that extends `OverrideMoveEffectAttr`. The problem would be that `OverrideMoveEffectAttr`s get applied before any accuracy checks, and Present has 90% accuracy, so the move could heal when it should otherwise miss.

I think the next idea to try would be to replicate what Pollen Puff does to heal the user's ally without damaging the user's ally.

innerthunder:
I think a better way to fix it would be to somehow stop `MoveEffectPhase.applyMove` early if `baseDamage <= 0` (Present's base power is -1 internally, and doesn't get overridden if the move ends up healing)

flx-sta (dismissed):

flx-sta (changes requested):
Heal & Damage at the same time

https://github.com/user-attachments/assets/85800e5a-e662-4d35-914d-867976a3dfaa

<details><summary>[APPLIED CHANGES FOR TESTING]</summary>
<p>

I had this
```ts
.attr(
        ChangeToStatusCategoryAttr,
        (user, _target, _move) => user.turnData.hitCount === user.turnData.hitsLeft && user.randSeedInt(100) < 20,
        (user, _target, move) => {
          user.turnData.hitCount = 1;
          user.turnData.hitsLeft = 1;
          move.addAttr(new HealAttr(0.25, true, false));
        },
      )
```
replaced with 
```ts
.attr(
        ChangeToStatusCategoryAttr,
        (user, _target, _move) => user.turnData.hitCount === user.turnData.hitsLeft && user.randSeedInt(100) < 50, // ⚠️
        (user, _target, move) => {
          user.turnData.hitCount = 1;
          user.turnData.hitsLeft = 1;
          move.addAttr(new HealAttr(0.25, true, false));
        },
      )
```

</p>
</details>

frutescens:
Very confused. I tested it with Singles and Doubles. 

I changed 
image


https://github.com/user-attachments/assets/a8e2a714-442d-47b9-a875-7e4b491545c8

src/data/move-attrs/change-to-status-category-attr.ts
```diff
@@ -0,0 +1,41 @@
...
+import type { NumberHolder } from "#app/utils";
+import type { Move } from "#app/data/move";
+import { VariableMoveCategoryAttr } from "#app/data/move-attrs/variable-move-category-attr";
+
+/**
+ * Change the move category to status when used on the user's ally.
+ * Used for {@link https://bulbapedia.bulbagarden.net/wiki/Pollen_Puff_(move) | Pollen Puff} and {@link https://bulbapedia.bulbagarden.net/wiki/Present_(move) | Present}
+ * @extends VariableMoveCategoryAttr
+ */
+export class ChangeToStatusCategoryAttr extends VariableMoveCategoryAttr {
```
PigeonBar: I think there is a problem where `VariableMoveCategoryAttr` gets applied multiple times per turn (e.g., when an enemy Pokemon is simulating its move choices). This results in multiple independent RNG rolls that can cause Present to heal multiple times, or cause Present to heal and then do damage, etc..

src/data/all-moves.ts
```diff
@@ -886,6 +886,15 @@ export function initMoves() {
       .target(MoveTarget.PARTY),
     new AttackMove(Moves.RETURN, Type.NORMAL, MoveCategory.PHYSICAL, -1, 100, 20, -1, 0, 2).attr(FriendshipPowerAttr),
     new AttackMove(Moves.PRESENT, Type.NORMAL, MoveCategory.PHYSICAL, -1, 90, 15, -1, 0, 2)
+      .attr(
+        ChangeToStatusCategoryAttr,
+        (user, _target, _move) => user.turnData.hitCount === user.turnData.hitsLeft && user.randSeedInt(100) < 20,
+        (user, _target, move) => {
+          user.turnData.hitCount = 1;
+          user.turnData.hitsLeft = 1;
+          move.addAttr(new HealAttr(0.25, true, false));
```
PigeonBar: Also, we should not be adding a HealAttr to the move, since adding a new attr is permanent and will cause future uses of Present to heal even if they are supposed to do damage instead.

PigeonBar (changes requested):

src/data/move-attrs/present-power-attr.ts
```diff
@@ -19,9 +22,9 @@ export class PresentPowerAttr extends VariablePowerAttr {
      * If this move is multi-hit, and this attribute is applied to any hit
      * other than the first, this move cannot result in a heal.
      */
-    const firstHit = user.turnData.hitCount === user.turnData.hitsLeft;
+    const isFirstHit = user.turnData.hitCount === user.turnData.hitsLeft;
 
-    const powerSeed = randSeedInt(firstHit ? 100 : 80);
+    const powerSeed = user.randSeedInt(isFirstHit ? 100 : 80);
     if (powerSeed <= 40) {
```
PigeonBar:
```suggestion
    if (powerSeed < 40) {
```
...and so on for the other RNG thresholds

PigeonBar:
I also recommend adding a `if (power < 0) {return -1}` to the function `getBaseDamage()`.

<sup>Since the formula for base damage adds 2 damage at the end, it is possible under extreme circumstances for a move with -1 power to deal positive damage.</sup>

src/data/move-attrs/present-power-attr.ts
```diff
@@ -19,16 +22,16 @@ export class PresentPowerAttr extends VariablePowerAttr {
      * If this move is multi-hit, and this attribute is applied to any hit
      * other than the first, this move cannot result in a heal.
      */
-    const firstHit = user.turnData.hitCount === user.turnData.hitsLeft;
+    const isFirstHit = user.turnData.hitCount === user.turnData.hitsLeft;
```
PigeonBar:
```suggestion
    const isFirstHit = user.turnData.hitCount === user.turnData.hitsLeft && user.turnData.hitsLeft > 0;
```
(An extra safeguard to prevent simulated calls from healing the target)

PigeonBar (dismissed):

src/data/move-attrs/present-power-attr.ts
```diff
@@ -19,16 +22,16 @@ export class PresentPowerAttr extends VariablePowerAttr {
...
-    const firstHit = user.turnData.hitCount === user.turnData.hitsLeft;
+    const isFirstHit = user.turnData.hitCount === user.turnData.hitsLeft && user.turnData.hitsLeft > 0;
 
-    const powerSeed = randSeedInt(firstHit ? 100 : 80);
-    if (powerSeed <= 40) {
+    const powerSeed = user.randSeedInt(isFirstHit ? 100 : 80);
+    if (powerSeed < 40) {
       power.value = 40;
-    } else if (40 < powerSeed && powerSeed <= 70) {
+    } else if (40 <= powerSeed && powerSeed < 70) {
```
Tempo-anon: nit: Is the `40 <= powerSeed` check necessary since this part will only be entered if the previous `powerSeed < 40` returns false

```ts
if (powerSeed < 40) {
//blah
} else if (powerSeed < 70) {
//blah
}

File: src/data/move-attrs/present-power-attr.ts [hunk 10]
```diff
@@ -19,16 +22,16 @@ export class PresentPowerAttr extends VariablePowerAttr {
...
       power.value = 80;
-    } else if (70 < powerSeed && powerSeed <= 80) {
+    } else if (70 <= powerSeed && powerSeed < 80) {
       power.value = 120;
-    } else if (80 < powerSeed && powerSeed <= 100) {
+    } else if (80 <= powerSeed && powerSeed <= 100) {
```
Tempo-anon: Is the `powerSeed <= 100` needed?

```ts
if (powerSeed < 40) {
  power.value = 40;
} else if (powerSeed < 70) {
  power.value = 80;
} else if (powerSeed < 80) {
  power.value = 120;
} else {
  //do heal stuff
}
```

src/field/pokemon.ts
```diff
@@ -3064,11 +3064,18 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     );
     applyMoveAttrs(VariableDefAttr, source, this, move, targetDef);
 
+    /** This prevents a move with negative power from possibly dealing positive damage.
+     * The issue can occur because the base damage is the result of the below equation plus 2.
```
Tempo-anon, line 3103: Should explain the +2. Is this just for moves with a set damage of -1?

src/field/pokemon.ts
```diff
@@ -3295,28 +3302,28 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+      * typeMultiplier
+      * burnMultiplier.value
+      * screenMultiplier.value
+      * hitsTagMultiplier.value
+      * mistyTerrainMultiplier
+      * tintedLensMultiplier.value
+      * receivedDamageMultiplier.value
+      * alliedFieldDamageMultiplier.value;
+    /** If damage is nullified by a form-ability (Eiscue's Ice Face, Mimikyu's Disguise) or the attack has a non-damaging outcome (Present), then damage is set to 0 instead */
+    if (damage.value <= 0) {
```
Tempo-anon, line 3357: nit: instead of `<= 0` would `< 1` be clearer?

Tempo-anon (dismissed):
Looks good to me, just some styling comments

src/field/pokemon.ts
frutescens: I think so. If a calculation goes very very awry, it could produce a result between -2 and 0 and that can get bumped up into positive damage.

src/field/pokemon.ts
frutescens: No, that would affect allowed damage that is between 0 and 1 like... 0.5 should still be rounded up to 1 damage.

PigeonBar (dismissed):

torranx (dismissed):

PigeonBar (approved):

flx-sta (approved):

PR #338, opened 2025-01-21
[Test] `toNextTurn` now takes double battles into account

DayKev:
N/A

To make writing tests easier.

`await game.toNextTurn(true);` will account for the extra `CommandPhase` in double battles and properly advance to the next turn.

## How to test the changes?
One of the Arena Trap tests was updated as a proof of concept: `npm run test arena_trap`

## Checklist

### DIFF
--- a/test/testUtils/gameManager.ts
+++ b/test/testUtils/gameManager.ts
@@ -420,15 +420,10 @@ export class GameManager {
     };
   }
 
-  /**
-   * Transition to the next upcoming {@linkcode CommandPhase}
-   * @param doubles - If `true`, will run `.to(CommandPhase)` twice to account for the extra `CommandPhase` in doubles
-   */
-  async toNextTurn(doubles: boolean = false) {
+  /** Transition to the next upcoming {@linkcode CommandPhase}. */
+  async toNextTurn() {
+    await this.phaseInterceptor.to(TurnInitPhase);
     await this.phaseInterceptor.to(CommandPhase);
-    if (doubles) {
-      await this.phaseInterceptor.to(CommandPhase);
-    }
   }
 
   /** Emulate selecting a modifier (item) and transition to the next upcoming {@linkcode CommandPhase} */
--- a/test/testUtils/gameManager.ts
+++ b/test/testUtils/gameManager.ts
@@ -420,9 +420,15 @@ export class GameManager {
     };
   }
 
-  /** Transition to the next upcoming {@linkcode CommandPhase} */
-  async toNextTurn() {
+  /**
+   * Transition to the next upcoming {@linkcode CommandPhase}
+   * @param doubles - If `true`, will run `.to(CommandPhase)` twice to account for the extra `CommandPhase` in doubles
+   */
+  async toNextTurn(doubles: boolean = false) {
     await this.phaseInterceptor.to(CommandPhase);
+    if (doubles) {
+      await this.phaseInterceptor.to(CommandPhase);
+    }
   }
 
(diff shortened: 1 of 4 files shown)


PigeonBar (dismissed):

test/testUtils/gameManager.ts
```diff
@@ -420,9 +420,15 @@ export class GameManager {
     };
   }
 
-  /** Transition to the next upcoming {@linkcode CommandPhase} */
-  async toNextTurn() {
+  /**
+   * Transition to the next upcoming {@linkcode CommandPhase}
+   * @param doubles - If `true`, will run `.to(CommandPhase)` twice to account for the extra `CommandPhase` in doubles
+   */
+  async toNextTurn(doubles: boolean = false) {
```
innerthunder: Does this need to be a parameter? the game manager can just look at `this.scene.currentBattle.double` to check if the extra command phase is needed.

DayKev: ...Good point.

DayKev: Except that'll be more work for me when a bunch of tests break. :despair:
Oh well.

DayKev: Hm, that does mean if the player only has a single pokemon in the test it would break, requiring the use of manually calling `phaseInterceptor.to()`. Is this an issue?

DayKev: Actually this means it breaks if any set-up that isn't pure "select move, to next turn" situation happens, such as this code that I found:
```ts
    game.move.select(Moves.SPLASH);
    game.move.select(Moves.SPLASH, 1);
    await game.forceEnemyMove(Moves.TACKLE, BattlerIndex.PLAYER);
    await game.forceEnemyMove(Moves.SPLASH);
    await game.toNextTurn();
```
Which breaks because `forceEnemyMove` awaits to `EnemyCommandPhase`.

flx-sta: @DayKev What is the output? I can't really follow the issue. Why does it break if the player has only 1 pokemon?

<sub> I do agree with the statement that it shouldn't be a param though</sub>

PigeonBar: Just realized that we could instead try something like:
```
  async toNextTurn() {
    await this.phaseInterceptor.to("TurnInitPhase");
    await this.phaseInterceptor.to("CommandPhase");
  }
```

DayKev: > @DayKev What is the output? I can't really follow the issue. Why does it break if the player has only 1 pokemon?
> 
> I do agree with the statement that it shouldn't be a param though

There would only be one `CommandPhase` to go through if the player had only one pokemon in a double battle, but the function would assume there are two and thus progress too many phases. Pigeon's suggestion might work, gonna try it out now.

DayKev: Seems to have worked, only had to fix one test (Powder).

innerthunder (dismissed):
PR title should be changed, but lgtm

test/testUtils/gameManager.ts
```diff
@@ -420,8 +420,9 @@ export class GameManager {
     };
   }
 
-  /** Transition to the next upcoming {@linkcode CommandPhase} */
+  /** Transition to the next upcoming {@linkcode CommandPhase}. */
```
PigeonBar:
```suggestion
  /** Transition to the first {@linkcode CommandPhase} of the next turn. */
```

PigeonBar (approved):

flx-sta (approved):

PR #334, opened 2025-01-21
[Move] Add low hanging fruit gmax move effects

Tempo-anon:
## What are the changes the user will see?

None

## Why am I making these changes?

Some G-Max move effects can be implemented without any new attr's

## What are the changes from a developer perspective?
Added docs to remaining unimplemented gmax moves for how to implement

* ~~G-Max Gold Rush = Pay day~~
* G-Max Resonance = Baddy Bad but with aurora veil instead
* G-Max Wind Rage = Defog
* ~~G-Max Depletion = Spite~~
* G-Max Gravitas = Gravity
* G-Max Drum Solo/fireball/Hydrosnipe = Sunsteel Strike
* G-Max One Blow/Rapid Flow = Feint

## Screenshots/Videos

n/a

## How to test the changes?

Override the moves

## Checklist

- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [ ] Have I tested the changes manually?

- [ ] Are all unit tests still passing? (`npm run test:silent`)
  - [ ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?
- [ ] Have I provided screenshots/videos of the changes (if applicable)?
  - [ ] Have I made sure that any UI change works for both UI themes (default and legacy)?

Are there any localization additions or changes? If so:

- [ ] Has a locales PR been created on the [locales](https://github.com/despair-games/poketernity-locales) repo?
  - [ ] If so, please leave a link to it here:
- [ ] Has the translation team been contacted for proofreading/translation?

PR #345, opened 2025-01-22
[Move] Implement g-max depletion

Tempo-anon:

None

#283 and #334 

* Added a new param to `AttackReducePpMoveAttr`
* Moved the move check and pp reduction to a private helper method
* ~~Created `getActiveOpps(Pokemon)` in `battle-scene.ts` and used that in `stat-stage-change-all-opps-attr.ts`~~
* Updated pokemon.getOpponents to take in a boolean onField parameter when filtering on isActive

## Screenshots/Videos

n/a

## How to test the changes?

```ts
const overrides = {
  MOVESET_OVERRIDE: [Moves.G_MAX_DEPLETION, Moves.EERIE_IMPULSE]
}
```

## Checklist


src/data/move-attrs/attack-reduce-pp-move-attr.ts
```diff
@@ -1,30 +1,48 @@
...
  * Attribute to reduce the PP of the target's last move after attacking.
  * Used for {@linkcode https://bulbapedia.bulbagarden.net/wiki/Eerie_Spell_(move) | Eerie Spell}.
+ * and {@linkcode https://bulbapedia.bulbagarden.net/wiki/G-Max_Depletion_(move) | G-Max Max_Depletion}
  * @extends ReducePpMoveAttr
  */
 export class AttackReducePpMoveAttr extends ReducePpMoveAttr {
-  constructor(reduction: number) {
+  private affectBothOpponents: boolean;
+
+  constructor(reduction: number, affectBothOpponents: boolean = false) {
```
innerthunder: G-max depletion's effect needs to be self-targeted for the same reasons as the stat drop gmax attrs:
1. To apply the effect even if the target faints
2. To bypass the target's substitute (if it has one)

This attribute and its parent are set up as non-self-targeting effects

src/battle-scene.ts
```diff
@@ -833,6 +833,25 @@ export default class BattleScene extends SceneBase {
...
+    let allOpps: Pokemon[];
+    if (target.isPlayer()) {
+      allOpps = this.getPlayerField();
+    } else {
+      allOpps = this.getEnemyField();
+    }
+    allOpps = allOpps.filter((p) => p.isActive(true));
+    return allOpps;
+  }
+
```
innerthunder: Turns out `Pokemon.getOpponents` already does this if updated to
```ts
getOpponents(): Pokemon[] {
  return this.getOpposingField().filter((p) => p.isActive(true));
}
```

Tempo-anon: Super useful utility function buried somewhere with only 2 other calls :pikastare:

src/data/move-attrs/attack-reduce-pp-move-attr.ts
```diff
@@ -6,23 +6,46 @@ import type { MoveConditionFunc } from "../move-conditions";
...
   }
 
+  /**
+   * Attempts to reduce the pp of the target (and its ally if affectsBothOpponents is true)'s
+   * last used move
+   * @param user the user of the move
+   * @param target the target of the attack
+   * @param move the move (eerie spell or g-max depletion)
+   * @returns true
+   */
```
torranx, line 27: I think we've agreed that `apply` methods (or other child methods) does not need to be documented if it is already on the base class to prevent redundancy. The description could be left out tho.

src/data/move-attrs/attack-reduce-pp-move-attr.ts
```diff
@@ -6,23 +6,46 @@ import type { MoveConditionFunc } from "../move-conditions";
...
+        this.reducePP(user, opp, move);
+      });
+    } else {
+      this.reducePP(user, target, move);
+    }
+
+    return true;
+  }
+
+  private reducePP(user: Pokemon, target: Pokemon, move: Move): void {
```
flx-sta, line 41: NIT: Looking at `movesetMove?.getPpRatio` this should be 
```suggestion
  private reducePp(user: Pokemon, target: Pokemon, move: Move): void {
```
for consistency 
<sub>No must!</sub>

flx-sta (approved):

Tempo-anon:
#348 will simplify this a lot

PR #333, opened 2025-01-21
[Move] Implement stackable critboost and chistrike

Tempo-anon:
## What are the changes the user will see?

None

## Why am I making these changes?

There is no attr for single target damage that also applies a BattlerTag to the user and its ally

## What are the changes from a developer perspective?

* Created a new battlerTag
* Updated the crit check in `Pokemon.ts`
* Create `chi-strike-attr`

## Screenshots/Videos

n/a

## How to test the changes?

```ts
const overrides = {
  MOVESET_OVERRIDE: [Moves.G_MAX_CHI_STRIKE]
}
```

## Checklist

- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?

- [x] Are all unit tests still passing? (`npm run test:silent`)
  - [x] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?
- [ ] Have I provided screenshots/videos of the changes (if applicable)?
  - [ ] Have I made sure that any UI change works for both UI themes (default and legacy)?

Are there any localization additions or changes? If so:

- [ ] Has a locales PR been created on the [locales](https://github.com/despair-games/poketernity-locales) repo?
  - [ ] If so, please leave a link to it here:
- [ ] Has the translation team been contacted for proofreading/translation?

PR #348, opened 2025-01-22
[Move] Add a Gmax flag

Tempo-anon:

None

PigeonBar came up with a solution to modify `MoveEffectPhase` instead of creating bespoke attr's for all the g-max moves

* Created a new `G_MAX_MOVE` move flag
* Created a `gmaxMove()` function that sets the above mentioned flag to true and also sets target to NEAR_ENEMY (g-max moves cannot target the user's ally)
* Deleted `stat-stage-change-all-opps-attr.ts`

## Screenshots/Videos

N/A

## How to test the changes?

`npm run test gmax_opp_stat_stage_change`

## Checklist


src/data/move.ts
```diff
@@ -570,6 +570,17 @@ export abstract class Move implements Localizable {
     return this;
   }
 
+  /**
+   * Sets the {@linkcode MoveFlags.G_MAX_MOVE} for the move
+   * and {@linkcode moveTarget} to NEAR_ENEMY (g-max moves cannot target allies)
+   * @returns The {@linkcode Move} that called this function
+   */
+  gmaxMove(): this {
```
flx-sta: NIT: Shouldn't it be:
```suggestion
  gMaxMove(): this {
```
Since it's also `G_MAX_MOVE`?

src/phases/move-effect-phase.ts
```diff
@@ -294,7 +295,19 @@ export class MoveEffectPhase extends HitCheckPhase {
...
+     * ignores substitutes
+     *
+     * TODO: G-Max move edge cases like snooze and the random status ones
+     */
+    if (move.checkFlag(MoveFlags.G_MAX_MOVE, user, target)) {
+      this.applyOnTargetEffects(user, target, hitResult, firstTarget);
+      if (target.getAlly()?.isActive(true)) {
+        this.applyOnTargetEffects(user, target.getAlly(), hitResult, firstTarget);
+      }
+    } else if (!move.hitsSubstitute(user, target)) {
```
flx-sta: NIT: Due to the heavy documentation here it might be worth it to put it into a private method
e.g. `applyGMaxEffects` or so

flx-sta (dismissed):

src/phases/move-effect-phase.ts
```diff
@@ -320,6 +312,19 @@ export class MoveEffectPhase extends HitCheckPhase {
     }
   }
 
+  /**
+   * G-Max Moves apply their effects to both opponents (even if the target faints) and also
+   * ignores substitutes
+   *
+   * TODO: G-Max move edge cases like G-max Snooze
```
flx-sta: NIT:
```suggestion
   * @todo G-Max move edge cases like G-max Snooze
```

flx-sta (dismissed):

src/phases/move-effect-phase.ts
```diff
@@ -308,6 +312,19 @@ export class MoveEffectPhase extends HitCheckPhase {
...
+   */
+  private applyGMaxEffects(user: Pokemon, target: Pokemon, hitResult: HitResult, firstTarget: boolean) {
+    this.applyOnTargetEffects(user, target, hitResult, firstTarget);
+    if (target.getAlly()?.isActive(true)) {
+      this.applyOnTargetEffects(user, target.getAlly(), hitResult, firstTarget);
```
PigeonBar:
```suggestion
      this.triggerMoveEffects(MoveEffectTrigger.POST_APPLY, user, target.getAlly(), firstTarget, false);
```

I've realized that `applyOnTargetEffects` applies too many effects (e.g., it also applies the user's post-attack abilities against the target's ally). We only want to be applying the G-Max move's move attrs.

src/phases/move-effect-phase.ts
```diff
@@ -295,7 +296,10 @@ export class MoveEffectPhase extends HitCheckPhase {
     const hitResult = this.applyMove(target, effectiveness);
 
     this.triggerMoveEffects(MoveEffectTrigger.POST_APPLY, user, target, firstTarget, true);
```
PigeonBar: This line applies self-targeting move attrs to the user. We may also want to change this line to make it also apply self-targeting attrs to the user's ally if the move is a G-Max move. (Right now, such a change would also overlap with #333.)

Tempo-anon: Oh neat, I'll add the if check here as well then when this PR gets merged in I'll update the chistrike PR to use the new move-effect-phase

src/phases/move-effect-phase.ts
```diff
@@ -308,6 +316,29 @@ export class MoveEffectPhase extends HitCheckPhase {
     }
   }
 
+  /**
+   * G-Max Moves applying positive effects to both the user and their active ally
+   */
+  private applyGMaxUserEffects(user: Pokemon, target: Pokemon, firstTarget: boolean) {
+    this.triggerMoveEffects(MoveEffectTrigger.POST_APPLY, user, target, firstTarget, true);
+    if (target.getAlly()?.isActive(true)) {
+      this.triggerMoveEffects(MoveEffectTrigger.POST_APPLY, user, user.getAlly(), firstTarget, false);
```
PigeonBar:
```suggestion
    if (user.getAlly()?.isActive(true)) {
      this.triggerMoveEffects(MoveEffectTrigger.POST_APPLY, user.getAlly(), target, firstTarget, true);
```

src/phases/move-effect-phase.ts
```diff
@@ -308,6 +316,29 @@ export class MoveEffectPhase extends HitCheckPhase {
...
+  }
+
+  /**
+   * G-Max Moves apply their effects to both opponents (even if the target faints) and also
+   * ignores substitutes
+   *
+   * @todo: G-Max move edge cases like G-max Snooze
+   */
+  private applyGMaxTargetEffects(user: Pokemon, target: Pokemon, hitResult: HitResult, firstTarget: boolean) {
+    this.applyOnTargetEffects(user, target, hitResult, firstTarget);
```
PigeonBar:
```suggestion
    const move = this.move.getMove();

    if (move.hitsSubstitute(user, target)) {
      this.triggerMoveEffects(MoveEffectTrigger.POST_APPLY, user, target, firstTarget, false)
    } else {
      this.applyOnTargetEffects(user, target, hitResult, firstTarget);
    }

```

src/phases/move-effect-phase.ts
Tempo-anon: Doesn't this mean that it's the user's ally calling a self targetting move effect trigger on itself? I assume there's some niche interactions where we want to keep the actual user as the Pokemon that used the gmax move

src/phases/move-effect-phase.ts
Tempo-anon: I believe all gmax effects that affect opponents bypass substitute anyway

PigeonBar: These suggested changes make it so that:
- If the move hits a substitute, only the gmax move attr applies.
- If the move does not hit a substitute, then all effects, including post-attack abilities, apply to the target.

src/phases/move-effect-phase.ts
PigeonBar: Although this is a bit hacky, I think it's the only way to make a self-targeting attr apply to the user's ally, since if an attr is self-targeting, it generally applies the effect to whichever Pokemon is passed to the `user` param.

(Passing `false` into `selfTarget` here does not tell the attr to stop being self-targeting. It instead tells the code to apply all non-self-targeting attrs to the user's ally.)

src/phases/move-effect-phase.ts
```diff
@@ -308,6 +316,36 @@ export class MoveEffectPhase extends HitCheckPhase {
...
+    if (user.getAlly()?.isActive(true)) {
+      this.triggerMoveEffects(MoveEffectTrigger.POST_APPLY, user.getAlly(), target, firstTarget, true);
+    }
+  }
+
+  /**
+   * G-Max Moves apply their effects to both opponents (even if the target faints) and also
+   * ignores substitutes
+   *
+   * @todo: G-Max move edge cases like G-max Snooze
```
DayKev, line 333:
```suggestion
   * @todo G-Max move edge cases like G-max Snooze
```

innerthunder (commented):
The flag name might be misleading since it doesn't technically apply to *all* G-max moves (G-max Snooze is the exception)

Tempo-anon:
I assume we'll want the flag for other purposes (like tweaking BP or setting targets) but since gmax snooze is the only move with this weird property I think a comment in the move-effect-phase to make mention of it when it's implemented should be fine

PigeonBar (approved):

PR #374, opened 2025-01-26
[Move] Gmax moves now have base powers

Tempo-anon:

None

Need to come up with a way for g-max moves to have damage.

G-Max moves have 80bp but get a flat +50 if the user is g-max'd (the gen 8 moves have 110bp)

<details><summary>Previous version</summary>
<p>

* The higher end bp of g-max moves should be 130 for non fighting/poison types and 100 for fighting/poison
* The moves should still be usable as attacks even when not g-maxed
* I think it would be cool if there was a slight bonus granted if the Pokemon using the move is its signature move

For this I came up with the following formula:
* G-max = +50%
* G-max and signature = +50% +10

 |                 | default bp | max bp | max and signature |
 |-----------------|------------|--------|-------------------|
 | Poison/Fighting | 60         | 90     | 100               |
 | Gen 8 starter   | 100        | 150    | 160               |
 | All other types | 80         | 120    | 130               |

</p>
</details> 

* All G-Max moves do not make contact
* All G-Max moves use the photon geyser attr (which has been renamed) to use the higher of the user's atk/spA
* Created `GMaxPowerAttr`
* Set the base powers of all the gmax moves as well as 3pp and also physical/special based off their original pokemon (purely cosmetic)

## Screenshots/Videos

image

`base damage 7.333333333333334 G-Max Wildfire 25 16 6`
base power is 25 here (10 * 1.5 + 10) which shows the formula is correct

## How to test the changes?

```ts
const overrides = {
  STARTER_SPECIES_OVERRIDE: Species.CHARIZARD,
  STARTER_FORM_OVERRIDES: {[Species.CHARIZARD]: 3},
  MOVESET_OVERRIDE: [Moves.G_MAX_WILDFIRE]
}
```

## Checklist


PigeonBar:
A few questions to start off:

- Should Poison/Fighting G-Max moves still have lower power? My understanding is that Max Ooze/Knuckle have lower power to balance out their secondary effects of increasing the user's Sp. Atk and Atk by 1 stage, respectively. There isn't a corresponding reason for the G-Max moves to have lower power, and it could cause the Poison/Fighting G-Max moves to be underwhelming compared to the other G-Max moves.
- In mainline, G-Max Drum Solo, Fireball, and Hydrosnipe have 160 power. We can incorporate this using the same formula by setting default = 100, g-max = 150, g-max + signature = 160.
- When checking whether or not a spliced Pokemon is G-Max and/or the signature species, should we also check the secondary Pokemon?

Tempo-anon:
> Should Poison/Fighting G-Max moves still have lower power? My understanding is that Max Ooze/Knuckle have lower power to balance out their secondary effects of increasing the user's Sp. Atk and Atk by 1 stage, respectively. There isn't a corresponding reason for the G-Max moves to have lower power, and it could cause the Poison/Fighting G-Max moves to be underwhelming compared to the other G-Max moves.

That's a good question. G-Max moves mirror the same base powers as the max moves so I figured I'd imitate that by having them be a bit weaker. As for secondary effects, Chi-Strike has the effect of stacking crit boost (+0 is 1/24, +1 is 1/8, +2 is 1/2, +3 is 100%) while malodor has a poisoning both opponents. I guess it's something I want to hear some more opinions of.

> In mainline, G-Max Drum Solo, Fireball, and Hydrosnipe have 160 power. We can incorporate this using the same formula by setting default = 100, g-max = 150, g-max + signature = 160.

That sounds like a good idea. They get stronger damage in exchange for having a more underwhelming secondary effect.

> When checking whether or not a spliced Pokemon is G-Max and/or the signature species, should we also check the secondary Pokemon?

I believe the `isMax()` function checks for if either part of the Pokemon is max
```ts
 isMax(): boolean {
    const maxForms = [
      SpeciesFormKey.GIGANTAMAX,
      SpeciesFormKey.GIGANTAMAX_RAPID,
      SpeciesFormKey.GIGANTAMAX_SINGLE,
      SpeciesFormKey.ETERNAMAX,
    ] as string[];
    return (
      maxForms.includes(this.getFormKey()) || (!!this.getFusionFormKey() && maxForms.includes(this.getFusionFormKey()!))
    );
  }
```

src/data/move-attrs/gmax-power-attr.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  constructor(signatureSpecies: Species) {
+    super();
+    this.signatureSpecies = signatureSpecies;
+  }
+
+  override apply(user: Pokemon, _target: Pokemon, _move: Move, power: NumberHolder): boolean {
+    if (user.isMax()) {
+      power.value *= 1.5;
+      if (user.species.speciesId === this.signatureSpecies) {
```
innerthunder: If the signature species boost is kept, this should probably also check for fusion species

src/data/move-attrs/gmax-power-attr.ts
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Attribute to modify a G-Max move's base power
+ * ```
+ * |                 | default bp | max bp | max and signature |
+ * |-----------------|------------|--------|-------------------|
+ * | Poison/Fighting | 60         | 90     | 100               |
+ * | Gen 8 starter   | 100        | 150    | 160               |
+ * | All other types | 80         | 120    | 130               |
+ * ```
```
innerthunder: imo the Poison/Fighting G-Max moves shouldn't have the power reduction because
1. They were originally reduced to match their Max Move counterparts, but the Max Moves aren't up for consideration here.
2. The effects themselves aren't strong enough to justify the power reduction in a vacuum
3. It's important to have a consistent pattern between all the G-max moves since this part of the implementation is custom.

Also, I'm not sure how I feel about the signature species aspect of this. I think simplicity is important when it comes to custom implementations, and the components of the current power boost might be difficult to succinctly communicate to the player.

src/data/all-moves.ts
```diff
@@ -3267,110 +3267,110 @@ export function initMoves() {
         StatusEffect.BURN,
         StatusEffect.SLEEP,
       ]),
-    new AttackMove(Moves.G_MAX_WILDFIRE, Type.FIRE, MoveCategory.PHYSICAL, 10, -1, 10, -1, 0, 8)
-      .gMaxMove()
+    new AttackMove(Moves.G_MAX_WILDFIRE, Type.FIRE, MoveCategory.SPECIAL, 80, -1, 3, -1, 0, 8)
```
innerthunder: How is the base move category determined for these moves? Just vibes? (Bulba just says the category varies for these moves)

src/data/move.ts
```diff
@@ -572,11 +575,17 @@ export abstract class Move implements Localizable {
...
+   * and have the {@linkcode PhotonGeyserCategoryAttr} to use the higher of the attacking stat
+   * and have the move use the {@linkcode GMaxPowerAttr}
    * @returns The {@linkcode Move} that called this function
    */
-  gMaxMove(): this {
+  gMaxMove(signatureSpecies: Species): this {
     this.setFlag(MoveFlags.G_MAX_MOVE, true);
     this.moveTarget = MoveTarget.NEAR_ENEMY;
+    this.makesContact(false);
+    this.attr(PhotonGeyserCategoryAttr);
```
innerthunder: nit: Rename this attr to something more generalized (`StatBasedCategoryAttr`?)

src/data/all-moves.ts
Tempo-anon: I looked at every single g-maxxable Pokemon and took the higher of their attacking stats (except Venusaur since its main attacking grass move is power whip which is physical) yes vibes

Snailman11:
The power of the Stat-Raising Max Moves should depend on how often the Player can use them. I think that at 100 or 110 base power, and a restriction of 3 uses would be reasonable enough. Snowballing Stat boosts is a problem and we should probably anticipate a "fair" gameplay feature to limit this utility, especially in "harder" battles. Keep in mind that a KO with a faster Stat-Boost is essentially a Stat-boost with a free turn.

Tempo-anon:
> The power of the Stat-Raising Max Moves should depend on how often the Player can use them. I think that at 100 or 110 base power, and a restriction of 3 uses would be reasonable enough. Snowballing Stat boosts is a problem and we should probably anticipate a "fair" gameplay feature to limit this utility, especially in "harder" battles. Keep in mind that a KO with a faster Stat-Boost is essentially a Stat-boost with a free turn.

This PR is only for deciding the base power of gmax moves, not generic max moves. No GMax moves increase the user's stats (closest is probably chi-strike which just adds +1 crit ratio to the user and allies). The generic fighting/poison max moves have lower BP to compensate for the fact that they give +1 atk/spA on use. Generic flying max move gives the user +1 speed but only fighting/poison have the bp penalty

src/data/move-attrs/gmax-power-attr.ts
Tempo-anon: Whatever, how about we just do the 80+50 like you suggested and we can always tweak it in the future?

src/battle-scene.ts
```diff
@@ -994,7 +994,7 @@ export default class BattleScene extends SceneBase {
       ENEMY_IVS_OVERRIDE_VALIDATED = new Array(6).fill(Overrides.ENEMY_IVS_OVERRIDE);
     }
     if (ENEMY_IVS_OVERRIDE_VALIDATED.length === 6) {
-      pokemon.ivs = ENEMY_IVS_OVERRIDE_VALIDATED.map(iv => Phaser.Math.Clamp(iv, 0, 31));
+      pokemon.ivs = ENEMY_IVS_OVERRIDE_VALIDATED.map((iv) => Phaser.Math.Clamp(iv, 0, 31));
```
flx-sta: NIT: `0` and `31` should be constants 
e.g.
```ts
/** The min value for IVs */
const IV_MIN = 0;
/** The max value for IVs */
const IV_MAX = 31;
```

src/data/move-attrs/gmax-power-attr.ts
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Attribute to modify a G-Max move's base power
+ * Adds a flat +50 base power if the user is G-Max'd
+ * @extends VariablePowerAttr
+ */
+export class GMaxPowerAttr extends VariablePowerAttr {
+  /**
+   * Unused, but keeping it in case we want to use it in the future
+   */
+  signatureSpecies: Species;
```
flx-sta: NIT: Missing scope

src/data/move-attrs/use-higher-attacking-stat-attr.ts
```diff
@@ -11,7 +11,7 @@ import { AbilityApplyMode } from "#enums/ability-apply-mode";
  * Used for {@link https://bulbapedia.bulbagarden.net/wiki/Photon_Geyser_(move) | Photon Geyser}.
```
flx-sta, line 11: This should probably be updated since the class was named more generic!

src/data/move.ts
```diff
@@ -581,11 +584,17 @@ export abstract class Move implements Localizable {
   /**
    * Sets the {@linkcode MoveFlags.G_MAX_MOVE} for the move
    * and {@linkcode moveTarget} to NEAR_ENEMY (g-max moves cannot target allies)
+   * and make the move not make contact
+   * and have the {@linkcode UseHigherAttackingStatAttr} to use the higher of the attacking stat
+   * and have the move use the {@linkcode GMaxPowerAttr}
    * @returns The {@linkcode Move} that called this function
    */
```
flx-sta: NIT: and, and and and......and and... phew..
```suggestion
/**
 * Modifies the move with the following properties:
 * - Sets the {@linkcode MoveFlags.G_MAX_MOVE}.
 * - Sets {@linkcode moveTarget} to NEAR_ENEMY (G-Max moves cannot target allies).
 * - Prevents the move from making contact.
 * - Applies {@linkcode UseHigherAttackingStatAttr} to use the higher attacking stat.
 * - Assigns the move the {@linkcode GMaxPowerAttr}.
 * 
 * @returns The {@linkcode Move} that called this function.
 */
```

src/data/move.ts
```diff
@@ -581,11 +584,17 @@ export abstract class Move implements Localizable {
...
-  gMaxMove(): this {
+  gMaxMove(_signatureSpecies: Species): this {
     this.setFlag(MoveFlags.G_MAX_MOVE, true);
     this.moveTarget = MoveTarget.NEAR_ENEMY;
+    this.makesContact(false);
+    this.attr(UseHigherAttackingStatAttr);
+    this.attr(GMaxPowerAttr, _signatureSpecies);
```
flx-sta: NIT: the argument shouldn't have the `_` prefix. This is associated with argument that aren't being used. It's used in line 597 though.

```suggestion
  gMaxMove(signatureSpecies: Species): this {
    this.setFlag(MoveFlags.G_MAX_MOVE, true);
    this.moveTarget = MoveTarget.NEAR_ENEMY;
    this.makesContact(false);
    this.attr(UseHigherAttackingStatAttr);
    this.attr(GMaxPowerAttr, signatureSpecies);
```

test/data/all_moves.json
flx-sta, line 1: Is there a reference, link to where those `power` and `pp` numbers are coming from?

src/battle-scene.ts
Tempo-anon: I don't even know where this came from... probably chalk it up to some git or beta weirdness

test/data/all_moves.json
Tempo-anon: Mostly just #266

flx-sta: @Tempo-anon is this discussion basically “done”? If so can we have a TLDR at the top so it’s easy to find ?

src/constants.ts
```diff
@@ -93,3 +93,7 @@ export const ME_AVERAGE_ENCOUNTERS_PER_RUN_TARGET = 12;
 export const ME_ANTI_VARIANCE_WEIGHT_MODIFIER = 15;
 
 export const PARTY_UI_NO_EFFECT_MSG_i18N_KEY = "partyUiHandler:anyEffect";
+
+export const IV_MIN = 0;
```
flx-sta, line 101: NIT: constants should always have a comment describing its purpose ( i know i know… if forgot  that above too)

flx-sta (dismissed):

src/data/all-moves.ts
```diff
@@ -3648,110 +3648,110 @@ export function initMoves() {
...
       .attr(TrapAttr, BattlerTagType.G_MAX_SAND_TOMB),
-    new AttackMove(MoveId.G_MAX_SNOOZE, ElementalType.DARK, MoveCategory.PHYSICAL, 10, -1, 10, -1, 0, 8)
-      .gMaxMove()
+    new AttackMove(MoveId.G_MAX_SNOOZE, ElementalType.DARK, MoveCategory.PHYSICAL, 80, -1, 3, -1, 0, 8)
+      .gMaxMove(Species.GRIMMSNARL)
       .attr(AddBattlerTagAttr, BattlerTagType.DROWSY, false, { effectChanceOverride: 50 })
       .edgeCase(), // The 50% chance incorrectly gets overridden by Shield Dust, Sheer Force, etc.
-    new AttackMove(MoveId.G_MAX_TARTNESS, ElementalType.GRASS, MoveCategory.PHYSICAL, 10, -1, 10, -1, 0, 8)
-      .gMaxMove()
+    new AttackMove(MoveId.G_MAX_TARTNESS, ElementalType.GRASS, MoveCategory.SPECIAL, 80, -1, 3, -1, 0, 8)
```
PigeonBar:
```suggestion
    new AttackMove(MoveId.G_MAX_TARTNESS, ElementalType.GRASS, MoveCategory.PHYSICAL, 80, -1, 3, -1, 0, 8)
```

src/data/move-attrs/use-higher-attacking-stat-attr.ts
```diff
@@ -9,9 +9,10 @@ import { AbilityApplyMode } from "#enums/ability-apply-mode";
 /**
  * Attribute to change move category to match the user's highest effective offensive stat.
  * Used for {@link https://bulbapedia.bulbagarden.net/wiki/Photon_Geyser_(move) | Photon Geyser}.
+ * And all G-Max Moves
  * @extends VariableMoveCategoryAttr
  */
-export class PhotonGeyserCategoryAttr extends VariableMoveCategoryAttr {
+export class UseHigherAttackingStatAttr extends VariableMoveCategoryAttr {
   override apply(user: Pokemon, target: Pokemon, move: Move, category: NumberHolder): boolean {
     if (
```
PigeonBar, line 18: This attr now also needs:
```
    else {
      category.value = MoveCategory.SPECIAL;
      return true;
    }
```

test/data/all_moves.json
```diff
@@ -14055,12 +14055,12 @@
...
     "type_id": 12,
-    "power": 10,
-    "pp": 10,
+    "power": 80,
+    "pp": 3,
     "accuracy": -1,
     "priority": 0,
     "target_id": 10,
-    "damage_class_id": 2,
+    "damage_class_id": 3,
```
PigeonBar:
```suggestion
    "damage_class_id": 2,
```

PigeonBar (approved):

PR #375, opened 2025-01-26
[Test] Add tests for Present

PigeonBar:

None

There were no tests for Present.

Added the following tests:
- Present should use correct probabilities, without off-by-one errors.
- If Present heals, it should end multi-hit Present (e.g., from Parental Bond), and it should also not deal damage.

## How to test the changes?

`npm run test present`

## Checklist

- ~~[ ] **I'm using `beta` as my base branch**~~ (This PR is an add-on to #337)

### DIFF
--- a/test/moves/present.test.ts
+++ b/test/moves/present.test.ts
@@ -1,0 +1,107 @@
+import { allMoves } from "#app/data/all-moves";
+import { PresentPowerAttr } from "#app/data/move-attrs/present-power-attr";
+import { NumberHolder } from "#app/utils";
+import { Abilities } from "#enums/abilities";
+import { Moves } from "#enums/moves";
+import { Species } from "#enums/species";
+import { GameManager } from "#test/testUtils/gameManager";
+import Phaser from "phaser";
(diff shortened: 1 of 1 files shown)


test/moves/present.test.ts
```diff
@@ -0,0 +1,107 @@
...
+      .disableCrits()
+      .enemySpecies(Species.MAGIKARP)
+      .enemyAbility(Abilities.BALL_FETCH)
+      .enemyMoveset(Moves.SPLASH);
+  });
+
+  it.each([
+    { descriptor: "first hit", hitsLeft: 2, totalOutcomes: 100, expectedHeals: 20 },
+    { descriptor: "subsequent hits", hitsLeft: 1, totalOutcomes: 80, expectedHeals: 0 },
+  ])("should have correct probabilities on $descriptor", async ({ hitsLeft, totalOutcomes, expectedHeals }) => {
```
torranx, line 39: is it possible to describe the test like so? "should have x% probability of dealing y damage" or similar

PigeonBar: Doing this makes the test descriptions too long, and Vitest would just cut off the test descriptions:

image

torranx (approved):

PR #378, opened 2025-01-27
[Refactor] Consolidate all `try/catch` for animations in a single place

DayKev:
N/A

To simplify development.

All `try/catch` statements around sprite animations are removed, and centralized into the existing replacement function in `src/ui/ui-theme.ts`. Important changes are in `src/field/pokemon.ts` and `src/ui/ui-theme.ts`.

Try/catch removed in: `src/data/mystery-encounters/encounters/global-trade-system-encounter.ts`, `src/data/mystery-encounters/utils/encounter-transformation-sequence.ts`, `src/phases/abstract-form-change-base-phase.ts`, `src/phases/egg-hatch-phase.ts`, `src/phases/evolution-phase.ts`, `src/phases/form-change-phase.ts`, `src/phases/quiet-form-change-phase.ts`, `src/ui/summary-ui-handler.ts`.

## Screenshots/Videos

https://github.com/user-attachments/assets/84c9d5c3-dd9c-48e5-81d0-6ebef67bd62d

## Checklist


src/phases/quiet-form-change-phase.ts
```diff
@@ -54,11 +54,7 @@ export class QuietFormChangePhase extends BattlePhase {
       sprite.setOrigin(0.5, 1);
 
       const spriteKey = this.pokemon.getBattleSpriteKey();
-      try {
-        sprite.play(spriteKey).stop();
-      } catch (err: unknown) {
-        console.error(`Failed to play animation for ${spriteKey}`, err);
-      }
+      sprite.play(spriteKey).stop();
```
PigeonBar, line 59: Based on some playtesting, the `stop()` function can still crash here. This can be fixed by also overriding `stop()` to use a try-catch.

src/ui/ui-theme.ts
```diff
@@ -161,6 +161,18 @@ export function addUiThemeOverrides(): void {
...
+      key: string | Phaser.Animations.Animation | Phaser.Types.Animations.PlayAnimationConfig,
+      ignoreIfPlaying?: boolean,
+    ): Phaser.GameObjects.Sprite {
+      try {
+        return originalPlay.apply(this, [key, ignoreIfPlaying]);
+      } catch (err: unknown) {
+        console.error(`Failed to play animation for ${key}`, err);
+        return this;
+      }
+    };
```
flx-sta, line 176: If we were to add our own extension of `Phaser` classes we can instead overwrite methods inside there. 
E.g.
```ts
class GameSprite extends Phaser.GameObjects.Sprite {

  override play([...]) {
    try {
      super.play(); //is that how it worked?
      // [...]
    } catch (err) {
      // [...]
    }
  }
}
```

That might be a more intrusive but also cleaner solution.
I was aiming for something like this in the long run anyway. Same for `Phaser.GameObjects.Container` -> `GameContainer` to override the `type` with a string-template type

DayKev: That was my initial idea (see #371) but I realized it wouldn't be as easy as I had hoped. This can be easily replaced with that later.

flx-sta (approved):

MokaStitcher:
The merge of #419 means that the globalScene.add functions overrides are no longer present on beta

So not sure where to go from there for this PR. Maybe keep the current solution but put it elsewhere, or try to make things a bit cleaner with our own custom sprite object.
I think it should be possible to extend Sprite with our own custom class and replace all `scene.add.Sprite(..)` with `sprite = new PktySprite(..); scene.add.existing(sprite)`. Or make a new `scene.add.PktySprite` that does this
That is, assuming this is all that scene.add.sprite does. Which I'm sadly not 100% sure of

flx-sta:
@DayKev are you still actively working on this?

DayKev:
Since the function replacement I was using was removed, something else will have to be figured out. For now I'm working on other stuff.

flx-sta:
Closing for now.
Feel free to reopen at a point of return

PR #388, opened 2025-01-28
[Bug][Mystery Encounters] Use proper BST calculation

DayKev:
The following MEs will properly calculate the BST of Pokémon for their effects: Global Transfer System, Weird Dream, The Strong Stuff, and The Expert Pokemon Breeder

These MEs are incorrectly looking at the modified BST of the Pokémon rather than the actual BST of the Pokémon.
Fixes PR5198

`pokemon.calculateBaseStats()` replaced with `pokemon.getSpeciesForm().getBaseStatTotal()` in the aforementioned MEs.

## Screenshots/Videos
<details><summary>Before PR fix</summary>
<p>

image

</p>
</details> 
<details><summary>After PR fix</summary>
<p>

image

</p>
</details> 

## How to test the changes?
See instructions in PR5198 

## Checklist


Tempo-anon (approved):

src/data/mystery-encounters/encounters/global-trade-system-encounter.ts
```diff
@@ -503,7 +503,7 @@ function getPokemonTradeOptions(): Map<number, EnemyPokemon[]> {
       });
       tradeOptionsMap.set(pokemon.id, tradeOptions);
     } else {
-      const originalBst = pokemon.calculateBaseStats().reduce((a, b) => a + b, 0);
+      const originalBst = pokemon.getSpeciesForm().getBaseStatTotal();
```
PigeonBar, line 499: This solution does not account for fused species.

I think a better solution would be to add a new param `ignoreModifiers` to `calculateBaseStats()`, and then pass `ignoreModifiers = true` when computing base stats for ME's.

DayKev: This is what's used in other parts of the MEs, so does that mean the MEs are bugged for fusions?

PigeonBar: I just checked, and all previous usages of `getBaseStatTotal()` seem to be for new Pokemon that the ME gives you, which are all unfused. So, they should still be correct.

PigeonBar: Since we are removing fusions according to the Discord poll, this is no longer an issue

PigeonBar (approved):

PR #356, opened 2025-01-24
[AI-Rework] Add starter code for move scoring functions

innerthunder:
> [!NOTE]
> I'm breaking up the AI Rework into multiple PRs targeting the [`ai-rework` branch](https://github.com/Despair-Games/poketernity/tree/ai-rework) to make the review process a bit more iterative and accessible. These changes will not be pushed into the `beta` branch until the AI Rework is complete and on the "parent" branch.

## What are the changes the user will see?

Nothing so far; the scoring functions introduced are currently unused

## Why am I making these changes?

Building blocks for the AI Rework

## What are the changes from a developer perspective?

Introduced functions covering the 3 core components of scoring for enemy command selection:
- Attack Score (AS): evaluates the value of a move against a given target based on the estimated damage dealt.
- Matchup Score (MUS): evaluates the Pokemon's overall viability against a given opponent based on each Pokemon's attack scores. This will mainly be used to determine if enemy Trainers will switch Pokemon.
- Effect Score (ES): the combined score bonus for a move collected from its attributes and conditions. Used to evaluate non-damage-based move effects.

For more information on how these scores are calculated and applied, please refer to the [AI Rework Proposal](https://docs.google.com/document/d/1olQcsBcgoY4oid2-3nko__AlBzG9BnIeEtTTRHp59DU/edit?usp=sharing).

## Screenshots/Videos

N/A

## How to test the changes?

> [!CAUTION]
> Only code-reviews are expected 

## Checklist

- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [?] Have I tested the changes manually?

- [x] Are all unit tests still passing? (`npm run test:silent`)
  - [x] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?
- [n/a] Have I provided screenshots/videos of the changes (if applicable)?
  - [n/a] Have I made sure that any UI change works for both UI themes (default and legacy)?

PR #389, opened 2025-01-28
[AI-Rework] Rewrite enemy command selection

innerthunder:
> [!NOTE]
> This PR is a component of the AI Rework and is directed at the rework's dedicated branch. This PR alone should not be considered a complete feature. 

The AI now has a more rigid heuristic when it comes to evaluating actions for any given turn. With this change, the AI should:
- switch less often between two Pokemon with similarly bad matchups against a player Pokemon
- never switch out when it can safely KO a player Pokemon
- select its strongest attack more consistently compared to other attacks
- virtually never select status moves (effect scores are not yet implemented)
- no longer formulate a move action if forced to skip their turn because of their Commander ability or the Fun and Games Mystery Encounter. This should mean that Sucker Punch and Thunderclap now always fail during the Fun and Games minigame.

See the [AI Rework Proposal](https://docs.google.com/document/d/1olQcsBcgoY4oid2-3nko__AlBzG9BnIeEtTTRHp59DU/edit?usp=sharing) for more details on the new command selection heuristic.

The current AI is dumb when it comes to selecting turn actions.

- Much of the code in `EnemyCommandPhase` has been cleaned up and moved into `EnemyPokemon#getNextCommand`
- The logic to determine if an enemy should switch has been simplified and uses the new matchup scoring heuristic from #356 
- Move command selection has been rewritten to use `EnemyPokemon#getMoveScore` (also from 356)

## Screenshots/Videos

All tests are passing 🔥 

image

### Move Selection changes

The following data from `enemy_command.test.ts` is pulled from the following game state:
- Enemy is a Level 100 Snorlax with Covet, Thief, Flame Wheel, and Splash. Its AI type is `SMART_RANDOM`.
- Opponent (player) is a Level 100 Duskull. None of Duskull's moves are revealed to the enemy

In this game state, the enemy has 2 moves that effectively do nothing: Covet because Duskull is immune to it and Splash because, well, it's Splash. Out of its other 2 moves, Thief is the clear best move to use since its effectiveness is twice that of Flame Wheel. However, running this test on `beta` yields the following move choices over 300 trials.

image

The current AI in this state wastes its turn about 10 percent of the time, and fails to choose its strongest move 30 percent of the time overall.

With the changes in this PR, though, the AI is much more precise in its move selection, and it never picks a move that does nothing if it can be helped. However, because of the variance in attack score output, it doesn't _always_ choose the absolute best move for the situation.

image

## How to test the changes?

`npm run test enemy_command` - 1 new test (old tests are still relevant and passing)

## Checklist

- [n/a] **I'm using `beta` as my base branch**


innerthunder:
welp, I messed up the commit history...

New PR for this is at https://github.com/Despair-Games/poketernity/pull/403

PR #397, opened 2025-01-30
[Dev] Add npm script `test:full` (= `typecheck` + `test:silent`)

PigeonBar:

Added the option to run `npm run test:full`, which is equivalent to running `npm run typecheck` and then `npm run test:silent`.

The typecheck is an optional extra step that can help with automatically detecting more errors.

Title.

## How to test the changes?

`npm run test:full`

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,9 @@
     "test:silent": "vitest run --silent --no-isolate",
     "posttest:silent": "git rev-parse --short HEAD",
 
+    "test:full": "tsc --noEmit && vitest run --silent --no-isolate",
+    "posttest:full": "git rev-parse --short HEAD",
+
     "typecheck": "tsc --noEmit",
     "posttypecheck": "git rev-parse --short HEAD",
 
--- a/package.json
+++ b/package.json
@@ -13,22 +13,16 @@
     "preview": "vite preview",
 
     "test": "vitest run --no-isolate",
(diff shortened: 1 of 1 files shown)


MokaStitcher (dismissed):
Unfortunately this is not super ideal because having overrides makes typecheck fail so the tests don't run even if having overrides is normally ok because they get ignored during the tests
Still, it works, and if it's something that would be useful to others I'm not opposed to it

package.json
```diff
@@ -24,6 +24,9 @@
     "test:silent": "vitest run --silent --no-isolate",
     "posttest:silent": "git rev-parse --short HEAD",
 
+    "test:full": "tsc --noEmit && vitest run --silent --no-isolate",
+    "posttest:full": "git rev-parse --short HEAD",
```
flx-sta: NIT: I think these `post<>:` should be dropped again. 
They only exist because we didn't have any CI before the repo was public.
No necessity for it anymore

flx-sta (dismissed):

package.json
MokaStitcher: oh right. Might as well remove them all in this PR, even though another PR would be fine too

flx-sta: I'll leave it to the author 👍

MokaStitcher (approved):

flx-sta (approved):

PR #364, opened 2025-01-24
[Feature] Add named double battles for Champs

CodeTappert:
## What are the changes the user will see?
No changes currently. Since they dont spawn in. We still need to decide where and when to spwan them

## Why am I making these changes?
I love this feature. And we all agreed to do it. NOTE: There is currently no way they can spawn in the game. Since we didnt decide onto where and how yet.

## What are the changes from a developer perspective?
I added 2 new fields to trainer (spriteName1 and spriteName2). 
Changed getName a little. Otherwise using most of the existing functions.
I added the option that you can give a array of music to initForChampion and it then picks one at random (we might change this). 
Otherwise only changes to make it work.

## Screenshots/Videos
- Blue_Red:
https://github.com/user-attachments/assets/c35426f0-63f8-40fe-a140-9eef4a84c4b0
- Lance_Clair:
https://github.com/user-attachments/assets/9f8c9f59-8bfd-4a19-b5b2-e7ba6afb5ad0
- Steven_Wallace
https://github.com/user-attachments/assets/7ffbdf3b-20a4-4f37-994c-b98c15bc4e55
- Cynthia_Diantha:
https://github.com/user-attachments/assets/8921b656-3e53-46ad-96a7-eba94530bc1c
- Iris_Alder:
https://github.com/user-attachments/assets/844bdf8d-7742-4315-b8f2-2472755f49e1
- Hau_Kukui:
https://github.com/user-attachments/assets/8ecd65d2-bdfd-4768-9c23-9acfd329f90f
- Leon_Hop:
https://github.com/user-attachments/assets/61911808-8b87-43c1-8665-8ff5acef7b26
- Geeta_Nemona:
https://github.com/user-attachments/assets/b68b5e8e-c1ca-4e83-bdc1-138b7f22203c
- Kieran_Carmine:
https://github.com/user-attachments/assets/01cb0805-bb4f-4267-bba6-188c8e1d0b96

## How to test the changes?
Edit FixedBattles and set them as TrainerType (make sure to set Variant to Double

## Checklist

- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [ ] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?

- [x] Are all unit tests still passing? (`npm run test:silent`)
  - [ ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?
[...]

PR #398, opened 2025-01-30
[Audio] Remove music preference option

DayKev:
The music preferences option is removed and the game will only use the previously named "mixed"/"all gens" option.

It's unnecessary.

All references to the old "consistent" option are gone.
The `MusicPreferences` enum was deleted.
All `if/else` blocks related to music had the "consistent" option half deleted.
All functions to get/set "consistent" battle music are deleted and only the "mixed" versions remain (with appropriate renaming).
The option to change music preferences is removed.

## How to test the changes?
Fight a trainer or wild legendary that would have had different music in the old system depending on the option (ie: any gym/E4/legendary/etc from gen 6+).

## Checklist


src/data/balance/trainer-configs/champion-trainer-configs.ts
```diff
@@ -10,7 +10,7 @@ import { Type } from "#enums/type";
 const DEFAULT_CHAMPION_THEME = "battle_champion_alder";
 const KANTO_CHAMPION_THEME = "battle_kanto_champion";
 const JOHTO_CHAMPION_THEME = "battle_johto_champion";
-const HOENN5_CHAMPION_THEME = "battle_hoenn_champion_g5";
+//const HOENN5_CHAMPION_THEME = "battle_hoenn_champion_g5";
```
Tempo-anon: Can this be removed? Should all the unused music be removed now?

Tempo-anon (dismissed):

PigeonBar:
https://github.com/Despair-Games/poketernity/blob/27ce618dfc17675feccbf7ab876947902ef1be94/src/ui/settings/settings-audio-ui-handler.ts#L11-L14
`6` needs to be updated to `5`.

MokaStitcher:
> `6` needs to be updated to `5`.

is that how the scrolling not working properly in this menu was fixed back then...
This shouldn't be even here, the input processing should just check the number of options in the menu 😭

MokaStitcher:
I was going to make a separate PR to fix the rowsToDisplay issue because I was anticipating it to be annoying. But actually it's a very small change so I went ahead and pushed it directly. Hope that's fine
<details><summary>Screenshots</summary>

before, pressing up from the top option in audio settings:
image

after, doing the same:
image

</details> 

We should probably make a locale PR to remove the now unused locales too

PigeonBar (dismissed):

src/data/balance/trainer-configs/champion-trainer-configs.ts
DayKev: Oops I meant to give one to Steven and the other to Wallace and got distracted by fixing something else, lemme update that.

DayKev:
Will resolve conflicts after the enum is renamed again lol.

torranx:
Can we change the title and labels? This doesn't count as refactor since it's technically removing a feature no?

DayKev:
> Can we change the title and labels? This doesn't count as refactor since it's technically removing a feature no?

I guess so. What should I mark it as, just "[Audio]"?

MokaStitcher:
> Will resolve conflicts after the enum is renamed again lol.

It doesn't look like this is happening for now? Also would it really cause conflicts with this PR?
The PR that is likely to have conflicts though is #364, I'm not sure which one of those would be best to get in first

DayKev:
> > Will resolve conflicts after the enum is renamed again lol.
> 
> It doesn't look like this is happening for now? Also would it really cause conflicts with this PR? The PR that is likely to have conflicts though is #364, I'm not sure which one of those would be best to get in first

The trainers are given signature types or w/e in the trainer config files, and since I edited those files the renaming of the `Type` enum causes conflicts.

CodeTappert:
> > Will resolve conflicts after the enum is renamed again lol.
> 
> It doesn't look like this is happening for now? Also would it really cause conflicts with this PR? The PR that is likely to have conflicts though is #364, I'm not sure which one of those would be best to get in first

well at the end its the same work then needs to be done on the changes of #364 
So it doesnt really matter in which order?

CodeTappert (dismissed):

PigeonBar (dismissed):
Fun fact: This PR also happens to fix a bug where the Weird Dream's fight option would play the default trainer music instead of the ME music

src/data/trainer-config.ts
```diff
@@ -1088,7 +1081,6 @@ export class TrainerConfig {
     this.setMoneyMultiplier(1.5);
     this.setBoss();
     this.setStaticParty();
-    this.setBattleBgm("battle_plasma_boss");
```
MokaStitcher, line 1091: with this there's no longer a default bgm for admins, so if they are not set there will be no music, is that fine? I wonder if the bgm should be passed as a parameter for all these, like for the evil team leader one. Although not really the scope of this PR I guess

(same a bit below with a few other configs)

MokaStitcher (approved):
lgtm (not tested)

src/data/trainer-config.ts
DayKev: Oh yeah, the inconsistent handling of trainers is kinda... lol. They should all have battle music set at the moment iirc.

DayKev: With no default set does that mean there will be no music at all (if we forget to assign one in the trainer config)?

MokaStitcher: From the quick look at the code I took it seems to mean the normal music will just keep going

DayKev: Alright, well that should at least make it obvious if we encounter a trainer without music specified, which is probably good (for now, until a refactor happens).

PigeonBar (approved):

PR #403, opened 2025-01-31
[AI-Rework] Rewrite enemy command selection

innerthunder:
> [!NOTE]
> This PR is a component of the AI Rework and is directed at the rework's dedicated branch. This PR alone should not be considered a complete feature. 

The AI now has a more rigid heuristic when it comes to evaluating actions for any given turn. With this change, the AI should:
- switch less often between two Pokemon with similarly bad matchups against a player Pokemon
- never switch out when it can safely KO a player Pokemon
- select its strongest attack more consistently compared to other attacks
- virtually never select status moves (effect scores are not yet implemented)
- no longer formulate a move action if forced to skip their turn because of their Commander ability or the Fun and Games Mystery Encounter. This should mean that Sucker Punch and Thunderclap now always fail during the Fun and Games minigame.

See the [AI Rework Proposal](https://docs.google.com/document/d/1olQcsBcgoY4oid2-3nko__AlBzG9BnIeEtTTRHp59DU/edit?usp=sharing) for more details on the new command selection heuristic.

The current AI is dumb when it comes to selecting turn actions.

- Much of the code in `EnemyCommandPhase` has been cleaned up and moved into `EnemyPokemon#getNextCommand`
- The logic to determine if an enemy should switch has been simplified and uses the new matchup scoring heuristic from #356 
- Move command selection has been rewritten to use `EnemyPokemon#getMoveScore` (also from 356)

## Screenshots/Videos

All tests are passing 🔥 

image

### Move Selection changes

The following data from `enemy_command.test.ts` is pulled from the following game state:
- Enemy is a Level 100 Snorlax with Covet, Thief, Flame Wheel, and Splash. Its AI type is `SMART_RANDOM`.
- Opponent (player) is a Level 100 Duskull. None of Duskull's moves are revealed to the enemy

In this game state, the enemy has 2 moves that effectively do nothing: Covet because Duskull is immune to it and Splash because, well, it's Splash. Out of its other 2 moves, Thief is the clear best move to use since its effectiveness is twice that of Flame Wheel. However, running this test on `beta` yields the following move choices over 300 trials.

image

The current AI in this state wastes its turn about 10 percent of the time, and fails to choose its strongest move 30 percent of the time overall.

With the changes in this PR, though, the AI is much more precise in its move selection, and it never picks a move that does nothing if it can be helped. However, because of the variance in attack score output, it doesn't _always_ choose the absolute best move for the situation.

image

## How to test the changes?

`npm run test enemy_command` - 1 new test (old tests are still relevant and passing)

## Checklist

- [n/a] **I'm using `beta` as my base branch**


src/field/pokemon.ts
```diff
@@ -5627,213 +5629,258 @@ export class EnemyPokemon extends Pokemon {
...
 
-          // Sort the move pool in decreasing order of move score
-          const sortedMovePool = movePool.slice(0);
-          sortedMovePool.sort((a, b) => {
-            const scoreA = moveScores[movePool.indexOf(a)];
-            const scoreB = moveScores[movePool.indexOf(b)];
-            return scoreA < scoreB ? 1 : scoreA > scoreB ? -1 : 0;
-          });
-          let r = 0;
-          if (this.aiType === AiType.SMART_RANDOM) {
```
PigeonBar, line 5802: Your new AI code seems to no longer differentiate between `SMART` (used for trainers and wild bosses) and `SMART_RANDOM` (used for regular wild Pokemon). Should we keep some aspect of trainers being smarter than wild Pokemon?

src/field/pokemon.ts
```diff
@@ -5627,213 +5629,258 @@ export class EnemyPokemon extends Pokemon {
...
+      };
+    } else if (move.moveTarget === MoveTarget.RANDOM_NEAR_ENEMY) {
+      /**
+       * Moves with random targeting resolve their final target within {@linkcode getMoveTargets},
+       * but calculate score based on the average move score between all legal targets
+       */
+      const averageScore =
+        this.getOpponents()
+          .map((p) => this.getMoveScore(p, move))
+          .reduce((total, score) => total + score, 0) / this.getOpponents.length;
```
PigeonBar:
```suggestion
          .reduce((total, score) => total + score, 0) / this.getOpponents().length;
```

src/field/pokemon.ts
innerthunder: I think their decision making process should be at least roughly the same (i.e. the logic in `getNextCommand` wouldn't change much between them). Instead, I plan on making more AI types that target the scoring formulas (attack score, matchup score, etc) once the base code is sorted out.

PigeonBar:
I experienced a crash due to an opponent off-field Pokemon's simulated damage calculations trying to use the Pokemon's `battleSummonData` before it has been initialized:

<details><summary>Crash log</summary>
<p>

image

</p>
</details> 

(Maybe we should add an `if (!this.summonData) {this.resetSummonData()}` to `Pokemon`'s constructor to fix all potential similar issues?)

Edit: It turns out that a `this.resetTurnData()` is likely also required.

test/ai/enemy_command.test.ts
```diff
@@ -99,4 +99,25 @@ describe("Enemy Commands - Move Selection", () => {
...
+    enemyPokemon.aiType = AiType.SMART_RANDOM;
+
+    const moveChoices: MoveChoiceSet = {};
+    const enemyMoveset = enemyPokemon.getMoveset();
+    enemyMoveset.forEach((mv) => (moveChoices[mv.moveId] = 0));
+    getEnemyMoveChoices(enemyPokemon, moveChoices);
+
+    expect(moveChoices[Moves.SPLASH]).toBe(0);
+    expect(moveChoices[Moves.COVET]).toBe(0);
+  });
```
PigeonBar, line 139:
```suggestion
  });

  it("should not crash from an off-field enemy Pokemon simulating every move", async () => {
    game.override.startingWave(5);
    await game.classicMode.startBattle([Species.FEEBAS]);
    
    const player = game.field.getPlayerPokemon();
    const offFieldEnemy = game.scene.getEnemyParty()[1];

    for (const move of Object.values(allMoves)) {
      const eas = offFieldEnemy.getExpectedAttackScore(player, move);
      expect(eas).toBeGreaterThanOrEqual(-1);
      expect(eas).toBeLessThanOrEqual(4);
    }
  });
```

innerthunder:
> (Maybe we should add an `if (!this.summonData) {this.resetSummonData()}` to `Pokemon`'s constructor to fix all potential similar issues?)

I'd rather not do that... I have no idea how much of the code relies on null checks for these things.

I just added a null check to `Pokemon#getMoveHistory`; hopefully that fixes the crash you're observing.

test/ai/enemy_command.test.ts
```diff
@@ -99,4 +99,39 @@ describe("Enemy Commands - Move Selection", () => {
...
+    game.override.startingWave(5);
+    await game.classicMode.startBattle([Species.FEEBAS]);
+    
+    const player = game.field.getPlayerPokemon();
+    const offFieldEnemy = game.scene.getEnemyParty()[1];
+
+    for (const move of Object.values(allMoves)) {
+      const eas = offFieldEnemy.getExpectedAttackScore(player, move);
+      expect(eas).toBeGreaterThanOrEqual(-1);
+      expect(eas).toBeLessThanOrEqual(4);
```
PigeonBar, line 134:
```suggestion
      expect(eas).toBeLessThanOrEqual(4);
      const eas2 = player.getExpectedAttackScore(offFieldEnemy, move);
      expect(eas2).toBeGreaterThanOrEqual(-1);
      expect(eas2).toBeLessThanOrEqual(4);
```

innerthunder:
The crash test now passes after the latest commit:

image

PigeonBar (approved):

src/data/move-attrs/turn-damaged-double-power-attr.ts
```diff
@@ -11,7 +11,7 @@ import { VariablePowerAttr } from "#app/data/move-attrs/variable-power-attr";
  */
 export class TurnDamagedDoublePowerAttr extends VariablePowerAttr {
   override apply(user: Pokemon, target: Pokemon, _move: Move, power: NumberHolder): boolean {
-    if (user.turnData.attacksReceived.find((r) => r.damage && r.sourceId === target.id)) {
+    if (user.turnData?.attacksReceived.find((r) => r.damage && r.sourceId === target.id)) {
```
flx-sta: NIT: use `some` (I know this wasn't your change)
```suggestion
    if (user.turnData?.attacksReceived.some((r) => r.damage && r.sourceId === target.id)) {
```

src/overrides.ts
```diff
@@ -164,6 +164,12 @@ class DefaultOverrides {
    * 2+: the Pokemon will be a boss with the given number of health segments
    */
   readonly ENEMY_HEALTH_SEGMENTS_OVERRIDE: number = 0;
+  /**
+   * Prevents enemy Trainers from switching their Pokemon.
+   * Setting this to `true` is highly recommended if you are overriding the
+   * enemy's moveset with only status moves.
+   */
+  readonly DISABLE_ENEMY_SWITCHING_OVERRIDE: boolean = false;
```
flx-sta: NIT: `ENEMY_` prefix should go first:
```suggestion
  readonly ENEMY_DISABLE_SWITCHING_OVERRIDE: boolean = false;
```

src/field/pokemon.ts
```diff
@@ -6124,6 +6171,12 @@ interface AbilityData {
   passive: boolean;
 }
 
+interface TargetScoreData {
+  move: Moves;
```
flx-sta:
```suggestion
  moveId: MoveId;
```

src/field/pokemon.ts
```diff
@@ -5627,213 +5629,258 @@ export class EnemyPokemon extends Pokemon {
...
+       * Moves with random targeting resolve their final target within {@linkcode getMoveTargets},
+       * but calculate score based on the average move score between all legal targets
+       */
+      const averageScore =
+        this.getOpponents()
+          .map((p) => this.getMoveScore(p, move))
+          .reduce((total, score) => total + score, 0) / this.getOpponents().length;
+
+      return {
+        move: move.id,
```
flx-sta:
```suggestion
        moveId: move.id,
```

src/field/pokemon.ts
```diff
@@ -5627,213 +5629,258 @@ export class EnemyPokemon extends Pokemon {
...
+       * Counter-attacks (e.g. Metal Burst) are scored entirely
+       * based on their effect score.
+       */
+      const score = move.getEffectScore(this);
 
-            // could make smarter by checking opponent def/spdef
-            moveScores[m] = moveScore;
-          }
+      return {
+        move: move.id,
```
flx-sta:
```suggestion
        moveId: move.id,
```

src/field/pokemon.ts
```diff
@@ -5627,213 +5629,258 @@ export class EnemyPokemon extends Pokemon {
...
+      (bi) => [bi, this.getMoveScore(globalScene.getFieldPokemonByBattlerIndex(bi)!, move)], // TODO: find a way to get rid of this bang
+    );
+
+    if (multiple) {
+      /**
+       * Multi-targeted moves use the full target set and the sum of move scores
+       * for each target.
+       */
+      return {
+        move: move.id,
```
flx-sta:
```suggestion
        moveId: move.id,
```

src/field/pokemon.ts
```diff
@@ -5627,213 +5629,258 @@ export class EnemyPokemon extends Pokemon {
...
-      }
+    /**
+     * The {@linkcode BattlerIndex | BattlerIndexes} of active Pokemon that
+     * can legally be targeted with this move.
+     */
+    const activeTargets = targets.filter((bi) => !isNullOrUndefined(globalScene.getFieldPokemonByBattlerIndex(bi)));
+    if (activeTargets.length === 0) {
+      /** Moves with no valid targets are given a "fail penalty" of (-5). */
+      return {
+        move: move.id,
```
flx-sta:
```suggestion
        moveId: move.id,
```

src/field/pokemon.ts
```diff
@@ -5627,213 +5629,258 @@ export class EnemyPokemon extends Pokemon {
...
-            let moveScore = moveScores[m];
-            const targetScores: number[] = [];
-
-            for (const mt of moveTargets[move.id]) {
-              // Prevent a target score from being calculated when the target is whoever attacks the user
-              if (mt === BattlerIndex.ATTACKER) {
-                break;
-              }
+    return {
+      move: optMoveAction.move,
```
flx-sta:
```suggestion
      moveId: optMoveAction.move,
```

src/field/pokemon.ts
```diff
@@ -5627,213 +5629,258 @@ export class EnemyPokemon extends Pokemon {
...
+    const movePool = this.getMoveset().filter((m) => m.isUsable(this));
 
-            const move = pkmnMove.getMove()!;
-            if (move.moveTarget === MoveTarget.ATTACKER) {
-              return false;
-            }
+    // If this Pokemon has no usable moves, it will use Struggle.
+    if (movePool.length === 0) {
+      return {
+        move: Moves.STRUGGLE,
```
flx-sta:
```suggestion
        moveId: MoveId.STRUGGLE,
```

src/field/pokemon.ts
```diff
@@ -5627,213 +5629,258 @@ export class EnemyPokemon extends Pokemon {
...
+    const queuedMove = this.getMoveQueue()[0];
     if (queuedMove) {
-      if (queuedMove.isUsable(this, this.getMoveQueue()[0].ignorePP)) {
+      const queuedMovesetMove = this.getMoveset().find((m) => m.moveId === queuedMove.move);
+      if (queuedMovesetMove?.isUsable(this, queuedMove.ignorePP)) {
         return {
-          move: queuedMove.moveId,
-          targets: this.getMoveQueue()[0].targets,
-          ignorePP: this.getMoveQueue()[0].ignorePP,
+          move: queuedMovesetMove.moveId,
```
flx-sta:
```suggestion
          moveid: queuedMovesetMove.moveId,
```

src/field/pokemon.ts
```diff
@@ -5627,213 +5629,258 @@ export class EnemyPokemon extends Pokemon {
...
+    }
+
+    const nextMove = this.getNextMove();
+    console.log(
+      `${BattlerIndex[this.getBattlerIndex()]}: selecting ${Moves[nextMove.move]} against ${nextMove.targets.map((i) => BattlerIndex[i])}`,
+    );
+
+    return {
+      command: BattleCommand.FIGHT,
+      move: nextMove,
```
flx-sta, line 5692:
```suggestion
      moveId: nextMove,
```

innerthunder:
@flx-sta Re: the change from `Moves` to `MoveId`, this branch isn't based in `beta`, so trying to merge in changes from beta will mess up this PR's commit history (I've done it before 💀 )

I usually just merge beta into the `ai-rework` branch after a PR is merged in to bring things up to date. Lmk if you have other suggestions on how to handle this, though.

MokaStitcher:
> I usually just merge beta into the `ai-rework` branch after a PR is merged in to bring things up to date. Lmk if you have other suggestions on how to handle this, though.

You can also merge beta into `ai-rework` then merge `ai-rework` into this branch, and that will do the trick. It's a bit cumbersome but at least you don't need to wait till this is merged to keep it up to date.
Although it's up to you which approach you prefer, since you're the one handling the conflicts in the end

Tempo-anon:
Does the 0% of using covet happen if the player has other Pokemon? Knowing an AI will never use a 0x move would make predictions and switching easier.

innerthunder:
> Does the 0% of using covet happen if the player has other Pokemon? Knowing an AI will never use a 0x move would make predictions and switching easier.

Right now, the AI doesn't account for inactive player Pokemon. However, I think making the AI's actions more predictable in this way is a good thing overall because
1. It feels more fair to lose against things that are more telegraphed. The old AI might make some wild "predictions" that swing the game, but I think the player can catch on that its actions are just random variance. The player may feel like they deserved to lose more if they lose against something predictable rather than something out of pocket that's ultimately a ~6% roll.
2. Most enemies' movesets are random, so the player still has to gather some information on the enemy before they can make accurate reads. The random egg moves and TMs are enough to keep the player on their toes.

Tempo-anon (approved):

PR #422, opened 2025-02-03
[Refactor][Move] Field-based move targeting no longer requires a target Pokemon

innerthunder:

Moves can now target the field, bypassing effectiveness checks. This means that
- All field moves (weather/terrain setting, hazards, and other arena tags) can be used while the target's side of the field is empty.
- Good As Gold is no longer (P) since its main incorrect interaction is now fixed.
- Pressure is now marked (P) since it does not affect field moves (among other edge cases, e.g. Dragon Darts)

This implements field-targeting effects more accurately to mainline.

`BattlerIndex` has 3 new values for move targeting:

```ts
export enum BattlerIndex {
  ATTACKER = -1,
  PLAYER,
  PLAYER_2,
  ENEMY,
  ENEMY_2,
  // NEW
  PLAYER_SIDE,
  ENEMY_SIDE,
  BOTH_SIDES,
}
```

Much like the `ATTACKER` index, these override the base move target resolution and execution in `MovePhase` and `MoveEffectPhase`. If a move has any of these indexes as a target, the move will bypass the usual `targets.length` check in `MovePhase` and execute its effect without performing a hit check against any active Pokemon in `MoveEffectPhase`.

## Screenshots/Videos

<details><summary>Stealth Rock is usable on an empty field</summary>
<p>

https://github.com/user-attachments/assets/73549713-007f-40d7-8961-0f6b9414ccdd

</p>
</details> 

## How to test the changes?

`npm run test spikes` (already includes tests where [Toxic] Spikes are laid on an empty field)
`npm run test good_as_gold`

## Checklist


src/data/move.ts
```diff
@@ -692,7 +693,7 @@ export abstract class Move implements Localizable {
   getTargetBenefitScore(user: Pokemon, target: Pokemon, move: Move): number {
     let score = 0;
 
-    if (target.getAlly()?.getTag(BattlerTagType.COMMANDED)?.getSourcePokemon() === target) {
+    if (target && target.getAlly()?.getTag(BattlerTagType.COMMANDED)?.getSourcePokemon() === target) {
```
DayKev, line 707:
```suggestion
    if (target?.getAlly()?.getTag(BattlerTagType.COMMANDED)?.getSourcePokemon() === target) {
```

src/data/move.ts
```diff
@@ -1119,6 +1120,7 @@ export function getMoveTargets(user: Pokemon, moveId: MoveId): MoveTargetSet {
   const opponents = user.getOpponents();
 
   let set: Pokemon[] = [];
+  let targets: BattlerIndex[] | undefined = undefined;
```
DayKev:
```suggestion
  let targets: BattlerIndex[] | undefined;
```

src/data/move.ts
innerthunder: This becomes `true` when `target` is undefined, so it's not equivalent.

torranx:
I noticed in the video that there is no target confirmation/highlights when selecting Stealth Rock, did this PR remove that or was it done before? 
Because if my memory serves me right, target highlight works for arena moves as well

src/data/move.ts
DayKev: Oh I see.

innerthunder:
> I noticed in the video that there is no target confirmation/highlights when selecting Stealth Rock, did this PR remove that or was it done before? Because if my memory serves me right, target highlight works for arena moves as well

I think you're right... I can try to fix it, but the change in battler index isn't easy to make compatible with the target select UI atm

src/data/all-moves.ts
```diff
@@ -3670,7 +3678,7 @@ export function initMoves() {
       .attr(AddBattlerTagAttr, BattlerTagType.TRAPPED),
     new AttackMove(MoveId.G_MAX_RESONANCE, ElementalType.ICE, MoveCategory.PHYSICAL, 10, -1, 10, -1, 0, 8)
       .gMaxMove()
-      .attr(AddArenaTagAttr, ArenaTagType.AURORA_VEIL, { turnCount: 5, selfSideTarget: true }),
+      .attr(AddArenaTagAttr, ArenaTagType.AURORA_VEIL, ArenaTagRelativeSide.USER, { turnCount: 5 }),
```
Tempo-anon, line 3680: I think some other gmax moves are missing the arena tag relative side field when changing the `AddArenaTagAttr`

innerthunder: They look ok to me. The attribute defaults to `ArenaTagRelativeSide.TARGET` when it isn't specified.

innerthunder:
> > I noticed in the video that there is no target confirmation/highlights when selecting Stealth Rock, did this PR remove that or was it done before? Because if my memory serves me right, target highlight works for arena moves as well
> 
> I think you're right... I can try to fix it, but the change in battler index isn't easy to make compatible with the target select UI atm

<details><summary>Fixed with the latest commit 🚀 </summary>
<p>

https://github.com/user-attachments/assets/c31f7f2c-ab0b-4dd8-bd7a-3285176d0f9d

</p>
</details>

src/ui/target-select-ui-handler.ts
```diff
@@ -131,17 +133,37 @@ export default class TargetSelectUiHandler extends UiHandler {
...
 
-    this.targetsHighlighted = this.isMultipleTargets ? multipleTargets : [singleTarget];
+  /** @returns the {@linkcode Pokemon} to highlight based on the move's targeting */
+  protected getHighlightedPokemon(cursor: number): Pokemon[] {
+    if (this.targets.includes(BattlerIndex.BOTH_SIDES)) {
+      return globalScene.getField(true);
+    } else if (this.targets.includes(BattlerIndex.ENEMY_SIDE)) {
+      return globalScene.getEnemyField().filter((p) => p.isActive(true));
+    } else if (this.targets.includes(BattlerIndex.PLAYER_SIDE)) {
+      return globalScene.getEnemyField().filter((p) => p.isActive(true));
```
DayKev: These are both `.getEnemyField()`, that seems like a mistake?

torranx (dismissed):

DayKev (dismissed):
Code looks good.

src/data/move.ts
```diff
@@ -1152,25 +1153,39 @@ export function getMoveTargets(user: Pokemon, moveId: MoveId): MoveTargetSet {
...
+      targets
+      ?? set
+        .filter((p) => p?.isActive(true))
+        .map((p) => p.getBattlerIndex())
+        .filter((t) => t !== undefined),
     multiple,
   };
 }
+
+export function isFieldTargeted(targets: BattlerIndex[]) {
```
PigeonBar, line 1197: nit: Is it possible to use `Move.isFieldTarget()` instead of this function, or would that lead to messy code due to needing to retrieve the move?

src/phases/move-effect-phase.ts
```diff
@@ -316,6 +321,44 @@ export class MoveEffectPhase extends HitCheckPhase {
...
+       * NOTE: this assumes all field effects are implemented with the
+       * `POST_APPLY` move effect trigger and are internally self-targeted.
+       */
+      this.triggerMoveEffects(MoveEffectTrigger.POST_APPLY, user, null);
+
+      // Log this move action as a success
+      user.pushMoveHistory({
+        moveId: this.move.moveId,
+        targets: this.targets,
+        result: MoveResult.SUCCESS,
```
PigeonBar, line 356:
```suggestion
        result: MoveResult.SUCCESS,
        virtual: this.move.virtual,
```

src/phases/move-effect-phase.ts
```diff
@@ -316,6 +321,44 @@ export class MoveEffectPhase extends HitCheckPhase {
...
   }
 
+  /**
+   * Applies all effects for moves that target one or both sides of the field.
+   * This assumes such effects are implemented with {@linkcode MoveEffectTrigger | POST_APPLY}
+   * effect triggers, and will try to play an animation even if no active Pokemon
+   * are affected.
+   * @param user the {@linkcode Pokemon} using the move
+   */
+  private applyFieldMoveEffects(user: Pokemon): void {
```
PigeonBar, line 329:
```suggestion
  private applyFieldMoveEffects(user: Pokemon): void {
    // Lapse `MOVE_EFFECT` effects (i.e. semi-invulnerability) when applicable
    user.lapseTags(BattlerTagLapseType.MOVE_EFFECT);

```
(Just in case we add any `MOVE_EFFECT` battler tags that aren't semi-invulnerability later on)

src/phases/move-phase.ts
```diff
@@ -195,12 +195,15 @@ export class MovePhase extends BattlePhase {
...
+   * Check for cancellation edge cases.
+   * Currently only checks if {@linkcode MoveId.NONE}
+   * is in the user's move queue
+   */
   protected resolveFinalPreMoveCancellationChecks(): void {
-    const targets = this.getActiveTargetPokemon();
     const moveQueue = this.pokemon.getMoveQueue();
 
-    if (targets.length === 0 || (moveQueue.length && moveQueue[0].moveId === MoveId.NONE)) {
+    if (moveQueue.length && moveQueue[0].moveId === MoveId.NONE) {
```
PigeonBar:
```suggestion
    const targets = this.getActiveTargetPokemon();
    if ((targets.length === 0 && !allMoves[this.move.moveId].isFieldTarget())
      || (moveQueue.length && moveQueue[0].moveId === MoveId.NONE)) {
```

src/data/move.ts
innerthunder: I made this mainly because the target select UI handler doesn't have direct access to the move being used; I wanted to keep it that way.

DayKev (approved):

torranx (approved):

PR #432, opened 2025-02-04
[Refactor] Remove `StatusEffect.FAINT`, add `Pokemon.faint()`

### LINKED ISSUE #497: [BUG] Splicing together an alive and a fainted pkmn will result in the fusion being alive (at avaraged HP) but displying FNT status
Copied over from PKR 2608

The fused pkmn can still be used in battle.

Step to reproduce: fuse together pkmn A (healthy) and B (fainted), in my case the second one was fainted but idk it that has nothing to do with the bug.
image

DayKev:
None, theoretically (except less bugs perhaps).

Being fainted is not a status effect, and should not be treated as such. This is/has been the cause of various bugs.

- All checks/references to `StatusEffect.FAINT` are removed.
- `Pokemon.isFainted()` was updated to remove the optional parameter that checked for `StatusEffect.FAINT` (it was only actually used 2-3 times, further proving its redundancy).
- `Pokemon.faint()` added. Currently it only sets the Pokémon's HP to 0 and removes statuses; it should be expanded in the future to handle the various side-effects of a Pokémon fainting instead of them being handled manually/inconsistently everywhere.
- The party UI and party summary handlers were updated to account for the change.
- Unused and single-use functions were removed from `status-effect.ts`.
- `Pokemon.resetStatus()` was updated to account for the change.

## Checklist


src/ui/party-ui-handler.ts
```diff
@@ -1379,9 +1379,10 @@ class PartySlot extends Phaser.GameObjects.Container {
       slotInfoContainer.add(splicedIcon);
     }
 
-    if (this.pokemon.status) {
+    if (this.pokemon.status || this.pokemon.isFainted()) {
+      const effectIcon = this.pokemon.isFainted() ? "faint" : StatusEffect[this.pokemon.status!.effect].toLowerCase();
```
MokaStitcher: the same thing probably needs to be done for the summary ui handler

src/field/pokemon.ts
```diff
@@ -531,11 +532,18 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
    */
-  public isFainted(checkStatus: boolean = false): boolean {
-    return this.hp <= 0 && (!checkStatus || this.status?.effect === StatusEffect.FAINT);
+  public isFainted(): boolean {
+    return this.hp <= 0;
+  }
+
+  /**
+   * Faints the pokemon.
+   * @todo Handle all fainting-related side-effects here
```
MokaStitcher, line 511: is this PR still draft?
When fainting we should remove all status effects so that the Pokemon doesn't keep them after getting revived (before this was handled automatically by having faint being a status effect :p)

src/phases/faint-phase.ts
```diff
@@ -242,9 +241,9 @@ export class FaintPhase extends PokemonPhase {
             .forEach((p) => p.removeTagsBySourceId(pokemon.id));
 
           pokemon.y -= 150;
-          pokemon.trySetStatus(StatusEffect.FAINT);
+          // pokemon.faint();
```
MokaStitcher: is this being commented normal?

MokaStitcher (commented):
very shallow look

src/field/pokemon.ts
DayKev: True, I'll add that to the function.

I was just replacing existing code without changing functionality (if possible) with this PR.

src/phases/faint-phase.ts
DayKev: Hm, should probably replace this with removing the status for now.

PigeonBar (dismissed):

src/field/pokemon.ts
flx-sta: I assume this is still open/pending?

Tempo-anon:
Got some conflicts but this should be fine. Playtesting will be the only real way to spot any tiny edgecases

DayKev:
Ah, will update later today.

src/phases/faint-phase.ts
```diff
@@ -244,9 +243,9 @@ export class FaintPhase extends PokemonPhase {
             .forEach((p) => p.removeTagsBySourceId(pokemon.id));
 
           pokemon.y -= 150;
-          pokemon.trySetStatus(StatusEffect.FAINT);
+          pokemon.status = null;
```
DayKev: Thinking about it, should `FaintPhase` set the pokemon's HP to 0 just in case?

public/locales
MokaStitcher, line 1: this PR doesn't update locales right? Can you bring this branch up to date with beta on that front?

src/phases/abstract-pokemon-phase.ts
```diff
@@ -32,6 +32,7 @@ export abstract class PokemonPhase extends FieldPhase {
   }
 
   public getPokemon(): Pokemon {
+    // TODO: change to `: PlayerPokemon | EnemyPokemon | nil`
```
MokaStitcher, line 35: I assume this is for another PR?

src/ui/party-ui-handler.ts
```diff
@@ -1366,9 +1366,10 @@ class PartySlot extends Phaser.GameObjects.Container {
...
```
MokaStitcher:
```suggestion
      const effectIconKey = this.pokemon.isFainted() ? "faint" : StatusEffect[this.pokemon.status!.effect].toLowerCase();
```

src/ui/party-ui-handler.ts
```diff
@@ -1366,9 +1366,10 @@ class PartySlot extends Phaser.GameObjects.Container {
...
       const statusIndicator = globalScene.add.sprite(0, 0, getLocalizedSpriteKey("statuses"));
-      statusIndicator.setFrame(StatusEffect[this.pokemon.status?.effect].toLowerCase());
+      statusIndicator.setFrame(effectIcon);
```
MokaStitcher:
```suggestion
      statusIndicator.setFrame(effectIconKey);
```

src/field/pokemon.ts
```diff
@@ -499,11 +500,19 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   */
+  public faint(): void {
+    this.hp = 0;
+    this.status = null;
```
MokaStitcher: I think we should call this.resetStatus() instead, specifically because of this, which resets the Pokemon animation speed if the previous status was sleep
https://github.com/Despair-Games/poketernity/blob/d1b4f6d1a4bd811d7d3ad698c5a343acb5e4cbd3/src/field/pokemon.ts#L4166-L4167
Maybe also because of the confusion tag removal?
```suggestion
    this.resetStatus(true);
```

DayKev: Good point.

src/phases/abstract-pokemon-phase.ts
DayKev: Yeah, I don't remember what issue I saw there but it wasn't related to this PR.

public/locales
DayKev: I don't really know why it says the submodule is being updated by the PR when it's not, I think we should just ignore it since it isn't actually changing anything afaict.

src/field/pokemon.ts
DayKev: It's something that needs to be followed up on after this PR still yeah.

src/ui/party-ui-handler.ts
```diff
@@ -1359,9 +1359,10 @@ class PartySlot extends Phaser.GameObjects.Container {
       slotInfoContainer.add(splicedIcon);
     }
 
-    if (this.pokemon.status) {
+    if (this.pokemon.status || this.pokemon.isFainted()) {
+      const effectIconKey = this.pokemon.isFainted() ? "faint" : StatusEffect[this.pokemon.status!.effect].toLowerCase();
       const statusIndicator = globalScene.add.sprite(0, 0, "status_icons");
-      statusIndicator.setFrame(StatusEffect[this.pokemon.status?.effect].toLowerCase());
+      statusIndicator.setFrame(effectIconKey);
```
MokaStitcher, line 1365: doesn't matter, but for reference the frame can be set directly as the 4th parameter of `globalScene.add.sprite`

src/ui/party-ui-handler.ts
```diff
@@ -1359,9 +1359,10 @@ class PartySlot extends Phaser.GameObjects.Container {
       slotInfoContainer.add(splicedIcon);
     }
 
-    if (this.pokemon.status) {
+    if (this.pokemon.status || this.pokemon.isFainted()) {
```
MokaStitcher, line 1362: agh wait, the KO status wasn't shown in party ui handler before apparently? So confused
It's not really needed I think because of the red background + empty HP bar, so we may want to revert this
<details><summary>pics</summary>

beta:
image
this branch: (looks a bit weird imo especially when the mon is highlighted)
image
image

</details>

src/ui/summary-ui-handler.ts
```diff
@@ -463,9 +463,17 @@ export default class SummaryUiHandler extends UiHandler {
 
     const fromSummary = args.length >= 2;
 
-    if (this.pokemon.status || this.pokemon.pokerus) {
+    let statusTextKey: string | undefined;
+    if (this.pokemon.isFainted()) {
+      statusTextKey = "faint";
```
MokaStitcher, line 468: Same thing, the faint status actually wasn't shown before it seems like. But I actually think it's better with it

before:
image
after:
image
</details>

src/modifier/modifier.ts
```diff
@@ -2103,7 +2103,7 @@ export class PokemonInstantReviveModifier extends PokemonHeldItemModifier {
     });
 
     // Remove the Pokemon's FAINT status
```
MokaStitcher:
```suggestion
    // Remove any status the Pokemon had before fainting
```
(I tested and confirmed that without the status reset they don't get cleared, so it was needed for all status, not just faint)

MokaStitcher (commented):
Updated the locales to latest beta, so don't forget to pull before doing changes

Did some testing, realized we may have gone a bit overboard with adding the status icon to the uis, otherwise things look good.

src/modifier/modifier.ts
DayKev: Whoops, I guess I forgot to update the comment.

src/ui/party-ui-handler.ts
DayKev: Wait, wtf? What am I missing here, it seems like it should have been...

MokaStitcher: I know I'm a bit confused too, but the branch is up to date with beta so I don't think this would be a beta issue, right?
I can try again on an older commit to make sure we haven't broken anything somehow at some point

DayKev: I'm gonna go check to see what it looks like on Pokerogue as well.

DayKev: Okay on Pokerogue it shows up in the party UI and the summary, so it's something we changed.
<details><summary>Party UI</summary>
<p>

image

</p>
</details> 
<details><summary>Summary UI</summary>
<p>

image

</p>
</details>

MokaStitcher: yes I was just testing that as well. I think what I did was kill the caterpie on this branch, switch to beta, and reload that same save. So maybe it was a bug with save loading?
Let me try and reproduce it 🤔

MokaStitcher: whoops, that was the reason, all is good actually
Since I made a save on your branch with a KO'd mon it didn't have the faint status in the save file. So when loading it on beta it wouldn't show 🥲

MokaStitcher (dismissed):

MokaStitcher:
Confirmed as fixing #497 (even if it doesn't really matter)

image


<sub> also conflicts just popped up xd

DayKev:
Will update in a bit lol.

MokaStitcher (approved):
(updated locales again, hopefully this'll get merged soon so it's no longer necessary)

flx-sta (approved):

PR #435, opened 2025-02-05
[P2] Form changes now trigger learning EVOLVE_MOVEs

Tempo-anon:

Pokemon that change forms will now be prompted to learn their EVOLVE_MOVEs (Zacian, Zamazenta, Urshifu, Charizard)

Previously Pokemon that changed forms were not prompted to learn the EVOLVE_MOVE

* Copied some code from `level-up-phase` into `form-change-phase`
```ts
const formChangeLearnMove = this.pokemon.getLevelMoves(EVOLVE_MOVE, true);
for (const lm of formChangeLearnMove) {
  globalScene.unshiftPhase(new LearnMovePhase(globalScene.getPlayerParty().indexOf(this.pokemon), lm[1]));
}
```
* Broke up `pokemon-form-level-moves` into a new file for `pokemon-form-level-moves`
* Documented `pokemon-form-level-moves` so the form index corresponds to a human readable format
* Added G-Max Charizard into `pokemon-form-level-moves` with G-Max wildfire, may add the rest in a separate PR

## Screenshots/Videos

image

## How to test the changes?

```ts
const overrides = {
  STARTER_SPECIES_OVERRIDE: Species.ZACIAN,
  STARTING_LEVEL_OVERRIDE:1000,
  STARTING_WAVE_OVERRIDE: 196,
  ITEM_REWARD_OVERRIDE: [{name: "RARE_FORM_CHANGE_ITEM"}],
  STARTING_MODIFIER_OVERRIDE: [{name: "LOCK_CAPSULE"}, {name: "GOLDEN_POKEBALL", count: 5}],
  WAIVE_SHOP_FEES_OVERRIDE: true,
}
```

## Checklist


DayKev (dismissed):
This also makes Mega evolutions try to learn the base form's evolution move (ie: Mega Venusaur tries to learn Petal Blizzard even though that's an evo move for regular Venusaur). If we don't care about that, then 👍.

MokaStitcher:
This may require a bit more complexity than that. Here's [how it's done for evolutions](<https://github.com/Despair-Games/poketernity/blob/591bde7221e8e42bca977b5df55c7db5d13b77ee/src/phases/evolution-phase.ts#L255>)
The different part of the fusion are taken into account so that only the part that changes gets its move.
What Kev reports is also a bit weird and could be handled similarly to what's done for evolutions, thanks to a new LearnMoveSituation specific to form changes if needed

Tempo-anon:
> This also makes Mega evolutions try to learn the base form's evolution move (ie: Mega Venusaur tries to learn Petal Blizzard even though that's an evo move for regular Venusaur). If we don't care about that, then 👍.

That's just a side effect of them not having form level up moves yet. For the Charizard example, when g-maxing a Charizard it only tries to learn gmax wildfire and not air slash (because I removed air slash as an EVOLVE_MOVE from g-max charizard's learnset)



> This may require a bit more complexity than that. Here's [how it's done for evolutions](https://github.com/Despair-Games/poketernity/blob/591bde7221e8e42bca977b5df55c7db5d13b77ee/src/phases/evolution-phase.ts#L255) The different part of the fusion are taken into account so that only the part that changes gets its move. What Kev reports is also a bit weird and could be handled similarly to what's done for evolutions, thanks to a new LearnMoveSituation specific to form changes if needed

chained ternaries and fusion logic... urgh... I guess I can copy some of the logic there to only check the part that changes otherwise it is a minor issue
image

src/phases/form-change-phase.ts
```diff
@@ -197,6 +199,17 @@ export class FormChangePhase extends FormChangeBasePhase {
   public override end(): void {
     const { ui } = globalScene;
 
+    /**
+     * Edgecase
+     *
+     * If a fused Pokemon has either of its halves change form then it will attempt to learn the
+     * EVOLVE_MOVE of both halves of the fusion
+     */
```
innerthunder:
```suggestion
    /**
     * @todo If a fused Pokemon has either of its halves change form then it will attempt to learn the
     * EVOLVE_MOVE of both halves of the fusion
     */
```

src/data/balance/pokemon-level-moves.ts
flx-sta, line 1: Oh oh.... This doesn't look good. It seems that git didn't make the connection that you renamed this file to `pokemon-form-level-moves` thus any future PRs might have trouble resolving conflicts around those 2 files. 
There is two options:
1. Wait for bigger PRs to be finished first
2. Fix it somehow (no ideas except remaking the branch)

src/phases/form-change-phase.ts
```diff
@@ -197,6 +199,17 @@ export class FormChangePhase extends FormChangeBasePhase {
...
+    const formChangeLearnMove = this.pokemon.getLevelMoves(EVOLVE_MOVE, true);
+    for (const lm of formChangeLearnMove) {
+      globalScene.unshiftPhase(new LearnMovePhase(globalScene.getPlayerParty().indexOf(this.pokemon), lm[1]));
+    }
```
flx-sta: NIT: use array destructure
```suggestion
    for (const [ , secondMoveToLearn ] of formChangeLearnMove) {
      globalScene.unshiftPhase(new LearnMovePhase(globalScene.getPlayerParty().indexOf(this.pokemon), secondMoveToLearn]));
    }
```

flx-sta (changes requested):

src/data/balance/pokemon-level-moves.ts
Tempo-anon: I didn't rename the file, I broke out this file into two different ones

src/phases/form-change-phase.ts
Tempo-anon: I can also make this change in the `level-up-phase` where I stole it from

src/data/balance/pokemon-level-moves.ts
flx-sta: Well that’s going to be fun for me to resolve if I’m  merged after you

flx-sta (approved):

PR #426, opened 2025-02-03
[Refactor] Eliminate circular imports | `3XX` -> `0`

flx-sta:
> [!CAUTION]
> PAY EXTRA ATTENTION TO `init-abilities`, `init-moves` etc. files!
> AS I HAVE TO MANUALLY ADOPT ANY CHANGES FROM THE ORIGINAL `all-*` FILES INTO THIS PR! 
  
## Why am I making these changes?

Eliminate the constant impact of circular dependenciesa

## What are the changes from a developer perspective?

* Please refer to the commit messages

> [!TIP]
> Added github action that executes and checks dependency injections

## How to test the changes?

- Game still starts & runs
- Tests succeed
- Run `npm run depcruise` to verify the given numbers

> [!TIP]
> `✔ no dependency violations found (1024 modules, 7619 dependencies cruised)`

## Checklist

- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?

- [x] Are all unit tests still passing? (`npm run test:silent`)

----

> [!NOTE]
> PR description copied from #302

PR #440, opened 2025-02-05
[Test] Fix reload test getting stuck on Map input

PigeonBar:

~~It appears that due to changes to the API code in #426, the code no longer considers itself to be connected to a server during tests. This means that daily runs during tests no longer use the API's mocked RNG seed, which can cause the reload tests to get stuck on a map input on certain days.~~

This API issue is getting fixed, but we should probably still have code for getting past the Map input just in case.

Added code to send an input to get past the Map prompt.

## How to test the changes?

`npm run test reload`
You can change the date used for generating the daily run over here:
https://github.com/Despair-Games/poketernity/blob/0bb0cbf5947334b83a6deba573c5bdf593386b22/src/phases/title-phase.ts#L300

## Checklist

- ~~[ ] **I'm using `beta` as my base branch**~~

### DIFF
--- a/test/rng/reload.test.ts
+++ b/test/rng/reload.test.ts
@@ -5,6 +5,8 @@ import { MoveId } from "#enums/move-id";
 import { Species } from "#enums/species";
 import { GameManager } from "#test/testUtils/gameManager";
 import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
+import { UiMode } from "#enums/ui-mode";
+import { Button } from "#enums/buttons";
 
 describe("Reload", () => {
   let phaserGame: Phaser.Game;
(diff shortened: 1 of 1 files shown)


Tempo-anon (approved):
While this patches this issue up, does this fix the underlining issue where unit tests for daily mode aren't on a set seed?

flx-sta:
@PigeonBar I appreciate this PR as it gives me insight into what is causing this issue. I'd like to investigate how to emulate the original behavior first before I'll use this solution

flx-sta:
We decided to go with a different approach.

https://github.com/Despair-Games/poketernity/pull/426/commits/32e1fd3a99bf3f47071b51ede36a44a29a7598a6

PR #442, opened 2025-02-06
[AI-Rework] Add move-property-based scoring

innerthunder:
> [!NOTE]
> This PR targets the `ai-rework` branch and should not be considered a complete feature 

The AI now properly evaluates priority, critical-hit ratio, and multi-hit moves, and it should use inaccurate moves less often.

Part of the AI Rework

Logic has been added for the AI to recognize and evaluate the following properties.

### Priority

If a move with increased priority is expected to KO, its Attack Score (AS) is now (+6) instead of (+4). This means that the AI should always prefer priority moves that KO over non-priority moves that KO. No other bonuses are given to priority moves.

### Critical Hits

Moves gain a bonus if:
1. The move is not expected to KO without a critical hit, but KOs with a critical hit.
2. The move has an increased critical hit ratio or is guaranteed to critically hit.

This bonus is equal to (+1), with a chance of an additional (+1) equal to the critical hit rate of the move (%, rounded down).

### Accuracy

Moves are given a penalty if their accuracy is low. The penalty is based on the move's perceived final accuracy:

| Accuracy | Penalty | Examples |
|--------|--------|--------|
| 80-100 | 0 | - |
| 70-79 | -1 | Focus Blast |
| 50-69 | -2 | Zap Cannon |
| 0-49 | -3 | Fissure |

Effects that cause the move to bypass accuracy (e.g. either the user or target's No Guard) negate this penalty, and if the target is semi-invulnerable, moves are given a (-5) penalty instead.

### Multi-Hit Moves

When the AI calculates damage, it now multiplies the damage of a single strike by a multiplier according to the move's multi-hit type. For multi-hit types that have a range in hit count, this multiplier is approximately equal to the expected hit count. For example, moves that hit 2-5 times are given a 3x multiplier (they hit 3.1 times on average)

For more information, see the Property-based Score Modifiers on the [AI Rework Effect Score Sheet](https://docs.google.com/spreadsheets/d/12UprXLHyLdw-ESls6pAQecyBFCG2tVg5cllAiUbp6KA/edit?usp=sharing)

## How to test the changes?

`npm run test ai/[priority | low_accuracy]`

## Checklist

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -47,6 +47,11 @@ import { HealStatusEffectAttr } from "./move-attrs/heal-status-effect-attr";
 import { ChargeAnim } from "#enums/charge-anim";
 import { allMoves } from "#app/data/all-moves";
 import { StatStageChangeAttr } from "#app/data/move-attrs/stat-stage-change-attr";
+import { AlwaysHitAbAttr } from "./ab-attrs/always-hit-ab-attr";
+import { AbilityApplyMode } from "#enums/ability-apply-mode";
+import { MultiHitPowerIncrementAttr } from "./move-attrs/multi-hit-power-increment-attr";
+import { MultiHitType } from "#enums/multi-hit-type";
+import { MaxMultiHitAbAttr } from "./ab-attrs/max-multi-hit-ab-attr";
 
 export abstract class Move implements Localizable {
   public id: MoveId;
@@ -724,7 +729,7 @@ export abstract class Move implements Localizable {
    * @returns a score value accumulated from effect score modifiers.
    * @todo Add Low Accuracy Penalty and Ally Target Penalty
    */
-  public getEffectScore(user: EnemyPokemon, target?: Pokemon): number {
+  public getEffectScore(user: EnemyPokemon, target: Pokemon): number {
     // penalize targeting Pokemon that are hidden by Commander
     if (
       (target && target.getAlly()?.getTag(BattlerTagType.COMMANDED)?.getSourcePokemon() === target)
@@ -739,9 +744,10 @@ export abstract class Move implements Localizable {
     /** The combined score from all conditions of the move */
     const conditionScores = this.conditions.map((cond) => cond.getConditionScore(user, target, this));
 
-    const totalScore = attrScores.concat(conditionScores).reduce((total, score) => total + score, 0);
+    /** The penalty given if the move is inaccurate */
+    const accuracyPenalty = this.getBattleAccuracyPenalty(user, target);
 
-    // @todo apply low accuracy penalty + ally target penalty to totalScore
+    const totalScore = attrScores.concat(conditionScores).reduce((total, score) => total + score, 0) + accuracyPenalty;
 
     return totalScore;
   }
@@ -782,6 +788,62 @@ export abstract class Move implements Localizable {
     }
 
     return moveAccuracy.value;
+  }
+
+  /**
+   * Calculates the score penalty granted to this move based on its accuracy, according
+   * to the table below:
+   *   ```
+   *   +--------+-------+
+   *   | Acc    | Score |
+   *   +--------+-------+
+   *   | 80-100 |     0 |
+   *   | 70-79  |    -1 |
+   *   | 50-69  |    -2 |
+   *   | 0-49   |    -3 |
+   *   +--------+-------+
+   *   ```
+   * This penalty is negated if any ongoing effect would cause the move to bypass accuracy checks,
+   * and it becomes (-5) if the target is semi-invulnerable.
+   * @param user the {@linkcode EnemyPokemon} using this move
+   * @param target the {@linkcode Pokemon} targeted by this move
+   * @returns the score penalty from accuracy
+   * @see {@linkcode getEffectScore}
+   */
+  protected getBattleAccuracyPenalty(user: EnemyPokemon, target: Pokemon): number {
+    /**
+     * If any ongoing effect would cause the move to bypass accuracy checks, assign no penalty.
+     * @todo the target's No Guard can be discovered prematurely here
+     */
+    if (
+      [user, target].some((p) => p.hasAbilityWithAttr(AlwaysHitAbAttr))
+      || user.getTag(BattlerTagType.IGNORE_ACCURACY)
+      || target.getTag(BattlerTagType.ALWAYS_GET_HIT)
+      || target.getTag(BattlerTagType.TELEKINESIS)
+    ) {
+      return 0;
+    }
+
+    /**
+     * If the user is faster than the target, and the target is semi-invulnerable,
+     * assign a (-5) penalty.
+     */
+    const userSpd = user.getEffectiveStat(Stat.SPD);
+    const targetSpd = target.getEffectiveStat(Stat.SPD, undefined, undefined, AbilityApplyMode.REVEALED);
+    if (target.isSemiInvulnerable() && userSpd > targetSpd) {
+      return -5;
+    }
+
+    const accuracy = this.calculateBattleAccuracy(user, target, true);
+    if (accuracy < 0 || accuracy >= 80) {
+      return 0;
+    } else if (accuracy >= 70) {
+      return -1;
+    } else if (accuracy >= 50) {
+      return -2;
+    } else {
+      return -3;
+    }
   }
 
   /**
@@ -857,6 +919,45 @@ export abstract class Move implements Localizable {
     }
 
     return power.value;
+  }
+
+  /**
+   * Obtains an approximate damage multiplier accounting for multi-hit
+   * move and ability modifiers relative to the move's first strike.
+   * @param user the {@linkcode Pokemon} using this move
+   * @returns a damage modifier to be used for move scoring
+   * @see {@linkcode Pokemon.getAttackScore}
+   */
+  public getMultiHitAttackScoreMultiplier(user: Pokemon): number {
+    let multiHitMultiplier = 1;
+    const skillLink = user.hasAbilityWithAttr(MaxMultiHitAbAttr);
+    if (this.hasAttr(MultiHitPowerIncrementAttr)) {
+      multiHitMultiplier *= 6; // assumes 3 hits: 1x + 2x + 3x
+    } else {
+      const multiHitType = this.getAttrs(MultiHitAttr)?.[0]?.getMultiHitType();
+      switch (multiHitType) {
+        case MultiHitType._2:
+          multiHitMultiplier *= 2;
+          break;
+        case MultiHitType._2_TO_5:
+          multiHitMultiplier *= skillLink ? 5 : 3; // Expected hit count is ~3.1
+          break;
+        case MultiHitType._3:
+          multiHitMultiplier *= 3;
+          break;
+        case MultiHitType._10:
+          multiHitMultiplier *= skillLink ? 10 : 6; // Population Bomb hits ~5.8 times on average
+          break;
+        case MultiHitType.BEAT_UP:
+          multiHitMultiplier *= user.getParty().length;
+      }
+    }
+
+    if (user.hasAbility(Abilities.PARENTAL_BOND)) {
+      multiHitMultiplier += 0.25;
+    }
+
+    return multiHitMultiplier;
   }
 
   getPriority(user: Pokemon, simulated: boolean = true) {
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -929,35 +929,28 @@ export abstract class Move implements Localizable {
    * @see {@linkcode Pokemon.getAttackScore}
    */
   public getMultiHitAttackScoreMultiplier(user: Pokemon): number {
-    let multiHitMultiplier = 1;
-    const skillLink = user.hasAbilityWithAttr(MaxMultiHitAbAttr);
+    const userHasSkillLink = user.hasAbilityWithAttr(MaxMultiHitAbAttr);
     if (this.hasAttr(MultiHitPowerIncrementAttr)) {
-      multiHitMultiplier *= 6; // assumes 3 hits: 1x + 2x + 3x
-    } else {
-      const multiHitType = this.getAttrs(MultiHitAttr)?.[0]?.getMultiHitType();
+      return 6; // assumes 3 hits: 1x + 2x + 3x
+    } else if (this.hasAttr(MultiHitAttr)) {
+      const multiHitType = this.getAttrs(MultiHitAttr)[0].getMultiHitType();
       switch (multiHitType) {
         case MultiHitType._2:
-          multiHitMultiplier *= 2;
-          break;
+          return 2;
         case MultiHitType._2_TO_5:
-          multiHitMultiplier *= skillLink ? 5 : 3; // Expected hit count is ~3.1
-          break;
+          return userHasSkillLink ? 5 : 3; // Expected hit count is ~3.1
         case MultiHitType._3:
-          multiHitMultiplier *= 3;
-          break;
+          return 3;
         case MultiHitType._10:
-          multiHitMultiplier *= skillLink ? 10 : 6; // Population Bomb hits ~5.8 times on average
-          break;
+          return userHasSkillLink ? 10 : 6; // Population Bomb hits ~5.8 times on average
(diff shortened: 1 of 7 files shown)


DayKev:
How'd you decide on these tiers for the accuracy?

Also, why not multiply the BP by the acc as an "expected BP" value?

src/data/move.ts
```diff
@@ -784,6 +790,62 @@ export abstract class Move implements Localizable {
...
   }
 
+  /**
+   * Calculates the score penalty granted to this move based on its accuracy, according
+   * to the table below:
+   *   ```
+   *   +--------+-------+
+   *   | Acc    | Score |
+   *   +--------+-------+
+   *   | 80-100 |     0 |
```
Tempo-anon, line 800: nit: Is there a reason this cut off is 80? I imagine most cases the real breakpoint is 85 (fire blast vs flamethrower). The only things in 80 are stone edge, gunk shot, head smash, cross chop (usually negated with no guard), hydro pump, and the storms. I know a +/-1 in score is huge but it's just a bit weird to see the 100acc reliable damage moves with the same score as the more damage but less acc ones

src/data/move.ts
```diff
@@ -784,6 +790,62 @@ export abstract class Move implements Localizable {
...
+   *   | 70-79  |    -1 |
+   *   | 50-69  |    -2 |
+   *   | 0-49   |    -3 |
```
Tempo-anon, line 803: random aside: it's funny just how few moves are in this range

src/data/move.ts
```diff
@@ -859,6 +921,45 @@ export abstract class Move implements Localizable {
...
+  /**
+   * Obtains an approximate damage multiplier accounting for multi-hit
+   * move and ability modifiers relative to the move's first strike.
+   * @param user the {@linkcode Pokemon} using this move
+   * @returns a damage modifier to be used for move scoring
+   * @see {@linkcode Pokemon.getAttackScore}
+   */
+  public getMultiHitAttackScoreMultiplier(user: Pokemon): number {
+    let multiHitMultiplier = 1;
+    const skillLink = user.hasAbilityWithAttr(MaxMultiHitAbAttr);
```
Tempo-anon: nit: since this is a boolean the name should instantly reflect that `userHasSkillLink`

src/data/move.ts
```diff
@@ -859,6 +921,45 @@ export abstract class Move implements Localizable {
...
+          multiHitMultiplier *= 2;
+          break;
+        case MultiHitType._2_TO_5:
+          multiHitMultiplier *= skillLink ? 5 : 3; // Expected hit count is ~3.1
+          break;
+        case MultiHitType._3:
+          multiHitMultiplier *= 3;
+          break;
+        case MultiHitType._10:
+          multiHitMultiplier *= skillLink ? 10 : 6; // Population Bomb hits ~5.8 times on average
```
Tempo-anon: Is this correct? Skill link just makes population bomb a single accuracy check rather than checking for accuracy each hit. I thought this should only be calculating the expected number of hits without accounting for accuracy as accuracy isn't accounted for in the other parts

src/data/move.ts
```diff
@@ -859,6 +921,45 @@ export abstract class Move implements Localizable {
...
+          break;
+        case MultiHitType.BEAT_UP:
+          multiHitMultiplier *= user.getParty().length;
```
Tempo-anon: Is there code elsewhere to calculate beat up power? The multi hit multiplier isn't needed because the exact bp of beat up can be calculated from the user's party

src/data/move.ts
```diff
@@ -859,6 +921,45 @@ export abstract class Move implements Localizable {
...
+      }
+    }
+
+    if (user.hasAbility(Abilities.PARENTAL_BOND)) {
```
Tempo-anon: Shouldn't this be in an `else` since parental bond does not affect multi hit moves?

src/field/pokemon.ts
```diff
@@ -2247,10 +2253,16 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
    * @see {@linkcode getAttackScore}
    */
   public getExpectedAttackScore(opponent: Pokemon, move: Move): number {
-    const { damage } = opponent.getAttackDamage(this, move, AbilityApplyMode.REVEALED);
+    const damage =
+      opponent.getAttackDamage(this, move, AbilityApplyMode.REVEALED).damage
+      * move.getMultiHitAttackScoreMultiplier(this);
```
Tempo-anon: seeing some repeated code with this and the `if (damage >= opponent.hp)` bit down below. Maybe all that can be moved to a helper function?

src/field/pokemon.ts
```diff
@@ -5642,7 +5654,65 @@ export class EnemyPokemon extends Pokemon {
...
+   */
+  protected getCriticalHitBonus(opponent: Pokemon, move: Move, attackScore?: number) {
+    const { damage: critDamage } = opponent.getAttackDamage(this, move, AbilityApplyMode.REVEALED, true);
+    if ((isNullOrUndefined(attackScore) || attackScore < 4) && critDamage >= opponent.hp) {
+      const critChance = this.getSimulatedCriticalHitChance(opponent, move);
+      /**
+       * Only grant a bonus if the calculated critical hit chance is over 10%
+       * (i.e. the user has 1 or more crit stages)
+       */
+      if (critChance > 10) {
```
Tempo-anon, line 5694: 10 is a weird cutoff since the only chances are 1/24 (4.17%) at +0, 1/8 (12.5%) at +1, 1/2 (50%) at +2, and 100% after

src/data/move.ts
innerthunder: There's `beatUpFunc` in `beat-up-attr.ts`... but that doesn't seem right lol. I tried to keep this simple, at least for now

src/data/move.ts
innerthunder: Right now the rest of the scoring code is only set up assuming a single accuracy check. This is fine enough at 3 hits for triple axel / triple kick, but population bomb is an extreme case -- you can't really assume it'll get the maximum number of hits most of the time. This is a simplified way to account for the accuracy checks

Tempo-anon: with skill link it should technically be 9 instead of 10 then 🤔

src/data/move.ts
Tempo-anon: `beatUpFunc` is such a badly named function too. I guess another helper function like `getExpectedTotalBasePower` can be added to the attr and used for calcs

src/data/move.ts
innerthunder: (-1) is a pretty severe penalty; in some cases it's enough to prefer a move that deals half as much damage or less. The moves you listed are run commonly as their respective mon's only STAB option, so I didn't want to harshly penalize them.

Tempo-anon: Yeah with the way it's set up -1 is too large. I would say maybe a fraction like -0.5 or -0.25 but I don't think that makes any functional difference

src/data/move.ts
```diff
@@ -859,6 +921,38 @@ export abstract class Move implements Localizable {
...
+      const multiHitType = this.getAttrs(MultiHitAttr)[0].getMultiHitType();
+      switch (multiHitType) {
+        case MultiHitType._2:
+          return 2;
+        case MultiHitType._2_TO_5:
+          return userHasSkillLink ? 5 : 3; // Expected hit count is ~3.1
+        case MultiHitType._3:
+          return 3;
+        case MultiHitType._10:
+          return userHasSkillLink ? 10 : 6; // Population Bomb hits ~5.8 times on average
```
Tempo-anon, line 950: nit: if this accounts for accuracy it should be `9 : 6` rather than `10 : 6` but very minor

src/data/move.ts
```diff
@@ -859,6 +921,38 @@ export abstract class Move implements Localizable {
...
+        case MultiHitType.BEAT_UP:
+          return user.getParty().length;
+      }
+    } else if (this.canBeMultiStrikeEnhanced(user) && user.hasAbility(Abilities.PARENTAL_BOND)) {
```
Tempo-anon, line 954: I think there's also the edge case of parental bond not activating for spread moves

Tempo-anon (approved):

src/data/move.ts
innerthunder: When it comes to move score I'd rather keep it an integer value overall. I suppose I could make it an X% chance of (-1)?

src/data/move.ts
innerthunder: `canBeMultiStrikeEnhanced` should cover this (I got rid of the spread moves exception for Multi-Lens in #123 )

src/field/pokemon.ts
innerthunder: I just left it as a nice round number that encompasses any positive crit stage

innerthunder:
> How'd you decide on these tiers for the accuracy?

Mostly considering the power differential between moves at different tiers. There's a clear difference in power between Zap Cannon and Focus Blast, for example, assuming both hit, and the same can be said for Zap Cannon vs. the OHKO moves.

> Also, why not multiply the BP by the acc as an "expected BP" value?

Expected values make more sense with low variance in outcomes, and hitting a move vs. missing is the opposite of that. It's also important to track whether a move can or cannot pass certain damage thresholds (mainly KOs)

src/data/move.ts
```diff
@@ -859,6 +921,38 @@ export abstract class Move implements Localizable {
...
+   * Obtains an approximate damage multiplier accounting for multi-hit
+   * move and ability modifiers relative to the move's first strike.
+   * @param user the {@linkcode Pokemon} using this move
+   * @returns a damage modifier to be used for move scoring
+   * @see {@linkcode Pokemon.getAttackScore}
+   */
+  public getMultiHitAttackScoreMultiplier(user: Pokemon): number {
+    const userHasSkillLink = user.hasAbilityWithAttr(MaxMultiHitAbAttr);
+    if (this.hasAttr(MultiHitPowerIncrementAttr)) {
+      return 6; // assumes 3 hits: 1x + 2x + 3x
```
DayKev: Seems like this is specifically [only] for Triple Axel? If so, should probably make a note of that in a comment.
Also Triple Axel is effectively a ~73% accuracy move (for all 3 hits) so that should probably be taken into account somewhere? (It's 90%/81%/72.9%)

src/field/pokemon.ts
```diff
@@ -5642,7 +5670,65 @@ export class EnemyPokemon extends Pokemon {
...
+
+    return attackScore + critBonus + move.getEffectScore(this, opponent);
+  }
+
+  /**
+   * Calculates the bonus granted to the given move based on its critical hit chance
+   * when used by this Pokemon against the given opponent.
+   * @param opponent the {@linkcode Pokemon} targeted by the move
+   * @param move the {@linkcode Move} being evaluated
+   * @param attackScore the pre
```
DayKev: Unfinished comment.

src/field/pokemon.ts
```diff
@@ -5642,7 +5670,65 @@ export class EnemyPokemon extends Pokemon {
...
+   */
+  protected getSimulatedCriticalHitChance(target: Pokemon, move: Move): number {
+    const defendingSide = this.getArenaTagSide();
+    const noCritTag = globalScene.arena.getTagOnSide(NoCritTag, defendingSide);
+
+    if (
+      noCritTag
+      || move.hasAttr(FixedDamageAttr)
+      || target.hasAbilityWithAttr(BlockCritAbAttr)
+      || Overrides.NEVER_CRIT_OVERRIDE
```
DayKev: Do we want this override to change the AI behavior?

test/ai/enemy_command.test.ts
```diff
@@ -1,35 +1,16 @@
...
 import { AiType } from "#enums/ai-type";
-import { randSeedInt } from "#app/utils";
 import { GameManager } from "#test/testUtils/gameManager";
 import Phaser from "phaser";
-import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeAll, beforeEach, describe, expect, it } from "vitest";
+import { getEnemyMoveChoices, type MoveChoiceSet } from "./utils/enemy_command_utils";
+import type BattleScene from "#app/battle-scene";
 
 let globalScene: BattleScene;
```
DayKev, line 13: I don't think this is necessary? I would think you can just pass `game.scene` directly, no need to do this (or if you really want to, import the "actual" `globalScene`).

test/ai/utils/enemy_command_utils.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const NUM_TRIALS = 300;
+
+export type MoveChoiceSet = { [key: number]: number };
+
+export function getEnemyMoveChoices(scene: BattleScene, pokemon: EnemyPokemon, moveChoices: MoveChoiceSet): void {
+  // Use an unseeded random number generator in place of the mocked-out randBattleSeedInt
+  vi.spyOn(scene, "randBattleSeedInt").mockImplementation((range, min?) => {
+    return randSeedInt(range, min);
+  });
```
DayKev, line 15: Shouldn't you run though the possible RNG values that can be generated and see what the output is for each?

src/field/pokemon.ts
```diff
@@ -5642,7 +5670,65 @@ export class EnemyPokemon extends Pokemon {
...
+  }
+
+  /**
+   * Calculates the chance (%, rounded down) of the given move critically hitting
+   * the given target when used by this Pokemon.
+   * @param target the {@linkcode Pokemon} targeted by the move
+   * @param move the {@linkcode Move} being evaluated
+   * @returns the chance of the move hitting
+   */
+  protected getSimulatedCriticalHitChance(target: Pokemon, move: Move): number {
```
PigeonBar, line 5709: This function shares a lot of code in common with `getCriticalHitResult()`. It may be possible to rewrite `getCriticalHitResult()` to call this hit chance function.

src/data/move.ts
Tempo-anon: Hmm maybe. Like a X% to penalize. 🤔 Not any deal breaker or anything since these can all be tweaked in the future. Maybe just leave a comment here about this idea for the future in case things do want to get tweaked

test/ai/enemy_command.test.ts
innerthunder: I tried removing it before, but this cuts down the test runtime by a lot

test/ai/utils/enemy_command_utils.ts
innerthunder: I don't expect there to be a consistent range of values for this when it can be called multiple times in a trial. I figure a sample size of 300 trials is good enough to get the big picture, anyway.

src/data/move.ts
innerthunder: Since Skill Link only enforces the first accuracy check, any penalty from the full damage would fall under a low accuracy penalty if that ever changes for 80-99 accuracy. I think just using the max value here is fine.

src/data/move.ts
PigeonBar: I think that (+/- 1) being such a large difference could be a problem itself, since it makes it harder to evaluate moves with attrs/properties that should only incentivize/penalize the move by a small amount. To solve this, we would have to either scale up move scores (e.g., allow normal attack scores to range from 0 to +10, with +10 for KO's), or rethink why we are forcing every move score to be an integer.

innerthunder: I'd rather keep the scale of move scores low because the integer conversion is a big part of how variance in move selection is enforced. The higher the granularity, the less variance there is in the enemy's decision-making.

src/field/pokemon.ts
innerthunder: I considered this, but there are a few minor differences that would make this difficult, e.g. the way Shell/Battle Armor are checked

test/ai/enemy_command.test.ts
DayKev: Weird. Oh well.

test/ai/utils/enemy_command_utils.ts
DayKev: Hm, alright.

DayKev (approved):

PR #450, opened 2025-02-08
[Ability][Checkup][Bug] Defeatist checkup + getHpRatio() precision improved

### LINKED ISSUE #449: [Ability][Checkup] Defeatist
## Ability Requirements
- [x] Documentation
- [x] Tests
- [x] Defeatist halves the user's Attack and Special Attack stats when their HP is equal to or less than half of its maximum. 

## Related PRs
- https://github.com/Despair-Games/poketernity/pull/450

## Related Bugs
- None as of 02-12-2025

## References
- [Bulbapedia - Defeatist](https://bulbapedia.bulbagarden.net/wiki/Defeatist_(Ability))

frutescens:

Getting damage for Flail will consider a HP ratio that has the same number of decimal places as listed in their conditions.
In random rewards, Potions and Hyper Potions will match the same number of decimal places as listed in their conditions too. Previously, the game cut all getHpRatio() requests to the hundredth place by default, causing issues for requests that were trying to match more precise values. 

Ability Checkup. 

`pokemon.getHpRatio()` has been replaced with `getHpBarScaleX()` in `battle-info.ts`, a helper function that suits the UI's need for a very precise decimal. `pokemon.getHpRatio()` now takes in an optional parameter `precision` that can increase the precision of the HP ratio calculated. It defaults to the hundredth place (precision value of 2) since most checks are limited to that, but now it is much more flexible and can be applied to other more precise checks. 

Test for Defeatist created. 

## How to test the changes?

`npm run test defeatist`

## Checklist


src/data/move-attrs/low-hp-power-attr.ts
```diff
@@ -11,7 +11,7 @@ import { VariablePowerAttr } from "#app/data/move-attrs/variable-power-attr";
  */
 export class LowHpPowerAttr extends VariablePowerAttr {
   override apply(user: Pokemon, _target: Pokemon, _move: Move, power: NumberHolder): boolean {
-    const hpRatio = user.getHpRatio();
+    const hpRatio = user.getHpRatio(4);
 
     switch (true) {
       case hpRatio < 0.0417:
```
Tempo-anon: Can you double check these values are right? From Smogon:
The power of this move is 20 if X is 33 to 48, 40 if X is 17 to 32, 80 if X is 10 to 16, 100 if X is 5 to 9, 150 if X is 2 to 4, and 200 if X is 0 or 1, where X is equal to (user's current HP * 48 / user's maximum HP), rounded down.

src/data/move-attrs/low-hp-power-attr.ts
```diff
@@ -11,7 +11,7 @@ import { VariablePowerAttr } from "#app/data/move-attrs/variable-power-attr";
  */
 export class LowHpPowerAttr extends VariablePowerAttr {
   override apply(user: Pokemon, _target: Pokemon, _move: Move, power: NumberHolder): boolean {
-    const hpRatio = user.getHpRatio();
+    const hpRatio = user.getHpRatio(4);
```
Tempo-anon: Is there a reason this is 4?

src/data/all-abilities.ts
```diff
@@ -608,10 +608,11 @@ export function initAbilities() {
       .attr(MoveAbilityBypassAbAttr),
     new Ability(Abilities.SUPER_LUCK, 4).attr(BonusCritAbAttr),
     new Ability(Abilities.AFTERMATH, 4).attr(PostFaintContactDamageAbAttr, 4).bypassFaint(),
-    new Ability(Abilities.ANTICIPATION, 4).attr(AnticipationAbAttr, (pokemon: Pokemon) =>
-      i18next.t("abilityTriggers:postSummonAnticipation", { pokemonNameWithAffix: getPokemonNameWithAffix(pokemon) }),
-    )
-    .edgeCase(), // Does not activate upon acquiring the Ability (e.g., via Skill Swap)
+    new Ability(Abilities.ANTICIPATION, 4)
```
Tempo-anon: not sure why this change is getting caught in this PR. Probably some linting

src/data/move-attrs/low-hp-power-attr.ts
frutescens: 0.0417... four decimal places

src/field/pokemon.ts
```diff
@@ -1337,8 +1337,13 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
   }
 
-  getHpRatio(precise: boolean = false): number {
-    return precise ? this.hp / this.getMaxHp() : Math.round((this.hp / this.getMaxHp()) * 100) / 100;
+  /**
+   * Helper function that returns a Pokemon's HP ratio rounded to a specific decimal value
+   * @param fractionDigits the number of decimal places desired, defaults to the hundredths place
+   * @returns the HP ratio rounded to the provided decimal place
+   */
+  getHpRatio(fractionDigits: number = 2): number {
```
innerthunder: I feel like this should have an option not to round (say, if `fractionDigits` is set to -1). That way `getHpBarScaleX` can be removed

PigeonBar: I think that this function should just never be rounding. I don't think there's a single place in the codebase where, for example, it is correct to treat a Pokemon at 50.4% HP as being at half health.

src/data/move-attrs/low-hp-power-attr.ts
Tempo-anon: I assume those are all fractional forms of X/48 but is there proof that the game rounds to 4 decimal places? 2/48 for example is 0.04166666666 which is not quite 0.0417.

src/field/pokemon.ts
frutescens: The unrounded version of the HP ratio is used to properly scale the HP bar in the UI. I don't mind either having getHpBarScaleX be its own function since UI-related things should be separate from battle stuff for clarity or having them merged for ease. Up to you two :D

src/data/move-attrs/low-hp-power-attr.ts
Tempo-anon: This is supposed to be divided and rounded down to an integer from 0-48 so getHpRatio isn't even needed to be called in this attr at all. #453

frutescens: Ok, if that PR gets merged before this one, I'll update accordingly. Or if this one gets merged before #453, then #453 can be updated.

src/modifier/modifier-type.ts
```diff
@@ -2190,7 +2190,7 @@ const modifierPool: ModifierPool = {
       modifierTypes.POTION,
       (party: Pokemon[]) => {
         const thresholdPartyMemberCount = Math.min(
-          party.filter((p) => p.getInverseHp() >= 10 && p.getHpRatio() <= 0.875 && !p.isFainted()).length,
+          party.filter((p) => p.getInverseHp() >= 10 && p.getHpRatio(3) <= 0.875 && !p.isFainted()).length,
```
Tempo-anon: Would this change anything if it was `7/8` instead of `0.875`?

torranx (dismissed):

src/modifier/modifier-type.ts
frutescens: As long as we keep the `3` in getHpRatio() it shouldn't.

test/abilities/defeatist.test.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  it.each([
+    { statName: "attack", stat: Stat.ATK },
+    { statName: "special attack", stat: Stat.SPATK },
+  ])("should halve the user's $statName if the user's HP is at or below 50%", async ({ stat }) => {
+    await game.classicMode.startBattle([Species.FEEBAS]);
+    const playerPokemon = game.scene.getPlayerPokemon()!;
+    const expectedStat = Math.floor(playerPokemon.getStat(stat as number) / 2);
+    playerPokemon.hp = 1;
+    const defeatistStat = playerPokemon?.getEffectiveStat(stat as number);
```
PigeonBar:
```suggestion
    const defeatistStat = playerPokemon.getEffectiveStat(stat as number);
```

test/abilities/defeatist.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  });
+
+  it.each([
+    { statName: "attack", stat: Stat.ATK },
+    { statName: "special attack", stat: Stat.SPATK },
+  ])("should have no effect on $statName if the user's HP is above 50%", async ({ stat }) => {
+    await game.classicMode.startBattle([Species.FEEBAS]);
+    const playerPokemon = game.scene.getPlayerPokemon()!;
+    const expectedStat = playerPokemon.getStat(stat as number);
+    const defeatistStat = playerPokemon?.getEffectiveStat(stat as number);
```
PigeonBar:
```suggestion
    const defeatistStat = playerPokemon.getEffectiveStat(stat as number);
```

test/abilities/defeatist.test.ts
```diff
@@ -0,0 +1,61 @@
...
+      .enemyAbility(Abilities.BALL_FETCH)
+      .enemyMoveset(MoveId.SPLASH);
+  });
+
+  it.each([
+    { statName: "attack", stat: Stat.ATK },
+    { statName: "special attack", stat: Stat.SPATK },
+  ])("should halve the user's $statName if the user's HP is at or below 50%", async ({ stat }) => {
+    await game.classicMode.startBattle([Species.FEEBAS]);
+    const playerPokemon = game.scene.getPlayerPokemon()!;
```
PigeonBar:
```suggestion
    const playerPokemon = game.field.getPlayerPokemon();
```

test/abilities/defeatist.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    expect(playerPokemon.getHpRatio()).toBeLessThanOrEqual(0.5);
+    expect(defeatistStat).toBe(expectedStat);
+  });
+
+  it.each([
+    { statName: "attack", stat: Stat.ATK },
+    { statName: "special attack", stat: Stat.SPATK },
+  ])("should have no effect on $statName if the user's HP is above 50%", async ({ stat }) => {
+    await game.classicMode.startBattle([Species.FEEBAS]);
+    const playerPokemon = game.scene.getPlayerPokemon()!;
```
PigeonBar:
```suggestion
    const playerPokemon = game.field.getPlayerPokemon();
```

PigeonBar (approved):

torranx (approved):

PR #463, opened 2025-02-10
[AI-Rework] Add Move Attribute scoring parameters under `MoveAttrOptions`

innerthunder:
> [!NOTE]
> This PR targets the `ai-rework` branch and should not be considered a complete feature. 

N/A

Part of the AI Rework

This PR implements additional parameters to move attribute scoring that help determine when attributes' scores contribute to overall move score. The parameters are as follows:
- `appliesScoreOnKO`: If `true`, allows the attribute to contribute to move score even when the move is known to KO its opponent.
- `appliesScoreOnFail`: If `true`, allows the attribute to contribute to move score even when the move would fail or have no effect on the target.
- `overridesAllyTargetPenalty`: Normally, targeting an ally with a move results in a -20 penalty. If this is `true` for at least one attribute, the attribute(s) override this penalty with their own effect scores.

Some new helper functions have also been made for effect scoring, including the logic for `ChanceBasedMoveEffectAttr`'s scoring calculations.

You can refer to the [AI Rework's Effect Scores Sheet](https://docs.google.com/spreadsheets/d/12UprXLHyLdw-ESls6pAQecyBFCG2tVg5cllAiUbp6KA/edit?gid=0#gid=0) to see how these additional parameters may be applied to different attributes.

## Screenshots/Videos

N/A

## How to test the changes?

Expecting code review only for this one.

## Checklist

- [?] Have I tested the changes manually?


src/data/move-attrs/chance-based-move-effect-attr.ts
```diff
@@ -53,23 +53,49 @@ export abstract class ChanceBasedMoveEffectAttr extends MoveEffectAttr {
...
+   * - The chance of the final score being `B + 1` is `D * 100 %`
+   */
+  public override getEffectScore(user: EnemyPokemon, target: Pokemon, move: Move): number {
+    /** @todo this chance calculation may prematurely reveal abilities */
+    const chance = this.getMoveChance(user, target, move);
+    const rawScore = this.getRawEffectScore(user, target, move);
+
+    const chanceWeightedScore = chance * rawScore;
+    const minScore = Math.floor(chanceWeightedScore);
+    const tierUpChance = Math.floor((chanceWeightedScore % 1) * 100);
```
Tempo-anon, line 95: Why is there a mod 1 needed?

src/data/move-attrs/chance-based-move-effect-attr.ts
```diff
@@ -53,23 +53,49 @@ export abstract class ChanceBasedMoveEffectAttr extends MoveEffectAttr {
...
+    if (!this.selfTarget) {
       applyAbAttrs(IgnoreMoveEffectsAbAttr, target, false, user, move, moveChance);
     }
     return moveChance.value;
   }
+
+  /**
+   * Calculates score based on raw score value from {@linkcode getRawEffectScore} and effect
+   * chance from {@linkcode getMoveChance}. If the product of the raw score and effect chance
+   * is equal to `B + D`, where `B` is the nearest integer less than the product and
```
Tempo-anon: the comment of B and D should match the naming scheme of the variables `minScore` and `tierUpChance`

src/data/move-attrs/move-attr.ts
```diff
@@ -11,9 +20,41 @@ import type { MoveCondition, MoveConditionFunc } from "../move-conditions";
...
+  }
+
+  /**
+   * Defines whether or not this attribute contributes to effect score even when the move
+   * KOs its target.
+   * @default false
+   * @see {@linkcode getEffectScore}
+   */
+  public get appliesScoreOnKO(): boolean {
+    return this.options?.appliesScoreOnKO ?? false;
```
Tempo-anon, line 39: Are the `?? false` needed? Could set all the variables to be false by default

src/data/move-attrs/move-attr.ts
```diff
@@ -68,7 +109,33 @@ export abstract class MoveAttr {
...
+  /**
+   * Uses a seeded chance roll to return either a maximum score or a minimum score
+   * @param user the {@linkcode EnemyPokemon} evaluating effect scores under which the chance roll is seeded
+   * @param chance the chance (%) to yield the maximum score. Assumed to be an integer in the range [0, 100].
+   * @param maxScore the score given if the chance roll is successful
+   * @param minScore the score given if the chance roll is unsuccessful. Defaults to `0`
+   * @returns either `maxScore` or `minScore`, depending on the chance roll's outcome.
+   */
+  protected getRandomScore(user: EnemyPokemon, chance: number, maxScore: number, minScore: number = 0): number {
+    return user.randSeedInt(100) < chance ? maxScore : minScore;
```
Tempo-anon: I assume this will be overriden in other move-attrs and may have more complicated calculations than a boolean ?

src/data/move-attrs/chance-based-move-effect-attr.ts
innerthunder: The mod 1 separates the decimal component of `chanceWeightedScore`

src/data/move-attrs/move-attr.ts
innerthunder: Not supported for interfaces:

image

Just following the conventions for other `options` interfaces in the attributes. Putting default values in the constructor also seems awkward imo

src/data/move-attrs/move-attr.ts
innerthunder: This is just a helper function, e.g. an attribute can call
```ts
this.getRandomScore(user, 50, 1, 0)
```
in their `getEffectScore` override to get a (+1) bonus 50% of the time

src/data/move-attrs/chance-based-move-effect-attr.ts
innerthunder: I feel like `minScore + (tierUpChance / 100)` might be more confusing? Or it at least makes it a bit less clear that the integer and decimal components are being split.

src/data/move-attrs/move-attr.ts
Tempo-anon: I'm just thinking that `minScore` and `maxScore` implies that the result will be some value in between those two inclusive. From the way you're describing it this is just a chance to get score1 or score2

src/data/move-attrs/move-attr.ts
```diff
@@ -68,7 +109,33 @@ export abstract class MoveAttr {
...
+  /**
+   * Defines the integer Effect Score bonus (or penalty) granted by this attribute based on the
+   * current game state. Effect Scores from a move's attributes are combined to form a move's
+   * overall Effect Score, which helps determine the Enemy AI's move selection.
+   *
+   * For Attack Moves, the total Effect Score should rarely exceed (+1) and should almost never
+   * exceed (+2). Status Moves' total Effect Score should rarely exceed (+3). You can also use
+   * {@linkcode getRandomScore} to add variance to a score, which can also be helpful if a non-integer
+   * score value seems appropriate.
+   * @param user the {@linkcode EnemyPokemon} evaluating the move
```
MokaStitcher, line 123: Mostly out of curiosity, is there a reason to have the user be an EnemyPokemon there and in other scoring related functions?
I know technically right now it's always the case. But we could imagine in the future having double battles with an npc on our side or something that would require them to use scoring to figure out their move as well

src/data/move-attrs/chance-based-move-effect-attr.ts
Tempo-anon: Ah okay. For some reason my brain thought you were flooring that before the * 100 and was wondering why that was needed when it would always be 0

src/data/move-attrs/chance-based-move-effect-attr.ts
Tempo-anon: I think `minScore + (tierUpChance / 100)` is actually clearer

src/data/move-attrs/move-attr.ts
DayKev:
```suggestion
    return !!this.options?.appliesScoreOnKO;
```
You could do this instead (same for the others).

src/field/pokemon.ts
```diff
@@ -5652,14 +5652,12 @@ export class EnemyPokemon extends Pokemon {
...
     const attackScore = this.getAttackScore(opponent, move);
+    const isKnockOut = attackScore >= 4;
+    const isFail = attackScore === -1 || !meetsConditions;
 
-    if (!meetsConditions || attackScore <= -1) {
-      return -5;
-    }
-
-    const critBonus = this.getCriticalHitBonus(opponent, move, attackScore);
+    const critBonus = isFail ? 0 : this.getCriticalHitBonus(opponent, move, attackScore);
```
```suggestion
    const critBonus = this.getCriticalHitBonus(opponent, move, attackScore);
```

src/field/pokemon.ts
```diff
@@ -5652,14 +5652,12 @@ export class EnemyPokemon extends Pokemon {
...
 
-    return attackScore + critBonus + move.getEffectScore(this, opponent);
+    return (isFail ? -5 : attackScore) + critBonus + move.getEffectScore(this, opponent, isKnockOut, isFail);
```
```suggestion
    return (isFail ? -5 : attackScore + critBonus) + move.getEffectScore(this, opponent, isKnockOut, isFail);
```

src/data/move-attrs/move-attr.ts
innerthunder: `getMoveScore` (and some other scoring functions) are `EnemyPokemon`-exclusive, so this is just to match. If there are plans for an NPC ally, they can be moved around, but for the time being I think this makes it a bit clearer that these functions are intended for the Enemy AI.

src/data/move-attrs/move-attr.ts
innerthunder: I like it as is a bit more because the default values are separated and can easily be tweaked.

src/data/move-attrs/move-attr.ts
Tempo-anon: NPC ally would be super lit

src/data/move-attrs/move-attr.ts
innerthunder: Yes, it's just one score or the other. I guess `min/maxScore` is a bit misleading? I might consider
- `scoreA` / `scoreB`
- `hitScore` / `missScore`
- `successScore` / `failScore`

but I'm open to other suggestions

Tempo-anon: I was thinking `baseScore` and something like `ImprovedBaseScore` or `baseScoreWithEffectBonus`

src/data/move-attrs/move-attr.ts
```diff
@@ -68,7 +109,33 @@ export abstract class MoveAttr {
...
+  /**
+   * Uses a seeded chance roll to return one of two score values.
+   * @param user the {@linkcode EnemyPokemon} evaluating effect scores under which the chance roll is seeded
+   * @param chance the chance (%) to yield the maximum score. Assumed to be an integer in the range [0, 100].
+   * @param hitScore the score given if the chance roll is successful
+   * @param missScore the score given if the chance roll is unsuccessful. Defaults to `0`
+   * @returns either `maxScore` or `minScore`, depending on the chance roll's outcome.
+   */
+  protected getRandomScore(user: EnemyPokemon, chance: number, hitScore: number, missScore: number = 0): number {
+    return user.randSeedInt(100) < chance ? hitScore : missScore;
```
DayKev: Would this ever want to be a difference greater than 1 between the possible return values? That would represent a huge amount of variance given the small scale of AI move scoring.

innerthunder: Probably very rarely, if ever, but I'd like to keep the option open for more volatile AI types

DayKev: Maybe `hitScore`'s default should be 1 then, and in the rare case larger values are wanted then they can be specified.

src/data/move.ts
```diff
@@ -726,32 +726,88 @@ export abstract class Move implements Localizable {
...
+       * from other non-overriding attributes are ignored.
+       */
+      const allyTargetAttrs = this.attrs.filter((attr) => attr.overridesAllyTargetPenalty);
+
+      if (allyTargetAttrs.length === 0) {
+        return -20;
+      } else {
+        return allyTargetAttrs
+          .map((attr) => attr.getEffectScore(user, target, this))
+          .reduce((total, score) => total + score);
```
PigeonBar:
```suggestion
          .reduce((total, score) => total + score, 0);
```

src/data/move-attrs/chance-based-move-effect-attr.ts
```diff
@@ -53,23 +53,60 @@ export abstract class ChanceBasedMoveEffectAttr extends MoveEffectAttr {
...
+    /** The minimum integer score this function can return */
+    const minScore = Math.floor(chanceWeightedScore);
+    /**
+     * The chance to return `minScore + 1` instead of `minScore`.
+     * This is the decimal component of `chanceWeightedScore` scaled
+     * up to a percent value, then rounded down.
+     */
+    const tierUpChance = Math.floor((chanceWeightedScore % 1) * 100);
+
+    return this.getRandomScore(user, tierUpChance, minScore + 1, minScore);
```
DayKev:
```suggestion
    return minscore + this.getRandomScore(user, tierUpChance);
```
Since `getRandomScore` was changed.

src/data/move-attrs/chance-based-move-effect-attr.ts
```diff
@@ -53,23 +53,60 @@ export abstract class ChanceBasedMoveEffectAttr extends MoveEffectAttr {
...
    * effect is guaranteed to apply.
    */
-  public getMoveChance(
-    user: Pokemon,
-    target: Pokemon,
-    move: Move,
-    selfEffect: boolean,
-    showAbility: boolean = false,
-  ): number {
+  public getMoveChance(user: Pokemon, target: Pokemon, move: Move, showAbility: boolean = false): number {
```
MokaStitcher, line 55: can't comment on that line, but the tsdoc needs to be updated (selfEffet removed and showAbility added)

also, should this be renamed to `getMove(Additionnal)EffectChance`? At first I thought it was related to the move's accuracy (may be just me)

src/data/move.ts
```diff
@@ -726,32 +726,88 @@ export abstract class Move implements Localizable {
    * should typically be limited to (+3) or lower.
    * @param user the {@linkcode Pokemon} using the move
    * @param target the {@linkcode Pokemon} targeted by the move
+   * @param isKnockOut `true` if the move is already known to KO the target (default `false`)
+   * @param isFail `true` if the move is already known to fail or have no effect (default `false`)
    * @returns a score value accumulated from effect score modifiers.
    * @todo Add Low Accuracy Penalty and Ally Target Penalty
```
MokaStitcher:
```suggestion
```
I think both of these have been addressed?

src/data/move.ts
```diff
@@ -726,32 +726,88 @@ export abstract class Move implements Localizable {
...
+       *
+       * NOTE: if at least one attribute overrides the Ally Target Penalty, ONLY
+       * the overriding attributes are accounted for in scoring. Score contributions
+       * from other non-overriding attributes are ignored.
+       */
+      const allyTargetAttrs = this.attrs.filter((attr) => attr.overridesAllyTargetPenalty);
+
+      if (allyTargetAttrs.length === 0) {
+        return -20;
+      } else {
```
MokaStitcher, line 782: since there's a `return` in the if statement we can reduce nesting by removing the else

src/data/move.ts
```diff
@@ -726,32 +726,88 @@ export abstract class Move implements Localizable {
...
+        return allyTargetAttrs
+          .map((attr) => attr.getEffectScore(user, target, this))
+          .reduce((total, score) => total + score);
+      }
+    } else {
```
MokaStitcher, line 787: same with this else

src/field/pokemon.ts
```diff
@@ -5652,14 +5652,12 @@ export class EnemyPokemon extends Pokemon {
...
-    if (!meetsConditions || attackScore <= -1) {
-      return -5;
-    }
+    const isKnockOut = attackScore >= 4;
+    const isFail = attackScore === -1 || !meetsConditions;
 
     const critBonus = this.getCriticalHitBonus(opponent, move, attackScore);
 
-    return attackScore + critBonus + move.getEffectScore(this, opponent);
+    return (isFail ? -5 : attackScore + critBonus) + move.getEffectScore(this, opponent, isKnockOut, isFail);
```
MokaStitcher, line 4763: not necessarily for this PR, but would it make sense to put this -5 as well as the various -20 penalty values in constants?

MokaStitcher (commented):
just some nits, otherwise looks good

src/data/move.ts
innerthunder: Is the nesting an issue? I feel like leaving the elses keeps it a bit more organized.

src/data/move-attrs/chance-based-move-effect-attr.ts
innerthunder: I'll leave this to another PR (targeting `beta`) if there's enough demand for it

src/field/pokemon.ts
innerthunder: I could do so; I'm just not sure what to name them...

src/data/move.ts
MokaStitcher: Removing them makes the code easier to read imo, but I'll leave the call to you

src/field/pokemon.ts
Tempo-anon: Can always use some filler names like PENALTY, BIG_PENALTY, NEGATIVE_FIVE_PENALTY, etc for now until you think of something better. Just add some comments explaining what the const is for

MokaStitcher (approved):

src/field/pokemon.ts
MokaStitcher: Probably something descriptive if possible, so in this case FAILING_MOVE_PENALTY or something like that

PR #466, opened 2025-02-10
[Balance] G-Max Pokemon take 1/3 less damage

Tempo-anon:

G-Max Pokemon get 33% (subject to change) damage reduction to damage taken

Since G-Max Pokemon no longer have increased stats, this mimics the similar effect of increased HP.

* Created a new constant
* Multiplied damage by the damage taken factor in `pokemon.ts` if the pokemon is Max 

## Screenshots/Videos

n/a

## How to test the changes?

```ts
const overrides = {
  STARTING_MODIFIER_OVERRIDE: [{name:"DYNAMAX_BAND"}],
  ITEM_REWARD_OVERRIDE: [{name:"RARE_FORM_CHANGE_ITEM"}]
}
```

## Checklist


src/field/pokemon.ts
```diff
@@ -3470,6 +3470,11 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
       }
     }
 
+    // Eternatus does not need the damage reduction as its emax form has increased hp/defenses
+    if (this.isMax() && this.species.speciesId !== Species.ETERNATUS) {
+      damage = Math.floor(damage * DYNAMAX_DAMAGE_TAKEN_FACTOR);
```
MokaStitcher: Others will know better but it seems to me that this is getting applied a bit too late down the chain
It should probably be done towards the end of `getAttackDamage` instead? Or as one of the many multipliers being applied here <https://github.com/Despair-Games/poketernity/blob/b8275372e77ba9d0bb3fe40de2f82acd12824c52/src/field/pokemon.ts#L3356-L3359>
This way it should be taken into account properly for sturdy/endure as well as AI calculations, and probably other things?

<sub> if not moved it should probably use `toDmgValue(damage * DYNAMAX_DAMAGE_TAKEN_FACTOR)` to ensure it's always at least 1, and be moved before the endure/sturdy thing

innerthunder: If this is only supposed to reduce attack damage and not indirect damage, then @MokaStitcher is correct. Otherwise, this should be moved to before enduring effects are taken into account (L3455). You might also have to check Sturdy for visual bugs if going for the latter option.

MokaStitcher: Ah right since it's meant to replace an HP increase it should probably apply to all damage

Tempo-anon: The intended implementation is supposed to reduce all damage taken to mimic the increased hp but I will also have to do more research for things like burn/poison damage as well as endeavor/pain split

flx-sta (commented):
@Tempo-anon I feel like the PR is missing some coverage through tests for this change

src/field/pokemon.ts
```diff
@@ -3432,6 +3432,11 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     }
     const surviveDamage = new BooleanHolder(false);
 
+    // Eternatus does not need the damage reduction as its emax form has increased hp/defenses
+    if (this.isMax() && this.species.speciesId !== Species.ETERNATUS) {
```
PigeonBar, line 3436: nit: Eternatus will likely also be an exception for a few other `isMax()` checks, should we add a param `includesEmax` to `isMax()`?

Tempo-anon: I believe E-Max Eternatus still retains the other max immunities like an immunity to flinch, phazing, weight moves, etc. Are there any in particular?

PigeonBar: Just in case we also want to make Eternatus an exception to some other G-max checks in the future

Tempo-anon: For now I don't think it makes too much sense to include it as a param if this spot is the only difference. If future differences crop up then it would make more sense to add it

test/battle/gmax_damage_calculation.test.ts
```diff
@@ -0,0 +1,92 @@
...
+      .enemyForms({ [Species.SNORLAX]: 1 })
+      .enemyAbility(Abilities.BALL_FETCH)
+      .enemyMoveset(MoveId.SPLASH)
+      .startingLevel(100)
+      .enemyLevel(100)
+      .disableCrits()
+      .moveset([MoveId.TACKLE, MoveId.DRAGON_RAGE, MoveId.HAIL]);
+  });
+
+  it("Tackle deals expected damage after gmax damage reduction", async () => {
```
innerthunder, line 36: nit: I'd prefer if these tests actually used pokemon in G-max form over mocking `isMax` to reduce coupling

innerthunder (dismissed):

test/battle/gmax_damage_calculation.test.ts
Tempo-anon: The tests do use actual g-max opponents. I just forgot to remove the mock after I updated the opponents

torranx (approved):

PR #663, opened 2025-02-13
[Refactor][Misc] Cleaned up critical capture code.

frutescens:

None.

With this codebase, there's moments where you open a random file and see a stupid, chunky ternary. 

Ternary broken up into if-else statements for better readability. 

## How to test the changes?

Critical Captures should work as expected. 

## Checklist


src/data/pokeball.ts
```diff
@@ -78,19 +79,21 @@ export function getCriticalCaptureChance(modifiedCatchRate: number): number {
...
+  } else if (dexCount > 400) {
+    dexMultiplier = 1.5;
+  } else if (dexCount > 200) {
+    dexMultiplier = 1;
+  } else if (dexCount > 40) {
+    dexMultiplier = 0.5;
+  } else {
+    dexMultiplier = 0;
+  }
+  return Math.floor((catchingCharmMultiplier.value * dexMultiplier! * Math.min(255, modifiedCatchRate)) / 6);
```
Tempo-anon: Does the dexMultiplier need the ! at the end?

Tempo-anon (dismissed):
chained ternaries are a crime

src/data/pokeball.ts
frutescens: yes, otherwise typedocs throws an error because it thinks the variable is being used. before it is assigned a value.

src/data/pokeball.ts
```diff
@@ -68,6 +68,7 @@ export function getPokeballTintColor(type: PokeballType): number {
 /**
  * Gets the critical capture chance based on number of mons registered in Dex and modified {@link https://bulbapedia.bulbagarden.net/wiki/Catch_rate Catch rate}
  * Formula from {@link https://www.dragonflycave.com/mechanics/gen-vi-vii-capturing Dragonfly Cave Gen 6 Capture Mechanics page}
+ * The values used to calculate the Pokedex multiplier have been adjusted so that they reflect the current size of the national dex, 1025,
  * @param modifiedCatchRate the modified catch rate as calculated in {@linkcode AttemptCapturePhase}
```
DayKev:
```suggestion
 * Gets the critical capture chance based on number of mons registered in Dex and modified {@link https://bulbapedia.bulbagarden.net/wiki/Catch_rate Catch rate}.
 * Formula from {@link https://www.dragonflycave.com/mechanics/gen-vi-vii-capturing Dragonfly Cave Gen 6 Capture Mechanics page}.
 * The values used to calculate the Pokedex multiplier have been adjusted so that they reflect the current size of the national dex, 1025.
 * @param modifiedCatchRate - The modified catch rate as calculated in {@linkcode AttemptCapturePhase}
```

DayKev (dismissed):

src/data/pokeball.ts
DayKev: Weird, it shouldn't complain about that because of the `else` block.

innerthunder (dismissed):

DayKev (dismissed):

src/data/pokeball.ts
MokaStitcher: typedoc or typecheck? I'm not getting an error from either locally when removing the !
If it's really an issue, an alternative would be to initialize it to 0 at declaration and remove the else

MokaStitcher (commented):
I'm assuming the change of one threshold from 100 to 40 was intended?
If yes that also becomes a balance PR and i'd like to know why/where this was decided and have that information added in the PR description please

MokaStitcher (approved):

DayKev (approved):

PR #665, opened 2025-02-13
[Move] Add `edgeCase` to moves with "unselectable" conditions

innerthunder:

N/A

The current implementation for Gravity doesn't restrict move selection as it does in mainline, according to [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Gravity_(move)):

> Intense gravity also prevents Pokémon from using the moves [Bounce](https://bulbapedia.bulbagarden.net/wiki/Bounce_(move)), [Fly](https://bulbapedia.bulbagarden.net/wiki/Fly_(move)), [Flying Press](https://bulbapedia.bulbagarden.net/wiki/Flying_Press_(move)), [High Jump Kick](https://bulbapedia.bulbagarden.net/wiki/High_Jump_Kick_(move)), [Jump Kick](https://bulbapedia.bulbagarden.net/wiki/Jump_Kick_(move)), [Magnet Rise](https://bulbapedia.bulbagarden.net/wiki/Magnet_Rise_(move)), [Sky Drop](https://bulbapedia.bulbagarden.net/wiki/Sky_Drop_(move)), [Splash](https://bulbapedia.bulbagarden.net/wiki/Splash_(move)) and [Telekinesis](https://bulbapedia.bulbagarden.net/wiki/Telekinesis_(move)). If a Pokémon is in the semi-invulnerable turn of Fly or Bounce when Gravity is used, that move is canceled immediately. If a Pokémon has no other usable moves besides the moves blocked by Gravity, it will use [Struggle](https://bulbapedia.bulbagarden.net/wiki/Struggle_(move)).

Similarly, Gigaton Hammer and Blood Moon should not be selectable the turn after they are used. Stuff Cheeks is already marked with an `edgeCase` for similar reasons.

Added `.edgeCase()` to the following moves' initializations with comments:
- Gravity
- G-max Gravitas
- Gigaton Hammer
- Blood Moon

## Checklist

  - [?] Have I made sure that any UI change works for both UI themes (default and legacy)?

### DIFF
--- a/src/data/init-moves.ts
+++ b/src/data/init-moves.ts
@@ -1553,7 +1553,7 @@ export function initMoves() {
       .ignoresProtect()
       .attr(AddArenaTagAttr, ArenaTagType.GRAVITY, ArenaTagRelativeSide.ALL, { turnCount: 5 })
       .target(MoveTarget.BOTH_SIDES)
-      .partial(), // does not prevent Bounce, Fly, etc. from being selected; only causes the moves to fail.
+      .edgeCase(), // does not prevent Bounce, Fly, etc. from being selected; only causes the moves to fail.
     new StatusMove(MoveId.MIRACLE_EYE, ElementalType.PSYCHIC, -1, 40, -1, 0, 4)
       .attr(ExposedMoveAttr, BattlerTagType.IGNORE_DARK)
       .ignoresSubstitute(),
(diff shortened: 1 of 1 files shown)


src/data/init-moves.ts
```diff
@@ -1552,7 +1552,8 @@ export function initMoves() {
     new StatusMove(MoveId.GRAVITY, ElementalType.PSYCHIC, -1, 5, -1, 0, 4)
       .ignoresProtect()
       .attr(AddArenaTagAttr, ArenaTagType.GRAVITY, ArenaTagRelativeSide.ALL, { turnCount: 5 })
-      .target(MoveTarget.BOTH_SIDES),
+      .target(MoveTarget.BOTH_SIDES)
+      .partial(), // does not prevent Bounce, Fly, etc. from being selected; only causes the moves to fail.
```
Tempo-anon: I would argue that all the gravity and stuff cheek partials (I believe gigaton hammer and one other move may have similar deviations from the mainline) should be edgeCases rather than partial because functionally there isn't that big of a difference for a player between failing a move and being prevented from using it

Tempo-anon (dismissed):
Actually was wondering if you could add the .edgeCase() to the other 3 moves this applies to

* Gigaton Hammer
* Blood Moon
* Stuff Cheeks

Since these are all cases of the mainline games preventing the player from using a move while the webgame implementation just makes the move fail

Tempo-anon (approved):

DayKev (approved):

PR #667, opened 2025-02-13
[P2] Fix Captivate not checking oblivious or right target

### LINKED ISSUE #569: [Bug] Captivate only checks the Pokemon on the Left Side's Gender to Determine Success
Copied over from PKR 4976

### Describe the bug

Captivate only checks for the Pokémon on the left side's Gender before deciding whether it can apply to both Pokémon in a Double battle.

-Captivate is used in a battle against two wild Pokémon of differing Gender types (male/female, male/genderless, female/genderless)
-Depending on the gender compatibility of the user, Captivate will either fail to affect or succeeds and lowers the stats of **both** Pokémon.

### Reproduction

Use Captivate in a double battle against a compatible-gendered Pokémon and an incompatible-gendered one.

### Expected behavior

https://replay.pokemonshowdown.com/gen9doublescustomgame-2258371603
-Captivate should check Gender individually

### Screenshots / Videos

Captivate works if the Pokemon on the left side is the correct gender

https://github.com/user-attachments/assets/09fc7422-67fb-42b7-aa4c-0414f2271c73

Captivate does not work if the Pokemon on the left side is not the correct gender

https://github.com/user-attachments/assets/16b0218c-d62b-4566-b54b-b15b0812b9c1

Slot of the User doesn't seem to matter

https://github.com/user-attachments/assets/4be1ea81-7f76-4ac6-b873-c8b9579cb514

### Session export file

_No response_

### User data export file

_No response_

### Additional context

https://github.com/pagefaultgames/pokerogue/issues/3875
A little bit similar, where only the Pokémon on the right side's healing is used when considering the healing of Spread moves against Pokémon in Double battles.

### LINKED ISSUE #683: [Move][Checkup] Captivate
- [x] Affects both opponents
- [x] Checks if each opponent is the opposite gender of the user
- [x] Has no affect on oblivious targets
- [x] Just does nothing if user is genderless (showdown does not have the move failing if the user is genderless or does nothing against all targets)

Tempo-anon:

* Oblivious now blocks captivate's stat drop
* Captivate now properly checks both targets' genders instead of just the left one

Fixes https://github.com/Despair-Games/poketernity/issues/569

* Created a new `CaptivateAttr`

## Screenshots/Videos

https://cdn.discordapp.com/attachments/1308786510077890732/1339726933528940620/image.png?ex=67afc56f&is=67ae73ef&hm=693080055513108aa9969bf44c05028c78ed70f5ebae63445634421d183c6a4e&

## How to test the changes?

```ts
const overrides = {
  STARTER_SPECIES_OVERRIDE: Species.NIDOQUEEN, // Might still want to use a female starter though
  STARTING_LEVEL_OVERRIDE: 100,
  MOVESET_OVERRIDE: [MoveId.CAPTIVATE, MoveId.ICE_BEAM],
  ENEMY_SPECIES_OVERRIDE: Species.NIDOKING,
}
```

## Checklist


innerthunder (changes requested):
Can you add this condition to the new attribute?

> It always fails if the user is a gender-unknown Pokémon.

Also, what's supposed to happen if all targets are invalid? Should the move fail?

Tempo-anon:
> Can you add this condition to the new attribute?
> 
> > It always fails if the user is a gender-unknown Pokémon.
> 
> Also, what's supposed to happen if all targets are invalid? Should the move fail?

If the user is genderless then the `isOppositeGender` check will automatically return false.

Testing on showdown, it doesn't look like the move fails if the user is genderless (so bulbapedia might be wrong again) or if both targets aren't affected, it just has no effect: https://replay.pokemonshowdown.com/gen9doublescustomgame-2300703977

innerthunder (dismissed):
Please make a checkup issue for this before merging 👍

Tempo-anon:
Closes #683

src/data/move-attrs/captivate-attr.ts
```diff
@@ -0,0 +1,26 @@
+import type { Pokemon } from "#app/field/pokemon";
+import { globalScene } from "#app/global-scene";
+import { StatStageChangePhase } from "#app/phases/stat-stage-change-phase";
+import { Abilities } from "#enums/abilities";
+import { Stat } from "#enums/stat";
+import type { Move } from "../move";
+import { MoveEffectAttr } from "./move-effect-attr";
+
+/**
+ * Attribute used for captivate where all opponents that do not have the {@linkcode OBLIVIOUS} ability
```
MokaStitcher:
```suggestion
 * Attribute used for captivate where all opponents that do not have the {@linkcode Abilities.OBLIVIOUS} ability
```

src/data/move-attrs/captivate-attr.ts
```diff
@@ -0,0 +1,26 @@
...
+ * and are of opposite gender from the user has their SPATK stat dropped by 2 stages
+ */
+export class CaptivateAttr extends MoveEffectAttr {
+  constructor() {
+    super();
+  }
+
```
MokaStitcher:
```suggestion
```

MokaStitcher (commented):
~~With this the move does not fail even when both targets are immune, is that intended? (same in single battle with immune opponent)~~ nevermind i had missed the above comment

But there is also no message saying that a given target is immune, which feels weird. Shouldn't that be added?

<sub> (also can you try and avoid posting discord image links in the future please, because that one is already expired. Or use the link to the post itself so that we can access it from discord (assuming it's in the server))

PigeonBar:
Also needs a couple of unit tests

src/data/move-attrs/captivate-attr.ts
```diff
@@ -0,0 +1,29 @@
...
+import { MoveEffectAttr } from "./move-effect-attr";
+import { getPokemonNameWithAffix } from "#app/messages";
+
+/**
+ * Attribute used for captivate where all opponents that do not have the {@linkcode Abilities.OBLIVIOUS} ability
+ * and are of opposite gender from the user has their SPATK stat dropped by 2 stages
+ */
+export class CaptivateAttr extends MoveEffectAttr {
+  override applyEffect(user: Pokemon, target: Pokemon, _move: Move): boolean {
+    if (!target.hasAbility(Abilities.OBLIVIOUS) && target.isOppositeGender(user)) {
```
PigeonBar, line 18: (TODO: This doesn't show the ability flyout for Oblivious. Do we want the flyout to show?)

PigeonBar (dismissed):

src/data/move-attrs/captivate-attr.ts
Tempo-anon: Hm I think that would be nice. Is there a easy way of triggering the ability flyout to happen here?

PigeonBar: I think it would require adding a new `CaptivateImmunityAbAttr` and then calling `applyAbAttrs` on that attr.

PigeonBar: On second thought, another option could be to add a new param `showAbility` to `Pokemon.hasAbility()`.

Tempo-anon: Oh that's probably a much better place to put it

Tempo-anon: Maybe another PR though to consolidate flyouts

MokaStitcher (dismissed):

src/data/move-attrs/captivate-attr.ts
MokaStitcher: flyouts are inconsistent all over the place, definitely need a to be taken a general look at

PR #668, opened 2025-02-13
[Test] Fix flaky Rage test

PigeonBar:

The test could break due to the player Pokemon occasionally fainting.

The test now sets the player Pokemon's Defense to +6 stages to prevent it from fainting.

## How to test the changes?

Check that the player Pokemon still survives if you override the opponents to have max IV's and a higher level.

## Checklist

- [x] Are all unit tests still passing? (`npm run test:silent`

### DIFF
--- a/test/moves/rage.test.ts
+++ b/test/moves/rage.test.ts
@@ -28,6 +28,7 @@ describe("Moves - Rage", () => {
       .battleType("double")
       .startingLevel(100)
       .moveset([MoveId.RAGE, MoveId.SPLASH])
+      .disableCrits()
       .enemySpecies(Species.BLISSEY)
       .enemyLevel(100)
       .enemyAbility(Abilities.NO_GUARD)
--- a/test/moves/rage.test.ts
+++ b/test/moves/rage.test.ts
@@ -37,6 +37,7 @@ describe("Moves - Rage", () => {
   it("Rage should increase the user's attack by 1 for each time they are hit", async () => {
     await game.classicMode.startBattle([Species.BLISSEY]);
     const playerPokemon = game.field.getPlayerPokemon();
+    playerPokemon.setStatStage(Stat.DEF, 6); // Prevents Blissey from fainting due to its low Defense stat
 
     game.move.select(MoveId.RAGE, 0, BattlerIndex.ENEMY);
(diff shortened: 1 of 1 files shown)


frutescens (dismissed):

test/moves/rage.test.ts
```diff
@@ -37,6 +37,7 @@ describe("Moves - Rage", () => {
   it("Rage should increase the user's attack by 1 for each time they are hit", async () => {
     await game.classicMode.startBattle([Species.BLISSEY]);
     const playerPokemon = game.field.getPlayerPokemon();
+    playerPokemon.setStatStage(Stat.DEF, 6); // Prevents Blissey from fainting due to its low Defense stat
```
Tempo-anon, line 41: I forget, are crits disabled in test? I was thinking of just setting Blissey's hp back to max at the start of every turn

Tempo-anon (dismissed):
Thanks for helping me clean up my unit test 😭

test/moves/rage.test.ts
PigeonBar: There's a `disableCrits()` in the create-test boilerplate, but `rage.test.ts` is missing it for some reason
https://github.com/Despair-Games/poketernity/blob/958f353b088a20de0b29c5f788823f8226f96ebb/create-test-boilerplate.js#L138-L148

Tempo-anon: makes sense, I make my tests by copying/pasting a random test that already exists instead of using the boilerplate

test/moves/rage.test.ts
```diff
@@ -37,6 +38,7 @@ describe("Moves - Rage", () => {
   it("Rage should increase the user's attack by 1 for each time they are hit", async () => {
     await game.classicMode.startBattle([Species.BLISSEY]);
```
DayKev, line 39: Maybe should have used Shuckle instead of Blissey. 🤔

DayKev (approved):

test/moves/rage.test.ts
Tempo-anon: You would think base 10 attack would be low enough...

Tempo-anon (approved):

PR #672, opened 2025-02-14
[Test] Display caution message when a Pokemon is damaged to low HP

PigeonBar:

Unit tests now display a caution message if a Pokemon get damaged to less than 20% HP (plus a few other conditions). This message gets displayed both after the tests are finished (so that the developer can notice it more easily) and also during the test runtime (so that the developer can search for which attack caused the low HP).

(I chose 20% based on vibes, would a different threshold be better?)

A measure to prevent flaky tests from Pokemon accidentally fainting.

1. `MockConsoleLog` has added functions to queue and print warning messages after all tests are finished.
2. The test framework now mocks `Pokemon.damage()` so that if the damage causes the Pokemon to drop to low HP, then a caution message gets displayed.
3. Fixed a Grudge test being broken due to the enemy Pokemon having Wonder Guard.

## Screenshots/Videos

20250213 Test Low HP

## How to test the changes?

`npm run test freeze_dry` -> caution message
`npm run test grudge` -> no caution, since the enemy Pokemon is intentionally damaged to 1 HP by False Swipe

## Checklist


test/testUtils/gameWrapper.ts
```diff
@@ -40,6 +46,43 @@ export class GameWrapper {
...
+       * - The Pokemon was not damaged by a move that intentionally involves low HP (False Swipe, Hard Press, etc.)
+       */
+      if (
+        ret > 0
+        && pokemon.getHpRatio() < 0.2
+        && !pokemon.isFainted()
+        && !pokemon.getTag(BattlerTagType.ENDURING)
+        && !isLowHpMove
+      ) {
+        const warning1 = `Caution: ${side} ${pokemon.name} was damaged to low HP (${pokemon.hp}/${pokemon.getMaxHp()}) by the move ${moveName}!`;
```
Tempo-anon: Is there a way to make the text color yellow or red? Would make it stand out from all the other console logging messages

Tempo-anon (commented):
`test/gameModes/classic_final_boss.test.ts` is being flaky hmm

test/testUtils/gameWrapper.ts
PigeonBar: <details><summary>Given that the test console outputs all colour-formatting code as raw strings, I don't think it would be possible.</summary>
<p>

image

</p>
</details>

test/testUtils/gameWrapper.ts
```diff
@@ -40,6 +46,43 @@ export class GameWrapper {
       if (cb) cb();
     };
 
+    Pokemon.prototype.damage = function (...args) {
+      const pokemon: Pokemon = this;
+      const ret = GameWrapper.originalDamage.apply(pokemon, args);
+
+      const side = pokemon.isPlayer() ? "Player" : "Enemy";
+      const lowHpMoves = ["False Swipe", "Hard Press"];
```
MokaStitcher: Would it be better to use MoveId?

test/testUtils/mocks/mockConsoleLog.ts
```diff
@@ -14,6 +14,13 @@ export class MockConsoleLog {
   private logs: any[] = [];
   private notified: any[] = [];
 
+  /**
+   * A list of warnings that are queued to be displayed after all tests are finished.
```
MokaStitcher: do we need to clarify that it's not "all tests" but "all tests in a file"? (same in other doc below)

MokaStitcher (dismissed):

test/testUtils/gameWrapper.ts
PigeonBar: I am checking the move names due to convenience since the caution message needs to display the move name anyway. I could change it to check the move ID if desired.

test/testUtils/mocks/mockConsoleLog.ts
PigeonBar: It turns out that the warnings would accumulate in a weird way where each thread would only accumulate warnings previously printed by the same thread, so I have also changed it to clear the warnings queue at the end of every file.

MokaStitcher (dismissed):

test/testUtils/gameWrapper.ts
```diff
@@ -40,6 +47,46 @@ export class GameWrapper {
...
+    Pokemon.prototype.damage = function (...args) {
+      const pokemon: Pokemon = this;
+      const ret = GameWrapper.originalDamage.apply(pokemon, args);
+
+      const side = pokemon.isPlayer() ? "Player" : "Enemy";
+      const lowHpMoves = [MoveId.FALSE_SWIPE, MoveId.HARD_PRESS];
+      const currentPhase = globalScene.getCurrentPhase();
+      let moveName = "N/A";
+      let moveId = MoveId.NONE;
+      if (currentPhase instanceof MoveEffectPhase) {
```
innerthunder:
```suggestion
      if (currentPhase.is<MoveEffectPhase>(PhaseType.MOVE_EFFECT)) {
```
should let you import `MoveEffectPhase` as a type

MokaStitcher (approved):

Tempo-anon (approved):

PR #673, opened 2025-02-14
[P2] Healing Wish and Lunar Dance now checks user's party instead of always using player's, also checks for fainted allies

### LINKED ISSUE #586: [Bug] Healing Wish/Lunar Dance KOs the user even when it has no non-fainted teammates in the back
Copied over from PKR 5033

### Describe the bug

Using Healing Wish or Lunar Dance at any point will KO the user. This includes occasions where the user has no teammates, all of the user's teammates are fainted, or the only teammate is next to the user in a Double battle.

How the move is supposed to work, is that it only succeeds if the user has at least one conscious party member in the back. If the user has no teammates, only fainted teammates, or its only teammate is currently its partner in a Double battle, both Healing Wish and Lunar Dance should fail.

Link to Discord's Bug Report (includes Showdown replays and cartridge footage displaying correct behaviour) https://discord.com/channels/1125469663833370665/1319936324618424320/1319936324618424320

### Reproduction

Use Healing Wish or Lunar Dance with only one non-fainted Pokémon in your party (namely, the HW/LD user).

In the Doubles case, use Healing Wish or Lunar Dance when the user has only one teammate, but it's already sent out. Then use Healing Wish/Lunar Dance.

To reproduce on a wild Pokémon, spawn a wild Pokémon that only has Healing Wish or Lunar Dance in its moveset. Alternatively, for Lunar Dance specifically, spawn a wild Pokémon with Dancer and use Lunar Dance. The wild Pokémon will copy the move and faint.

### Expected behavior

In all cases, Healing Wish/Lunar Dance should fail, displaying a "But it failed!" message and not fainting the user.

### Screenshots / Videos

(will add later)

### Session export file

(will add later)

### User data export file

_No response_

### Additional context

Regarding the Lunar Dance/Dancer interaction: if Lunar Dance fails for any of the reasons given, Dancer should not copy it. If Dancer successfully copies Lunar Dance, but the Dancer user has no living teammates etc., Lunar Dance will fail for the Dancer user.

### LINKED ISSUE #671: [Bug] Healing Wish and Lunar Blessing Use the Player's Party to Determine Success, Healing
### Describe the bug

Healing Wish and Lunar Blessing use the Player's Party rather than the Opponent's when the move is used by an Opponent.

### Reproduction

Begin with 2 or more Pokemon. 

Refresh and release a Pokemon, notice that the Opponent's Healing Wishes are now failing,
 ```
	STARTING_WAVE_OVERRIDE: 5,

	ENEMY_LEVEL_OVERRIDE: 5,

	ENEMY_MOVESET_OVERRIDE: [Moves.HEALING_WISH],
```

### Expected behavior

The Opponent should consider their own Party when using Healing Wish, and not the Player's.

### Screenshots / Videos

Player's Healing Wish

https://github.com/user-attachments/assets/d37d37cb-8231-4cd8-ba18-aae929da3155

Opponent's Healing Wish (Player has more than one eligible Pokemon)

https://github.com/user-attachments/assets/c343dc08-0e50-49f2-95b9-fbb6ab9fc4ac

Opponent's Healing Wish (Player only has one eligible Pokemon)

https://github.com/user-attachments/assets/a5d712e0-34ab-48b8-9d0f-3ea902d37ca8

### Session export file

_No response_

### User data export file

_No response_

### Additional context

Fixed by Pokerogue PR 5207, Originally Reported on Discord.

### PR DESCRIPTION
**Pull request by Tempo-anon**

* Enemies using Healing Wish and Lunar Dance will no longer check the player's party
* These moves fail if the user does not have at least one fainted ally

* Fixes #671
* Fixes #586

* Removed the hardcoded `getPlayerParty` and used `user.getParty()` instead
* Updated the condition
```
.condition(
        (user, _target, _move) =>
      user.getParty().filter((p) => p.isActive()).length > globalScene.currentBattle.getBattlerCount();
      )
```
that checks to make sure there is at least 1 unfainted ally out of battle

## Screenshots/Videos

n/a

## How to test the changes?

```ts
ENEMY_LEVEL_OVERRIDE: 5,

ENEMY_MOVESET_OVERRIDE: [Moves.HEALING_WISH],
```

## Checklist


src/data/init-moves.ts
```diff
@@ -1568,6 +1568,10 @@ export function initMoves() {
       .attr(GyroBallPowerAttr)
       .bulletMove(),
     new SelfStatusMove(MoveId.HEALING_WISH, ElementalType.PSYCHIC, -1, 10, -1, 0, 4)
+      .condition(
+        (user, _target, _move) =>
+          user.getParty().filter((p) => !p.isFainted()).length > (globalScene.currentBattle.double ? 1 : 0),
```
MokaStitcher: I'm pretty sure getParty does not filter out mons that are invalid for a challenge. So we need to check for that
```suggestion
          user.getParty().filter((p) => !p.isFainted() && p.isAllowedInChallenge()).length > (globalScene.currentBattle.double ? 1 : 0),
```

A comment clarifying the condition could be nice too, it's not super easy to understand at a glance i think

src/data/init-moves.ts
```diff
@@ -1946,6 +1950,10 @@ export function initMoves() {
       HighCritAttr,
     ),
     new SelfStatusMove(MoveId.LUNAR_DANCE, ElementalType.PSYCHIC, -1, 10, -1, 0, 4)
+      .condition(
+        (user, _target, _move) =>
+          user.getParty().filter((p) => !p.isFainted()).length > (globalScene.currentBattle.double ? 1 : 0),
```
MokaStitcher:
```suggestion
          user.getParty().filter((p) => !p.isFainted() && p.isAllowedInChallenge()).length > (globalScene.currentBattle.double ? 1 : 0),
```

(same about the comment)

src/data/move-attrs/sacrificial-full-restore-attr.ts
```diff
@@ -27,8 +27,10 @@ export class SacrificialFullRestoreAttr extends SacrificialAttr {
 
   override applyEffect(user: Pokemon, target: Pokemon, move: Move): boolean {
     // We don't know which party member will be chosen, so pick the highest max HP in the party
-    const maxPartyMemberHp = globalScene
-      .getPlayerParty()
+
+    const party = user.getParty();
+
```
MokaStitcher:
```suggestion
    const party = user.getParty();
```

MokaStitcher (commented):
challenges :pikasmile:

A test for the move failing when it should would be nice to have too

src/data/init-moves.ts
innerthunder: There's a helper function for this. Also, this probably needs to filter out active Pokemon:

```suggestion
          user.getParty().filter((p) => !p.isActive(true) && p.isAllowedInBattle()).length > 0,
```

MokaStitcher (dismissed):

torranx (dismissed):

src/data/move-attrs/sacrificial-full-restore-attr.ts
```diff
@@ -26,12 +26,18 @@ export class SacrificialFullRestoreAttr extends SacrificialAttr {
   }
 
   override applyEffect(user: Pokemon, target: Pokemon, move: Move): boolean {
+    const party = user.getParty();
+
     // We don't know which party member will be chosen, so pick the highest max HP in the party
-    const maxPartyMemberHp = globalScene
-      .getPlayerParty()
+    const maxPartyMemberHp = party
```
innerthunder: nit: Something like this might be simpler (github won't let me make a suggestion over the whole thing)

```ts
const maxPartyMemberHp = Math.max(...party.map((p) => p.getMaxHp()));
```

src/data/move-attrs/sacrificial-full-restore-attr.ts
```diff
@@ -26,12 +26,18 @@ export class SacrificialFullRestoreAttr extends SacrificialAttr {
...
       .map((p) => p.getMaxHp())
       .reduce((maxHp: number, hp: number) => Math.max(hp, maxHp), 0);
 
+    /**
+     * @todo If the incoming Pokemon does not get any HP healed, status healed, or PP restored,
+     * There should be a battler tag applied to the Pokemon which should expire whenever the heal
+     * would be needed
+     */
```
innerthunder:
```suggestion
    /**
     * @todo If the incoming Pokemon does not get any HP healed, status healed, or PP restored,
     * There should be an arena tag applied to the field which should expire whenever the heal
     * would be needed
     */
```
nit: An arena tag would be more fitting since the stored heal targets a field position rather than a specific Pokemon

src/data/move-attrs/sacrificial-full-restore-attr.ts
```diff
@@ -45,8 +51,12 @@ export class SacrificialFullRestoreAttr extends SacrificialAttr {
     return -20;
   }
 
+  /**
+   * Only works if there is at least 1 unfainted allowed Pokemon in the party and not already in battle
+   * @returns `true` if the condition is met, `false` otherwise
```
innerthunder: This is technically inaccurate; it returns the condition function to add to `Move` objects with this attribute

src/data/move-attrs/sacrificial-full-restore-attr.ts
```diff
@@ -45,8 +51,12 @@ export class SacrificialFullRestoreAttr extends SacrificialAttr {
...
+   */
   override getCondition(): MoveConditionFunc {
-    return (_user, _target, _move) =>
-      globalScene.getPlayerParty().filter((p) => p.isActive()).length > globalScene.currentBattle.getBattlerCount();
+    return (user, _target, _move) =>
+      user.getParty().filter((p) => p.isActive()).length > globalScene.currentBattle.getBattlerCount();
```
innerthunder, line 58:
```suggestion
      user.getParty().filter((p) => p.isActive(true)).length > globalScene.currentBattle.getBattlerCount();
```
Accounting for challenges

MokaStitcher: I don't think that's correct, `isActive(true)` returns the active pokemon on the field if I recall correctly
what we need is to make sure that there is at least one active mon that is not on the field

I think the current implementation worked when I playtested it

Tempo-anon: Yes. `isActive` takes in a boolean of whether or not the Pokemon is on the field or not. By default `isActive` has that param as false so it's checking for Pokemon that are unfainted, allowed in the challenge, and not on the field

PR #690, opened 2025-02-16
[P3] Fix settings giving a warning due to unset player gender

MokaStitcher:

No more warning in console (and tests) when running the game with no set player gender

When launching the game for the first time the PlayerGender's default value is UNSET, so that the game knows to offer the option to the player.
However since all ui handlers are initialized at launch, the settings ui handler attempts to draw everything at that point, and since the UNSET value is not one that is offered to players it throws a warning in the console
This warning shouldn't happen in this situation

- Remove the code outputting the warning from `setup` and set the `optionsCursors` to a default value.
The code's purpose is to highlight all the currently selected settings, but the same code is being run when the settings are opened, in `show`, so it does not need to be done in setup.
- Remove the code that highlights the currently selected option from `setup`, since it is handled in `show` as well

## Screenshots/Videos

before:
warning shown when running tests (from PR 672's branch, because currently warnings do not get logged during tests)
413628069-b12ed769-7f26-4fe7-a2c2-4ba8aedc090b
warning shown in console
image

after:
The correct options are still highlighted, but no more warnings
image
image
 

## How to test the changes?

- Run the game, check that the correct settings are highlighted
- Run the game in a private window to have an unset gender
- No warning should be shown in the console,

## Checklist

### DIFF
--- a/src/ui/settings/abstract-settings-ui-handler.ts
+++ b/src/ui/settings/abstract-settings-ui-handler.ts
@@ -136,18 +136,9 @@ export default class AbstractSettingsUiHandler extends MessageUiHandler {
       this.optionsContainer.add(this.settingLabels[i]);
       this.optionValueLabels.push(
         uiItem.options.map((option) => {
-          const valueLabel = addTextObject(
-            0,
-            0,
-            option.label,
-            option.value === settingsManager[this.category][uiItem.key]
-              ? TextStyle.SETTINGS_SELECTED
-              : TextStyle.SETTINGS_VALUE,
-          );
+          const valueLabel = addTextObject(0, 0, option.label, TextStyle.SETTINGS_VALUE);
           valueLabel.setOrigin(0, 0);
-
           this.optionsContainer.add(valueLabel);
-
           return valueLabel;
         }),
       );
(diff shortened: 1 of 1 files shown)


Tempo-anon (dismissed):

PigeonBar:
<details><summary>A new bug that occurs if you press the speed up / speed down keys before displaying the settings:</summary>
<p>

20250217 Two Game Speeds

</p>
</details> 

I think this can be fixed by changing the following block of code:
https://github.com/Despair-Games/poketernity/blob/c2f04107d86a96795a69e94b31e612788d0d898d/src/ui/settings/abstract-settings-ui-handler.ts#L441-L452
to:
```
    for (const lastValueLabel of this.optionValueLabels[settingIndex]) {
      setTextColor(lastValueLabel, TextStyle.SETTINGS_VALUE);
    }
```
since `lastCursor` is no longer guaranteed to be correct before `show()` has been called.

MokaStitcher:
> A new bug that occurs if you press the speed up / speed down keys before displaying the settings:

ugh!! Good catch, thank you.
I actually ended up removing even more code from `setup` as it shouldn't need to set the color there either. Things should be good now

torranx (approved):

Tempo-anon (approved):
lol I have never (intentionally) used those buttons

PigeonBar (approved):

Issue #512, opened 2025-02-11
[Bug] No Indication for Status Move Failure on Misty Terrain


Opaque02:
Copied over from PKR 3898

### Describe the bug

https://replay.pokemonshowdown.com/gen9customgame-2190484548-11eut1hfmal1b0y1s7o2cjauij8qrzepw

When using a status move with Misty terrain active no message is displayed when it fails to inflict status.

### Expected behavior

When using a status move with Misty Terrain on the field, there should be a message that shows a pokemon was protected;
"xxxxx surrounds itself with a protective mist"

-Message should not appear in cases where a Pokémon would be given status by the secondary effect of a move.

### Screenshots / Videos

https://github.com/user-attachments/assets/e03dad4a-4ea6-4587-bba8-3f568771b590

### What OS did you observe the bug on?

Other

### If other please specify

gameplay

### Which browser do you use?

Other

### If other please specify

gameplay

### Additional context

Related to #3722 ([Bug] No Indication for Sleep Move Failure on Electric Terrain)

SangaraSorama:
Might be like this :
[Fix [BUG] #5081 Wrong visual text when using priority moves on Psychic Terrain #5091](https://github.com/pagefaultgames/pokerogue/pull/5091/files)

SangaraSorama:
Might need a case for it in `isMoveTerrainCancelled` in [terrain.ts](https://github.com/Despair-Games/poketernity/blob/beta/src/data/terrain.ts#L72-L86) :
```
  isMoveTerrainCancelled(user: Pokemon, targets: BattlerIndex[], move: Move): boolean {
    switch (this.terrainType) {
      case TerrainType.PSYCHIC:
        if (!move.hasAttr(ProtectAttr)) {
          // Cancels move if the move has positive priority and targets a Pokemon grounded on the Psychic Terrain
          return (
            move.getPriority(user) > 0 &&
            user.getOpponents().some(o => targets.includes(o.getBattlerIndex()) && o.isGrounded())
          );
        }
    }
```

Additionally, Misty Terrain effects seem to be defined in `canSetStatus` in [field/pokemon.ts](<https://github.com/Despair-Games/poketernity/blob/beta/src/field/pokemon.ts#L3625>)

PR #699, opened 2025-02-18
[Move] Fully Implement Uproar

innerthunder:

[Uproar](https://bulbapedia.bulbagarden.net/wiki/Uproar_(move)) is now fully implemented.

📈 

- Added the new battler tag `UproarTag` which inherits from `MoveLockTag` to implement Uproar's execution and sleep-preventing effects
- Uproar has 2 new attributes:
  - An `AddBattlerTagAttr` to add the aforementioned uproar tag onto the user
  - A `MessageHeaderAttr` to play "{pokemon} is making an uproar!" at the start of the turn when the user's uproar is in effect.
- Torment no longer interrupts any moves while the afflicted Pokemon is under the effects of a `MoveLockTag`. With this, Torment's `edgeCase` is removed.

## Screenshots/Videos

WIP

## How to test the changes?

`npm run test uproar`

## Checklist

  - [x] If so, please leave a link to it here: https://github.com/Despair-Games/poketernity-locales/pull/34
- [?] Has the translation team been contacted for proofreading/translation?


PigeonBar:
One final Frenzy check to change to MoveLockTagTypes:
https://github.com/Despair-Games/poketernity/blob/52916fabbd0de9f3cccd65c9a17d854a8a124baa/src/phases/check-switch-phase.ts#L61-L68

PigeonBar:
Every new battler tag needs to be added to this list:
https://github.com/Despair-Games/poketernity/blob/52916fabbd0de9f3cccd65c9a17d854a8a124baa/src/data/move-attrs/add-battler-tag-attr.ts#L70-L71

src/data/battler-tags.ts
```diff
@@ -1180,6 +1181,78 @@ export class FrenzyTag extends MoveLockTag {
...
+   */
+  override onAdd(pokemon: Pokemon): void {
+    // "{pokemonNameWithAffix} caused an uproar!"
+    globalScene.queueMessage(
+      i18next.t("battlerTags:uproarOnAdd", { pokemonNameWithAffix: getPokemonNameWithAffix(pokemon) }),
+    );
+
+    // Wake up all sleeping Pokemon on the field
+    globalScene.getField(true).forEach((p) => {
+      if (p.status?.effect === StatusEffect.SLEEP) {
```
Tempo-anon: Does this have any interaction with comatose?

src/field/pokemon.ts
```diff
@@ -4056,7 +4057,12 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
         }
         break;
       case StatusEffect.SLEEP:
-        if (this.isGrounded() && globalScene.arena.hasTerrain(TerrainType.ELECTRIC)) {
+        const preventSleep = new BooleanHolder(false);
+        globalScene
+          .getField(true)
+          .forEach((p) => p.getTag<UproarTag>(BattlerTagType.UPROAR)?.apply(p, quiet, this, preventSleep));
+
+        if (preventSleep.value || (this.isGrounded() && globalScene.arena.hasTerrain(TerrainType.ELECTRIC))) {
```
Tempo-anon, line 3678: Not related to this PR but another bug is https://github.com/Despair-Games/poketernity/issues/512 where terrains blocking something result in no message. Maybe the `hasTerrain` check can also be something like tweaking a boolean holder and displaying a message for a specific case

src/field/pokemon.ts
```diff
@@ -4099,7 +4105,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     turnsRemaining: number = 0,
     sourceText: string | null = null,
   ): boolean {
-    if (!this.canSetStatus(effect, asPhase, false, sourcePokemon)) {
+    if (!this.canSetStatus(effect, !asPhase, false, sourcePokemon)) {
```
Tempo-anon, line 3721: What does flipping the boolean value of `asPhase` do here?

test/moves/uproar.test.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    const player = game.field.getPlayerPokemon();
+
+    game.move.use(MoveId.UPROAR);
+    await game.toNextTurn();
+
+    expect(player.getTag(BattlerTagType.UPROAR)?.turnCount).toBe(2);
+    expect(player.getMoveQueue()[0]).toMatchObject({
+      moveId: MoveId.UPROAR,
+      ignorePP: true,
```
Tempo-anon, line 50: Is there a way to also check PP here? Bulbapedia says that PP is only deducted on the first initial use of uproar

src/data/battler-tags.ts
innerthunder: No, but it's changed now to ignore it.

src/field/pokemon.ts
innerthunder: `asPhase` is used to schedule an `ObtainStatusEffectPhase` instead of modifying the Pokemon's status (it's extremely convoluted, I know). Before this change, the flow for this process would be

```
trySetStatus > check canSetStatus quietly > if true, schedule ObtainStatusEffectPhase
```

`try/canSetStatus` would then be called again within the phase, except not quietly that time. The reason `asPhase` is flipped here is because if the phase is never queued, the logic in `canSetStatus` would only be run quietly, i.e. Uproar's prevention message wouldn't show.

src/field/pokemon.ts
innerthunder: Possibly? I thought there was a fix for Electric Terrain a while ago...

Tempo-anon (dismissed):

Tempo-anon (dismissed):

src/data/init-moves.ts
```diff
@@ -1123,9 +1123,15 @@ export function initMoves() {
       .attr(FlinchAttr)
       .condition(new FirstMoveCondition()),
     new AttackMove(MoveId.UPROAR, ElementalType.NORMAL, MoveCategory.SPECIAL, 90, 100, 10, -1, 0, 3)
+      .attr(AddBattlerTagAttr, BattlerTagType.UPROAR, true)
+      .attr(MessageHeaderAttr, (user, _move) =>
+        !!user.getTag(BattlerTagType.UPROAR)
+          ? // "{pokemonNameWithAffix} is making an uproar!"
+            i18next.t("moveTriggers:isMakingAnUproar", { pokemonNameWithAffix: getPokemonNameWithAffix(user) })
```
MokaStitcher: this gives `pokemonNameWithAffix` but the text expects just `pokemonName`
image

MokaStitcher (dismissed):
The uproar never stops if the wave finishes before it expires. It will also always fail in the next wave, resulting in an infinite loop
<details><summary>Steps to reproduce</summary>

 ```
  STARTING_LEVEL_OVERRIDE: 50,
  MOVESET_OVERRIDE: MoveId.UPROAR,
  ENEMY_MOVESET_OVERRIDE: MoveId.SPLASH,
```
- start a run
- use uproar
- go to the next wave

</details>

**Comment by innerthunder** [comment]
> The uproar never stops if the wave finishes before it expires. It will also always fail in the next wave, resulting in an infinite loop
> 
> Steps to reproduce
> ```
>  STARTING_LEVEL_OVERRIDE: 50,
>  MOVESET_OVERRIDE: MoveId.UPROAR,
>  ENEMY_MOVESET_OVERRIDE: MoveId.SPLASH,
> ```
> 
> * start a run
> * use uproar
> * go to the next wave

This seems to be an issue with frenzy moves as well... As far as I can tell it's related to `MoveLockTag` trying to resolve the target of the next queued move immediately after using the move. Random target selection only considers battler indexes with active Pokemon at the time, so all moves queued afterward have an empty `targets` array.

I'll put this in draft and make another PR with a fix since the issue isn't exclusive to Uproar

File: src/data/init-moves.ts [hunk 6]
```diff
@@ -1124,9 +1124,15 @@ export function initMoves() {
       .attr(FlinchAttr)
       .condition(new FirstMoveCondition()),
     new AttackMove(MoveId.UPROAR, ElementalType.NORMAL, MoveCategory.SPECIAL, 90, 100, 10, -1, 0, 3)
+      .attr(AddBattlerTagAttr, BattlerTagType.UPROAR, true)
+      .attr(MessageHeaderAttr, (user, _move) =>
+        !!user.getTag(BattlerTagType.UPROAR)
+          ? // "{pokemonName} is making an uproar!"
+            i18next.t("moveTriggers:isMakingAnUproar", { pokemonName: getPokemonNameWithAffix(user) })
```
Tempo-anon: Should this still be using `getPokemonNameWithAffix` if it's supposed to just be pokemonName?

MokaStitcher: imo it's the locale that is wrong, but since it was merged a new PR needs to be made for it

innerthunder: I reverted it. Now waiting for https://github.com/Despair-Games/poketernity-locales/pull/40 to go through.

PR #702, opened 2025-02-18
[Move] Fully Implement Rollout and Ice Ball

innerthunder:

Rollout and Ice Ball's move-locking effect is now implemented

📈 

- Added a new battler tag, `RollingTag`, which inherits from `MoveLockTag` to implement Rollout and Ice Ball's move-locking properties. The new tag is nearly identical to the base class, except that it also provides a power multiplier getter based on its remaining turn count.
- Added a new move attribute extending `MovePowerMultiplierAttr` which calculates Rollout and Ice Ball's power based on the user's `RollingTag` and move history (to check if Defense Curl has been used).
- Removed some code in `ConsecutiveUsePowerMultiplierAttr` specific to the old Rollout/Ice Ball implementations.

## Screenshots/Videos

WIP

## How to test the changes?

`npm run test rollout`

## Checklist

No locale changes needed :>


src/enums/battler-tag-type.ts
```diff
@@ -105,4 +105,6 @@ export enum BattlerTagType {
   SKY_DROP,
   CRIT_BOOST_STACKABLE,
   RAGE,
+  ROLLOUT,
+  ICE_BALL,
```
Tempo-anon: Does this need two different moves? Can't it just be a single one that just takes in the sourceMove when it's created?

innerthunder: Is it safe to assume Rollout / Ice Ball will never be used by the same user within the same turn? The main reason I made them separate tags was so that they don't feed off of each other (wrt. power calculation) in that case.

Tempo-anon: I'm struggling to think of a case where something like this could happen. Dancer grants extra moves but only for dance moves. I also know it's possible to do with some game overriding stuff like MEs and setting up multiple moves, but I don't think rollout/iceball are ever going to be a consideration (especially since both moves aren't that good)

src/data/move-attrs/rolling-power-multiplier-attr.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Multiplies the move's power by 2 for every turn spent under the move's
+ * execution
+ * @extends MovePowerMultiplierAttr
+ * @see {@linkcode RollingTag}
+ */
+export class RollingPowerMultiplierAttr extends MovePowerMultiplierAttr {
+  constructor() {
+    super((user, _target, _move) => {
+      const baseMultiplier = user.getTag<RollingTag>(BattlerTagType.ROLLING)?.powerMultiplier ?? 1;
+      const defenseCurlMultiplier = user.getMoveHistory().some((turnMove) => turnMove.move.id === MoveId.DEFENSE_CURL)
```
Tempo-anon: Hmm, and here I was thinking it'd make more sense for defense curl to apply a CURLED_UP battlerTag to the user or something, like what we do with MINIMIZE

src/ui/party-ui-handler.ts
```diff
@@ -1360,7 +1360,9 @@ class PartySlot extends Phaser.GameObjects.Container {
     }
 
     if (this.pokemon.status || this.pokemon.isFainted()) {
-      const effectIconKey = this.pokemon.isFainted() ? "faint" : StatusEffect[this.pokemon.status!.effect].toLowerCase();
+      const effectIconKey = this.pokemon.isFainted()
```
Tempo-anon: random linting of random files despair

test/moves/rollout.test.ts
```diff
@@ -23,57 +24,90 @@ describe("Moves - Rollout", () => {
...
-    game.override.enemySpecies(Species.BIDOOF);
-    game.override.enemyAbility(Abilities.BALL_FETCH);
-    game.override.startingLevel(100);
-    game.override.enemyLevel(100);
-    game.override.enemyMoveset(MoveId.SPLASH);
+    game.override
+      .disableCrits()
+      .battleType("single")
+      .ability(Abilities.BALL_FETCH)
+      .enemySpecies(Species.STAKATAKA)
```
Tempo-anon: Will Stakataka automatically be a boss? Just a shoutout since I used to use Regirock for tests when I needed a high def target but passives can cause tests to be flaky. If you're going to be mocking all its stats anyway, you can just use something like Aron for the typing

test/moves/rollout.test.ts
```diff
@@ -23,57 +24,90 @@ describe("Moves - Rollout", () => {
...
+      .enemyLevel(100)
+      .enemyMoveset(MoveId.SPLASH);
   });
 
-  it("should double it's dmg on sequential uses but reset after 5", async () => {
-    game.override.moveset([MoveId.ROLLOUT]);
-    vi.spyOn(allMoves[MoveId.ROLLOUT], "accuracy", "get").mockReturnValue(100); //always hit
+  it("should double its power on sequential uses for up to 5 uses", async () => {
+    const moveObj = allMoves[MoveId.ROLLOUT];
+    vi.spyOn(moveObj, "accuracy", "get").mockReturnValue(100); //always hit
```
Tempo-anon: Can also just override on no guard

test/moves/rollout.test.ts
```diff
@@ -23,57 +24,90 @@ describe("Moves - Rollout", () => {
...
+    game.override
+      .disableCrits()
+      .battleType("single")
+      .ability(Abilities.BALL_FETCH)
+      .enemySpecies(Species.STAKATAKA)
+      .enemyAbility(Abilities.BALL_FETCH)
+      .startingLevel(100)
+      .enemyLevel(100)
+      .enemyMoveset(MoveId.SPLASH);
   });
```
Tempo-anon, line 37: also missing a test for defense curl interactions

src/data/move-attrs/rolling-power-multiplier-attr.ts
innerthunder: The only difference between this and a tag is that move history also resets at the end of battle. If we reset state at the end of each wave, that shouldn't be an issue, though 🙃

src/ui/party-ui-handler.ts
innerthunder: Should be removed now? Prettier was tweakin

src/data/move-attrs/rolling-power-multiplier-attr.ts
Tempo-anon: Yeah this isn't a big thing or whatever, I was just thinking defense curl and minimize should probably be implemented in a similar way just for code consistency since they function very similarly

Tempo-anon (dismissed):

MokaStitcher:
I am getting the same infinite loop issue as for #699 so I'm putting it in draft as well for now
(also I was offered to switch mon at the beginning of the next wave while locked into rollout, I don't know if that's correct or not)

innerthunder:
> I am getting the same infinite loop issue as for #699 so I'm putting it in draft as well for now (also I was offered to switch mon at the beginning of the next wave while locked into rollout, I don't know if that's correct or not)

Should be fixed via #771

src/utils/battler-tag-type-utils.ts
```diff
@@ -17,7 +17,7 @@ export const ProtectionBattlerTagTypes = Object.freeze([
   BattlerTagType.OBSTRUCT,
 ]);
 
-export const MoveLockTagTypes = Object.freeze([BattlerTagType.FRENZY]);
+export const MoveLockTagTypes = Object.freeze([BattlerTagType.FRENZY, BattlerTagType.ROLLING]);
```
PigeonBar: This line should also be updated to use `...MoveLockTagTypes`:
https://github.com/Despair-Games/poketernity/blob/8fe50a3351860ff5c70fc1e622601a175a37bc4e/src/data/battler-tags.ts#L3272

src/data/move-attrs/rolling-power-multiplier-attr.ts
```diff
@@ -0,0 +1,26 @@
...
+import { MoveId } from "#enums/move-id";
+import type { RollingTag } from "../battler-tags";
+import { MovePowerMultiplierAttr } from "./move-power-multiplier-attr";
+
+/**
+ * Attribute implementing the power-multiplying properties of
+ * {@link https://bulbapedia.bulbagarden.net/wiki/Rollout_(move) | Rollout}
+ * and {@link https://bulbapedia.bulbagarden.net/wiki/Ice_Ball_(move) | Ice Ball}.
+ * Multiplies the move's power by 2 for every turn spent under the move's
+ * execution
```
PigeonBar:
```suggestion
 *
 * Multiplies the move's power by 2 for every turn spent under the move's execution.
 * Also multiplies the move's power by 2 once if the Pokemon has previously used Defense Curl.
```

test/moves/rollout.test.ts
```diff
@@ -23,57 +24,104 @@ describe("Moves - Rollout", () => {
...
+
+  it("should stop its execution if an attack is unsuccessful", async () => {
+    game.override.ability(Abilities.BALL_FETCH);
+
+    await game.classicMode.startBattle([Species.FEEBAS]);
+
+    const player = game.field.getPlayerPokemon();
+
+    game.move.use(MoveId.ROLLOUT);
+    await game.setTurnOrder([BattlerIndex.PLAYER, BattlerIndex.ENEMY]);
```
PigeonBar:
```suggestion
    game.setTurnOrder([BattlerIndex.PLAYER, BattlerIndex.ENEMY]);
```

test/moves/rollout.test.ts
```diff
@@ -23,57 +24,104 @@ describe("Moves - Rollout", () => {
...
+    await game.move.forceHit();
+
+    await game.toNextTurn();
+
+    expect(player.getTag(BattlerTagType.ROLLING)).toBeDefined();
+    expect(player.getMoveQueue()[0]?.move.id).toBe(MoveId.ROLLOUT);
+
+    await game.setTurnOrder([BattlerIndex.PLAYER, BattlerIndex.ENEMY]);
```
PigeonBar:
```suggestion
    game.setTurnOrder([BattlerIndex.PLAYER, BattlerIndex.ENEMY]);
```

test/moves/rollout.test.ts
```diff
@@ -23,57 +24,104 @@ describe("Moves - Rollout", () => {
...
+  });
+
+  it("should double its power on sequential uses for up to 5 uses", async () => {
+    const moveObj = allMoves[MoveId.ROLLOUT];
+    const spy = vi.spyOn(moveObj, "calculateBattlePower");
+
+    await game.classicMode.startBattle([Species.FEEBAS]);
+
+    const powerResults: number[] = [];
+
```
PigeonBar:
```suggestion
```

test/moves/rollout.test.ts
```diff
@@ -23,57 +24,104 @@ describe("Moves - Rollout", () => {
...
+    for (let i = 0; i < 6; i++) {
+      game.move.use(MoveId.ROLLOUT);
+      await game.toNextTurn();
+
+      const lastReturnedPower = spy.mock.results.at(-1)?.value;
+      if (lastReturnedPower) {
+        powerResults.push(lastReturnedPower);
+      }
```
PigeonBar:
```suggestion
```

test/moves/rollout.test.ts
```diff
@@ -23,57 +24,104 @@ describe("Moves - Rollout", () => {
...
+    }
+
+    expect(powerResults).toHaveLength(6);
+    powerResults
+      .slice(1, 5)
+      .forEach((power, i) => expect(power).toBe(Math.floor(Math.pow(2, i + 1) * powerResults[0])));
+    expect(powerResults[5]).toBe(powerResults[0]);
```
PigeonBar:
```suggestion
    const powerResults = spy.mock.results.map(result => result.value);
    expect(powerResults).toStrictEqual([30, 60, 120, 240, 480, 30]);
```

test/moves/rollout.test.ts
```diff
@@ -23,57 +24,94 @@ describe("Moves - Rollout", () => {
...
+    for (let i = 0; i < 6; i++) {
+      game.move.use(MoveId.ROLLOUT);
+      await game.toNextTurn();
+    }
 
-    const variance = 5;
-    const turns = 6;
-    const dmgHistory: number[] = [];
+    const powerResults = spy.mock.results.map((result) => result.value);
+    expect(powerResults).toStrictEqual([30, 60, 120, 240, 480, 30]);
```
DayKev: The current implementation of Rollout/Ice Ball is bugged and the damage doesn't increase after it resets back to 30 base power, so there should be a test checking for that too (just adding an extra turn to this test and adding that to the expect should probably be good enough).

PigeonBar (approved):

DayKev (approved):

PR #703, opened 2025-02-18
[Tests] Decouple BerryPhase from tests, modernize test script

torranx:
none

in comparison to `BerryPhase`, `TurnEndPhase` is more intuitive, and its execution timing is unlikely to change

- updated template content for the `npm run test:create`
- removed `BerryPhase` from tests, replaced with a **new util `toEndOfTurn()`**
- cleaned up Fairy Lock tests

## How to test the changes?
`npm run test`

## Checklist

### DIFF
--- a/create-test-boilerplate.js
+++ b/create-test-boilerplate.js
@@ -138,20 +138,22 @@ describe("${description}", () => {
   beforeEach(() => {
     game = new GameManager(phaserGame);
     game.override
-      .moveset([ MoveId.SPLASH ])
       .ability(Abilities.BALL_FETCH)
       .battleType("single")
       .disableCrits()
       .enemySpecies(Species.MAGIKARP)
       .enemyAbility(Abilities.BALL_FETCH)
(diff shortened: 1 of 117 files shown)


DayKev:
There should probably be a `.toEndOfTurn()` helper (like `.toNextTurn()`) and then that gets used instead of `.to("TurnEndPhase")`.

torranx:
> There should probably be a `.toEndOfTurn()` helper (like `.toNextTurn()`) and then that gets used instead of `.to("TurnEndPhase")`.

I've thought of this too but not really a fan of turning a one liner to a helper. But I'd agree it'd make it short to type, and probably easier to follow for new devs

DayKev:
The `.toNextTurn()` helper used to be a one-liner too (until it was updated to better account for double battles). In this case I think it's appropriate as this also lets us change which phase it goes to in case the phases are changed around.

DayKev (approved):

innerthunder (approved):

PR #706, opened 2025-02-18
[Refactor] Refactor timed events

MokaStitcher:

- Timer before an event ends is now localized

- Needed to clean up the pkr event banner images from the repo
- Some event related gameplay changes like extra rewards for the rival fights are currently enabled if *any* event is ongoing. But it should get enabled on a per event basis, not by default
- The way event banners are handled in the loading scene is not great. Currently when adding an event you need to put it in the timedEvent file, and also update the loading scene with the banner information. The loading scene also doesn't use the TimedEvent "availableLangs" field to check which languages have a dedicated banner for that event. (And people don't think to remove that code from the loading scene after an event is done so the last event's banner always gets loaded).

Anyway here I am a full refactor later. Hopefully it'll give a better structure to add new event specific effects in the game in the future

- Removed banners for past events and add a couple placeholder banner images for testing
- Break up the timed-event-manager file into multiple files:
   - `src/@types/TimedEvent.ts` for the TimedEvent interface declaration
   - `src/data/all-timed-events.ts` to put the actual event information
   - `src/ui/timed-event-display.ts` for the UI container displaying the event banners
   - `src/timed-event-manager.ts` with the manager itself
- Rework the TimedEvent type:
   - put all banner-related attributes in the EventBanner interface
   - put all modifications an event can add to the game in a new EventModifiers interface
- Rework the TimedEventManager:
   - decouple the manager from the globalScene and have it accessible from everywhere
   - add helper function to find out if there is an active event, or an upcoming event
   - add helper functions `isEventActive` to check if there is an active event with the given "modifier" (currently: shiny, friendship candy, special battle rewards)
   - Call it everywhere in the code a special event one of these modifiers is applied
- Loading Scene:
   - loading scene now ask the timedEventManager for the current active event with a banner if any, or the next upcoming event with a banner, if any.
   - it uses the information from the EventBanner interface to load the proper image for that event's banner. Otherwise, no loading. No need to update the loading scene anymore when adding a new event.
- Tests
   - Remove previous mock of the timedEventManager and replace it with an override in the GameManager
   - Add tests for the event related code (except the special battle reward because it fully needs to be reworked along with modifiers)

## Screenshots/Videos

Silly dev art placeholder banner and localized remaining time text
image
<sub>I removed the scaling from the banner display code, no more mixels allowed!

## How to test the changes?

`npm run test:silent test/events`

- Go to the `data/all-timed-events` file
- Play around with the event configuration
- Run the game and make sure things work as intended

## Checklist

  - [X] Have I created new automated tests (`npm run test:create`) or updated existing tests related to the PR's changes?
  - [X] Have I made sure that any UI change works for both UI themes (dark and light)?

Are there any localization additions or changes?- 
- [X] Yes. They were merged some time ago in https://github.com/Despair-Games/poketernity-locales/pull/12


public/images/events/august-variant-update.png
Tempo-anon, line 1: The fact that all these were a thing is....hmm.....

Like I guess it's good that there is a record somewhere, but git is meant as keeping a snapshot fro everything. Makes me think if there should just be a events/banner_de.png, events/banner_en.png, ...etc and then just have them replaced with whatever new png is needed

test/events/shiny-chance-event.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    expect(game.field.getEnemyPokemon().isShiny()).toBeTruthy();
+  });
+
+  it("should not apply the active event's shiny chance modifier to eggs", async () => {
+    expect(timedEventManager.getActiveEvent()).toBeDefined();
+    expect(timedEventManager.isEventActive(EventModifierType.WILD_SHINY_CHANCE)).toBeTruthy();
+    expect(timedEventManager.getWildShinyChanceMultiplier()).toBe(9000);
+
+    let nonShinies = 0;
+    for (let i = 0; i < 100; i++) {
```
Tempo-anon: If the shiny chance multiplier is high enough it should be able to guarantee a shiny so there's no need to try 100 times

MokaStitcher: this check is testing that the multiplier does NOT affect eggs
but yeah I agree that 10 checks should be more than enough, I just did 100 for extra safety

MokaStitcher:
```suggestion
    for (let i = 0; i < 10; i++) {
```

src/loading-scene.ts
```diff
@@ -195,12 +196,15 @@ export class LoadingScene extends SceneBase {
...
+    // Load the banner for the current or next event with a banner, if any
+    const eventBanner = timedEventManager.getActiveOrNextEventBanner();
+    if (eventBanner) {
+      if (eventBanner.availableLangs) {
+        const bannerLang = eventBanner.availableLangs.includes(lang) ? lang : "en";
+        this.loadImage(eventBanner.key, ImagesFolder.BANNERS, { languageKey: bannerLang });
+      } else {
+        this.loadImage(eventBanner.key, ImagesFolder.BANNERS);
+      }
     }
```
torranx:
```suggestion
    if (eventBanner?.availableLangs.length > 0) {
      const bannerLang = eventBanner.availableLangs.includes(lang) ? lang : "en";
      this.loadImage(eventBanner.key, ImagesFolder.BANNERS, { languageKey: bannerLang });
    } else {
      this.loadImage(eventBanner.key, ImagesFolder.BANNERS);
    }
```

torranx: also question on this, what is loaded on the `else` statement? I'm thinking `availableLangs` should always include "en" since that is the default

MokaStitcher: It's to account for cases of a banner with no text. In this case there's no need for a localized banner at all and "availableLangs" shouldn't be defined or else it will look for an image that ends in _en even though that would make no sense

public/images/events/august-variant-update.png
MokaStitcher: I see your point. Right now I think I'd rather leave things more open rather than define a specific process and format for events / banners.
In the future we want to handle events server side so this will have to be changed regardless, I think I'd rather have a less rigid structure for when we get there

src/timed-event-manager.ts
```diff
@@ -1,193 +1,89 @@
...
 
-interface TimedEvent extends EventBanner {
-  name: string;
-  eventType: EventType;
-  shinyMultiplier?: number;
-  friendshipMultiplier?: number;
-  startDate: Date;
-  endDate: Date;
+function isActiveOrUpcoming(event: TimedEvent) {
+  return event.startDate >= new Date() || new Date() < event.endDate;
```
PigeonBar:
```suggestion
  return new Date() < event.endDate;
```
Should be equivalent, assuming that `startDate` is always before `endDate`

src/data/all-timed-events.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Array of the {@linkcode TimedEvent} used in the game.
+ * This should *not* be accessed directly, all event-related logic should go through the TimedEventManager.
+ *
+ * Note that the event manager filters for outdated events, and the loading scene only load banner images
+ * for the currently active event, or the next event if any.
+ * This means it is fine to keep past events in this array even when adding new ones.
+ */
+export const allTimedEvents: TimedEvent[] = [
```
PigeonBar, line 14:
```suggestion
 *
 * CAUTION: All month numbers are 0-indexed (0 = January, 1 = February, etc.). The year and date numbers are still 1-indexed as normal.
 * Example: Date.UTC(2025, 2, 15) = March 15, 2025.
 */
export const allTimedEvents: TimedEvent[] = [
```

src/ui/timed-event-display.ts
```diff
@@ -0,0 +1,116 @@
...
+  private timeToGo(date: Date) {
+    // Utility to add leading zero
+    function z(n) {
+      return (n < 10 ? "0" : "") + n;
+    }
+    const now = new Date();
+    let diff = Math.abs(date.getTime() - now.getTime());
+
+    // Allow for previous times
+    diff = Math.abs(diff);
```
PigeonBar:
```suggestion
    const diff = Math.max(date.getTime() - now.getTime(), 0); // Make sure the timer does not keep counting past zero
```

src/timed-event-manager.ts
```diff
@@ -1,193 +1,89 @@
...
-    let multiplier = 1;
-    const shinyEvents = timedEvents.filter((te) => te.eventType === EventType.SHINY && this.isActive(te));
-    shinyEvents.forEach((se) => {
-      multiplier *= se.shinyMultiplier ?? 1;
-    });
-
-    return multiplier;
+  private setEvents(events: TimedEvent[]) {
+    // Filter out any expired event from the list
+    this.events = events.filter((te: TimedEvent) => te.endDate >= new Date());
```
PigeonBar:
```suggestion
    this.events = events.filter((te: TimedEvent) => isActiveOrUpcoming(te));
```

PigeonBar:
For reference, this can be used to playtest event expiration:

`endDate: new Date(new Date().getTime() + 30000), // Makes the event end 30 seconds after booting up the game`

src/timed-event-manager.ts
```diff
@@ -1,193 +1,89 @@
...
+    // Sort so that active events appear before upcoming ones
+    this.events.sort((event1, event2) => event2.startDate.getDate() - event1.startDate.getDate());
```
PigeonBar:
```suggestion
    this.events.sort((event1, event2) => event1.startDate.getTime() - event2.startDate.getTime());
```

src/field/pokemon.ts
```diff
@@ -2084,8 +2086,8 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
 
     const shinyThreshold = new NumberHolder(BASE_SHINY_CHANCE);
     if (thresholdOverride === undefined) {
-      if (globalScene.eventManager.isEventActive()) {
-        shinyThreshold.value *= globalScene.eventManager.getShinyMultiplier();
+      if (timedEventManager.isEventActive(EventModifierType.WILD_SHINY_CHANCE)) {
+        shinyThreshold.value *= timedEventManager.getWildShinyChanceMultiplier();
```
PigeonBar, line 2090: Are these `isEventActive()` checks redundant now, since `getWildShinyChanceMultiplier()` and related functions now filter out inactive events?

PigeonBar:
Probably not caused by this PR, but while testing the shiny multiplier, I encountered a rare variant of Persian, despite the fact that Persian's variants were removed in #383:
<details><summary>Screenshot</summary>
<p>

image

</p>
</details>

MokaStitcher:
> Probably not caused by this PR, but while testing the shiny multiplier, I encountered a rare variant of Persian, despite the fact that Persian's variants were removed in #383:


Indeed not related to this PR. I think the issue is that it is still in the masterlist, and that the check for variant existance does not actually look for the values in the masterlist, but only if there's a key for this Pokemon (also explains why mons with replacement shiny and no variants still get variants)
<sub> For reference for myself or whoever tackles this, the check in Pokemon.generateShinyVariant is a duplicate from what's done in PokemonSpecies.hasVariants, it should just call that (and that check should get fixed)

src/timed-event-manager.ts
MokaStitcher: omg thank you... i was sure it was working correctly since I had scrambled the event order in the tests and all
Fun fact... There was a stray `)` in the test that should've failed so it never ran properly and I thought things were fine

src/field/pokemon.ts
MokaStitcher: They are yes, but I think it's probably best practice to have these checks everywhere events are involved.
It really makes it clear when looking at the code that this code only happens when an event is active and hopefully will help to avoid things like that being forgotten for potential future events with more complex behavior than just applying a multiplier.
I'm willing to remove them if you think it's better without though

src/timed-event-manager.ts
MokaStitcher: Good point. I added a check for start date and end date in the initial filtering of the events list just to be safe rather than assume!

src/field/pokemon.ts
PigeonBar: I agree with keeping these checks then.

PigeonBar (approved):

Tempo-anon (approved):

PR #713, opened 2025-02-19
[P3] Fix map and gender select menu configurations

MokaStitcher:

- Can no longer exit out of the gender menu without making a selection (it would select "girl" by default in that case)
- Adds a 1 second input delay before being able to validate the choice
- Biome selection menu also has a 1 second input deal and prevents cancelled

It shouldn't be possible to press cancel to exit those menus, since there is no "cancel" options. The input delay is here to prevent accidental selection cause by mashing.
In #238 I renamed a few fields in the mode's configuration and since the type wasn't being enforced everywhere some configs didn't get properly updated as a result (`delay` was renamed to `inputDelay` and `noCancel` to `blockCancelButton`)

- use `blockCancelButton` instead of the previous name `noCancel` in Select Gender Phase
- add `blockCancelButton` in SelectBiomePhase and changed `delay` to `inputDelay`
- went over the remaining option select configs to make sure they didn't present the same issue. 
- Took the opportunity to somewhat reduce the nesting in the starter select "Manage Moves" option. Behavior should still be the same though (does not fix an existing bug where reordering moves does not work for some Pokemon).

## Screenshots/Videos

Gender Select input delay

https://github.com/user-attachments/assets/d0ea531d-9aeb-4c37-8b8e-cd3feb14c9cd

Map input delay

https://github.com/user-attachments/assets/b29eb435-247b-467c-bbf0-67f4b43d2008

Starter Select move selection still working

https://github.com/user-attachments/assets/c4444f02-c9d8-465d-8724-61e912a366a7

</details> 

## How to test the changes?

Gender selection
- Open the game in a private window to have a fresh file and get offered the gender choice
- During the choice, try to cancel with X / whatever button you're usually using to cancel

Map:
```
  STARTING_LEVEL_OVERRIDE: 50,
  STARTING_WAVE_OVERRIDE: 10,
  STARTING_BIOME_OVERRIDE: Biome.PLAINS,
  STARTING_HELD_ITEMS_OVERRIDE: [ {name: "DYNAMAX_BAND"}, {name: "MAP"}],
```

Starter Select:
Play around with the "Manage Move" menus, behavior should be as before (better to be done on a save file with at least some unlocks)

## Checklist

### DIFF
--- a/src/ui/login-form-ui-handler.ts
+++ b/src/ui/login-form-ui-handler.ts
@@ -7,7 +7,7 @@ import i18next from "i18next";
 import { addTextObject } from "./text";
 import { TextStyle } from "#enums/text-style";
 import { addWindow } from "./ui-theme";
-import type { OptionSelectItem } from "#app/ui/interfaces/option-select-config";
+import type { OptionSelectItem, OptionSelectModeConfig } from "#app/ui/interfaces/option-select-config";
 import { api } from "#app/plugins/api/api";
 import { globalScene } from "#app/global-scene";
 import JSZip from "jszip";
@@ -230,12 +230,13 @@ export default class LoginFormUiHandler extends FormModalUiHandler {
             },
           });
         }
-        globalScene.ui.setOverlayMode(UiMode.OPTION_SELECT, {
+        const optionSelectConfig: OptionSelectModeConfig = {
           options: options,
-          delay: 1000,
+          inputDelay: 1000,
           xOffset: GAME_WIDTH,
           yOffset: GAME_HEIGHT - this.usernameInfoImage.displayHeight - 16 * dataKeys.length - 22,
-        });
+        };
+        globalScene.ui.setOverlayMode(UiMode.OPTION_SELECT, optionSelectConfig);
       } else {
         if (dataKeys.length > 2) {
           return onFail(this.ERR_TOO_MANY_SAVES);
--- a/src/ui/login-form-ui-handler.ts
+++ b/src/ui/login-form-ui-handler.ts
@@ -232,7 +232,6 @@ export default class LoginFormUiHandler extends FormModalUiHandler {
         }
         const optionSelectConfig: OptionSelectModeConfig = {
           options: options,
-          inputDelay: 1000,
           xOffset: GAME_WIDTH,
           yOffset: GAME_HEIGHT - this.usernameInfoImage.displayHeight - 16 * dataKeys.length - 22,
         };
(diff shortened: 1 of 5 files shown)


PigeonBar:
The Map input also has the same bugs (no delay before waiting for input, and pressing Cancel simply chooses the last biome option).
https://github.com/Despair-Games/poketernity/blob/effce69ffecf51286cb5a9224a746ef1aa44708d/src/phases/select-biome-phase.ts#L74-L76

MokaStitcher:
> The Map input also has the same bugs (no delay before waiting for input, and pressing Cancel simply chooses the last biome option).

Ugh, thank you, I'll fix that tomorrow and make a pass over all Option select modes for similar issues. I searched for other 'noCancel's but didn't think of checking 'delay'

There may be other places that could use preventing cancelling too... For the gender thing I added it during option select refactor because it had been bothering me for months

MokaStitcher:
Of course this could've been caught if mode setting had proper type checking for parameters
Someday... I will do it...

DayKev (dismissed):

src/ui/login-form-ui-handler.ts
```diff
@@ -230,12 +230,13 @@ export default class LoginFormUiHandler extends FormModalUiHandler {
             },
           });
         }
-        globalScene.ui.setOverlayMode(UiMode.OPTION_SELECT, {
+        const optionSelectConfig: OptionSelectModeConfig = {
           options: options,
-          delay: 1000,
+          inputDelay: 1000,
```
PigeonBar:
```suggestion
```
This is the text box for displaying usernames for password resets, it probably doesn't need a delay

PigeonBar (approved):

DayKev (approved):

src/ui/starter-select-ui-handler.ts
```diff
@@ -1601,87 +1605,94 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
...
+              moves: MoveId[],
+              selectHandler: (moveId: MoveId, index: number, currentMoveId?: MoveId, currentIndex?: number) => boolean,
+              cancelHandler: () => boolean,
+              currentMoveId?: MoveId,
+              currentIndex?: number,
+            ): OptionSelectModeConfig => {
+              const options: OptionSelectItem[] = moves.map((moveId: MoveId, index: number): OptionSelectItem => {
+                return {
+                  label: allMoves[moveId].name,
+                  handler: () => selectHandler(moveId, index, currentMoveId, currentIndex),
```
MokaStitcher, line 1621:
```suggestion
                  handler: () => return selectHandler(moveId, index, currentMoveId, currentIndex),
```
Think I forgot that. Will double check

torranx: you dont need `return` in this case

DayKev: The `return` is implied because there's no `{}`.

PR #714, opened 2025-02-19
[Refactor][Bug] Add `canSetWeather` utility + Fix Chilly Reception

### LINKED ISSUE #710: [Bug] Chilly Reception Replaces Legendary Weather (Primordial Sea, Desolate Land, Delta Stream)
### Describe the bug

On Usage, Chilly Reception replaces the weather (Harsh Sun, Heavy Rain, Heavy Winds) set by the Primal Weather Abilities.

### Reproduction

Use Chilly Reception while a Legendary weather is active.

### Expected behavior

Sandstorm and other weather-setting fails while Legendary Weather is active.

https://github.com/user-attachments/assets/da130bd1-6ab5-4bb3-be3c-67422135379d

> heavy rain causes the weather-modifying moves Sunny Day, Rain Dance, Sandstorm, and Hail to fail if used, and causes the Abilities Drought, Drizzle, Sand Stream, and Snow Warning to fail to activate.
-Bulba

Chilly Reception should still switch the Pokemon despite the weather not being able to be changed, unlike Parting Shot.

https://replay.pokemonshowdown.com/gen9customgame-2303972860

### Screenshots / Videos

Chilly and Delta Stream

https://github.com/user-attachments/assets/1f9b72da-4e64-424f-a94e-335593875389

Chilly and Desolate Land

https://github.com/user-attachments/assets/f7764b84-8837-4e24-adce-481512e7f659

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

frutescens:

Fixes #710 

The game would benefit from a `canSetWeather` function

Created a `canSetWeather` function that will check if a new weather can. be set on the field before overwriting the current weather in `trySetWeather`
Changed `isImmutable()` to `isPrimal()` because primal weather can be overwritten by other primal weather, so it is not 'immutable' and `isPrimal()` clarifies the weather hierarchy better. 
Created a 'PRIMAL_WEATHER' constant that is used for reference in `canSetWeather()` and `isPrimal()` 
Cleaned up `isDamaging()` to match `isPrimal()` 

## How to test the changes?

Chilly Reception should no longer overwrite the 'immutable' weathers like Strong Winds/Harsh Sun/Heavy Rain

## Checklist

### DIFF
--- a/src/data/move-attrs/weather-change-attr.ts
+++ b/src/data/move-attrs/weather-change-attr.ts
@@ -23,8 +23,6 @@ export class WeatherChangeAttr extends MoveEffectAttr {
   }
 
   override getCondition(): MoveConditionFunc {
-    return (_user, _target, _move) =>
-      !globalScene.arena.weather
-      || (globalScene.arena.weather.weatherType !== this.weatherType && !globalScene.arena.weather.isImmutable());
+    return (_user, _target, _move) => globalScene.arena.canSetWeather(this.weatherType);
   }
 }
--- a/src/data/weather.ts
+++ b/src/data/weather.ts
@@ -58,8 +58,7 @@ export class Weather {
    * @returns true for sandstorm or hail, false otherwise
    */
   isDamaging(): boolean {
-    const DAMAGING_WEATHER = Object.freeze([WeatherType.SANDSTORM, WeatherType.HAIL]);
-    return DAMAGING_WEATHER.includes(this.weatherType);
+    return [WeatherType.SANDSTORM, WeatherType.HAIL].includes(this.weatherType);
   }
 
   /**
--- a/src/data/weather.ts
+++ b/src/data/weather.ts
@@ -27,7 +27,7 @@ export class Weather {
 
   constructor(weatherType: WeatherType, turnsLeft?: number) {
     this.weatherType = weatherType;
-    this.turnsLeft = !this.isImmutable() ? turnsLeft || 0 : 0;
+    this.turnsLeft = !this.isPrimal() ? turnsLeft || 0 : 0;
   }
 
   /**
@@ -35,7 +35,7 @@ export class Weather {
    * @returns false if turnsLeft is set to 0. True otherwise
    */
   lapse(): boolean {
-    if (this.isImmutable()) {
+    if (this.isPrimal()) {
       return true;
     }
     if (this.turnsLeft) {
@@ -49,7 +49,7 @@ export class Weather {
    * Checks if the weather is immutable (heavy rain, harsh sun, or strong winds)
    * @returns true if {@linkcode WeatherType} is immutable, false otherwise
    */
-  isImmutable(): boolean {
+  isPrimal(): boolean {
     return PRIMAL_WEATHER.includes(this.weatherType);
   }
 
@@ -149,7 +149,7 @@ export class Weather {
           ? pokemon.getPassiveAbility().getAttrs<SuppressWeatherEffectAbAttr>(AbAttrFlag.SUPPRESS_WEATHER_EFFECT)[0]
           : null;
       }
-      if (suppressWeatherEffectAbAttr && (!this.isImmutable() || suppressWeatherEffectAbAttr.affectsImmutable)) {
+      if (suppressWeatherEffectAbAttr && (!this.isPrimal() || suppressWeatherEffectAbAttr.affectsPrimal)) {
         return true;
       }
     }
--- a/src/data/weather.ts
+++ b/src/data/weather.ts
@@ -12,6 +12,11 @@ import { WeatherType } from "#enums/weather-type";
 import { AbAttrFlag } from "#enums/ab-attr-flag";
 
 /**
+ * Weather types that are associated with the primal forms of the Generation III cover legendaries and cannot be overwritten by weaker weather types
+ */
+export const PRIMAL_WEATHER = Object.freeze([WeatherType.HARSH_SUN, WeatherType.HEAVY_RAIN, WeatherType.STRONG_WINDS]);
+
+/**
  * Class representing Weather effects
  * @var weatherType - The {@linkcode WeatherType} that is being represented
  * @var turnsLeft - How many turns the weather still has left (0 if immutable)
@@ -45,14 +50,7 @@ export class Weather {
    * @returns true if {@linkcode WeatherType} is immutable, false otherwise
    */
   isImmutable(): boolean {
-    switch (this.weatherType) {
-      case WeatherType.HEAVY_RAIN:
-      case WeatherType.HARSH_SUN:
-      case WeatherType.STRONG_WINDS:
-        return true;
-    }
-
-    return false;
+    return PRIMAL_WEATHER.includes(this.weatherType);
   }
 
   /**
@@ -60,13 +58,8 @@ export class Weather {
    * @returns true for sandstorm or hail, false otherwise
    */
   isDamaging(): boolean {
-    switch (this.weatherType) {
-      case WeatherType.SANDSTORM:
-      case WeatherType.HAIL:
-        return true;
-    }
-
-    return false;
+    const DAMAGING_WEATHER = Object.freeze([WeatherType.SANDSTORM, WeatherType.HAIL]);
+    return DAMAGING_WEATHER.includes(this.weatherType);
   }
 
   /**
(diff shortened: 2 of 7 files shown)


src/data/move-attrs/weather-change-attr.ts
```diff
@@ -23,8 +23,6 @@ export class WeatherChangeAttr extends MoveEffectAttr {
   }
 
   override getCondition(): MoveConditionFunc {
-    return (_user, _target, _move) =>
-      !globalScene.arena.weather
-      || (globalScene.arena.weather.weatherType !== this.weatherType && !globalScene.arena.weather.isImmutable());
+    return (_user, _target, _move) => globalScene.arena.canSetWeather(this.weatherType);
```
Tempo-anon, line 26: This would make the entire move fail if the weather can't be changed right? The Pokemon using chilly reception should still attempt to switch out even if the weather changing fails. i.e. If the current weather is snow using chilly reception still switches the user out

src/field/arena.ts
```diff
@@ -406,7 +434,7 @@ export class Arena {
       return this.trySetWeatherOverride(Overrides.WEATHER_OVERRIDE);
     }
 
-    if (this.weather?.weatherType === (weather || undefined)) {
+    if (!this.canSetWeather(weather)) {
```
Tempo-anon: nit: something about `if (!...)` just bothers me but I don't know how other people feel about it or if `if (this.canSetWeather(weather) === false)` is too much

Tempo-anon (dismissed):
Looks like some tests are failing.

src/data/move-attrs/weather-change-attr.ts
frutescens: This isn't Chilly Reception's attr. It's Sunny Day/Rain Dance/etc's move attribute.
Chilly Reception's is `ChillyReceptionAttr`

Tempo-anon: Oh whoops, my bad

src/field/arena.ts
frutescens: Think that should be a larger question since it is used commonly across the codebase + would require a change across the codebase

src/field/arena.ts
```diff
@@ -395,6 +395,38 @@ export class Arena {
...
   }
 
+  /**
+   * Helper function that checks if it is possible to set a new weather given current weather conditions
+   * @param newWeather the weather that the game is attempting to set
+   * @returns `true` if the weather can be set given current conditions | `false` if not
+   */
+  canSetWeather(newWeather: WeatherType): boolean {
+    if (this.weather) {
+      // Checks if the new weather is identical to the current weather
```
torranx: I don't think the comments on top of the `if` blocks add any value since the conditions are already straightforward

src/field/arena.ts
```diff
@@ -395,6 +395,38 @@ export class Arena {
...
+          newWeather === WeatherType.HARSH_SUN
+          || newWeather === WeatherType.HEAVY_RAIN
+          || newWeather === WeatherType.STRONG_WINDS
+          || newWeather === WeatherType.NONE
+        ) {
+          return true;
+        } else {
+          return false;
+        }
+      }
```
torranx: the 2 `if` statements can be merged, and using array.includes might be cleaner here

src/field/arena.ts
```diff
@@ -395,6 +395,38 @@ export class Arena {
...
+      // Checks if there is no current weather and the game is attempting to set the weather to none
+    } else if (isNullOrUndefined(this.weather) && newWeather === WeatherType.NONE) {
```
torranx: nit: you can remove the `if-else if` conditions by returning `false` early given the conditions here. much more readable that way imo

src/field/arena.ts
torranx: I'd still go with `if (!...)` considering the function returns boolean

DayKev: > I'd still go with `if (!...)` considering the function returns boolean

Agreed. +1 from Moka as well.
image

src/field/arena.ts
```diff
@@ -395,6 +395,38 @@ export class Arena {
...
+      return false;
+    }
+    return true;
+  }
```
```suggestion
   * @param weather - The weather that the game is attempting to set
   * @returns `true` if the weather can be set given current conditions | `false` if not
   */
  public canSetWeather(weather: WeatherType): boolean {
    if (this.weather) {
      if (weather === this.weather.weatherType) {
        return false;
      }
      if (this.weather.isImmutable()) {
        if ([WeatherType.NONE, ...primalWeathers].includes(weather)) {
          return true;
        }
        return false;
      }
    } else if (weather === WeatherType.NONE) {
      return false;
    }
    return true;
  }
```

DayKev: As part of this suggestion you should also add this to `constants.ts` (and import it into `arena.ts`):
```ts
export const primalWeathers = Object.freeze([WeatherType.HARSH_SUN, WeatherType.HEAVY_RAIN, WeatherType.STRONG_WINDS]);
```

DayKev: This also lets you change `Weather.isImmutable()` to `return primalWeathers.includes(this.weatherType);`

src/field/arena.ts
```diff
@@ -395,6 +395,25 @@ export class Arena {
     return true;
   }
 
+  /**
+   * Helper function that checks if it is possible to set a new weather given current weather conditions
+   * @param weather - The weather that the game is attempting to set
+   * @returns `true` if the weather can be set given current conditions | `false` if not
+   */
+  public canSetWeather(weather: WeatherType): boolean {
```
torranx: did you mean to remove the `isNullOrUndefined` check?

frutescens: See Kev's comment here https://github.com/Despair-Games/poketernity/pull/714#discussion_r1962840474

DayKev: That check is implied by the first if failing.

src/data/weather.ts
```diff
@@ -22,15 +27,15 @@ export class Weather {
 
   constructor(weatherType: WeatherType, turnsLeft?: number) {
```
DayKev:
```suggestion
  constructor(weatherType: WeatherType, turnsLeft: number = 0) {
```
Huh. Well, since your changes revealed this might as well fix it lol.

src/data/weather.ts
```diff
@@ -22,15 +27,15 @@ export class Weather {
...
     this.weatherType = weatherType;
-    this.turnsLeft = !this.isImmutable() ? turnsLeft || 0 : 0;
+    this.turnsLeft = !this.isPrimal() ? turnsLeft || 0 : 0;
```
DayKev:
```suggestion
    this.turnsLeft = !this.isPrimal() ? turnsLeft : 0;
```

src/field/arena.ts
torranx:
```
else if (isNullOrUndefined(this.weather) && newWeather === WeatherType.NONE) {
      return false;
    }
```
before it returns `false`, now if `this.weather` is falsy, it defaults to return `true`

src/field/arena.ts
```diff
@@ -395,6 +395,25 @@ export class Arena {
...
+    if (this.weather) {
+      if (weather === this.weather.weatherType) {
+        return false;
+      }
+      if (this.weather.isPrimal() && [WeatherType.NONE, ...PRIMAL_WEATHER].includes(weather)) {
+        return true;
+      }
```
DayKev: You can't do it this way because that lets any weather override primal weather.

src/field/arena.ts
frutescens: I don't think so because then if `this.weather` is falsy, it then checks if the `newWeather` is `WeatherType.NONE` first before defaulting to `true`

torranx: yes but it doesnt check if the `weather` is undefined or null, and will return `true` by default
or am I missing something here

frutescens: if `this.weather` is undefined or null, it's checked in `if (this.weather)` + it should return `true` by default

DayKev:
```ts
  public canSetWeather(newWeather: WeatherType): boolean {
    if (this.weather) {
      // ...
    }
    // this.weather must be falsy now, because the above if
    // did not evaluate this.weather as truthy, and we've reached
    // the else-if block
    else if (newWeather === WeatherType.NONE) {
      return false;
    }
    return true;
  }
```

DayKev: The `else if` should never run if `this.weather` is truthy, either, afaik.

DayKev:
Looks good, gonna add a test for the primal weathers and then approve. 👍

DayKev (approved):

torranx (approved):

PR #717, opened 2025-02-20
[Misc] Set version to `0.1.0`, add version update commands

DayKev:
N/A

The version number was incorrect.
Updating the version number and the locales submodule will now be easier/more convenient.

- New commands `update-version:minor` and `update-version:patch` are added to update the game version (version is in the format `MAJOR.MINOR.PATCH`).
- New command `update-locales:remote` is added to update the locales submodule to point to the latest remote commit.
- Deprecated data fields removed from `PokemonData`.
- Obsolete save data migrators deleted.
- Template added for creating save data migrators.
- Module `compare-versions` installed for improved version comparison checks.
- Existing save migration code disabled.

## Checklist


MokaStitcher:
I agree with this, but I think we should remove the outdated save migration code at the same time
There are also attributes in Pokemon and PokemonData that were kept only for save migration purposes and should be removed
There are also legacy fields in eggs that I have marked with todos

src/system/version_migration/version_converter.ts
```diff
@@ -21,12 +33,12 @@ const LATEST_VERSION = version.split(".").map((value) => parseInt(value));
  * @see {@link SystemVersionConverter}
  */
 export function applySystemVersionMigration(data: SystemSaveData) {
-  const curVersion = data.gameVersion.split(".").map((value) => parseInt(value));
+  const prevVersion = data.gameVersion;
 
-  if (!curVersion.every((value, index) => value === LATEST_VERSION[index])) {
+  if (compareVersions(prevVersion, version) === -1) {
```
MokaStitcher:
```suggestion
  if (compareVersions(prevVersion, version) < 0) {
```
I think I prefer this?

src/system/version_migration/version_converter.ts
```diff
@@ -41,12 +53,12 @@ export function applySystemVersionMigration(data: SystemSaveData) {
  * @see {@link SessionVersionConverter}
  */
 export function applySessionVersionMigration(data: SessionSaveData) {
-  const curVersion = data.gameVersion.split(".").map((value) => parseInt(value));
+  const prevVersion = data.gameVersion;
 
-  if (!curVersion.every((value, index) => value === LATEST_VERSION[index])) {
+  if (compareVersions(prevVersion, version) === -1) {
```
MokaStitcher:
```suggestion
  if (compareVersions(prevVersion, version) < 0) {
```

src/system/version_migration/version_converter.ts
```diff
@@ -61,13 +73,12 @@ export function applySessionVersionMigration(data: SessionSaveData) {
  * @see {@link SettingsVersionConverter}
  */
 export function applySettingsVersionMigration(data: Object) {
-  const gameVersion: string = data.hasOwnProperty("meta") ? (data["meta"]["gameVersion"] ?? "1.0.0") : "1.0.0";
-  const curVersion = gameVersion.split(".").map((value) => parseInt(value));
+  const prevVersion: string = data.hasOwnProperty("meta") ? (data["meta"]["gameVersion"] ?? "1.0.0") : "1.0.0";
```
MokaStitcher: This 1.0.0 default would mess things up since we are not yet at 1.0, right?

src/system/version_migration/version_converter.ts
```diff
@@ -61,13 +73,12 @@ export function applySessionVersionMigration(data: SessionSaveData) {
...
 
-  if (!curVersion.every((value, index) => value === LATEST_VERSION[index])) {
+  if (compareVersions(prevVersion, version) === -1) {
```
MokaStitcher:
```suggestion
  if (compareVersions(prevVersion, version) < 0) {
```

MokaStitcher (commented):
will take a closer look later to check if more things should be removed

src/system/version_migration/version_converter.ts
DayKev: The compare function can only return `1 | 0 | -1`, where `1` is "the first version is higher", `0` is "the versions are equal", and `-1` is "the second version is higher". It works the same as various built-in JS/TS functions for comparison.

src/system/version_migration/version_converter.ts
DayKev: Oh yeah maybe that should probably be removed, all settings should be assigned a version now right? I think this was for old saves where the settings weren't versioned or something.

package.json
```diff
@@ -22,7 +22,10 @@
     "docs": "typedoc",
     "depcruise": "depcruise src",
     "depcruise:graph": "depcruise src --output-type dot | node dependency-graph.js > dependency-graph.svg",
-    "postinstall": "npx lefthook install && npx lefthook run post-merge"
+    "postinstall": "npx lefthook install && npx lefthook run post-merge",
+    "update-version:patch": "npm version patch --force --no-git-tag-version",
+    "update-version:minor": "npm version minor --force --no-git-tag-version",
```
torranx, line 27: nit: to make this shorter, maybe call it `bump:{n}`?

src/system/version_migration/version_converter.ts
torranx: the `1.0.0`/default version could be made into constant too

package.json
DayKev: I'd rather the command be clear in what it does.

DayKev: Also since these commands won't be run very often (only when there is a game version update) it's okay that they're a bit longer imo.

src/system/version_migration/version_converter.ts
f-fsantos: can you assign this to a variable with an appropriate name for context and use that variable in the condition?

```suggestion
  var isCurrentVersionHigher = compareVersions(prevVersion, version) === -1;
  if (isCurrentVersionHigher) {
```

MokaStitcher: For me `<0` / `>0` is easier to understand than `=== (-)1`, but using a const is actually much clearer

src/system/version_migration/version_converter.ts
MokaStitcher: settings use the same version as the game now

src/system/version_migration/version_converter.ts
DayKev: I looked at the other functions the module provides and there's also `compare(versionA, versionB, comparisonOperator)` which lets you do a specific comparison (ie: `compare(versionA, versionB, "<")` would return `true` if `versionA` is less than `versionB`). Would that be better or should I go with the const?

MokaStitcher: That' seems fine yeah, I don't have any strong preference for one or the other

src/system/version_migration/version_converter.ts
```diff
@@ -176,22 +161,7 @@ class SystemVersionConverter extends VersionConverter {
...
-          this.callMigrators(data, v1_0_4.settingsMigrators);
-        }
-      }
-      if (curMinor <= 1) {
-        console.log("Applying v1.1.0 settings data migration!");
-        this.callMigrators(data, v1_1_0.settingsMigrators);
-      }
-    }
-
+  override applyMigration(_data: Object, _curVersion: string): void {
```
PigeonBar:
```suggestion
  override applyMigration(_data: Object, _prevVersion: string): void {
```

src/system/version_migration/version_converter.ts
```diff
@@ -150,22 +150,7 @@ class SessionVersionConverter extends VersionConverter {
...
-          this.callMigrators(data, v1_0_4.systemMigrators);
-        }
-      }
-      if (curMinor <= 1) {
-        console.log("Applying v1.1.0 system data migration!");
-        this.callMigrators(data, v1_1_0.systemMigrators);
-      }
-    }
-
+  override applyMigration(_data: SystemSaveData, _curVersion: string): void {
```
PigeonBar:
```suggestion
  override applyMigration(_data: SystemSaveData, _prevVersion: string): void {
```

src/system/version_migration/version_converter.ts
```diff
@@ -124,22 +139,7 @@ class SessionVersionConverter extends VersionConverter {
...
-          this.callMigrators(data, v1_0_4.sessionMigrators);
-        }
-      }
-      if (curMinor <= 1) {
-        console.log("Applying v1.1.0 session data migration!");
-        this.callMigrators(data, v1_1_0.sessionMigrators);
-      }
-    }
-
+  override applyMigration(_data: SessionSaveData, _curVersion: string): void {
```
PigeonBar:
```suggestion
  override applyMigration(_data: SessionSaveData, _prevVersion: string): void {
```

src/system/version_migration/version_converter.ts
```diff
@@ -110,7 +125,7 @@ abstract class VersionConverter {
    *                   [1] Current minor version
    *                   [2] Current patch version
    */
-  abstract applyMigration(data: any, curVersion: number[]): void;
+  abstract applyMigration(data: any, curVersion: string): void;
```
PigeonBar:
```suggestion
  abstract applyMigration(data: any, prevVersion: string): void;
```

src/system/version_migration/version_converter.ts
```diff
@@ -61,13 +76,13 @@ export function applySessionVersionMigration(data: SessionSaveData) {
  * @see {@link SettingsVersionConverter}
  */
 export function applySettingsVersionMigration(data: Object) {
-  const gameVersion: string = data.hasOwnProperty("meta") ? (data["meta"]["gameVersion"] ?? "1.0.0") : "1.0.0";
-  const curVersion = gameVersion.split(".").map((value) => parseInt(value));
+  const prevVersion: string = settings.meta.gameVersion;
+  const isCurrentVersionHigher = compareVersions(prevVersion, version) === -1;
```
PigeonBar:
```suggestion
  const prevVersion = data.meta?.gameVersion;
  const isCurrentVersionHigher = prevVersion && compareVersions(prevVersion, version) === -1;
```

src/system/version_migration/version_converter.ts
```diff
@@ -61,13 +76,13 @@ export function applySessionVersionMigration(data: SessionSaveData) {
  * @see {@link SettingsVersionConverter}
  */
 export function applySettingsVersionMigration(data: Object) {
```
PigeonBar:
```suggestion
export function applySettingsVersionMigration(data: Partial<Settings>) {
```

PigeonBar:
Should we also un-comment this line?
https://github.com/Despair-Games/poketernity/blob/4de712ab775161c0bf529a65fb31c3e4c4f99fac/src/system/settings/settings-manager.ts#L164-L165

PigeonBar:
Also, pkr has sometimes forgotten to update the version number during hotfixes. Should we add a reminder in the PR template checklist (or somewhere else)?

DayKev:
> Should we also un-comment this line?
> 
> https://github.com/Despair-Games/poketernity/blob/4de712ab775161c0bf529a65fb31c3e4c4f99fac/src/system/settings/settings-manager.ts#L164-L165

Possibly, though it looks like it needs to be changed a bit since `lsSettings` is of type `Partial<UserFacingSettings>` and not `Partial<Settings>`, dunno if just changing the type hint is enough; @flx-sta would have to weigh in on that since he made the `SettingsManager`.

DayKev:
> Also, pkr has sometimes forgotten to update the version number during hotfixes. Should we add a reminder in the PR template checklist (or somewhere else)?

Good point, adding an item into the checklist would probably be good.

PigeonBar (dismissed):

MokaStitcher:
> > Should we also un-comment this line?
> > https://github.com/Despair-Games/poketernity/blob/4de712ab775161c0bf529a65fb31c3e4c4f99fac/src/system/settings/settings-manager.ts#L164-L165
> 
> Possibly, though it looks like it needs to be changed a bit since `lsSettings` is of type `Partial<UserFacingSettings>` and not `Partial<Settings>`, dunno if just changing the type hint is enough; @flx-sta would have to weigh in on that since he made the `SettingsManager`.

I was the one who added this and had to comment it out because one of the system data converter in the same file as the settings converter was causing things to be loaded out of order because of circular dependencies.
I looked at it quickly though and I think we'll need to make a separate PR to make it work properly since the version is not stored the same way in settings now. I'll be making an issue so it's not forgotten

PigeonBar (dismissed):

.github/pull_request_template.md
```diff
@@ -66,7 +66,8 @@ Do the reviewers need to do something special in order to test your changes?
 
 ## Checklist
 
-- [ ] **I'm using `beta` as my base branch**
+- [ ] ⚠️ If this is a PR for `main` (such as a hotfix), has the game version been updated?
```
torranx:
```suggestion
- [ ] ⚠️ If this is a PR for `main` (such as a hotfix), has the game version been updated (`npm run update-version:patch` / `npm run update version:minor`?
```

torranx (approved):

PR #726, opened 2025-02-22
[Misc] Update logging in `EncounterPhase`

DayKev:
N/A

Additional logging when playing on beta or locally for use in development.

At the start of an encounter, more of the enemy pokemon's information will be output to the console.

## Screenshots/Videos
<details><summary>An example team from wave 190</summary>
<p>

image

</p>
</details> 

## How to test the changes?
Start any wave and look at the console output.

## Checklist

### DIFF
--- a/src/phases/encounter-phase.ts
+++ b/src/phases/encounter-phase.ts
@@ -5,22 +5,16 @@ import { type NextEncounterPhase } from "#app/phases/next-encounter-phase";
 /* eslint-enable @typescript-eslint/no-unused-vars */
 // -- end tsdoc imports --
 
-import { BattlerIndex } from "#enums/battler-index";
-import { BattleType } from "#enums/battle-type";
-import { PLAYER_PARTY_MAX_SIZE } from "#app/constants";
+import { ME_WEIGHT_INCREMENT_ON_SPAWN_MISS, PLAYER_PARTY_MAX_SIZE } from "#app/constants";
 import { applyAbAttrs } from "#app/data/apply-ab-attrs";
-import { loadEncounterAnimAssets } from "#app/utils/anim-utils";
-import { initEncounterAnims } from "#app/data/init-encounter-anims";
 import { getCharVariantFromDialogue } from "#app/data/dialogue";
-import { ME_WEIGHT_INCREMENT_ON_SPAWN_MISS } from "#app/constants";
+import { initEncounterAnims } from "#app/data/init-encounter-anims";
 import { getEncounterText } from "#app/data/mystery-encounters/utils/encounter-dialogue-utils";
 import { doTrainerExclamation } from "#app/data/mystery-encounters/utils/encounter-phase-utils";
 import { getGoldenBugNetSpecies } from "#app/data/mystery-encounters/utils/encounter-pokemon-utils";
-import { TrainerSlot } from "#enums/trainer-slot";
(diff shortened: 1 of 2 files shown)


src/phases/encounter-phase.ts
```diff
@@ -195,13 +195,31 @@ export class EncounterPhase extends BattlePhase {
...
+      enemyPokemon.getMoveset().forEach((move) => {
+        moveset.push(move!.getName());
+      });
+
       console.log(
         `Pokemon: ${getPokemonNameWithAffix(enemyPokemon)}`,
-        `Species ID: ${enemyPokemon.species.speciesId}`,
-        `Stats: ${enemyPokemon.stats}`,
+        `| Species ID: ${enemyPokemon.species.speciesId}`,
+        `| Nature: ${Nature[enemyPokemon.getNature()]}`,
```
Tempo-anon: nit: If this would also log the +/- of the nature that would be even better

src/phases/encounter-phase.ts
```diff
@@ -195,13 +195,31 @@ export class EncounterPhase extends BattlePhase {
...
+      );
+      console.log(`Stats (IVs): ${stats}`);
+      console.log(
         `Ability: ${enemyPokemon.getAbility().name}`,
-        `Passive Ability: ${enemyPokemon.getPassiveAbility().name}`,
+        `| Passive Ability${enemyPokemon.isBoss() ? "" : " (inactive)"}: ${enemyPokemon.getPassiveAbility().name}`,
```
Tempo-anon: Not sure if `isBoss` is the best check for passive since they're not intrinsically tied. i.e. Eternatus is hardcoded to not have its passive despite being a boss and this might be something we want to change in the future. Pokemon has a `  public passive: boolean;` that can be used instead

Tempo-anon (changes requested):
More logging is always good but the passive check isn't quite accurate. Happy to approve once that's changed

src/phases/encounter-phase.ts
DayKev: Ah, good point.

DayKev: Checking for that doesn't work:
image

DayKev: Okay it seems I can use `.hasPassive()` instead (which is mostly a check for `.isBoss` lol).

Tempo-anon: oh what the heck :pikastare:

DayKev:
Updated PR description with new screenshot.

Tempo-anon (dismissed):

PigeonBar:
There is duplicate code for ME's that also needs to be updated :(
https://github.com/Despair-Games/poketernity/blob/80c0f5f6ee997b662542302249d0488da84bc23a/src/data/mystery-encounters/utils/encounter-phase-utils.ts#L396-L402

DayKev:
> There is duplicate code for ME's that also needs to be updated :(

TBH the ME code desperately needs to be deduplicated. I guess for now I can copy-paste this there.

DayKev:
<details><summary>Expert Pokemon Breeder ME</summary>
<p>

image

</p>
</details>

PigeonBar (approved):

Tempo-anon (approved):

PR #727, opened 2025-02-22
[Test] The sprite test will now output useful error messages on fail

DayKev:
N/A

The sprite check test didn't give helpful error messages when something was wrong, now it will.

Changed the `expect()`s to go from checking the length of the error message array to checking that it equals an empty array, which makes it output the error message array to the console if it's not empty.

## How to test the changes?
`npm run test pokemonSprite` after breaking something

## Checklist


Tempo-anon (approved):
Should include in the PR desc an example of what before/after looks like

MokaStitcher (approved):
ooh, hadn't thought I that. I guess I could change my own tests checking length to do something similar 🤔

PR #728, opened 2025-02-22
[Balance] Multi-hit moves now use gen 5+ behavior

DayKev:
Changed behavior of multi-hit moves so that the odds are 35/35/15/15 for 2/3/4/5 hits (gen 5+) instead of 37.5/37.5/12.5/12.5 for 2/3/4/5 hits (gens 1-4).
cf https://bulbapedia.bulbagarden.net/wiki/Multistrike_move#Variable_number_of_strikes

Multi-hit moves don't need to be that bad.

The game now rolls a d20 and assigns the number of hits as follows:

| Hits | RNG rolls | Chance | %  |
|------|-----------|--------|----|
| 2    | 13-19     | 7/20   | 35 |
| 3    | 6-12      | 7/20   | 35 |
| 4    | 3-5       | 3/20   | 15 |
| 5    | 0-2       | 3/20   | 15 |

## Checklist

### DIFF
--- a/src/data/move-attrs/multi-hit-attr.ts
+++ b/src/data/move-attrs/multi-hit-attr.ts
@@ -1,12 +1,12 @@
-import { MultiHitType } from "#enums/multi-hit-type";
-import type { Pokemon } from "#app/field/pokemon";
-import { NumberHolder } from "#app/utils";
 import { applyAbAttrs } from "#app/data/apply-ab-attrs";
 import { type Move } from "#app/data/move";
-import { applyMoveAttrs } from "#app/utils/move-utils";
 import { ChangeMultiHitTypeAttr } from "#app/data/move-attrs/change-multi-hit-type-attr";
 import { MoveAttr } from "#app/data/move-attrs/move-attr";
+import type { Pokemon } from "#app/field/pokemon";
+import { NumberHolder } from "#app/utils";
+import { applyMoveAttrs } from "#app/utils/move-utils";
 import { AbAttrFlag } from "#enums/ab-attr-flag";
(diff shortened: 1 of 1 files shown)


innerthunder (dismissed):

Tempo-anon (dismissed):

src/data/move-attrs/multi-hit-attr.ts
```diff
@@ -65,14 +65,14 @@ export class MultiHitAttr extends MoveAttr {
   getHitCount(user: Pokemon, _target: Pokemon): number {
```
MokaStitcher, line 65: > odds are 35/35/15/15 for 2/3/4/5

The check being done with 20 doesn't make it easy to parse, can we put these number in a comment or the tsDoc?

Tempo-anon: I was going to suggest this but I don't know if it would make sense for this function since this handles all multi hit checks and the 35/35/15/15 only applies for the 2-5 moves. (Maybe move the 2-5 thing into its own func?)

Anyway here's the chart if it would help
```
| Hits | RNG rolls | Chance | %  |
|------|-----------|--------|----|
| 2    | 13-19     | 7/20   | 35 |
| 3    | 6-12      | 7/20   | 35 |
| 4    | 3-5       | 3/20   | 15 |
| 5    | 0-2       | 3/20   | 15 |
```

DayKev: Added the chart to the attr.

Tempo-anon (approved):

PR #469, opened 2025-02-10
[Refactor] Remove fusions from the game

DayKev:
## What are the changes the user will see?
Fusions are removed from the game, as well as everything related to them.

## Why am I making these changes?
Design decision.

## What are the changes from a developer perspective?
All fusion-related code is deleted.

## How to test the changes?
Uh... play the game and see that it still works?

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test:silent`)

PR #732, opened 2025-02-22
Small cleanup of pokemon.evolve

MokaStitcher:
For #469 

- Fix merge conflicts
- Remove preEvolution parameter from Pokemon.evolve (did not bother with adding tsdoc since I'm doing it in another PR)
- Replace the outdated ability count check with proper check for the absence of HA for the evolved Pokemon

whoops, with the merge conflicts fix it's showing 100 changes 😭 
just check the individual commits I guess x')

image


src/field/pokemon.ts
```diff
@@ -4220,16 +4222,17 @@ export class PlayerPokemon extends Pokemon {
         this.formIndex = formIndex;
       }
       this.generateName();
-      const abilityCount = this.getSpeciesForm().getAbilityCount();
-      const preEvoAbilityCount = preEvolution.getAbilityCount();
       if ([0, 1, 2].includes(this.abilityIndex)) {
-        // Handles cases where a Pokemon with 3 abilities evolves into a Pokemon with 2 abilities (ie: Eevee -> any Eeveelution)
-        if (this.abilityIndex === 2 && preEvoAbilityCount === 3 && abilityCount === 2) {
-          this.abilityIndex = 1;
+        // Handles cases where a Pokemon with HA evolves into a Pokemon with no HA
```
Tempo-anon, line 4226: Are there any cases still like this? Ideally I think we should just get rid of all of them. I know Ferroseed was given a HA to account for this

src/field/pokemon.ts
```diff
@@ -4220,16 +4222,17 @@ export class PlayerPokemon extends Pokemon {
...
+          console.warn(
+            `${preEvolutionSpecies.getName()} with HA evolved into a Pokemon without HA, please report this.`,
+          );
+          this.abilityIndex = 0;
         }
       } else {
         // Prevent pokemon with an illegal ability value from breaking things
-        console.warn("this.abilityIndex is somehow an illegal value, please report this");
+        console.warn("this.abilityIndex is somehow an illegal value, please report this.");
         console.warn(this.abilityIndex);
```
Tempo-anon: Can these two console messages be combined into a single one so the output reads like `"Pokemon species Charizard's abilityIndex is somehow 5, an illegal value. Please report this."` Feel like logging the pokemon species would also be helpful

src/field/pokemon.ts
```diff
@@ -4220,16 +4222,17 @@ export class PlayerPokemon extends Pokemon {
         this.formIndex = formIndex;
       }
       this.generateName();
-      const abilityCount = this.getSpeciesForm().getAbilityCount();
-      const preEvoAbilityCount = preEvolution.getAbilityCount();
       if ([0, 1, 2].includes(this.abilityIndex)) {
-        // Handles cases where a Pokemon with 3 abilities evolves into a Pokemon with 2 abilities (ie: Eevee -> any Eeveelution)
-        if (this.abilityIndex === 2 && preEvoAbilityCount === 3 && abilityCount === 2) {
-          this.abilityIndex = 1;
```
Tempo-anon, line 4178: ugh abilityIndexes as a whole are confounding with the way sometimes abilityIndex 1 refers to the HA and sometimes it doesn't...

Tempo-anon (approved):
Not actually that much to review when looking at the specific commits, just some comments and a nitpick about the logging

torranx (approved):

src/field/pokemon.ts
DayKev: Iirc index 2 is always the HA now.

src/field/pokemon.ts
MokaStitcher: Ferrosseed is different, it's a mon without HA evolving into a mon with HA, so we had to give it a HA to be able to have it evolve into its HA version
This check is for Pokemon like Trapinch which has an HA and evolves into mons without HAs
The cases where this would cause an issue is then if that mon could change form or evolve again into a Pokemon with HA. So maybe we want to get rid of them, I'm not sure? If yes I guess it could be part of the data validation that is planned for the pokemon species jsonificatilon

src/field/pokemon.ts
MokaStitcher: Indeed

PR #736, opened 2025-02-23
[Balance] Change OHKO acc formula, go to bosses' next unbroken bar

Tempo-anon:

* G-Max Pokemon are immune to OHKO moves
* The accuracy of OHKO moves increases by 1% for each level the user is above the target
* OHKO moves do damage to the next unbroken hp bar segment to bosses instead of 200 bp
* the OHKO message will only display if it KOs the opponent

We want to move away from specific custom Pokerogue implementations especially since a lot were implemented by Sam for specific interactions (endless)

* Replaced the acc increase of ohko moves from `100 * (1 - target.level / user.level)` to `user.level - target.level`
* Added a `!target.isMax()` check in the one-hit-ko-attr condition
* Created `getBossSegments` abstract method in `Pokemon` so the Pokemon base class can access that information from EnemyPokemon
* Made OHKO moves now do damage to the next unbroken hp bar segment
* Created unit tests

## Screenshots/Videos

image

## How to test the changes?

```ts
const overrides = {
  ENEMY_SPECIES_OVERRIDE:Species.ARCEUS,
  ENEMY_MOVESET_OVERRIDE: [MoveId.SPLASH],
  ENEMY_LEVEL_OVERRIDE:100,
  STARTING_LEVEL_OVERRIDE:200,
  MOVESET_OVERRIDE: [MoveId.GUILLOTINE],
  // ABILITY_OVERRIDE: Abilities.NO_GUARD,
}
```

`npm run test ohko`

## Checklist


test/moves/ohko.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  afterEach(() => {
+    game.phaseInterceptor.restoreOg();
+  });
+
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    game.override
+      .battleType("single")
+      .startingLevel(100)
```
PigeonBar, line 28:
```suggestion
      .startingLevel(100)
      .ability(Abilities.BALL_FETCH)
```
Test is flaky due to occasionally rolling No Guard Machamp.

Tempo-anon: oh lol I was wondering about this, nice catch.

src/data/move-attrs/one-hit-ko-accuracy-attr.ts
```diff
@@ -15,7 +15,7 @@ export class OneHitKOAccuracyAttr extends VariableAccuracyAttr {
     if (user.level < target.level) {
       accuracy.value = 0;
     } else {
-      accuracy.value = Math.min(Math.max(30 + 100 * (1 - target.level / user.level), 0), 100);
+      accuracy.value = Math.min(30 + user.level - target.level, 100);
```
innerthunder, line 18:
```suggestion
      accuracy.value = Math.max(Math.min(move.accuracy + user.level - target.level, 100), 0);
```
A final accuracy of -1 causes the move to always hit, so this also needs a lower bound. This also doesn't assume a base accuracy of 30.

I'm also wondering if there's a way to reduce redundant code for Sheer Cold's attribute...

src/data/move-attrs/sheer-cold-accuracy-attr.ts
```diff
@@ -16,7 +16,7 @@ export class SheerColdAccuracyAttr extends OneHitKOAccuracyAttr {
       accuracy.value = 0;
     } else {
       const baseAccuracy = user.isOfType(ElementalType.ICE) ? 30 : 20;
-      accuracy.value = Math.min(Math.max(baseAccuracy + 100 * (1 - target.level / user.level), 0), 100);
+      accuracy.value = Math.min(baseAccuracy + user.level - target.level, 100);
```
innerthunder, line 20:
```suggestion
      accuracy.value = Math.max(Math.min(baseAccuracy + user.level - target.level, 100), 0);
```

src/data/move-attrs/sheer-cold-accuracy-attr.ts
```diff
@@ -16,7 +16,7 @@ export class SheerColdAccuracyAttr extends OneHitKOAccuracyAttr {
       accuracy.value = 0;
     } else {
       const baseAccuracy = user.isOfType(ElementalType.ICE) ? 30 : 20;
```
innerthunder, line 19: more hardcoded accuracy 😶

innerthunder (dismissed):

src/data/move-attrs/sheer-cold-accuracy-attr.ts
Tempo-anon: Speaking of this, apparently I've read somewhere that despite sheer cold's base acc starting at 20, it still displays in game as 30? :plains:

src/data/move-attrs/one-hit-ko-accuracy-attr.ts
Tempo-anon: This should never be in the negatives because this is in the `else` block so user.level will always be >= target.level

src/data/move-attrs/sheer-cold-accuracy-attr.ts
innerthunder: That seems to have been fixed based on the data in `init-moves`

src/data/move-attrs/one-hit-ko-attr.ts
```diff
@@ -34,7 +31,7 @@ export class OneHitKOAttr extends MoveAttr {
     return (user, target, _move) => {
       const cancelled = new BooleanHolder(false);
       applyAbAttrs(AbAttrFlag.BLOCK_ONE_HIT_KO, target, false, cancelled);
-      return !cancelled.value && user.level >= target.level;
+      return !cancelled.value && user.level >= target.level && !target.isMax();
```
PigeonBar:
```suggestion
      return !cancelled.value && user.level >= target.level && !target.isMax(false);
```
<details><summary>On Showdown, Eternamax does not grant immunity to OHKO moves.</summary>
<p>

image

</p>
</details>

PigeonBar:
If the enemy's health segment is partially damaged, then the OHKO move's damage will spill over to the next health segment. Is this intended?

Also, should the move still say "It's a one-hit KO"? 

https://github.com/user-attachments/assets/0151acfa-0625-41bc-b1a8-800e09ae79ac

test/moves/ohko.test.ts
PigeonBar, line 1: There's already an existing `fissure.test.ts`. We should merge the two test files.

Tempo-anon:
> If the enemy's health segment is partially damaged, then the OHKO move's damage will spill over to the next health segment. Is this intended?

I think aesthetically and theme wise it's a bit weird, but gameplay wise it makes sense. 🤔 

> Also, should the move still say "It's a one-hit KO"?

This should probably only show if the damage done is >= the Pokemon's current HP. Argh p3

PigeonBar:
It turns out that with some precise damage values (Eternatus with 31 IVs has exactly 832 HP, I used a 207-damage Seismic Toss after depleting the second health segment), it is possible for the OHKO damage spillover to faint Eternatus in stage 1, which breaks stage 2:

https://github.com/user-attachments/assets/ed48b0f6-7706-4590-b01e-b22d408e2f14

test/moves/ohko.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    await game.classicMode.startBattle([Species.MACHAMP]);
+    const enemy = game.scene.getEnemyParty()[0];
+    expect(enemy.isMax()).toBe(true);
+
+    game.move.select(MoveId.GUILLOTINE);
+    await game.toNextTurn();
+    expect(enemy.hp).toBe(enemy.getMaxHp());
+  });
+
+  it("OHKO moves should do 1 HP bar for boss Pokemon", async () => {
```
DayKev, line 136: Can you add a test for when a boss bar is broken and then the boss heals up afterwards, to check that it properly reduces the boss's HP to the first unbroken boss bar?

MokaStitcher:
> > If the enemy's health segment is partially damaged, then the OHKO move's damage will spill over to the next health segment. Is this intended?
> 
> I think aesthetically and theme wise it's a bit weird, but gameplay wise it makes sense. 🤔


Personally I think it's weird in both regards. I'd rather have it bring it down to the next bar always. Maybe we should ask everyone's opinion on this

DayKev:
> > > If the enemy's health segment is partially damaged, then the OHKO move's damage will spill over to the next health segment. Is this intended?
> > 
> > 
> > I think aesthetically and theme wise it's a bit weird, but gameplay wise it makes sense. 🤔
> 
> Personally I think it's weird in both regards. I'd rather have it bring it down to the next bar always. Maybe we should ask everyone's opinion on this

Oh is that what it does? It should always deal the exact damage necessary to drop down to the next unbroken boss bar imo.

src/field/pokemon.ts
```diff
@@ -2820,14 +2822,16 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
       };
     }
 
-    // If the attack is a one-hit KO move, return a result with damage equal to this Pokemon's HP
+    /**
+     * If the attack is a one-hit KO move, return a result equal to the Pokemon's HP bar (or one segment if a boss)
```
MokaStitcher: like we've discussed in the comments, I think we'd prefer to have the damage only go down to the next segment for bosses. Feel free to make a poll if you're not sure about that

MokaStitcher (commented):
Just a quick pass over the code. Some comments from other reviews still need to be addressed, otherwise seems fine

src/data/move-attrs/sheer-cold-accuracy-attr.ts
MokaStitcher: If this is not getting fixed in this PR please add a TODO for it

src/field/pokemon.ts
```diff
@@ -4903,17 +4929,12 @@ export class EnemyPokemon extends Pokemon {
...
 
-  getBossSegmentIndex(): number {
-    const segments = (this as EnemyPokemon).bossSegments;
-    const segmentSize = this.getMaxHp() / segments;
-    for (let s = segments - 1; s > 0; s--) {
-      const hpThreshold = Math.round(segmentSize * s);
-      if (this.hp > hpThreshold) {
-        return s;
-      }
-    }
```
Tempo-anon, line 4908: I removed this because this was not being used anywhere. Looking back, I could have probably just used this function if the code was replaced with `return this.bossSegmentIndex`

test/moves/ohko.test.ts
```diff
@@ -0,0 +1,175 @@
...
+  });
+
+  it("OHKO moves should always fail if the opponent is higher level", async () => {
+    game.override.ability(Abilities.NO_GUARD).enemyLevel(101);
+    await game.classicMode.startBattle([Species.ALAKAZAM]);
+    const enemyPokemon = game.field.getEnemyPokemon();
+
+    game.move.select(MoveId.GUILLOTINE);
+    await game.toNextTurn();
+    expect(enemyPokemon.hp).toBe(enemyPokemon.getMaxHp());
```
DayKev, line 109: Iirc there is a `pokemon.isFullHp()` function, though it doesn't really matter here.

DayKev (approved):

PR #739, opened 2025-02-23
[P2] Crits bypass screens

Tempo-anon:
## What are the changes the user will see?

Critical hits ignore the damage reduction from screens

## Why am I making these changes?

Fixes #738

## What are the changes from a developer perspective?

Moved the chunk of applying `WeakenMoveScreenArenaTagTypes` inside a `if (!isCritical)` check

## Screenshots/Videos

Testing Giga Impact without screens (150bp) and comparing to Crabhammer crit with screens (100bp * 1.5)
image
image

0 Atk 18 IVs Typhlosion Giga Impact vs. 0 HP 12 IVs / 0- Def 4 IVs Snorlax: 164-194 (37.1 - 43.8%) -- guaranteed 3HKO
Possible damage amounts: (164, 166, 168, 170, 172, 174, 176, 178, 180, 182, 184, 186, 188, 190, 192, 194)

0 Atk 18 IVs Typhlosion Crabhammer vs. 0 HP 12 IVs / 0- Def 4 IVs Snorlax on a critical hit: 165-195 (37.3 - 44.1%) -- guaranteed 3HKO
Possible damage amounts: (165, 167, 169, 171, 173, 175, 177, 179, 181, 183, 185, 187, 189, 191, 193, 195)

## How to test the changes?

```ts
const overrides = {
  ENEMY_SPECIES_OVERRIDE: Species.SNORLAX,
  ENEMY_LEVEL_OVERRIDE: 100,
  ENEMY_MOVESET_OVERRIDE: [MoveId.REFLECT],
  STARTER_SPECIES_OVERRIDE: Species.TYPHLOSION,
  STARTING_LEVEL_OVERRIDE: 100,
  MOVESET_OVERRIDE: [MoveId.GIGA_IMPACT, MoveId.FOCUS_ENERGY, MoveId.CRABHAMMER]
}
```

## Checklist

- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?

- [ ] Are all unit tests still passing? (`npm run test:silent`)
  - [ ] Have I created new automated tests (`npm run test:create`) or updated existing tests related to the PR's changes?
- [ ] Have I provided screenshots/videos of the changes (if applicable)?
  - [ ] Have I made sure that any UI change works for both UI themes (dark and light)?

Are there any localization additions or changes? If so:

[...]

PR #746, opened 2025-02-23
[Test] Overhaul of Reflect, Light Screen, and Aurora Veil tests

PigeonBar:

These tests needed an overhaul. For example, they did not properly check how much the target actually gets damaged.

1. Consolidated the three test files into a single file.
2. The tests now check how much the target actually gets damaged in battle instead of calling `getMockedMoveDamage()`.
3. Added tests that each screen does not reduce the damage of a critical hit, a fixed-damage move, or confusion damage. Also added tests that Aurora Veil fails if there is no snow or hail, and that Aurora Veil does not stack damage reduction with Reflect or Light Screen.

## How to test the changes?

`npm run test screens`.

## Checklist

- [ ] **I'm using `beta` as my base branch** (Base branch is #739)


test/moves/screens.test.ts
```diff
@@ -0,0 +1,216 @@
...
+  describe("Moves - Aurora Veil", () => {
+    beforeEach(() => {
+      game.override.enemyMoveset(MoveId.AURORA_VEIL).weather(WeatherType.SNOW);
+    });
+
+    it("should reduce damage of physical attacks by half in a single battle", async () => {
+      await testDamageMultiplier("single", MoveId.EARTHQUAKE, singleBattleMultiplier);
+    });
+
+    it("should reduce of physical attacks by a third in a double battle", async () => {
```
innerthunder:
```suggestion
    it("should reduce damage of physical attacks by a third in a double battle", async () => {
```

test/moves/screens.test.ts
```diff
@@ -0,0 +1,216 @@
...
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    game.override
+      .ability(Abilities.BALL_FETCH)
+      .startingLevel(1000)
+      .enemyAbility(Abilities.STALL) // So that player always outspeeds enemy's screen move on turn 1
+      .enemyLevel(1000)
+      .enemySpecies(Species.MAGIKARP);
+
+    vi.spyOn(allMoves[MoveId.FROST_BREATH], "accuracy", "get").mockReturnValue(100);
```
Tempo-anon: Using no guard as an ability is also an option for guaranteeing this hits.

test/moves/screens.test.ts
```diff
@@ -0,0 +1,216 @@
...
+import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
+import type { BattleStyle } from "#app/overrides";
+import { WeatherType } from "#enums/weather-type";
+import { allMoves } from "#app/data/data-lists";
+
+describe("Moves - Screen Moves", () => {
+  let phaserGame: Phaser.Game;
+  let game: GameManager;
+  const singleBattleMultiplier = 0.5;
+  const doubleBattleMultiplier = 2732 / 4096;
```
Tempo-anon: Hmmm. Wonder if maybe these constants should be moved somewhere else to like `constants` or as a constant in `arena-tag`

test/moves/screens.test.ts
```diff
@@ -0,0 +1,216 @@
...
+    it("should not reduce damage of a fixed-damage move", async () => {
+      await testDamageMultiplier("single", MoveId.NIGHT_SHADE, 1);
+    });
+
+    it("should not reduce confusion damage", async () => {
+      await testConfusionDamage();
+    });
+  });
+
+  describe("Moves - Aurora Veil", () => {
```
Tempo-anon, line 192: Should also be a test for it not stacking with reflect/light screen

Tempo-anon (approved):

torranx (approved):

PR #753, opened 2025-02-24
[Refactor] Update damage and heal functions in the `Pokemon` class

DayKev:
The damage number pop-up when healing is converted to damage (ie: due to Liquid Ooze) will no longer be green (which normally indicates healing).

`Pokemon.damage` and `Pokemon.damageAndUpdate` had many optional parameters, and when changing them to use destructuring I noticed a bunch of other jank.

- `Pokemon.damage` and `Pokemon.damageAndUpdate` now use destructuring for optional parameters.
- `Pokemon.damage` is now `protected`.
- `EnemyPokemon.damage` also updated to match the changes.
- `Pokemon.heal` now handles the heal number popup.
- `HpSplitAttr` simplified.
- Tsdocs were updated.
- Nesting reduced in `PostTurnStatusEffectPhase` and `WeatherEffectPhase` by using early returns.
- Bug in Present test fixed.
- All instances of `pokemon.turnData.damageTaken += ...` were consolidated into `Pokemon.damage()`.
- Added `ignoreDynamaxReduction` param to `Pokemon.damageAndUpdate` which causes the damage not to be reduced by the custom dynamax damage reduction factor if `true`.

## Checklist


src/field/pokemon.ts
```diff
@@ -3045,91 +3046,129 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-   * @param ignoreSegments boolean, not currently used
-   * @param preventEndure  used to update damage if endure or sturdy
-   * @param ignoreFaintPhase  flag on wheter to add FaintPhase if pokemon after applying damage faints
-   * @returns integer representing damage
+   * Deals damage to the pokemon. Called by {@linkcode damageAndUpdate}
+   * @param amount - Amount of damage that should be dealt
+   * @param ignoreSegments - Only used by {@linkcode EnemyPokemon.damage}
+   * @param preventEndure - If `true`, bypasses the effects of Endure, Sturdy, etc. Default `false`
+   * @param ignoreFaintPhase - If `true`, won't push a {@linkcode FaintPhase}. Default `false`
+   * @param ignoreDynamaxReduction - If `true`, dynamax damage reduction will be ignored. Default `false`
```
Tempo-anon: nit: the order of these should match the order of the params

src/field/pokemon.ts
```diff
@@ -3045,91 +3046,129 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-    damage: number,
-    _ignoreSegments: boolean = false,
-    preventEndure: boolean = false,
-    ignoreFaintPhase: boolean = false,
+  protected damage(
+    amount: number,
+    {
+      preventEndure = false,
+      ignoreFaintPhase = false,
+      ignoreDynamaxReduction = false,
```
Tempo-anon, line 3132: Is there a reason these 3 are now bundled together?

src/field/pokemon.ts
```diff
@@ -3045,91 +3046,129 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-    if (this.isMax(false)) {
-      damage = toDmgValue(damage * DYNAMAX_DAMAGE_TAKEN_FACTOR);
+    if (this.isMax(false) && !ignoreDynamaxReduction) {
+      amount = toDmgValue(amount * DYNAMAX_DAMAGE_TAKEN_FACTOR);
     }
 
-    if (!preventEndure && this.hp - damage <= 0) {
+    const surviveDamage = new BooleanHolder(false);
+
+    if (!preventEndure && this.hp - amount <= 0) {
```
Tempo-anon: nit: Change `this.hp - amount <= 0` to `amount >= this.hp` or `this.hp <= amount` for simplicity

src/field/pokemon.ts
```diff
@@ -3045,91 +3046,129 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-   * Called by damageAndUpdate()
-   * @param damage integer
-   * @param ignoreSegments boolean, not currently used
-   * @param preventEndure  used to update damage if endure or sturdy
-   * @param ignoreFaintPhase  flag on wheter to add FaintPhase if pokemon after applying damage faints
-   * @returns integer representing damage
+   * Deals damage to the pokemon. Called by {@linkcode damageAndUpdate}
+   * @param amount - Amount of damage that should be dealt
+   * @param ignoreSegments - Only used by {@linkcode EnemyPokemon.damage}
+   * @param preventEndure - If `true`, bypasses the effects of Endure, Sturdy, etc. Default `false`
```
Tempo-anon: If possible should be some comments explaining when these happen such as recoil, status, weather, etc

src/data/battler-tags.ts
```diff
@@ -1872,8 +1871,12 @@ export class PerishSongTag extends BattlerTag {
         }),
       );
     } else {
-      // The 2 here is just a number big enough to overcome the G-Max damage reduction
-      pokemon.damageAndUpdate(2 * pokemon.hp, HitResult.ONE_HIT_KO, false, true, true);
+      pokemon.damageAndUpdate(pokemon.hp, {
+        result: HitResult.ONE_HIT_KO,
+        ignoreSegments: true,
+        preventEndure: true,
+        ignoreDynamaxReduction: true,
```
Tempo-anon, line 2081: A much more elegant way of handling it. Are there any other times when this should be true though?

src/field/pokemon.ts
DayKev: In hindsight it wasn't necessary to do this for `Pokemon.damage` since it's only supposed to be used by `Pokemon.damageAndUpdate`, but I've already done it so... 🤷

src/field/pokemon.ts
DayKev: They do.

src/data/battler-tags.ts
DayKev: All OHKO instances most likely. I'll do a quick check.

DayKev: Added it to all sources of ratio-based damage (ie: Steel Beam should deal half the user's HP in damage, this shouldn't be reduced by the dynamax damage reduction) and OHKOs.

I wonder if it would be better to just implement the dynamax HP boost as an actual HP boost instead of a damage reduction. You could add a multiplier into the code that calculates a pokemon's HP if that pokemon is dynamaxed, and then whenever a pokemon dynamaxes or reverts from dynamax update their hp and call `pokemon.calculateStats()` to update the UI.

Tempo-anon: I thought modifying HP would have been more complex than just a damage modifier but maybe I was incorrect :joltdead:

src/field/pokemon.ts
```diff
@@ -3045,91 +3046,134 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+      ignoreDynamaxReduction = false,
+    }: {
+      result?: DamageResult;
+      isCritical?: boolean;
+      ignoreSegments?: boolean;
+      preventEndure?: boolean;
+      ignoreFaintPhase?: boolean;
+      source?: Pokemon;
+      ignoreDynamaxReduction?: boolean;
+    } = {},
```
torranx: nit: move this into an interface (and the other params)

DayKev: A single-use interface that has to be defined away from the function isn't necessary.

torranx: I agree but it improves readability

DayKev: I forgot about the in-editor tsdocs viewing, and this does improve that, so I added it.

src/field/pokemon.ts
```diff
@@ -3046,91 +3048,126 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-    source?: Pokemon,
+  public damageAndUpdate(
+    amount: number,
+    {
+      result = HitResult.EFFECTIVE,
+      isCritical = false,
+      ignoreSegments = false,
+      preventEndure = false,
+      ignoreFaintPhase = false,
+      source,
```
MokaStitcher: feels weird to have a param with no default value between 2 params with a default value

src/data/ab-attrs/form-block-damage-ab-attr.ts
```diff
@@ -54,7 +54,12 @@ export class FormBlockDamageAbAttr extends ReceivedMoveDamageMultiplierAbAttr {
         multiplier.value *= this.multiplier;
```
MokaStitcher: I can't comment on the specific line, but the tsdoc for this function seems outdated, do you mind updating it? (`args` should be replaced by `multiplier`)

MokaStitcher (commented):
Can you update the description to explain the introduction of the `ignoreDynamaxReduction` parameter, just so that we have a trace that doesn't require reading through the discussion

otherwise, not playtested yet but lgtm

MokaStitcher:
(new conflict to solve)

Also I seem to be getting decimal damage numbers in some cases (here, burn damage)
image
image

DayKev:
> Also I seem to be getting decimal damage numbers in some cases (here, burn damage)

Found the issue, when I replaced the bitshifts with normal division operators in `PostTurnStatusEffectPhase` that revealed that the damage values weren't wrapped in `toDmgValue()` so I added them in.

DayKev:
> > Also I seem to be getting decimal damage numbers in some cases (here, burn damage)
> 
> Found the issue, when I replaced the bitshifts with normal division operators in `PostTurnStatusEffectPhase` that revealed that the damage values weren't wrapped in `toDmgValue()` so I added them in.

Actually I think maybe that shouldn't even be necessary, it should probably be handled by `Pokemon.damageAndUpdate()`. Maybe I'll do that as a follow-up to this PR.

src/data/ab-attrs/form-block-damage-ab-attr.ts
MokaStitcher: I guess that's a valid way to not make it outdated :plains:

MokaStitcher (dismissed):

MokaStitcher (approved):

flx-sta (approved):

src/data/battler-tags.ts
```diff
@@ -981,7 +984,7 @@ export class SeedTag extends BattlerTag {
             new CommonAnimPhase(source.getBattlerIndex(), pokemon.getBattlerIndex(), CommonAnim.LEECH_SEED),
           );
 
-          const damage = pokemon.damageAndUpdate(toDmgValue(pokemon.getMaxHp() / 8));
+          const damage = pokemon.damageAndUpdate(toDmgValue(pokemon.getMaxHp() / 8), { ignoreDynamaxReduction: true });
```
PigeonBar, line 987: 20250322 Gmax Percent Damage

According to Showdown and Bulbapedia, effects that deal a percentage of a G-max Pokemon's health still follow dynamax damage reduction. (G-max Grimmsnarl took around 1/16 instead of 1/8 of its HP in damage from each of Leech Seed and Wrap.)

DayKev: GameFreak please...

src/field/pokemon.ts
```diff
@@ -3112,91 +3114,126 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   * Heals a pokemon by the amount specified.
+   * @param amount - The amount of HP to heal.
+   * @param quiet - If `true`, won't display a number above the pokemon. Default `false`
+   * @returns The actual amount of HP healed.
+   */
+  public heal(amount: number, quiet: boolean = false): number {
     const healAmount = Math.min(amount, this.getMaxHp() - this.hp);
     this.hp += healAmount;
+    if (!quiet && this.isOnField()) {
+      globalScene.damageNumberHandler.add(this, amount, HitResult.HEAL);
```
PigeonBar, line 3235: It turns out `healAmount` should be used here instead of `amount`, so that the displayed heal number does not show the Pokemon healing to more than full HP.

src/data/battler-tags.ts
DayKev: Does this include things like Substitute? Is there even a way for a Dynamax'd pokemon to use Substitute?

PigeonBar: I think Dynamax'd Pokemon in mainline are unable to use any moves other than Max Moves, so stuff like Substitute or Steel Beam would not be testable in mainline.

DayKev: What should we do about that?

src/field/pokemon.ts
DayKev: Put up a draft PR #867 with fixes to the issues noticed so far.

Issue #740, opened 2025-02-23
[Bug] KO'ing both Pokemon in a trainer double battle simultaneously causes a wave skip


PigeonBar:
### Describe the bug

In a trainer double battle, if you KO both opponent Pokemon simultaneously using a spread move, the game will skip the next wave's battle and instead give you 2 rewards screens in a row.

This is confirmed to occur for the Twins and for the Clowning Around ME.

### Reproduction

In a trainer battle, KO both opponent Pokemon simultaneously with a spread move.

### Expected behavior

The next wave should occur as normal.

### Screenshots / Videos

https://github.com/user-attachments/assets/0d921af5-56fb-49d7-8e23-e885c967db08

### Additional context

Probably caused by #432. Here is the check for when to push a new BattleEndPhase:
https://github.com/Despair-Games/poketernity/blob/b255e51d6b4e0c262db2d95dfd5b1082cf49074a/src/phases/victory-phase.ts#L51
Previously, the code would call `isFainted(true)` to check whether both Pokemon have the FNT status. This meant that the check would pass only if both FaintPhases have finished, so the check would only pass once after the second Pokemon faints. After the refactor, the check now passes for both Pokemon, so the game runs two BattleEndPhases in a row.

There is also duplicate ME code that likely has the same issue:
https://github.com/Despair-Games/poketernity/blob/b255e51d6b4e0c262db2d95dfd5b1082cf49074a/src/data/mystery-encounters/utils/encounter-phase-utils.ts#L839-L843

PR #757, opened 2025-02-25
[Misc] Delete stale pokemon objects in `BattleEndPhase`

DayKev:
PR text copied from cat's original PR.

The user should ideally see improvements in game performance.

It turns out that the game was not destroying enemy Pokémon objects after each battle and left them sitting in scene, meaning that a pile of enemy Pokémon assets were piling up wave after wave. This would often force a refresh. 

After modifiers are cleared from enemy Pokémon, the enemy Pokémon object is destroyed.

## How to test the changes?
Transition between waves and observe whether the game functions normally or not.

## Checklist

### DIFF
--- a/src/phases/battle-end-phase.ts
+++ b/src/phases/battle-end-phase.ts
@@ -59,6 +59,12 @@ export class BattleEndPhase extends BattlePhase {
 
     globalScene.clearEnemyHeldItemModifiers();
 
+    try {
+      globalScene.getEnemyParty().forEach((p) => p.destroy());
+    } catch {
+      console.warn("Unable to destroy stale pokemon objects in BattleEndPhase.");
+    }
+
     const lapsingModifiers = globalScene.findModifiers(
       (m) => m.isLapsingPersistentModifier() || m.isLapsingPokemonHeldItemModifier(),
     ) as (LapsingPersistentModifier | LapsingPokemonHeldItemModifier)[];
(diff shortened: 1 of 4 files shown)


Tempo-anon (dismissed):
You would think unused objects would get cleaned up by the javascript garbage collector...

DayKev:
Ran through a few waves with no issue, also added a test to run from wave 8 to wave 11 (so that covers a trainer battle, a wild battle, and a wild boss battle, plus a biome transition).

MokaStitcher:
> You would think unused objects would get cleaned up by the javascript garbage collector...

It's not unused it's a Phaser Container so if not removed from the scene it is still tied to it

Tempo-anon (approved):

PigeonBar (approved):
Tested that the following scenarios do not crash the game:
- Winning against Expert Pokemon Breeder
- Losing against Expert Pokemon Breeder
- Winning in Training Session
- Winstrates
- Stealing an enemy item with Pickup after ending a battle
- Triggering #740 (the extra Pokemon sprites from the skipped wave never experience a BattleEndPhase, so they remain on-screen without getting destroyed)
- Catching a Pokemon
- Fun and Games

PigeonBar:
By the way, out of curiosity, is there an easy way to test whether or not a Pokemon object has been garbage collected?

DayKev:
> By the way, out of curiosity, is there an easy way to test whether or not a Pokemon object has been garbage collected?

I think you'd have to put a console log that lists the [length of the] enemy pokemon array (whatever it is) and have it print at the start of every wave.

PR #759, opened 2025-02-26
[Bug] Several fixes for moves that call other moves

PigeonBar:

1. Two Pokemon using Mirror Move on each other no longer causes an infinite loop.
2. Moves that call other moves can no longer call G-max moves.
3. Nature Power now targets the intended target instead of picking a random opponent.

Fixing a few bugs introduced by #205.

1. Mirror Move now actually checks its list of invalid moves.
2. Added all max moves to each list of invalid moves.
3. Nature Power's attr now sets `hasTarget = true`.
4. Added some tests.

## Screenshots/Videos

<details><summary>Before the changes: Mirror Move infinite loop</summary>
<p>

https://github.com/user-attachments/assets/fb386c72-4978-4687-8835-7b47a9fe0a4c

</p>
</details>
<details><summary>After the changes: Mirror Move fails without causing an infinite loop</summary>
<p>

https://github.com/user-attachments/assets/98eb5d39-8538-4bde-b45a-c86cbed21e97

</p>
</details> 

## How to test the changes?

Mirror Move: Have all Pokemon on the field use only Mirror Move or Copycat.
Nature Power: Start a double battle, and check that Nature Power always hits the intended target, even if the target is the user's ally.
Metronome: Harder to test manually due to the RNG factor. Run the automated test instead.

## Checklist

### DIFF
--- a/src/utils/move-utils.ts
+++ b/src/utils/move-utils.ts
@@ -60,12 +60,13 @@ export function applyMoveChargeAttrs<TAttr extends MoveAttr>(
 export function getGmaxMoveList(): MoveId[] {
   const ret: MoveId[] = [];
   for (const move_name of getEnumKeys(MoveId)) {
-    if (move_name.slice(0, 6) === "G_MAX_") {
+    if (move_name.startsWith("G_MAX_")) {
       ret.push(MoveId[move_name]);
     }
   }
   return ret;
 }
+
 /**
  * Returns a list of all Max Moves, including G-max Moves, based on the move's name.
  *
@@ -75,7 +76,7 @@ export function getGmaxMoveList(): MoveId[] {
 export function getMaxMoveList(): MoveId[] {
   const ret = getGmaxMoveList();
   for (const move_name of getEnumKeys(MoveId)) {
-    if (move_name.slice(0, 4) === "MAX_") {
+    if (move_name.startsWith("MAX_")) {
       ret.push(MoveId[move_name]);
     }
   }
--- a/src/utils/move-utils.ts
+++ b/src/utils/move-utils.ts
@@ -5,10 +5,11 @@ import type { Pokemon } from "#app/field/pokemon";
 import type { PokemonMove } from "#app/field/pokemon-move";
 import { globalScene } from "#app/global-scene";
 import { getPokemonNameWithAffix } from "#app/messages";
-import { BooleanHolder, toDmgValue, type AbstractConstructor } from "#app/utils";
+import { BooleanHolder, getEnumKeys, toDmgValue, type AbstractConstructor } from "#app/utils";
 import { AbAttrFlag } from "#enums/ab-attr-flag";
 import { BattlerIndex } from "#enums/battler-index";
 import { HitResult } from "#enums/hit-result";
+import { MoveId } from "#enums/move-id";
 import { t } from "i18next";
 
 //#region Exports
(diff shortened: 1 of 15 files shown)


test/moves/metronome.test.ts
```diff
@@ -111,4 +112,23 @@ describe("Moves - Metronome", () => {
 
     await game.phaseInterceptor.to("CommandPhase");
   });
+
+  it("should never call a G-Max move", async () => {
```
Tempo-anon, line 116: Wouldn't this test make sense to just test that all the moves in the invalidMetronomeMoves are never called rather than just the g-max moves?

Tempo-anon (dismissed):

test/moves/metronome.test.ts
PigeonBar: Maybe, but this test is also testing that invalidMetronomeMoves includes all g-max moves.

src/utils/move-utils.ts
```diff
@@ -50,6 +51,37 @@ export function applyMoveChargeAttrs<TAttr extends MoveAttr>(
...
+export function getGmaxMoveList(): MoveId[] {
+  const ret: MoveId[] = [];
+  for (const move_name of getEnumKeys(MoveId)) {
+    if (move_name.slice(0, 6) === "G_MAX_") {
+      ret.push(MoveId[move_name]);
+    }
+  }
+  return ret;
+}
+/**
```
MokaStitcher, line 70:
```suggestion

/**
```

src/utils/move-utils.ts
```diff
@@ -50,6 +51,37 @@ export function applyMoveChargeAttrs<TAttr extends MoveAttr>(
...
+ * Returns a list of all G-max Moves, based on the move's name.
+ *
+ * Note that this function deliberately does not check for a move flag, since it is
+ * designed to be called before `initMoves()` has finished.
+ */
+export function getGmaxMoveList(): MoveId[] {
+  const ret: MoveId[] = [];
+  for (const move_name of getEnumKeys(MoveId)) {
+    if (move_name.slice(0, 6) === "G_MAX_") {
```
MokaStitcher:
```suggestion
    if (move_name.startsWith("G_MAX_")) {
```

src/utils/move-utils.ts
```diff
@@ -50,6 +51,37 @@ export function applyMoveChargeAttrs<TAttr extends MoveAttr>(
...
+/**
+ * Returns a list of all Max Moves, including G-max Moves, based on the move's name.
+ *
+ * Note that this function deliberately does not check for a move flag, since it is
+ * designed to be called before `initMoves()` has finished.
+ */
+export function getMaxMoveList(): MoveId[] {
+  const ret = getGmaxMoveList();
+  for (const move_name of getEnumKeys(MoveId)) {
+    if (move_name.slice(0, 4) === "MAX_") {
```
MokaStitcher:
```suggestion
    if (move_name.startsWith("MAX_")) {
```

MokaStitcher (dismissed):
Apologies for not double checking that the mirror move infinite loop was fixed in the original PR
I did playtest this time. Looks good, just nits (enemies still target each other with mirror move but I don't know if that's an issue with the move or AI)

MokaStitcher (approved):

Tempo-anon (approved):

PR #765, opened 2025-03-01
[Misc] Further cleanup of `PokemonSpecies.getSpeciesForLevel()`

DayKev:
N/A

Code cleanup.

- Now-redundant variables `evolutionChance` and `noEvolutionChance` removed from `PokemonSpecies.getSpeciesForLevel()`.
- The array of possible evolutions will now properly be chosen from again.

## Checklist

### DIFF
--- a/src/data/pokemon-species.ts
+++ b/src/data/pokemon-species.ts
@@ -4,7 +4,7 @@ import { PartyMemberStrength } from "#enums/party-member-strength";
 import { Species } from "#enums/species";
 import i18next from "i18next";
 import type { GameMode } from "#app/game-mode";
-import { randSeedInt, randSeedGauss } from "#app/utils";
+import { randSeedGauss, randSeedItem } from "#app/utils";
 import type { GrowthRate } from "#enums/growth-rates";
 import type { EvolutionLevel } from "#app/data/balance/pokemon-evolutions";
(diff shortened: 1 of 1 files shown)


src/data/pokemon-species.ts
```diff
@@ -170,15 +170,13 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
 
     const evolutionPool: Map<number, Species> = new Map();
```
Tempo-anon: Changing this from a Map<number, Species> to an array or Set of Species and then line 192 `evolutionPool.set(evolutionChance, ev.speciesId);` to a push will also fix the issue of only a single Species in evolutionPool

src/data/pokemon-species.ts
```diff
@@ -196,7 +193,7 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
       }
     }
 
-    if (noEvolutionChance === 1) {
+    if (evolutionChance === 0) {
       return this.speciesId;
     }
```
Tempo-anon, line 192: Lines 200-214 can then be replaced with something like
```ts
const randValue = randSeedInt(evolutionPool.length);
    console.log(evolutionPool);
    return getPokemonSpecies(evolutionPool[randValue]).getSpeciesForLevel(
      level, true, forTrainer, strength, currentWave
    );
```

DayKev:
Superseded by #767

PR #761, opened 2025-02-28
[Refactor] Refactor Pokemon-evolutions into seperate files, (gen 1)

Tempo-anon:
## What are the changes the user will see?

* Happiny and Cleffa now evolve with 70 happiness in line with the rest of the babies
* Lickitung no longer needs to be level 32 to evolve
* Some Pokemon might appear earlier/later

## Why am I making these changes?

pokemon-evolutions was too large of a file and a lot of the Pokemon were not grouped together logically

## What are the changes from a developer perspective?
* Created a `pokemon-evolution-alt-levels` to store constants for altLevels (when enemy Pokemon evolve)
* Created a `init-pokemon-evolutions` and `gen1-pokemon-evolutions` where I add all the evolutions associated with Pokemon that were introduced in gen 1. I will be doing this for all other gens as well

## Screenshots/Videos
n/a

## How to test the changes?

`npm run test`

## Checklist

- [] ⚠️ If this is a PR for `main` (such as a hotfix), has the game version been updated (`npm run update-version:patch` / `npm run update version:minor`?
- [x] Otherwise: **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?

- [ ] Are all unit tests still passing? (`npm run test:silent`)
  - [ ] Have I created new automated tests (`npm run test:create`) or updated existing tests related to the PR's changes?
- [ ] Have I provided screenshots/videos of the changes (if applicable)?
  - [ ] Have I made sure that any UI change works for both UI themes (dark and light)?

Are there any localization additions or changes? If so:

- [ ] Has a locales PR been created on the [locales](https://github.com/despair-games/poketernity-locales) repo?
  - [ ] If so, please leave a link to it here:
- [ ] Has the translation team been contacted for proofreading/translation?

PR #749, opened 2025-02-24
[Balance] Delete species wild evolution delay

Tempo-anon:
## What are the changes the user will see?

* Certain Pokemon that require moves to evolve no longer have a minimum level requirement
* When enemy Pokemon evolve will be different
* Scyther and Onix no longer require knowing a steel move to evolve

## Why am I making these changes?

* `SpeciesWildEvolutionDelay` and the way the calculation was being used in `pokemon-species` was very confusing to read and understand
* Certain Pokemon would appear too low level to be evolved or too high level and unevolved

## What are the changes from a developer perspective?

* Created an `altLevel` field that determines when a Pokemon should evolve if it's not by standard level
* Created some temporary values like `GENERIC_ITEM_EVO_LEVEL` and `HAPPINESS_EVO_LEVEL`. These values can and should be changed later, but for now these are the values being used for altLevels (there are cases of Pokemon whose natural level evolutions are later than the item evo level right now 💀 )
* Simplified pokemon-species a lot so now enemy Pokemon should always evolve at that given level or never evolve if under it. We can tweak the likelihood of this in the future, but for now it is a hard cutoff

## Screenshots/Videos

n/a

## How to test the changes?

`npm run start`

## Checklist

- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [ ] Have I tested the changes manually?

- [ ] Are all unit tests still passing? (`npm run test:silent`)
  - [ ] Have I created new automated tests (`npm run test:create`) or updated existing tests related to the PR's changes?
- [ ] Have I provided screenshots/videos of the changes (if applicable)?
  - [ ] Have I made sure that any UI change works for both UI themes (dark and light)?

Are there any localization additions or changes? If so:

- [ ] Has a locales PR been created on the [locales](https://github.com/despair-games/poketernity-locales) repo?
  - [ ] If so, please leave a link to it here:
- [ ] Has the translation team been contacted for proofreading/translation?

PR #767, opened 2025-03-01
[Bug] Fix wild Pokemon not picking a random evolution species

PigeonBar:

1. If an enemy Pokemon spawn has multiple possible evolutions, it now properly selects a random one.
2. For certain species, if there is a possible pre-evolution and the Pokemon is a low enough level, it now properly selects the pre-evolution.

Fixing a bug introduced by #749.

1. The related functions have been consolidated into one function `getEnemySpeciesForLevel`.
2. This function now picks a random evolution using `randSeedInt()`. The previous `totalWeight` system, which #749 had temporarily broken, is now completely removed.
3. Renamed `altLevel` to `enemyEvolveLevel`. Its default value is now equal to the evolution's regular level instead of 0.
4. Updated `getPrevolutionLevels()` to use `enemyEvolveLevel` (discovered by Temp). (TODO: `getPrevolutionLevels()` is also called by `getSimulatedEvolutionChain()`, which is sometimes used in moveset generation. Hopefully this change does not break moveset generation.)
5. Also deleted the unused function `getEvolutionLevels()` and the unused params `allowEvolving`, `strength`, and `currentWave`.
6. Added some tests.

## Screenshots/Videos

<details><summary>Before the changes: All Eevees evolve into Glaceon</summary>
<p>

20250228 Before Fix Eeveelutions

</p>
</details> 
<details><summary>After the changes: On the same seed, Eevees evolve into multiple different evolutions</summary>
<p>

20250228 After Fix Eeveelutions

</p>
</details> 

## How to test the changes?

Edit this line to spawn only Eevee:
https://github.com/Despair-Games/poketernity/blob/5849f8873c6287c48dc94c2e784e6a06e7456f4f/src/field/arena.ts#L175

## Checklist

- [ ] There is no overlap with another PR? (#765, #761)

### DIFF
--- a/src/data/balance/pokemon-evolutions.ts
+++ b/src/data/balance/pokemon-evolutions.ts
@@ -30,11 +30,17 @@ export class SpeciesFormEvolution {
   public level: number;
   public item: EvolutionItem | null;
   public condition: SpeciesEvolutionCondition | null;
-  /** A numerical level for Pokemon that don't evolve with level
-   * Is 0 if it is a level evolution
-   */
-  public altLevel: number;
+  public enemyEvolveLevel: number;
 
+  /**
+   * @param speciesId The ID of the species that the Pokemon will evolve into.
+   * @param preFormKey The form key that a Pokemon must have before being eligible for this evolution.
+   * @param evoFormKey The form key for the form that the Pokemon will evolve into.
+   * @param level The minimum level that a Pokemon must have for this evolution.
+   * @param item If applicable, the evolution item that the Pokemon must use for this evolution.
+   * @param condition If applicable, an extra condition that the Pokemon must satisfy for this evolution.
+   * @param enemyEvolveLevel The level at which enemy spawns will undergo this evolution. Default: Equal to `level`.
+   */
   constructor(
     speciesId: Species,
     preFormKey: string | null,
@@ -42,7 +48,7 @@ export class SpeciesFormEvolution {
     level: number,
     item: EvolutionItem | null,
     condition: SpeciesEvolutionCondition | null,
-    altLevel: number = 0,
+    enemyEvolveLevel: number = level,
   ) {
     this.speciesId = speciesId;
     this.preFormKey = preFormKey;
@@ -50,7 +56,7 @@ export class SpeciesFormEvolution {
     this.level = level;
     this.item = item || EvolutionItem.NONE;
     this.condition = condition;
-    this.altLevel = altLevel;
+    this.enemyEvolveLevel = enemyEvolveLevel;
   }
 }
 
@@ -60,9 +66,9 @@ export class SpeciesEvolution extends SpeciesFormEvolution {
     level: number,
     item: EvolutionItem | null,
     condition: SpeciesEvolutionCondition | null,
-    altLevel: number = 0,
+    enemyEvolveLevel: number = level,
   ) {
-    super(speciesId, null, null, level, item, condition, altLevel);
+    super(speciesId, null, null, level, item, condition, enemyEvolveLevel);
   }
 }
 
--- a/src/data/pokemon-species.ts
+++ b/src/data/pokemon-species.ts
@@ -173,7 +173,8 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
     const eligibleEvolutions: Species[] = [];
 
     for (const ev of evolutions) {
-      if (ev.level > level) {
+      // TODO: Should enemy Pokemon have a random chance of evolving if they are close to the level threshold?
+      if (ev.enemyEvolveLevel > level) {
         continue;
       }
 
@@ -182,9 +183,7 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
 
       // Random non-trainer spawns are not eligible for regional evolutions (e.g. Alolan Raichu)
       if (forTrainer || !isRegionalEvolution) {
-        if (level > ev.enemyEvolveLevel) {
-          eligibleEvolutions.push(ev.speciesId);
-        }
+        eligibleEvolutions.push(ev.speciesId);
       }
     }
 
@@ -221,7 +220,7 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
     return prevolutionLevels;
   }
 
-  // This could definitely be written better and more accurate to the getSpeciesForLevel logic, but it is only for generating movesets for evolved Pokemon
+  // TODO: This could definitely be written better and more accurate to the getSpeciesForLevel logic, but it is only for generating movesets for evolved Pokemon
   getSimulatedEvolutionChain(
     currentLevel: number,
     forTrainer: boolean = false,
@@ -241,9 +240,9 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
           prevolutionLevels[l][0],
           Math.min(
             Math.max(
-              evolution?.level! + Math.round(randSeedGauss(0.5, 1 + levelDiff * 0.2) * 0.5 * 5) - 1,
+              evolution?.enemyEvolveLevel! + Math.round(randSeedGauss(0.5, 1 + levelDiff * 0.2) * 0.5 * 5) - 1,
               2,
-              evolution?.level!,
+              evolution?.enemyEvolveLevel!,
             ),
             currentLevel - 1,
           ),
@@ -259,7 +258,7 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
           Math.max(
             lastPrevolutionLevel + Math.round(randSeedGauss(0.5, 1 + levelDiff * 0.2) * 0.5 * 5),
             lastPrevolutionLevel + 1,
-            evolution?.level!,
+            evolution?.enemyEvolveLevel!,
           ),
           currentLevel,
         ),
--- a/src/data/pokemon-species.ts
+++ b/src/data/pokemon-species.ts
@@ -200,7 +200,7 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
     return prevolutionLevels;
   }
 
-  // TODO: This could definitely be written better and more accurate to the getSpeciesForLevel logic, but it is only for generating movesets for evolved Pokemon
+  // TODO: This could definitely be written better and more accurate to the getEnemySpeciesForLevel logic, but it is only for generating movesets for evolved Pokemon
   getSimulatedEvolutionChain(
     currentLevel: number,
     forTrainer: boolean = false,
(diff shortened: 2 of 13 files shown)


src/data/balance/pokemon-evolutions.ts
```diff
@@ -30,27 +30,33 @@ export class SpeciesFormEvolution {
...
+   */
   constructor(
     speciesId: Species,
     preFormKey: string | null,
     evoFormKey: string | null,
     level: number,
     item: EvolutionItem | null,
     condition: SpeciesEvolutionCondition | null,
-    altLevel: number = 0,
+    enemyEvolveLevel: number = level,
```
Tempo-anon, line 51: This is a much better idea than setting it to 0. Also a better name too. (What was I thinking?)

src/data/pokemon-species.ts
```diff
@@ -126,31 +124,37 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
...
-      (isBoss ? PartyMemberStrength.WEAKER : PartyMemberStrength.AVERAGE) + (gameMode?.isEndless ? 1 : 0),
-    );
+  /**
+   * Calculates the correct evolution stage for a wild enemy Pokemon based on its level,
+   * applying pre-evolutions and evolutions as necessary.
+   * @param level The level of the Pokemon.
+   * @param allowEvolving Whether or not to allow further evolution.
+   * @returns The {@linkcode Species | species ID} of the desired evolution stage.
+   */
+  getWildSpeciesForLevel(level: number, allowEvolving: boolean): Species {
```
Tempo-anon: This is only called in 2 places
* arena.ts `    const newSpeciesId = ret.getWildSpeciesForLevel(level, true);`
* evolution.test `      actualEvolutions.add(getPokemonSpecies(Species.EEVEE).getWildSpeciesForLevel(100, true));`

allowEvolving is always true

src/data/pokemon-species.ts
```diff
@@ -126,31 +124,37 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
...
-  ): Species {
-    return this.getSpeciesForLevel(level, allowEvolving, true, strength, currentWave);
+  /**
+   * Calculates the correct evolution stage for an enemy trainer's Pokemon based on its level,
+   * applying pre-evolutions and evolutions as necessary.
+   * @param level The level of the Pokemon.
+   * @param allowEvolving Whether or not to allow further evolution. Default: `false`.
+   * @returns The {@linkcode Species | species ID} of the desired evolution stage.
+   */
+  getTrainerSpeciesForLevel(level: number, allowEvolving: boolean = false): Species {
```
Tempo-anon: `allowEvolving` is false only once in the codebase in trainer.ts's block of code:
```ts
// If useNewSpeciesPool is true, we need to generate a new species from the new species pool, otherwise we generate a random species
        let species = useNewSpeciesPool
          ? getPokemonSpecies(newSpeciesPool[Math.floor(randSeedInt(newSpeciesPool.length))])
          : template.isSameSpecies(index) && index > offset
            ? getPokemonSpecies(battle.enemyParty[offset].species.getTrainerSpeciesForLevel(level, false))
            : this.genNewPartyMemberSpecies(level, strength);
```
Wonder if we can also do something about that

src/data/pokemon-species.ts
```diff
@@ -167,75 +171,30 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
     }
 
     const evolutions = pokemonEvolutions[this.speciesId];
-
-    const evolutionPool: Map<number, Species> = new Map();
-    const totalWeight = 0;
-    let noEvolutionChance = 1;
+    const eligibleEvolutions: Species[] = [];
```
Tempo-anon, line 153: Another consideration I had was whether to use an array vs a set but I don't think it matters too much

src/data/pokemon-species.ts
```diff
@@ -167,75 +171,30 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
...
 
     for (const ev of evolutions) {
       if (ev.level > level) {
```
Tempo-anon: Should this be `ev.level` or `ev.wildEvolveLevel`?

Also another nit: I think it makes sense for me to have the check be `if (level < ev.wildEvolveLevel)` because later on the if check has `level` on the left instead of the right. Also a comment here to explain what's going on (we're skipping an evolution if the pokemon's level is not high enough

src/data/pokemon-species.ts
```diff
@@ -167,75 +171,30 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
...
       const evolutionSpecies = getPokemonSpecies(ev.speciesId);
       const isRegionalEvolution = !this.isRegional() && evolutionSpecies.isRegional();
 
-      if (!forTrainer && isRegionalEvolution) {
-        evolutionChance = 0;
-      } else {
-        if ((ev.altLevel !== 0 && level > ev.altLevel) || level > ev.level) {
-          evolutionChance = 1;
-          noEvolutionChance = 0;
+      // Random non-trainer spawns are not eligible for regional evolutions (e.g. Alolan Raichu)
```
Tempo-anon, line 164: Unrelated to this PR, but I wouldn't mind something like this happening with a rare chance, though I can see arguments for/against. The way Pokerogue does it, Hisui Typhlosion will only ever spawn in the graveyard and never in the volcano, but normal Cyndaquil can also show up in the graveyard if the levels are low enough

src/data/pokemon-species.ts
```diff
@@ -167,75 +171,30 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
...
+      if (forTrainer || !isRegionalEvolution) {
+        if (level > ev.enemyEvolveLevel) {
```
Tempo-anon: This should probably be a `>=` rather than a `>`

Can also add a `todo` here if we want to add in some % to evolve or not rather than a direct cutoff point

Tempo-anon (commented):
I really appreciate the added documentation and bug squashing

src/data/pokemon-species.ts
PigeonBar: It seems like `getTrainerSpeciesForLevel()` here is only being used if `isSameSpecies()` is true, in which case we should probably just be copying the previously used species instead of calling `getTrainerSpeciesForLevel()` on it again.

src/data/pokemon-species.ts
```diff
@@ -283,9 +240,9 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
           prevolutionLevels[l][0],
           Math.min(
             Math.max(
-              evolution?.level! + Math.round(randSeedGauss(0.5, 1 + levelDiff * 0.2) * 0.5 * 5) - 1,
+              evolution?.enemyEvolveLevel! + Math.round(randSeedGauss(0.5, 1 + levelDiff * 0.2) * 0.5 * 5) - 1,
```
Tempo-anon, line 223: Hmm I should have definitely documented better what I changed here when deleting `wildDelay`. Basically the second `0.5` term here used to be a `Math.max(evolution?.wildDelay!, 0.5)` and wildDelay was a enum ranging from 0-5 so this term will be bigger for when it used to be NONE (the default) but smaller for everything else. But ugh... undocumented complex sam code...

src/field/trainer.ts
```diff
@@ -413,21 +413,12 @@ export default class Trainer extends Phaser.GameObjects.Container {
...
           : template.isSameSpecies(index) && index > offset
-            ? getPokemonSpecies(
-                battle.enemyParty[offset].species.getTrainerSpeciesForLevel(
-                  level,
-                  false,
-                  template.getStrength(offset),
-                  globalScene.currentBattle.waveIndex,
-                ),
-              )
+            ? battle.enemyParty[offset].species
```
Tempo-anon: nice cleanup!

src/field/trainer.ts
```diff
@@ -413,21 +413,12 @@ export default class Trainer extends Phaser.GameObjects.Container {
...
             : this.genNewPartyMemberSpecies(level, strength);
 
         // If the species is from newSpeciesPool, we need to adjust it based on the level and strength
```
Tempo-anon: This comment says `newSpeciesPool` while the `if` is checking `useNewSpeciesPool`

Tempo-anon (dismissed):

src/field/trainer.ts
PigeonBar: Checking `useNewSpeciesPool` is correct because even if the species is not from `newSpeciesPool`, `newSpeciesPool` will be an empty array, which is still truthy.

Tempo-anon: I meant the comment should be updated to say `useNewSpeciesPool` rather than `newSpeciesPool`

src/data/pokemon-species.ts
```diff
@@ -126,33 +124,36 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
...
-    currentWave: number = 0,
-  ): Species {
-    return this.getSpeciesForLevel(level, allowEvolving, true, strength, currentWave);
+  /**
+   * Calculates the correct evolution stage for an enemy trainer's Pokemon based on its level,
+   * applying pre-evolutions and evolutions as necessary.
+   * @param level The level of the Pokemon.
+   * @returns The {@linkcode Species | species ID} of the desired evolution stage.
+   */
+  getTrainerSpeciesForLevel(level: number): Species {
```
DayKev: Tbh these extra functions aren't even necessary, just pass the `forTrainer` param at the callsite instead of using another function to do that for you.

Tempo-anon (dismissed):

src/data/trainer-config.ts
```diff
@@ -1600,16 +1600,11 @@ export function getRandomPartyMemberFunc(
   trainerSlot: TrainerSlot = TrainerSlot.TRAINER,
   ignoreEvolution: boolean = false,
   postProcess?: (enemyPokemon: EnemyPokemon) => void,
-) {
-  return (level: number, strength: PartyMemberStrength) => {
+): PartyMemberFunc {
+  return (level: number, _strength: PartyMemberStrength) => {
```
DayKev:
```suggestion
  return (level: number) => {
```

src/data/trainer-config.ts
```diff
@@ -1634,12 +1629,10 @@ export function getSpeciesFilterRandomPartyMemberFunc(
     return (allowLegendaries || notLegendary) && !species.isTrainerForbidden() && originalSpeciesFilter(species);
   };
 
-  return (level: number, strength: PartyMemberStrength) => {
+  return (level: number, _strength: PartyMemberStrength) => {
```
DayKev:
```suggestion
  return (level: number) => {
```

DayKev (dismissed):

PR #776, opened 2025-03-03
[Refactor] Re-implement save migration system

DayKev:
N/A

Making it easier for developers to write save data migrators in the future.

Version comparison for the data migrators is now handled automatically, developers only need to write the migrators (in `src/system/version_migration/versions/vA_B_C.ts` like before, but also assign them a version now) and then add them to the appropriate array in `src/system/version_migration/version_converter.ts` (if it hasn't been done already).
All the classes were removed and replaced with a set of functions.
An example is provided in `src/system/version_migration/version_converter.ts`.

## Checklist


src/system/version_migration/version_converter.ts
```diff
@@ -7,21 +7,87 @@ import { version } from "../../../package.json";
...
+
+export const vA_B_C_SettingsMigrators: SettingsSaveMigrator[] = [settingsMigratorA] as const;
 */
 
 // --- vA.B.C PATCHES --- //
-// import * as vA_B_C from "./versions/vA_B_C";
+// import { vA_B_C_SystemMigrators, vA_B_C_SessionMigrators, vA_B_C_SettingsMigrators } from "./versions/vA_B_C";
+
+/** Current game version */
+const LATEST_VERSION = version;
```
xsn34kzx, line 51: This will likely have to get duplicated based the amount of possible saves (in this case, 3 for system, session, and settings) if the move going forward is to decouple from game version and have separate versioning for each save type so as to avoid the scenario of game versioning affecting how migrators are applied.

src/system/version_migration/version_converter.ts
```diff
@@ -7,21 +7,87 @@ import { version } from "../../../package.json";
...
+const settingsMigrators: SettingsSaveMigrator[] = [].concat();
+
+/** Sorts migrators by their stated version, ensuring they are applied in order from oldest to newest */
+const sortMigrators = (migrators: SaveMigrator[]): void => {
+  migrators.sort((a, b) => compareVersions(a.version, b.version));
+};
+
+sortMigrators(systemMigrators);
+sortMigrators(sessionMigrators);
+sortMigrators(settingsMigrators);
```
xsn34kzx, line 79: The idea to avoid extra steps for adding new migrators to their arrays and the potential for putting them out of order makes sense, but I feel like the extra runtime spent on sorting them every time (which will grow over time as game scope increases equate to save data scope increases) can be avoided by making sure people put them in order at the review level and/or having a unit test to verify.

DayKev: Yeah it's mostly a safeguard to prevent issues due to mistakes. The sorting shouldn't be that computationally expensive since I doubt we're going to have that many save migrators. Making a test instead might be a good idea, then we can offload the computation entirely to GitHub.

src/system/version_migration/version_converter.ts
DayKev: Yeah if we make that change then this will also have to be changed. For now I just kept the existing behavior but replaced the direct access of `version` from `package.json` with checking this `const`.

src/system/version_migration/version_converter.ts
DayKev: After looking into it, writing a test for this would require me to export a bunch of things that aren't supposed to be exported, so I think leaving the sorting is fine since it shouldn't have a noticeable performance impact (if we need to write that many save data migrators maybe we've made a mistake lol).

innerthunder (dismissed):

src/system/version_migration/version_converter.ts
```diff
@@ -1,27 +1,91 @@
...
+  migrate: (data: SessionSaveData): void => {
+    // migration code goes here
+  },
+};
+
+export const vA_B_C_SessionMigrators: SessionSaveMigrator[] = [sessionMigratorA] as const;
+
+const settingsMigratorA: SettingsSaveMigrator = {
+  version: "A.B.C",
+  migrate: (data: Settings): void => {
```
MokaStitcher:
```suggestion
  migrate: (data: Partial<Settings>): void => {
```
I think?

MokaStitcher (commented):
lgtm.
Don't have the bandwidth to test this fully at the moment unfortunately. 
I don't know if it's for this PR but I think having a test template for data migration might be good. Every time we add a migrator we should probably add a corresponding save file in the test folder and a test loading it. We could also have a more generic one test loading all migration related save files and just ensuring they have all up to date attributes (which would means all required migrators have run).

Of note, the settings migrator is currently not being called in the code at the moment. Back during the settings refactor it was disabled because of circular import issues and now I'm not 100% sure anymore where/when the migration should happen. I know I left a todo in the code but the things I ended up doing with the gameVersion might clash with that spot now. So, probably for another PR as well unless you want to have a look at it

src/system/version_migration/version_converter.ts
DayKev: Good catch.

innerthunder (approved):

src/system/version_migration/version_converter.ts
```diff
@@ -1,27 +1,91 @@
...
 import { compareVersions } from "compare-versions";
 import { version } from "../../../package.json";
 
 /*
 // template for save migrator creation
 // versions/vA_B_C.ts
-const systemMigratorA = (data: SystemSaveData): void => {};
 
-export const systemMigrators = [systemMigratorA] as const;
+// The version for each migrator should match the filename, ie: `vA_B_C.ts` -> `version: "A.B.C"
```
PigeonBar, line 14: This is a bit late, but this comment should probably also mention that "A.B.C" should be the destination version of the migrator. (e.g., If migrating from 0.1.0 to 0.2.0, the migrator's version should be 0.2.0, not 0.1.0.)

PR #791, opened 2025-03-06
[Refactor] Created new scene for unlocking Achievements

frutescens:

None. 

First part of an attempt to break up the giant global scene object. 

A new scene has been created `UnlockAchievementsManager` that determines if any related achievements have been unlocked at a specific moment in the game. If an Achievement has been unlocked, it updates the player's save data with the new updates and queues the achievement banners. It then is stopped until the next time it is needed. 

Achievements have been refactored to reflect the more-context driven nature of their validation. Calls to achievements now use a single function that looks at the current point of the game instead. 

Have not touched Vouchers yet, so vouchers still use the old achievements system.

### Notes:
- It would be nice to have this be an overall `UnlockManager` instead so achievements can be associated with certain items instead and items could be unlocked through that, but right now, in `GameOverPhase`, the check to unlock  items is past end card. It should be possible to do so, but I think there's some things that need to be done to the message ui.
- Should be easy to convert gym leader/special trainers victories to this system once we figure out what is to be done about vouchers. 

## Screenshots/Videos

https://github.com/user-attachments/assets/074f2dc6-8e3c-4963-af19-b8d12eb37e36

## How to test the changes?

Unlock any of the various achievements available in the game at the moment. 

## Checklist


src/battle-scene-bus.ts
```diff
@@ -0,0 +1 @@
+export const battleSceneBus = new Phaser.Events.EventEmitter();
```
innerthunder, line 1: Is this meant to be global, or would the built-in `BattleScene.events` also work?

src/battle-scene.ts
```diff
@@ -2930,6 +2932,20 @@ export default class BattleScene extends SceneBase {
     return true;
   }
 
+  validAchievements(achievementCategory: AchvCategory, data: any) {
```
innerthunder: Should be `validateAchievements` instead.

Also, it might be better to define this as

```ts
validateAchievements(achievementCategory: AchvCategory, ...data: unknown[]) {
```

Then add overloads to enforce different types of data for each category, e.g.

```ts
validateAchievements(achievementCategory: AchvCategory.PARTY, party: Pokemon[]);
```

src/battle-scene-bus.ts
frutescens: It is meant to be 'global' since the messages it is handling are related to scene status (i.e. `scene/achievement_manager/ready`) or game data manipulations but it is a placeholder event handler until more gets separated :)

src/system/achievements/unlocking/unlock-achievements-manager.ts
```diff
@@ -0,0 +1,84 @@
...
+import { newAchvs } from "../achievements";
+import AchvBanner from "./achievements-banner";
+import { achievementsBus } from "../achievements-events";
+
+type UnlockAchievementsManagerData = {
+  context: AchvCategory;
+  unlockedAchievements: string[];
+};
+
+export class UnlockAchievementsManager extends SceneBase {
```
torranx, line 13: was this intentionally named `...Manager` even tho we're referencing this as a scene?

flx-sta:
Closing this for now.
Feel free to take over where @frutescens left of if you want to

PR #793, opened 2025-03-06
[Test] Fix flaky Sky Drop test

PigeonBar:

https://github.com/Despair-Games/poketernity/actions/runs/13704319361/job/38325809014

Increased the player Pokemon's level from 100 to 200 so that it does not faint to the opponent's Thrash.

## How to test the changes?

Run the test. It should not fail, and it also should not display warning messages about the player Pokemon reaching low HP.

## Checklist

### DIFF
--- a/test/moves/sky_drop.test.ts
+++ b/test/moves/sky_drop.test.ts
@@ -436,6 +436,7 @@ describe("Moves - Sky Drop", () => {
   });
 
   it("should stop the target's consecutive uses of frenzy moves", async () => {
+    game.override.startingLevel(200);
     await game.classicMode.startBattle([Species.FEEBAS]);
 
     const player = game.field.getPlayerPokemon();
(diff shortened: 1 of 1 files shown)


test/moves/sky_drop.test.ts
```diff
@@ -436,6 +436,7 @@ describe("Moves - Sky Drop", () => {
   });
 
   it("should stop the target's consecutive uses of frenzy moves", async () => {
+    game.override.startingLevel(200);
     await game.classicMode.startBattle([Species.FEEBAS]);
```
Tempo-anon: Can also making the user be Regirock/Aggron for high defense and resistance. Should add a comment here explaining the change here, be it level increase or pokemon changed

Tempo-anon (dismissed):

Tempo-anon (approved):

DayKev (approved):
Maybe it would be a good idea to start using Sturdy instead of Ball Fetch when we don't want the pokemon in the tests to faint.

PR #808, opened 2025-03-09
[Refactor][UI] Refactor Text Style

MokaStitcher:
This is hopefully the final big UI cleanup PR before we can update the game resolution and UI.

- Japanese text is no longer scaled down compared to the text of other languages. This may cause some overflowing text or readability issues in some UIs until the rework, but fixes many issues of misalignment and inaccurate word wrapping values
- Fixed most (if not all) instances of non readable text colors in some UIs, especially in light mode

Dev side:
- Regroup the various font style (sizes, shadow, etc) used in the game under a single enum
- Rewrote the TextStyles to be a combination of colors (dependent on theme) and font styles (dependent on language)

- The way the TextStyle enum is handled by the ui makes it very hard to use. Color information and text size and style are currently handled in different methods in giant switch cases, making it impossible to know at a glance what a specific Text Style entails.
- There is no uniformity of text sizes across the UI, which causes several issues:
   - no coherence in the text styles in the UI. Some screens like starter selection use a huge variety of text sizes
   - because we are using pixel fonts, they are not meant to be displayed in a lot of different sizes, resulting in blurry anti aliased text in a lot of places
- Even though TextStyle encode information on both color and text size and shadow placement, many UIs use TextStyle only to represent text color, and create a text object with a text style that matches the color they want, then alter its size and sometimes other attributes. This creates even more style inconsistency because not all TextStyles use the same font. It also messes up the way the text shadow looks for these text objects

- All text sizes in the game are scaled by 1/6, except Japanese that is scaled by 5/36. While I am not totally sure why this was done, this difference is causing issues all over the place
   - Japanese text is blurry even at sizes where other languages text is not
   - The cursor and text in Mystery Encounters are misaligned in Japanese
   - All spots in the UI that set a specific word wrap width are also wrong in Japanese if they don't take this specific case into account
   - Many places in the code need to specify a different line spacing for Japanese because of some other Japanese-specific code when text objects are created

Removing that scaling does not seem to be causing any major issues at a glance and keeping it is not sustainable, so this is being removed. We should strive to make the future UI readable and with more breathing room in all languages, without resorting to such messy solutions.

The main goal of this PR is to prepare the terrain for changing font and limiting the possible font sizes that the UI will use to sizes where the font looks crisp by cleaning up the TextStyle-related code. The actual font sizes and UI designs are mostly unaltered except in a few select places, **this will all have to be done when designing the new UI, along with the game resolution and font changes**.

These seem like a lot of changes, but most added lines are just regrouping TextStyle and FontStyle information in a single place. The rest of it is updating the UIs to use the new TextStyle. The most important changes are:

- `text.ts`: remove the giant font style information switch case
- `font-style.ts`: create FontStyle enum and FontStyleOptions to hold information on font sizes, shadow and stroke. Regroup currently used font sizes in that enum.
- `text-style.ts`: instead of doing it in separate places, in this file each TextStyle is associated with:
   - a text color, or associations of `ui theme -> text color`
   - a font style, or associations of `language -> font style`

This is used automatically to retrieve and create the text objects from a TextStyle, based on the current UI Theme and language, like before
- All instances of text objects setting their own fontsize have been replaced with creating a new FontStyle and/or TextStyle if needed, and using that. *Those specific font styles are not meant to stay in the long run, but the general structure will. Having done that inventory of all font sizes and styles should make cleaning them up much easier when we get there*

Japanese text changes:
- removed the Japanese text scale exception, create a universal TEXT_SCALE constant
- use TEXT_SCALE wherever a word wrap width number is set, so that it stays correct if/when we change that value for a potential new font in the future

## Screenshots/Videos

<sub>please excuse the blurriness of the text in some of these as usual I'm running the game at a lower scale to spare my dying computer

<details><summary>Text color fixes for light mode - Egg Summary</summary>

Before, light theme:
image
Before, dark theme:
image

After, light theme:
image
After, dark theme:
image

After, IV text colors still show properly in catch panel despite the change in egg summary:
image
image

</details>

<details><summary>Japanese text scaling</summary>

Dialogue before/after:
image
image

Settings before/after:
image
image

ME option cursor placement before/after:
image
image

</details> 

## How to test the changes?

Run the game either in dark or light mode, most things should be unchanged.
Can try running it in Japanese to see how things look.
Can also try languages like Brazilian Portuguese or German to check that some UIs still have different text sizes in those languages

## Checklist


src/enums/font-style.ts
```diff
@@ -0,0 +1,32 @@
+export enum FontStyle {
+  // note: please don't judge me for the naming of these, they will not stay when we rework the UI
```
Tempo-anon, line 2: No judgment here. Ideally this would be something documented on a wiki or something with examples of each one but at least the each of these names is pretty clear on what they mean

src/ui/form-change-scene-handler.ts
```diff
@@ -45,7 +45,7 @@ export default class FormChangeSceneHandler extends MessageUiHandler {
     const message = addTextObject(0, 0, "", TextStyle.MESSAGE, {
       maxLines: 2,
       wordWrap: {
-        width: 1780,
+        width: (GAME_WIDTH - 24) * TEXT_SCALE,
```
Tempo-anon, line 48: Should have a comment here explaining why it's -24

src/ui/mystery-encounter-ui-handler.ts
```diff
@@ -419,6 +436,7 @@ export default class MysteryEncounterUiHandler extends UiHandler {
       }
 
       if (text) {
+        optionText.setLineSpacing(-8); // TODO: this is broken
```
Tempo-anon: what does "broken" mean? Like does it not show up at all? Or does it just look bad?

src/ui/registration-form-ui-handler.ts
```diff
@@ -48,10 +39,14 @@ export default class RegistrationFormUiHandler extends FormModalUiHandler {
       error = error.slice(0, colonIndex);
     }
     switch (error) {
```
Tempo-anon, line 41: nit and out of scope for this PR, but using strings to decide code flow always felt weird to me. Is there a way to not use strings but something else like a variable?

Tempo-anon (dismissed):
I don't know much about UI but this looks to be mainly refactors. Pulled the code and played a few waves, looked ok

src/ui-constants.ts
```diff
@@ -17,3 +17,13 @@ export const CANVAS_SCALE = 6;
...
+
+/**
+ * This is used as an internal value to scale text objects so that the end result looks crisp.
+ * This value should not be modified unless the fonts and font styles used are modified.
+ *
+ * This scale should be used when setting a text object's wrapping width. For example,
+ * for a text object that should be at most 100 pixels wide (not taking in account CANVAS_SCALE),
+ * the wrap width should be set to `100 * TEXT_SCALE`.
+ */
+export const TEXT_SCALE = 6;
```
torranx, line 33: how did we come up with `6`? is this the same value used in `TEMP_SCALE_ADJUSTEMENT` above (also I think that should be `ADJUSTMENT`?)

src/ui/ball-ui-handler.ts
```diff
@@ -20,8 +20,7 @@ export default class BallUiHandler extends UiHandler {
 
   private cursorObj: Phaser.GameObjects.Image | null;
 
-  // TODO scaling: find a way to improve this. currently needed for japanese
-  private scale: number = 0.1666666667;
+  private scale: number = 1 / TEXT_SCALE;
```
torranx, line 23: the comment `currently needed for japanese` is helpful still imo

src/enums/font-style.ts
MokaStitcher: Oh yeah that's awful but I think right now the string is what the server returns instead of an error code, so we have to work with that :')

src/ui/form-change-scene-handler.ts
MokaStitcher: Margin of 12 pixels on each side of the text
All those actual numbers are subject to change when the UI gets reworked so for now I don't think there's much benefit to making note of them
The whole thing is full of magic numbers (which is kind of expected with UI but, it's more than what is strictly needed in this case)

edit, more details on how I got to these values: https://github.com/Despair-Games/poketernity/pull/808#issuecomment-2709039153

torranx (commented):
37/55 files viewed.

how did you come up with the numbers/equations found in {variable} +/-/*// {number} ie `(GAME_WIDTH - 24) * TEXT_SCALE`?

src/ui/mystery-encounter-ui-handler.ts
MokaStitcher: means it was broken in Japanese and I didn't finish that sentence apparently lol
And also that I did not remove the comment once it was fixed by removing the Japanese text scaling, so I will do that

src/ui/ball-ui-handler.ts
MokaStitcher: No, I removed the Japanese part of the issue in this PR
This scale is where the 6 comes from
And it's needed for the text to look crisp

Like, we set the font size to 96 and the scale to 1/6, and the text looks crisp
You'd think simply setting the font size to 16 with no scaling would do the thing, but if you do that the text is *extremely* blurry and almost unreadable
I tested it with many other values and only the 6 worked
I'll try to figure out exactly why when changing font but for now that's what has to be done to get things working with the current font

src/ui-constants.ts
MokaStitcher: See the other comment for where the 6 comes from
Good catch for adjustment. These words always get me xd

src/ui/ball-ui-handler.ts
torranx: > You'd think simply setting the font size to 16 would work, but if you do that the text is extremely blurry and almost unreadable
I tested it with many other values and only the 6 worked

can we place this in the comments?

MokaStitcher:
> 37/55 files viewed.
> 
> how did you come up with the numbers/equations found in {variable} +/-/*// {number} ie `(GAME_WIDTH - 24) * TEXT_SCALE`?

This is the width at which the text should wrap around
So say the value before was 1870
I know this is scaled up by 6 because of the TEXT_SCALE I explained in other comments. So the actual number we care about is 1870/6 = 311,66666... I round it up to 310 or something

Now in actuality what that number means is "the text should take almost the full width of the game, minus some margin"
So instead of putting `310 * TEXT_SCALE` I put the GAME_WIDTH in the equation to hopefully make a bit clearer where the number comes from. Since the game's width is 320 it ends up being `(GAME_WIDTH - 10) * TEXT_SCALE`
This way when we change the width there is one less magic number to update / try to interpret 

I hope that makes sense (it's possible I messed up some of these conversations though, I'll try and make another pass on them tomorrow with a fresh mind)

src/ui/registration-form-ui-handler.ts
MokaStitcher: Oh yeah that's awful but I think right now the string is what the server returns instead of an error code, so we have to work with that :')

src/enums/font-style.ts
MokaStitcher: I definitely plan on making a guide to working with the UI once all is said and done, and that would be part of it
Artists would also need to know all the font sizes they can use when designing UIs, which is my next task to figure out, along with the new font

(And those names won't stay since we'll be removing the font and pruning down all these font sizes to a much more manageable number)

src/ui/ball-ui-handler.ts
MokaStitcher: I kind of did in the const's documentation already, but I'll try to give more details
But the next PR is to change the font, and I have some amount of hope that we won't need this anymore with the new font

src/ui/pokemon-info-container.ts
```diff
@@ -55,18 +56,21 @@ export default class PokemonInfoContainer extends Phaser.GameObjects.Container {
...
+  constructor(x: number = 372, y: number = 66, defaultTextStyle: TextStyle = TextStyle.CATCH_PANEL_LABEL) {
     super(globalScene, x, y);
     this.initialX = x;
+    this.defaultTextStyle = defaultTextStyle;
   }
 
   setup(): void {
     this.setName("pkmn-info");
-    const currentLanguage = i18next.resolvedLanguage!; // TODO: is this bang correct?
+    const currentLanguage = i18next.resolvedLanguage ?? DEFAULT_LANGUAGE_KEY; // TODO: is this bang correct?
```
torranx:
```suggestion
    const currentLanguage = i18next.resolvedLanguage ?? DEFAULT_LANGUAGE_KEY;
```

PR #813, opened 2025-03-10
[Misc] Remove unused `target` and `move` params from anonymous function in `IncrementMovePriorityAttr`

DayKev:
N/A

Removing a `// @ts-ignore`.

`null` is now valid to be passed in as the `target` param of `IncrementMovePriorityAttr.apply()`.

## Checklist

### DIFF
--- a/src/data/move-attrs/increment-move-priority-attr.ts
+++ b/src/data/move-attrs/increment-move-priority-attr.ts
@@ -3,19 +3,20 @@ import type { NumberHolder } from "#app/utils";
 import type { Move } from "#app/data/move";
 import { MoveAttr } from "#app/data/move-attrs/move-attr";
 
+type MoveIncrementFunc = (pokemon: Pokemon, target: Pokemon | null, move: Move) => boolean;
+
 /**
  * Attribute used for moves that change priority in a turn given a condition.
  * Used for {@link https://bulbapedia.bulbagarden.net/wiki/Grassy_Glide_(move) | Grassy Glide}.
  * @extends MoveAttr
  */
-
 export class IncrementMovePriorityAttr extends MoveAttr {
   /** The condition for a move's priority being incremented */
-  private moveIncrementFunc: (pokemon: Pokemon, target: Pokemon, move: Move) => boolean;
+  private moveIncrementFunc: MoveIncrementFunc;
   /** The amount to increment priority by, if condition passes. */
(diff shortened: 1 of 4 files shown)


src/data/move-attrs/increment-move-priority-attr.ts
```diff
@@ -3,19 +3,20 @@ import type { NumberHolder } from "#app/utils";
 import type { Move } from "#app/data/move";
 import { MoveAttr } from "#app/data/move-attrs/move-attr";
 
+type MoveIncrementFunc = (pokemon: Pokemon, target: Pokemon | null, move: Move) => boolean;
```
torranx: I think you can even remove the `target` and`move` totally since the condition function only uses `pokemon`

torranx (dismissed):

test/moves/grassy_glide.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    game.override.ability(terrainAbility);
+    await game.classicMode.startBattle([Species.SHUCKLE]);
+
+    game.move.use(MoveId.GRASSY_GLIDE);
+    await game.toEndOfTurn();
+
+    expect(game.scene.arena.getTerrainType()).toEqual(terrainType);
+    expect(game.field.getSpeedOrder()).toEqual([BattlerIndex.ENEMY, BattlerIndex.PLAYER]);
+    expect(game.field.getTurnOrder()).toEqual(game.field.getSpeedOrder());
+  });
```
PigeonBar, line 65:
```suggestion
  });

  it("should not have increased priority if the user in ungrounded in Grassy Terrain", async () => {
    game.override.ability(Abilities.GRASSY_SURGE);
    await game.classicMode.startBattle([Species.VESPIQUEN]);

    game.move.use(MoveId.GRASSY_GLIDE);
    await game.toEndOfTurn();

    expect(game.scene.arena.getTerrainType()).toEqual(TerrainType.GRASSY);
    expect(game.field.getSpeedOrder()).toEqual([BattlerIndex.ENEMY, BattlerIndex.PLAYER]);
    expect(game.field.getTurnOrder()).toEqual(game.field.getSpeedOrder());
  });
```

PigeonBar (approved):

Tempo-anon (approved):
This attr can also just be named GrassyGlideAttr 🤷

PR #820, opened 2025-03-12
[Move] Implement Me First

innerthunder:

[Me First](https://bulbapedia.bulbagarden.net/wiki/Me_First_(move)) is now fully implemented.

The new turn manager and `CallMoveAttr` refactors made this easy 😄 

- Added new `MeFirstAttr` move attribute covering Me First's full functionality
- Added new battler tag `MeFirstPowerBoostTag` to boost the follow-up move's power after Me First is used

## Screenshots/Videos

Provided on request

## How to test the changes?

`npm run test me_first`

## Checklist


src/data/moves/move-attrs/me-first-attr.ts
```diff
@@ -0,0 +1,58 @@
...
+  MoveId.METAL_BURST,
+  MoveId.METRONOME,
+  MoveId.MIRROR_COAT,
+  MoveId.MIRROR_MOVE,
+  MoveId.NATURE_POWER,
+  MoveId.SHELL_TRAP,
+  MoveId.SLEEP_TALK,
+  MoveId.SNATCH,
+  MoveId.STRUGGLE,
+  MoveId.THIEF,
```
Tempo-anon: Minor nit to add a comment just to make clear
```
Moves that call other moves are considered to be the original status move (not the move it calls), so Me First will always fail if the target has selected one of those moves
```

innerthunder: I added a comment to the test file

src/data/moves/move-attrs/me-first-attr.ts
```diff
@@ -0,0 +1,59 @@
...
+      user.addTag(BattlerTagType.ME_FIRST_POWER_BOOST);
+      return super.apply(user, target, targetMove, overridden);
+    }
+    return false;
+  }
+
+  override getCondition(): MoveConditionFunc {
+    return (_user, target, _move) => {
+      const targetMove = this.getTargetMove(target);
+      return !!targetMove && !invalidMeFirstMoves.has(targetMove.id);
```
PigeonBar:
```suggestion
      return !!targetMove?.isAttackMove() && !invalidMeFirstMoves.has(targetMove.id);
```
According to Bulbapedia, Me First fails if the target selected a non-damaging move.

test/moves/me-first.test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    const player = game.field.getPlayerPokemon();
+    const enemy = game.field.getEnemyPokemon();
+
+    game.setTurnOrder([BattlerIndex.PLAYER, BattlerIndex.ENEMY]);
+    game.move.use(MoveId.ME_FIRST);
+
+    await game.toEndOfTurn();
+
+    expect(player.getLastXMoves()[0]?.move.id).toBe(MoveId.SWORDS_DANCE);
```
PigeonBar: This should also be converted into a test that Me First fails if the opponent selected Swords Dance.

src/data/moves/move-attrs/me-first-attr.ts
```diff
@@ -0,0 +1,59 @@
...
+import { BattlerTagType } from "#enums/battler-tag-type";
+import { MoveId } from "#enums/move-id";
+
+/**
+ * Attribute to copy the target's selected (and not yet used)
+ * move for the turn and use it against the target.
+ * Used for {@link https://bulbapedia.bulbagarden.net/wiki/Me_First_(move) | Me First}.
+ * @extends CallMoveAttr
+ */
+export class MeFirstAttr extends CallMoveAttr {
```
PigeonBar, line 21:
```suggestion
export class MeFirstAttr extends CallMoveAttr {
  constructor() {
    super();
    this.hasTarget = true;
  }

```
(So that Me First hits the chosen target instead of choosing a random target)

test/moves/me-first.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    await game.move.forceEnemyMove(MoveId.TACKLE);
+
+    await game.phaseInterceptor.to("MoveEndPhase");
+    expect(tackle.calculateBattlePower).toHaveReturnedWith(60);
+
+    await game.phaseInterceptor.to("MoveEndPhase");
+    expect(tackle.calculateBattlePower).toHaveReturnedWith(40);
+  });
+
+  it("should put the user in a frenzy if Outrage is copied", async () => {
```
Tempo-anon: nit: instead of Outrage in particular, the test name should indicate it's testing for frenzy or move lock moves

test/moves/me-first.test.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    game.setTurnOrder([BattlerIndex.PLAYER, BattlerIndex.ENEMY]);
+    game.move.use(MoveId.ME_FIRST);
+    await game.move.forceEnemyMove(moveId);
+
+    await game.phaseInterceptor.to("MoveEndPhase");
+
+    expect(player.getLastXMoves()[0]?.result).toBe(MoveResult.FAIL);
+  });
+});
```
Tempo-anon, line 162: Bulbapedia also claims that
> Unlike other moves, if the target of Me First is a vacant slot or a Pokémon that already fainted before the Me First user's turn, Me First will fail instead of being redirected to an adjacent non-fainted opponent. 

There should be a test to account for it not redirecting

innerthunder: I wasn't sure if this should actually be implemented. It seems like a bug lol

PigeonBar (dismissed):

test/moves/me-first.test.ts
Tempo-anon: Should add a comment for it I guess. What does showdown do?

innerthunder: Showdown redirects it like other moves https://replay.pokemonshowdown.com/gen9nationaldexdoubles-2327396017-e7u72xf7qfk7on8ai72mc4wl2srx8ktpw

src/data/battler-tags/move-lock-tag.ts
```diff
@@ -61,7 +61,9 @@ export abstract class MoveLockTag extends BattlerTag {
     if (ret) {
       const move = allMoves.get(this.sourceMoveId);
       this.lastTargets = lastTargets;
-      pokemon.getMoveQueue().push({ move, targets: [], ignorePP: true, type: pokemon.getMoveType(move) });
+      pokemon
+        .getMoveQueue()
+        .push({ move, targets: [], ignorePP: true, type: pokemon.getMoveType(move), virtual: true });
```
DayKev, line 66: This is for Rollout/Outrage/etc isn't it? The forced turns count as virtual moves?

src/data/moves/move-attrs/me-first-attr.ts
```diff
@@ -0,0 +1,60 @@
...
+    };
+  }
+
+  protected getTargetMove(target: Pokemon): Move | undefined {
+    const { turnManager } = globalScene.currentBattle;
+    return turnManager.findCommandFromPokemon(target)?.turnMove?.move;
+  }
+}
+
+export const invalidMeFirstMoves: Set<MoveId> = new Set([
```
DayKev:
```suggestion
export const invalidMeFirstMoves: ReadonlySet<MoveId> = new Set([
```

src/data/moves/move.ts
```diff
@@ -852,6 +854,8 @@ export abstract class Move implements Localizable {
       power.value *= 1.5;
     }
 
+    applyBattlerTags<MeFirstPowerBoostTag>(BattlerTagType.ME_FIRST_POWER_BOOST, source, simulated, power);
```
DayKev, line 867: Is this the right timing for this?

innerthunder: According to https://bulbapedia.bulbagarden.net/wiki/Power, both this and the Helping Hand power boost are meant to apply earlier in the calculation. Should these be rearranged here? I figure keeping the two tags together makes sense for now.

DayKev: It's best to try and follow the official calculation order (to prevent more issues like the recently fixed Pixilate bug).

src/data/battler-tags/move-lock-tag.ts
innerthunder: If they're not virtual, then the forced turns don't execute when the moves are copied by Me First (and Metronome, etc.). `CommandPhase` automatically deletes any moves in the move queue that aren't virtual and aren't in the Pokemon's moveset, iirc.

src/data/moves/move.ts
innerthunder: Yes, but a lot of things are out of order in this function... The link also only covers multipliers, so I'm not sure how much that given order matters

DayKev: True, we just need to make sure the things that aren't multipliers are in the right spot.

innerthunder: Seems like it's fine wherever as long as it's placed after `VariablePowerAttrs` / `VariableMovePowerAbAttrs` are applied. I'll just leave it for now.

src/data/battler-tags/move-lock-tag.ts
DayKev: Well, as long as it doesn't cause issues for things that check move history then it's fine.

src/data/moves/move-attrs/me-first-attr.ts
```diff
@@ -0,0 +1,62 @@
...
+    return turnManager.findCommandFromPokemon(target)?.turnMove?.move;
+  }
+}
+
+export const invalidMeFirstMoves: ReadonlySet<MoveId> = Object.freeze(
+  new Set([
+    MoveId.BEAK_BLAST,
+    MoveId.BELCH,
+    MoveId.CHATTER,
+    MoveId.COUNTER,
```
PigeonBar, line 60:
```suggestion
    MoveId.COMEUPPANCE,
    MoveId.COUNTER,
```
Also, should all the Max Moves be uncopiable by Me First? (Me First only existed up to gen 7, so Game Freak never dealt with this interaction.)

DayKev: Max moves should be handled the same as Gamefreak handled Z moves. If it can't copy Z moves, it shouldn't copy Max moves.

Tempo-anon: but here G-Max moves are treated as normal moves, just that they get a bonus in damage when used by a max'd mon. No strong opinions one way or the other if they should be included

innerthunder: They should not be included after https://github.com/Despair-Games/poketernity/pull/820/commits/37cda212ad76000677667088c942c603fd274263

src/data/moves/move-attrs/me-first-attr.ts
```diff
@@ -0,0 +1,68 @@
...
+    };
+  }
+
+  protected getTargetMove(target: Pokemon): Move | undefined {
+    const { turnManager } = globalScene.currentBattle;
+    return turnManager.findCommandFromPokemon(target)?.turnMove?.move;
+  }
+}
+
+export const invalidMeFirstMoves: ReadonlySet<MoveId> = Object.freeze(
```
DayKev:
```suggestion
const invalidMeFirstMoves: ReadonlySet<MoveId> = Object.freeze(
```

src/data/moves/move-attrs/me-first-attr.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * NOTE: Unlike in mainline, Me First is currently redirected to the target's ally
+ * if the target faints or the target's index is otherwise vacant. This is intentional;
+ * we'd rather keep move behavior consistent than replicate what is likely a bug.
+ * @extends CallMoveAttr
+ */
+export class MeFirstAttr extends CallMoveAttr {
+  constructor() {
+    super();
+    this.hasTarget = true;
```
DayKev, line 24:
```suggestion
    this.hasTarget = true;
    this.invalidMoves = invalidMeFirstMoves;
```

src/data/moves/move-attrs/me-first-attr.ts
```diff
@@ -0,0 +1,68 @@
...
+    return false;
+  }
+
+  override getCondition(): MoveConditionFunc {
+    return (user, target, _move) => {
+      const targetMove = this.getTargetMove(target);
+      return (
+        !!targetMove?.isAttackMove()
+        && !targetMove.checkFlag(MoveFlags.G_MAX_MOVE, user, target)
+        && !invalidMeFirstMoves.has(targetMove.id)
```
DayKev:
```suggestion
        && !this.invalidMoves.has(targetMove.id)
```

DayKev (dismissed):

PR #839, opened 2025-03-15
[BUG] Fix Locked Reroll not using Luck Score

### LINKED ISSUE #526: [Bug] Formula used to calculate the chance of items upgrading in the shop is not the same with and without Lock Rarities selected
Copied over from PKR 4308

### Describe the bug

When item tier upgrade count is calculated for items initially presented in the shop and whenever the shop is rerolled without Lock Rarities enabled the following code is run (Starting at line 2151 in src/modifier/modifier-type.ts)
```ts
    if (player && tierValue) {
      const partyLuckValue = getPartyLuckValue(party);
      const upgradeOdds = Math.floor(128 / ((partyLuckValue + 4) / 4));
      let upgraded = false;
      do {
        upgraded = Utils.randSeedInt(upgradeOdds) < 4;
        if (upgraded) {
          upgradeCount++;
        }
      } while (upgraded);
    }
```
Whereas when item tier upgrade count is calculated whenever the shop is rerolled with Lock Rarities enabled it instead runs the next else if block which uses the following code to calculate upgrade count (Starting at line 2184 in src/modifier/modifier-type.ts)
```ts
    if (tier < ModifierTier.MASTER) {
      const partyShinyCount = party.filter(p => p.isShiny() && !p.isFainted()).length;
      const upgradeOdds = Math.floor(32 / ((partyShinyCount + 2) / 2));
      while (modifierPool.hasOwnProperty(tier + upgradeCount + 1) && modifierPool[tier + upgradeCount + 1].length) {
        if (!Utils.randSeedInt(upgradeOdds)) {
          upgradeCount++;
        } else {
          break;
        }
      }
      tier += upgradeCount;
    }
```

### Session export file

_No response_

### User data export file

_No response_

### Expected behavior

It seems odd that this is written using essentially two completely different methods. The first method uses the total party luck value that is based on the total value of all your shinies added together whereas the second only takes into account the number of your party pokemon that are shiny. In my opinion this should probably be changed so that both places use the first method so that reroll upgrade chances are the same.

### Screenshots / Videos

_No response_

### What OS did you observe the bug on?

PC/Windows

### If other please specify

_No response_

### Which browser do you use?

Chrome

### If other please specify

_No response_

### Additional context

After finding this bug while looking at the code I cloned main to my pc and ran it locally to test and make sure I was reading the code correctly below you can see the break points that I put in to see which section of code was running in each case

When rerolling without Lock Rarities only the first breakpoint is hit
image

When rerolling with Lock Rarities only the second breakpoint is hit
image

SangaraSorama:

When locking your reroll using Lock Capsule, it now correctly uses your Luck Value

It is currently simply counting the number of shiny pokemon you have

From : https://github.com/pagefaultgames/pokerogue/pull/5502

Fixes #526

Stop using partyShinyCount and instead use getPartyLuckValue.

```typescript
Math.floor(32 / ((partyShinyCount + 2) / 2));
```

VS

```typescript
Math.floor(128 / ((partyLuckValue + 4) / 4));
```

## Screenshots/Videos

N/A

## How to test the changes?

Check that locked rerolls are giving the proper upgrades. Testing without automation is impossible to check without just "feeling lucky"

## Checklist

### DIFF
--- a/src/modifier/modifier-type.ts
+++ b/src/modifier/modifier-type.ts
@@ -2078,10 +2078,10 @@ function getNewModifierTypeOption(
   } else if (upgradeCount === undefined && player) {
     upgradeCount = 0;
     if (tier < ModifierTier.MASTER && allowLuckUpgrades) {
-      const partyShinyCount = party.filter((p) => p.isShiny() && !p.isFainted()).length;
-      const upgradeOdds = Math.floor(32 / ((partyShinyCount + 2) / 2));
+      const partyLuckValue = getPartyLuckValue(party);
+      const upgradeOdds = Math.floor(128 / ((partyLuckValue + 4) / 4));
       while (modifierPool.hasOwnProperty(tier + upgradeCount + 1) && modifierPool[tier + upgradeCount + 1].length) {
-        if (!randSeedInt(upgradeOdds)) {
+        if (randSeedInt(upgradeOdds) < 4) {
           upgradeCount++;
         } else {
           break;


PigeonBar (approved):
These changes worked for me, although I wonder if it would be better to create a new helper function to perform this while loop so that there is less repeated code between using or not using the Lock Capsule.

DayKev (approved):

PR #843, opened 2025-03-15
[UI/UX] Autofill login text color adapts to UI theme

### LINKED ISSUE #547: [Bug] Autofilling username and password appears with black text, making it hard to read on the background
Copied over from PKR 4768

### Describe the bug

When you type text into the username / password fields to log in, the text is white. However, if you log in and use the autofill after saving your username / password, it appears in black / dark grey instead which is quite hard to read. The font also looks different (smaller? not sure if it's a different font).

### Reproduction

1. launch pokerogue
2. click username field
3. select "autofill" option from dropdown

### Expected behavior

I would expect consistent styling for username and password text regardless of whether you choose to type it manually or use autofill. 

### Screenshots / Videos

Typing normally: 
Screenshot 2024-10-30 095315

Autofill: Screenshot 2024-10-30 094953

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

SangaraSorama:

When a user autofills their username and password on the login page, the text will be gray instead of the browser's default black.

The black text on dark grey is bad for accessibility. It also stops you from being able to confirm if the autofill is correct, especially if you changed your information recently, because you can't actually read what got filled out.

From : [[UI/UX] Autofill login text color changed to gray #5523](https://github.com/pagefaultgames/pokerogue/pull/5523) (mirror of [#4473](https://github.com/pagefaultgames/pokerogue/pull/4773))

Fixes #547

Added a new CSS property to index.css for the webkit text fill colour in autofill inputs

(there was no code for this before, so it's an addition as opposed to a modification)

## Screenshots/Videos
Screenshots directly from the pkr PRs

<details><summary>Before</summary>
<p>

image

</p>
</details> 

<details><summary>After (default)</summary>
<p>

login-autofill-gray_default

</p>
</details>  

<details><summary>After (light)</summary>
<p>

login-autofill-gray_light

</p>
</details> 

## How to test the changes?

In `env.development` change the first line to `VITE_BYPASS_LOGIN=0` (to get the login screen on a local instance), then try to autofill a password.
The reviewer will need to use the browser default autofill or some other plugin.

## Checklist


MokaStitcher:
Alright, so like I explained in the original PR, the ideal solution would be to define a specific color depending on the theme.
Thankfully, the CSS code already gets access to the selected ui theme and window type. At the moment the UI Theme is not being used in there, only the window type, but both are set and updated as needed at runtime. Both would be used in the same way.

<details><summary>Here's how this is done currently</summary>

Here's how the CSS uses the window type to set specific colors that are then used for the touch control buttons
https://github.com/Despair-Games/poketernity/blob/f325f1b19d1ea15df503ecbd41f6df2f73a9296d/index.css#L22-L32

Here's how these colors are used to set the color of the touch control buttons, using `var()`
https://github.com/Despair-Games/poketernity/blob/f325f1b19d1ea15df503ecbd41f6df2f73a9296d/index.css#L258-L260

Here's how the game code sets the data-window-style value so that it is accessible from the CSS:
https://github.com/Despair-Games/poketernity/blob/f325f1b19d1ea15df503ecbd41f6df2f73a9296d/src/touch-controls.ts#L198

</details>

Now, if we look at the line just above, we see that the ui theme is also set in the same way
https://github.com/Despair-Games/poketernity/blob/f325f1b19d1ea15df503ecbd41f6df2f73a9296d/src/touch-controls.ts#L197

Which means, we can do the same thing that is done with the different window styles, but using `data-ui-theme` instead of `data-window-style`. We can define a color variable that will depend on the UI theme and be used for the autofilled text, and use it instead of a single middle of the road color.

There is a bit more complexity than that and following this suggestion will not fully work out of the box. But it's a good start to get this done, then figure out how to test it in game, and then we can work through how to fix any issues with it.

If using chrome, you can just click the little key in the address bar to register username/password, to be able to test it
image
If using another browser I don't know, but they all have a saving credentials option

</details>

flx-sta:
Adjusted the PR so it reacts to light/dark mode as suggested by @MokaStitcher 

<p>

### Dark Mode (Default)

image

### Light Mode

image

</p>
</details>

MokaStitcher:
> Adjusted the PR so it reacts to light/dark mode as suggested by @MokaStitcher

I hadn't thought of using the `color-light` and `color-dark` variables. That's a good idea but unfortunately I don't think this fully works because `--color-light` and `--color-dark` are changed based on the window type set in settings. The thing is the window type is only set in the document when using touch controls because that's the only situation where the colors are used at the moment (for the touch controls colors) so you wouldn't notice it testing normally

<details><summary>screenshots take too much room</summary>

image

Based on this I thought it would make the text bright blue for the teal windows, but apparently not it seems to fully break which is a bit weird. But it works fine when not in mobile mode so I'm almost certain the issue comes from that.

image

</details> 

I'm also not sure I like having main.ts accessing the settings, I'm a bit concerned about that.
I ended up pushing my own version of that fix for reference #897  (I stole your addition of `:autofill` :p ). It is in general more consistent with what was already done in the codebase (there was already code setting the ui theme, just it is only being called on devices with a touchscreen).
I don't know which approach is preferable so it's mostly for reference. Feel free to update this PR to whatever your think is best. The only thing I don't want to lose from that other PR is the addition of the caret color as well (and that it works on mobile lol).

flx-sta:
> I'm also not sure I like having main.ts accessing the settings, I'm a bit concerned about that.

Don't worry about that. My whole SettingsManager PR was all about having no issue with that as it's disconnected from the whole game logic a simple state (manager)

flx-sta:
Closing this for now. 
**Please reopen when work continues!**

PR #846, opened 2025-03-15
[Refactor] Rewrite biomes⚠️

Tempo-anon:

* A lot of biomes have been replaced with filler values and some have been temporarily removed

Complete rewrite of biomes. Hopefully this makes implementing new biomes in the future easier.

* Deleted a bunch from `biomes.ts`
* Created a `PEBiome` class (name pending)
* Created 3 temporary PEBiomes, Town, Cave, and Volcano
* Added an `allBiomes` to `data-lists.ts` that gets initialized in the new `init-biomes.ts` file
* Because I removed `biomes.biomeLinks`, I had to change some code in `encounter-phase-utils.ts` and `select-biome-phase.ts` Maybe it's okay to keep biomeLinks?

## Checklist


src/data/balance/biomes/PEBiome.ts
```diff
@@ -0,0 +1,44 @@
+import { Biome } from "#enums/biome";
+import type { BiomePoolTier } from "#enums/biome-pool-tier";
+import type { Species } from "#enums/species";
+import type { TerrainType } from "#enums/terrain-type";
+import type { TimeOfDay } from "#enums/time-of-day";
+import type { TrainerType } from "#enums/trainer-type";
+import type { WeatherType } from "#enums/weather-type";
+
+/**
+ * Calling it PEBiome for now to distinguish from Biome.ts and Biome the enum
```
DayKev: The `Biome` enum should probably be renamed to `Biomes` or `BiomeId`.

Tempo-anon: Yes it probably should. Some of the enums (like Species) really need renaming

src/data/balance/biomes/PEBiome.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+export class PEBiome {
+  public biomeType: Biome;
+  // TODO: consider `partial` and `Omit`
+  public pokemonPool: Record<BiomePoolTier, Record<TimeOfDay, Species[]>>;
+  public trainerPool: Record<BiomePoolTier, TrainerType[]>;
+  public weatherPool: Record<WeatherType, number>;
+  public terrainPool: Record<TerrainType, number>;
+  public outgoingPaths: Partial<Record<Biome, number>>;
```
Tempo-anon: Maybe I should just keep all biome links in its own file instead of on each individual Biome. Would certainly save me some headache in the places it's being used

src/data/balance/biomes/PEBiome.ts
Tempo-anon: #871

src/battle-scene.ts
```diff
@@ -2032,29 +2032,11 @@ export default class BattleScene extends SceneBase {
...
-        return biome;
-      }
-    }
-
-    return biomes[randSeedInt(biomes.length)];
+  /**
+   * TODO: Rewrite this later for weighting?
+   */
+  generateRandomBiome(_waveIndex: number): Biome {
+    return Array.from(allBiomes.keys())[randSeedInt(allBiomes.keys.length)];
```
DayKev, line 2039:
```suggestion
    return randSeedItem(allBiomes);
```

src/data/balance/biomes/biome-links.ts
```diff
@@ -0,0 +1,42 @@
+import { Biome } from "#enums/biome";
+
+interface BiomeLinks {
+  [key: number]: Biome | (Biome | [Biome, number])[];
+}
+
+export const biomeLinks: BiomeLinks = {
```
DayKev: This should also be a property of each biome instead of a separate object.

src/data/data-lists.ts
```diff
@@ -2,12 +2,16 @@ import type PokemonSpecies from "#app/data/pokemon-species";
 import { type Move } from "#app/data/moves/move";
 import { type MoveId } from "#enums/move-id";
 import type { Ability } from "#app/data/abilities/ability";
-interface MoveMap<K, V> extends Map<K, V> {
+import type { PEBiome } from "#app/data/balance/biomes/PEBiome";
+import type { Biome } from "#enums/biome";
```
DayKev:
```suggestion
import type { Biome } from "#enums/biome";

```

src/data/balance/biomes/biome-links.ts
```diff
@@ -0,0 +1,42 @@
+import { Biome } from "#enums/biome";
+
+interface BiomeLinks {
+  [key: number]: Biome | (Biome | [Biome, number])[];
+}
```
DayKev: Instead of this mess, it should be simplified (such as like this):
```ts
interface BiomeLinks {
  [key: number /* Biome */]: number;
}

const plainsBiomeLinks: BiomeLinks = {
  [Biome.GRASS]: 1,
  [Biome.METROPOLIS]: 1,
  [Biome.LAKE]: 1,
};

const mountainBiomeLinks: BiomeLinks = {
  [Biome.VOLCANO]: 1,
  [Biome.WASTELAND]: 2,
  [Biome.SPACE]: 3,
};
````

File: src/data/balance/biomes/PEBiome_volcano.ts [hunk 7]
```diff
@@ -0,0 +1,110 @@
...
+import { Species } from "#enums/species";
+import { TerrainType } from "#enums/terrain-type";
+import { TimeOfDay } from "#enums/time-of-day";
+import { TrainerType } from "#enums/trainer-type";
+import { WeatherType } from "#enums/weather-type";
+
+/**
+ * This file represents all the data of the voclano biome
+ */
+
```
DayKev:
```suggestion
```

src/data/balance/biomes/PEBiome_town.ts
```diff
@@ -0,0 +1,110 @@
...
+import { Species } from "#enums/species";
+import { TerrainType } from "#enums/terrain-type";
+import { TimeOfDay } from "#enums/time-of-day";
+import { TrainerType } from "#enums/trainer-type";
+import { WeatherType } from "#enums/weather-type";
+
+/**
+ * This file represents all the data of the town biome
+ */
+
```
DayKev:
```suggestion
```

src/data/balance/biomes/PEBiome_end.ts
```diff
@@ -0,0 +1,128 @@
+import { BiomePoolTier } from "#enums/biome-pool-tier";
+import { Species } from "#enums/species";
+import { TerrainType } from "#enums/terrain-type";
+import { TimeOfDay } from "#enums/time-of-day";
+import { WeatherType } from "#enums/weather-type";
+
+/**
+ * This file represents all the data of the end biome
+ */
+
```
DayKev:
```suggestion
```

src/data/balance/biomes/PEBiome_cave.ts
```diff
@@ -0,0 +1,110 @@
...
+import { Species } from "#enums/species";
+import { TerrainType } from "#enums/terrain-type";
+import { TimeOfDay } from "#enums/time-of-day";
+import { TrainerType } from "#enums/trainer-type";
+import { WeatherType } from "#enums/weather-type";
+
+/**
+ * This file represents all the data of the cave biome
+ */
+
```
DayKev:
```suggestion
```

src/battle-scene.ts
DayKev: Ah wait this isn't an array, one sec.

DayKev: Wait, what exactly is this doing? Just picking any random biome?
If so, this is simpler:
```suggestion
    return randSeedItem([...allBiomes.values()]);
```

Tempo-anon: Should be the `keys()` rather than `values` since `allBiomes` is a map of Biome (the enum) to PEBiome (the class) and this function needs the enum but thanks, this looks simpler

src/data/balance/biomes/biome-links.ts
Tempo-anon: I'm 50/50 on this because I can see arguments for and against. Time to put at TODO and think about it later

src/data/balance/biomes/biome-links.ts
Tempo-anon: Also going to do this later, but agree it should be a `[key: number /* Biome id */]: number`

src/data/balance/biomes/PEBiome.ts
```diff
@@ -0,0 +1,40 @@
+import type { BiomeId } from "#enums/biome-id";
+import type { BiomePoolTier } from "#enums/biome-pool-tier";
+import type { SpeciesId } from "#enums/species-id";
+import type { TerrainType } from "#enums/terrain-type";
+import type { TimeOfDay } from "#enums/time-of-day";
+import type { TrainerType } from "#enums/trainer-type";
+import type { WeatherType } from "#enums/weather-type";
+
+/**
+ * Calling it PEBiome for now to distinguish from Biome.ts and Biome the enum
```
MokaStitcher: what does PE stand for? If Poketernity, let's use PKTY since that's the abbreviation we agreed upon. Although I'd rather something clearer
Is this intended to replace the existing Biome class in the future and be renamed Biome then? Maybe just WIPBiome then

src/data/balance/biomes/PEBiome.ts
```diff
@@ -0,0 +1,40 @@
...
+export class PEBiome {
+  public biomeType: BiomeId;
+  // TODO: consider `partial` and `Omit`
+  public pokemonPool: Record<BiomePoolTier, Record<TimeOfDay, SpeciesId[]>>;
+  public trainerPool: Record<BiomePoolTier, TrainerType[]>;
+  // TODO: weatherPool and terrainPool are currently not used
+  public weatherPool: Record<WeatherType, number>;
+  public terrainPool: Record<TerrainType, number>;
+  public bgm: string;
+  // TODO: image assets should also be in here probably
```
MokaStitcher: actual graphical components/ Phaser GameObjects? Or their keys so that the UI can handle creating the needed gameobjects? (if the later, seems good yes)

src/data/balance/biomes/PEBiome_cave.ts
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * This file represents all the data of the cave biome
+ */
+export const cavePokemonPool = {
+  [BiomePoolTier.COMMON]: {
+    [TimeOfDay.DAWN]: [],
+    [TimeOfDay.DAY]: [],
+    [TimeOfDay.DUSK]: [],
+    [TimeOfDay.NIGHT]: [SpeciesId.ZUBAT],
+    // Instead of a TimeofDay.ALL, should these just be in the other TimeOfDays' lists?
```
MokaStitcher: you mean remove `TimeOfDay.ALL` altogether and put this mons here in `TimeOfDay.DAWN`, `TimeOfDay.DAY`,  `TimeOfDay.DUSK` and `TimeOfDay.NIGHT`?
I'd say no, the separation makes it easier to see things and prevent mistakes when making changes.
However we could do that when parsing this data so that internally the game doesn't have to bother with that ALL array, if it's causing issues

src/data/balance/biomes/biome-links.ts
```diff
@@ -0,0 +1,58 @@
+import { BiomeId } from "#enums/biome-id";
+
+/**
+ * TODO
+ * Having a BiomeId | (BiomeId | [BiomeId, number]) [] type is needlessly complex
+ *
+ * Change this to
+ * ```
+ * interface BiomeLinks {
+ *   [key: number /* BiomeId *\/ ]: number /* Weight *\/;
```
MokaStitcher: is that what had been defined by cat? They had done all this work of redefining a data format and porting existing biome data to it.

src/field/arena.ts
```diff
@@ -182,21 +179,6 @@ export class Arena {
       let species: SpeciesId;
       if (typeof entry === "number") {
         species = entry as SpeciesId;
-      } else {
-        // Try and evolve the Pokemon if possible
```
MokaStitcher, line 186: is this handled elsewhere?

src/data/balance/biomes/PEBiome.ts
```diff
@@ -0,0 +1,40 @@
...
+import type { TimeOfDay } from "#enums/time-of-day";
+import type { TrainerType } from "#enums/trainer-type";
+import type { WeatherType } from "#enums/weather-type";
+
+/**
+ * Calling it PEBiome for now to distinguish from Biome.ts and Biome the enum
+ *
+ * TODO: let's rename this Biome.ts in the future
+ */
+export class PEBiome {
```
MokaStitcher: this is probably missing a mysteryEncounterPool too
https://github.com/Despair-Games/poketernity/blob/8ba589f277b23e80e44cc0a4cb49d65f0a2b3711/src/data/mystery-encounters/mystery-encounters.ts#L189-L191

Not all biome-ME links are defined here, the rest of this is filled at runtime based on biome groupings, like human biomes:
https://github.com/Despair-Games/poketernity/blob/8ba589f277b23e80e44cc0a4cb49d65f0a2b3711/src/data/mystery-encounters/mystery-encounters.ts#L86-L88
https://github.com/Despair-Games/poketernity/blob/8ba589f277b23e80e44cc0a4cb49d65f0a2b3711/src/data/mystery-encounters/mystery-encounters.ts#L147-L149
Biomes can be in multiple of these categories

How would that fit in the rework?

src/data/balance/biomes/PEBiome.ts
Tempo-anon: Yes, this is intended to replace the `Biome.ts` file completely later.

src/data/balance/biomes/PEBiome.ts
Tempo-anon: I'm thinking the string keys for image assets like the background and props. Right now the way images are done is that arenas and props have to be specifically named a certain way which feels a bit clunky.

src/data/balance/biomes/PEBiome_cave.ts
Tempo-anon: Sure, sounds good. The way pokemon generation is done is that it takes the TimeOfDay.ALL array and just concatenates that with the specific timeofday array.

src/data/balance/biomes/biome-links.ts
Tempo-anon: I'm not sure, but it's what Kev suggested and makes the most sense. Right now the key is a BiomeId while the value is either a BiomeId, a list of BiomeId's, or a map of BiomeId->weight which is just unnecessary complexity.

src/field/arena.ts
Tempo-anon: Yes, Pokemon evolving is handled elsewhere. Arenas having code for "evolving" a Pokemon is unnecessary (and was also hardcoded)

src/data/balance/biomes/PEBiome.ts
Tempo-anon: Oh MEs causing headaches again.....

Ideally there should only be a single place the mapping appears, either `biome -> available MEs` or `ME -> array of biomes it can appear in`. I admit I'm not the most familiar with ME code so whichever of the two is easier and makes more sense should be how it's done

src/data/balance/biomes/PEBiome.ts
MokaStitcher: it's still good to have a specific format to enforce some kind of consistency in file naming imo, but there's still probably a good middle ground to reach. Either way I agree with having this information here, as long as it's not actual Phaser GameObjects :p

At the very least biomes should probably hold the information about them having props, rather than that information being stored in yet another biome array in another file https://github.com/Despair-Games/poketernity/blob/8ba589f277b23e80e44cc0a4cb49d65f0a2b3711/src/field/arena.ts#L1027-L1029

src/data/balance/biomes/PEBiome.ts
MokaStitcher: I think like other places when it comes to biomes, both mappings exists. Biome -> ME and ME -> Biome, one being generated based on the other. The second one was possibly wanted by people to be able to see easily which in biomes a given ME appears, but I'm not 100% sure anymore. There's definitely value to that mapping when creating a ME and trying to balance things out.
But code wise I think it makes the most sense to have Biome -> ME links, because when generating a random ME the code will start by looking at the current Biome.
So, both options have pros and cons

src/data/balance/biomes/PEBiome.ts
MokaStitcher: I realized the biome.ts file has very little content now, so I think it would make sense to rename that file to biome-utils.ts for now, so that we can name `PEBiome` `Biome` already and put it in a `biome.ts` file, maybe directly in the data folder since it's just the class definition 🤔 
That balance folder makes things complicated, but that's what was done for Pokemon evolutions and forms. Base class in `data` folder, and specific evolution data in the `balance` subfolder

DayKev (commented):
#878

DayKev (dismissed):

src/data/biome.ts
```diff
@@ -0,0 +1,48 @@
...
+ * - Move bgm loop point here as well
+ * - Move {@linkcode Arena.getTrainerChance} as well
+ */
+export class Biome {
+  public biomeType: BiomeId;
+  public pokemonPool: Record<BiomePoolTier, Record<TimeOfDay, SpeciesId[]>>;
+  public trainerPool: Record<BiomePoolTier, TrainerType[]>;
+  public weatherPool: Record<WeatherType, number>;
+  public terrainPool: Record<TerrainType, number>;
+  public bgm: string;
```
innerthunder: These could use docs

src/battle-scene.ts
```diff
@@ -2036,29 +2037,11 @@ export default class BattleScene extends SceneBase {
...
-        return biome;
-      }
-    }
-
-    return biomes[randSeedInt(biomes.length)];
+  /**
+   * TODO: Rewrite this later for weighting?
+   */
+  generateRandomBiome(_waveIndex: number): BiomeId {
+    return randSeedItem([...allBiomes.keys()]);
```
innerthunder:
```suggestion
    return randSeedItem(allBiomes.keys());
```

src/data/init/init-biomes.ts
```diff
@@ -0,0 +1,80 @@
...
+    slumBiome,
+    snowyForestBiome,
+    islandBiome,
+    laboratoryBiome,
+    endBiome,
+  ];
+
+  for (const pebiome of rawAllBiomes) {
+    allBiomes.set(pebiome.biomeType, pebiome);
+  }
```
innerthunder:
```suggestion
  for (const peBiome of rawAllBiomes) {
    allBiomes.set(peBiome.biomeType, peBiome);
  }
```
Maybe `pktyBiome` instead?

src/data/biomes/cave.ts
```diff
@@ -0,0 +1,134 @@
...
+const weatherPool = {
+  [WeatherType.NONE]: 1,
+  [WeatherType.SUNNY]: 0,
+  [WeatherType.RAIN]: 0,
+  [WeatherType.SANDSTORM]: 0,
+  [WeatherType.HAIL]: 0,
+  [WeatherType.SNOW]: 0,
+  [WeatherType.FOG]: 0,
+  [WeatherType.HEAVY_RAIN]: 0,
+  [WeatherType.HARSH_SUN]: 1,
```
innerthunder:
```suggestion
  [WeatherType.HARSH_SUN]: 0,
```
Not sure why a cave would have Desolate Land weather...

src/data/biomes/cave.ts
```diff
@@ -0,0 +1,134 @@
...
+import { BiomePoolTier } from "#enums/biome-pool-tier";
+import { SpeciesId } from "#enums/species-id";
+import { TerrainType } from "#enums/terrain-type";
+import { TimeOfDay } from "#enums/time-of-day";
+import { TrainerType } from "#enums/trainer-type";
+import { WeatherType } from "#enums/weather-type";
+
+/**
+ * This file represents all the data of the cave biome
+ */
```
innerthunder:
```suggestion
// This file represents all the data of the cave biome
```
To be consistent with the comment in `town.ts`

src/loading-scene.ts
```diff
@@ -32,6 +31,7 @@ import { initModifierTypes } from "#app/modifier/init-modifier-types";
 import { initModifierPools } from "#app/modifier/init-modifier-pools";
 import { timedEventManager } from "#app/timed-event-manager";
 import { DEFAULT_LANGUAGE_KEY } from "#app/system/settings/supported-languages";
+import { initBiomes } from "#app/data/init/init-biomes";
```
innerthunder: imports should be sorted (in VSCode you can right click > Source Action > Sort imports)

test/test-utils/testFileInitialization.ts
```diff
@@ -21,6 +20,7 @@ import { initAbilities } from "#app/data/init/init-abilities";
 import { initMoves } from "#app/data/init/init-moves";
 import { initModifierTypes } from "#app/modifier/init-modifier-types";
 import { initModifierPools } from "#app/modifier/init-modifier-pools";
+import { initBiomes } from "#app/data/init/init-biomes";
```
innerthunder: imports should be sorted

innerthunder (commented):
Some nits, otherwise lgtm

src/data/biomes/cave.ts
Tempo-anon: That's a typo lol

src/battle-scene.ts
Tempo-anon: allBiomes.keys() is a MapIterator of BiomeId so this doesn't quite work

src/battle-scene.ts
```diff
@@ -1474,7 +1475,7 @@ export default class BattleScene extends SceneBase {
   }
 
   newArena(biome: BiomeId): Arena {
-    this.arena = new Arena(biome, BiomeId[biome].toLowerCase());
+    this.arena = new Arena(biome, allBiomes.get(biome).bgm);
```
flx-sta: I'd prefer to have a helper method for that to lower the amount of random `allBiomes.` calls to a minimum across the codebase

----

e.g.
```ts
getBiomeBgm(biome: Biome) {
 ....
}

File: src/data/biome-utils.ts [hunk 25]
```diff
@@ -0,0 +1,41 @@
+import { BiomeId } from "#enums/biome-id";
+import type { SpeciesId } from "#enums/species-id";
+import type { TrainerType } from "#enums/trainer-type";
+import i18next from "i18next";
+
+export function getBiomeName(biome: BiomeId | -1) {
+  if (biome === -1) {
+    return i18next.t("biome:unknownLocation");
+  }
```
flx-sta: Instead you could define a BiomId for `-1` to be `UNKNOWN_LOCATION = -1`

src/data/biome.ts
```diff
@@ -0,0 +1,53 @@
...
+ * - Consider `partial` and `Omit` or an interface for `pokemonPool` and `trainerPool`
+ * - `weatherPool` and `terrainPool` are not used anywhere, move {@linkcode getRandomWeatherType} here
+ * - Add image assets here like `biomeWithProps` (in `arena.ts`)
+ * - Consider moving `mysteryEncounterByBiome` here as well
+ * - Move bgm loop point here as well
+ * - Move {@linkcode Arena.getTrainerChance} as well
+ */
+export class Biome {
+  /** The corresponding biomeId enum */
+  public biomeType: BiomeId;
```
flx-sta: Why `type` and not `id`? 
Is it overwritten anywhere? Otherwise readonly, especially for `public` fields

```suggestion
  public readonly biomeId: BiomeId;
```

src/data/biome.ts
```diff
@@ -0,0 +1,53 @@
...
+  /** A mapping of BiomePoolTier to TimeOfDay to SpeciesId representing the wild Pokemon that appear */
+  public pokemonPool: Record<BiomePoolTier, Record<TimeOfDay, SpeciesId[]>>;
```
flx-sta: Check if `readonly` can be used since it's set in the constructor

```suggestion
  public readonly pokemonPool: Record<BiomePoolTier, Record<TimeOfDay, SpeciesId[]>>;
```

src/data/biome.ts
```diff
@@ -0,0 +1,53 @@
...
+  /** A mapping of BiomePoolTier to a list of TrainerType representing the trainers that appear */
+  public trainerPool: Record<BiomePoolTier, TrainerType[]>;
```
flx-sta: Check if `readonly` can be used since it's set in the constructor

```suggestion
  public readonly trainerPool: Record<BiomePoolTier, TrainerType[]>;
```

src/data/biome.ts
```diff
@@ -0,0 +1,53 @@
...
+  /** weatherPool and terrainPool are currently unused, to be implemented in a future PR */
+  public weatherPool: Record<WeatherType, number>;
```
flx-sta: Check if `readonly` can be used since it's set in the constructor

```suggestion
  public readonly weatherPool: Record<WeatherType, number>;
```

src/data/biome.ts
```diff
@@ -0,0 +1,53 @@
...
+  public terrainPool: Record<TerrainType, number>;
```
flx-sta: Check if `readonly` can be used since it's set in the constructor

```suggestion
  public readonly terrainPool: Record<TerrainType, number>;
```

src/data/biome.ts
```diff
@@ -0,0 +1,53 @@
...
+  /** String representing the bgm of the biome */
+  public bgm: string;
```
flx-sta: Check if `readonly` can be used since it's set in the constructor

```suggestion
  public readonly bgm: string;
```

src/data/biome.ts
```diff
@@ -0,0 +1,53 @@
...
+    terrainPool: Record<TerrainType, number>,
+    bgm: string,
+  ) {
+    this.biomeType = biomeType;
+    this.pokemonPool = pokemonPool;
+    this.trainerPool = trainerPool;
+    this.weatherPool = weatherPool;
+    this.terrainPool = terrainPool;
+    this.bgm = bgm;
+  }
```
flx-sta: As previously mentioned: Why `type` and not `id`?

```suggestion
    biomeId: BiomeId,
    pokemonPool: Record<BiomePoolTier, Record<TimeOfDay, SpeciesId[]>>,
    trainerPool: Record<BiomePoolTier, TrainerType[]>,
    weatherPool: Record<WeatherType, number>,
    terrainPool: Record<TerrainType, number>,
    bgm: string,
  ) {
    this.biomeId = biomeId;
    this.pokemonPool = pokemonPool;
    this.trainerPool = trainerPool;
    this.weatherPool = weatherPool;
    this.terrainPool = terrainPool;
    this.bgm = bgm;
  }
```

src/data/biomes/cave.ts
```diff
@@ -0,0 +1,134 @@
...
+const pokemonPool = {
```
flx-sta, line 10: Missing type

src/data/biomes/cave.ts
```diff
@@ -0,0 +1,134 @@
...
+  [BiomePoolTier.COMMON]: {
+    [TimeOfDay.DAWN]: [],
```
flx-sta, line 12: Do they have to be present, even if they are empty? Couldn't that be omitted? 
(same for all following like that)

src/data/biomes/cave.ts
```diff
@@ -0,0 +1,134 @@
...
+  [BiomePoolTier.BOSS_ULTRA_RARE]: {
+    [TimeOfDay.DAWN]: [],
+    [TimeOfDay.DAY]: [],
+    [TimeOfDay.DUSK]: [],
+    [TimeOfDay.NIGHT]: [],
+    [TimeOfDay.ALL]: [SpeciesId.TERAPAGOS],
+  },
+};
+
+const trainerPool = {
```
flx-sta, line 98: NIT: Missing type

src/data/biomes/cave.ts
```diff
@@ -0,0 +1,134 @@
...
+  [BiomePoolTier.RARE]: [],
+  [BiomePoolTier.SUPER_RARE]: [],
+  [BiomePoolTier.ULTRA_RARE]: [],
+  [BiomePoolTier.BOSS]: [TrainerType.BROCK, TrainerType.ROXANNE, TrainerType.ROARK],
+  [BiomePoolTier.BOSS_RARE]: [],
+  [BiomePoolTier.BOSS_SUPER_RARE]: [],
+  [BiomePoolTier.BOSS_ULTRA_RARE]: [],
+};
+
+const weatherPool = {
```
flx-sta, line 110: NIT: Missing type

src/data/biomes/cave.ts
```diff
@@ -0,0 +1,134 @@
...
+  [WeatherType.SANDSTORM]: 0,
+  [WeatherType.HAIL]: 0,
+  [WeatherType.SNOW]: 0,
+  [WeatherType.FOG]: 0,
+  [WeatherType.HEAVY_RAIN]: 0,
+  [WeatherType.HARSH_SUN]: 0,
+  [WeatherType.STRONG_WINDS]: 0,
+};
+
+const terrainPool = {
```
flx-sta, line 123: NIT: Missing type

src/data/data-lists.ts
```diff
@@ -2,12 +2,17 @@ import type PokemonSpecies from "#app/data/pokemon-species";
 import { type Move } from "#app/data/moves/move";
 import { type MoveId } from "#enums/move-id";
 import type { Ability } from "#app/data/abilities/ability";
-interface MoveMap<K, V> extends Map<K, V> {
+import type { Biome } from "#app/data/biome";
+import { type BiomeId } from "#enums/biome-id";
+
+interface DataMap<K, V> extends Map<K, V> {
```
flx-sta, line 8: How is a `DataMap` now different from a `Map`?

src/data/mystery-encounters/utils/encounter-phase-utils.ts
```diff
@@ -1039,6 +1039,7 @@ export function calculateMEAggregateStats(baseSpawnWeight: number) {
         } else if (biomeLinks.hasOwnProperty(currentBiome)) {
           currentBiome = biomeLinks[currentBiome] as BiomeId;
         } else {
+          // Endless logic
```
flx-sta: I assume for Endless mode?

src/field/arena.ts
```diff
@@ -59,7 +55,7 @@ export class Arena {
     this.biomeType = biome;
     this.tags = [];
     this.bgm = bgm;
-    this.trainerPool = biomeTrainerPools[biome];
+    this.trainerPool = allBiomes.get(biome).trainerPool;
```
flx-sta: The constructor param should probably be renamed as `biome` would be associated with the class and not an id

src/field/arena.ts
```diff
@@ -122,10 +118,11 @@ export class Arena {
...
       this.pokemonPool = {};
-      for (const tier of Object.keys(biomePokemonPools[this.biomeType])) {
-        this.pokemonPool[tier] = Object.assign([], biomePokemonPools[this.biomeType][tier][TimeOfDay.ALL]).concat(
-          biomePokemonPools[this.biomeType][tier][timeOfDay],
-        );
+      for (const tier of Object.keys(allBiomes.get(this.biomeType).pokemonPool)) {
+        this.pokemonPool[tier] = Object.assign(
+          [],
+          allBiomes.get(this.biomeType).pokemonPool[tier][TimeOfDay.ALL],
+        ).concat(allBiomes.get(this.biomeType).pokemonPool[tier][timeOfDay]);
```
flx-sta: NIT: There is also a `Object.entries()` method which returns an array of `[key, value]` so you don't have to use `allBiomes.get(this.biomeType).pokemonPool[tier]` inside the loop

src/data/biomes/cave.ts
Tempo-anon: A `Record` requires all the fields to be present. One of the TODOs I have is to explore using `Omit` instead which do not require all the fields to be present. I just figured for now I could use records and just have everything

src/data/data-lists.ts
Tempo-anon: The only difference is that you don't need to add `!` when doing `allBiomes.get(...)`

src/data/biome-utils.ts
Tempo-anon: `switch` statements can't `case` on a -1 value

src/data/biomes/cave.ts
Tempo-anon: Is the type hinting really needed? I'll add type hinting to the `town.ts` file for now. If that's what we want we can add it for all the others but I'm also a bit lazy to do so in case we want to change the data structure or cut any biomes

src/field/arena.ts
flx-sta: `Object.assign` can also be often replaced with the `split` operator `...`

```suggestion
      for (const [tier, pool] of Object.entries(allBiomes.get(this.biomeType).pokemonPool)) {
        this.pokemonPool[tier] = [
          ...pool[TimeOfDay.ALL],
          ...pool[timeOfDay],
        ];
      }
```

src/data/biomes/end.ts
```diff
@@ -0,0 +1,131 @@
...
+import { BiomeId } from "#enums/biome-id";
+import { BiomePoolTier } from "#enums/biome-pool-tier";
+import { SpeciesId } from "#enums/species-id";
+import { TerrainType } from "#enums/terrain-type";
+import { TimeOfDay } from "#enums/time-of-day";
+import { WeatherType } from "#enums/weather-type";
+
+/**
+ * This file represents all the data of the end biome
+ */
```
DayKev:
```suggestion
```

src/data/biomes/cave.ts
DayKev:
```suggestion
```

src/data/biomes/town.ts
```diff
@@ -0,0 +1,208 @@
...
+import { SpeciesId } from "#enums/species-id";
+import { TerrainType } from "#enums/terrain-type";
+import { TimeOfDay } from "#enums/time-of-day";
+import { TrainerType } from "#enums/trainer-type";
+import { WeatherType } from "#enums/weather-type";
+
+/**
+ * This file represents all the data of the voclano biome
+ *
+ * TODO: Remove these exports. Currently exporting them as temporary
```
DayKev:
```suggestion
/**
 * @todo Remove these exports. Currently exporting them as temporary
```

src/data/biomes/volcano.ts
```diff
@@ -0,0 +1,165 @@
...
+import { BiomePoolTier } from "#enums/biome-pool-tier";
+import { SpeciesId } from "#enums/species-id";
+import { TerrainType } from "#enums/terrain-type";
+import { TimeOfDay } from "#enums/time-of-day";
+import { TrainerType } from "#enums/trainer-type";
+import { WeatherType } from "#enums/weather-type";
+
+/**
+ * This file represents all the data of the voclano biome
+ */
```
DayKev:
```suggestion
```

src/field/arena.ts
```diff
@@ -40,36 +36,39 @@ import { TrainerType } from "#enums/trainer-type";
...
-  public biomeType: BiomeId;
+  public biomeId: BiomeId;
   public weather: Weather | null;
   public terrain: Terrain | null;
   public tags: ArenaTag[];
   public bgm: string;
   public ignoreAbilities: boolean;
   public ignoringEffectSource: BattlerIndex | null;
 
+  /** Used to keep track of the previous TimeOfDay.
```
```suggestion
  /**
   * Used to keep track of the previous TimeOfDay.
```

src/field/arena.ts
```diff
@@ -118,14 +117,17 @@ export class Arena {
     return Array.isArray(timeOfDay) ? timeOfDay.includes(this.getTimeOfDay()) : this.getTimeOfDay() === timeOfDay;
   }
 
+  /**
+   * Updates the pokemonPool if the time of day changes.
+   * The pokemonPool is a combination of a biome's TimeOfDay.ALL pool
```
```suggestion
   * Updates the `pokemonPool` if the time of day changes.
   * The `pokemonPool` is a combination of a biome's `TimeOfDay.ALL` pool
```

src/field/arena.ts
```diff
@@ -1011,14 +998,20 @@ export class Arena {
...
+        console.warn(`missing bgm loop-point for biome "${BiomeId[this.biomeId]}" (=${this.biomeId})`);
         return 0;
     }
   }
 }
 
-export function getBiomeKey(biome: BiomeId): string {
-  return BiomeId[biome].toLowerCase();
+/**
+ * Todo: Make the key (biomeId in lower case) not what
```
```suggestion
 * @todo Make the key (`biomeId` in lower case) not what
```

src/data/biomes/volcano.ts
```diff
@@ -0,0 +1,162 @@
...
+    [TimeOfDay.DAWN]: [],
+    [TimeOfDay.DAY]: [],
+    [TimeOfDay.DUSK]: [],
+    [TimeOfDay.NIGHT]: [],
+    [TimeOfDay.ALL]: [
+      SpeciesId.CHARMANDER,
+      SpeciesId.CYNDAQUIL,
+      SpeciesId.CHIMCHAR,
+      SpeciesId.TEPIG,
+      SpeciesId.LITTEN,
```
PigeonBar, line 44: Is Fennekin supposed to no longer be here?

Tempo-anon: Probably just missed when copy/pasting but all biome values right now are still filler and subject to change

src/field/arena.ts
```diff
@@ -182,21 +185,6 @@ export class Arena {
       let species: SpeciesId;
       if (typeof entry === "number") {
         species = entry as SpeciesId;
```
PigeonBar: This type check should no longer be necessary since the entries in `tierPool` were changed to all be single `SpeciesId`s.

src/battle-scene.ts
```diff
@@ -2037,29 +2038,11 @@ export default class BattleScene extends SceneBase {
...
-        return biome;
-      }
-    }
-
-    return biomes[randSeedInt(biomes.length)];
+  /**
+   * TODO: Rewrite this later for weighting?
+   */
+  generateRandomBiome(_waveIndex: number): BiomeId {
+    return randSeedItem([...allBiomes.keys()]);
```
PigeonBar: This function should probably also exclude the Town and ??? biomes.

Tempo-anon: This function will likely be rewritten completely in the future but I can add in the framework for excluded biomes right now

flx-sta (approved):

PR #858, opened 2025-03-19
[P1] Make getAlly() check for player legal Pokemon

### LINKED ISSUE #840: [Bug] Fusion Flare and Fusion Bolt Crash in Challenges when used with Second Slot Ineligible Pokemon in Double Battles
### Describe the bug

Fusion Flare and Fusion Bolt will crash the game if used while the Player is in a Challenge Run with any ineligible Pokemon in the second slot during a Double Battle.

-Does not occur if the Pokemon in the second slot does not exist or is Fainted

### Reproduction

Begin a Mono-Fighting Challenge run with Mankey and Type: Null
Use Fusion Flare or Fusion Bolt while the Type: Null is in the Second Slot of a Double Battle.

```
	MOVESET_OVERRIDE: [MoveId.FUSION_FLARE, MoveId.FUSION_BOLT],
```

### Expected behavior

The game should not Crash when a Pokemon uses Fusion Flare or Fusion Bolt in Challenges while the Second Slot during a Double Battle is occupied by an Ineligible Pokemon.

### Screenshots / Videos

Image

Fusion Bolt Crash Second Slot Ineligible - Crash

https://github.com/user-attachments/assets/3e58893c-5246-4fe4-ba61-290c2bb82ebd

No Pokemon in Second Slot - No Crash

https://github.com/user-attachments/assets/801a36c6-b72b-4859-ae41-7f03ea30411e

Fainted in Second Slot - No Crash

https://github.com/user-attachments/assets/b1804881-56fe-45b1-a46f-d7505d118ee5

Water Pledge in Doubles Challenges Ineligible Pokemon in Second Slot - No Crash

https://github.com/user-attachments/assets/8c5a773a-e68b-4d5c-b4ee-a7ee2b8ecb3c

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

Tempo-anon:

Fusion flare/bolt will no longer crash the game if the user's ally is not allowed in battle

Fixes #840

Created a new `PlayerPokemon.getAlly()` function that overrides `Pokemon.getAlly()` and updated usages of it across the code base

## Screenshots/Videos

image

## How to test the changes?

Mono fighting challenge with mankey + type:null
```ts
MOVESET_OVERRIDE: [MoveId.FUSION_FLARE, MoveId.FUSION_BOLT],
```

## Checklist


src/data/moves/move-attrs/last-move-double-power-attr.ts
```diff
@@ -34,7 +34,7 @@ export class LastMoveDoublePowerAttr extends VariablePowerAttr {
       const userAlly = user.getAlly();
       const enemyAlly = enemy?.getAlly();
 
-      if (userAlly && userAlly.turnData.acted) {
+      if (userAlly.isAllowedInBattle() && userAlly.turnData.acted) {
```
MokaStitcher: hmm, wonder why `getAlly` doesn't make that check
I'd rather change that to avoid similar issues happening elsewhere in the code. We may need to double check that no call to getAlly expects to get invalid Pokemon, but that would be weird

Tempo-anon: Adding the `isAllowedInBattle()` check to `getAlly` would mean that it would also apply for enemies 🫠 players are the only side that's affected by challenge legality

flx-sta (dismissed):
Using `Mono-Type: Fight` Challenge with only Mankey in my team stil seems to break it

--------

<p>


## Fusion Bolt

https://github.com/user-attachments/assets/a869ad41-8f39-4242-aba0-26e732716c5e

## Fusion Flare

https://github.com/user-attachments/assets/be2ff25a-cdb0-4377-9d19-53cc691382ab

## Console Error

fusion-flare-bolt-error

</p>
</details>

src/data/moves/move-attrs/last-move-double-power-attr.ts
MokaStitcher: Then it should always return true for enemies if not fainted, not sure why that would be an issue?
Or the PlayerPokemon subclass should override getAlly to add the check

MokaStitcher: Also there may be a logic issue here. user and enemy should be named user and target probably, because there is nothing telling us that the user is a PlayerPokemon and the enemy an EnemyPokemon, it could very well be the other way around and in both cases it should make sure whatever mon getAlly returns is valid

src/field/pokemon.ts
```diff
@@ -4147,6 +4150,14 @@ export class PlayerPokemon extends Pokemon {
     return this.getFieldIndex();
   }
 
+  override getAlly(): Pokemon | undefined {
+    const ally = globalScene.getPlayerField()[this.getFieldIndex() ? 0 : 1];
```
```suggestion
    const ally = super.getAlly();
```
Technically it's the same but in case the logic behind Pokemon.getAlly ever gets changed this ensures it would get reflected here

DayKev (commented):
https://github.com/Despair-Games/poketernity/pull/861

src/field/pokemon.ts
```diff
@@ -3022,7 +3027,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     if (globalScene.currentBattle.double && this.getAlly()?.isActive(true)) {
       applyAbFunc(
         AbAttrFlag.ALLIED_FIELD_DAMAGE_REDUCTION,
-        this.getAlly(),
+        this.getAlly()!,
```
torranx: I think the bang can be removed by replacing the condition `this.getAlly()?.isActive(true)` with `this.getAlly() && this.getAlly().isActive(true)`

DayKev: It would need an indirection variable added (like I did for the other ones in my review, guess I missed this one).

src/data/abilities/ab-attrs/post-damage-force-switch-ab-attr.ts
```diff
@@ -133,8 +133,9 @@ class ForceSwitchOutHelper {
    * @param pokemon The {@linkcode Pokemon} attempting to switch out.
    * @returns `true` if the switch is successful
    */
-  public switchOutLogic(pokemon: Pokemon): boolean {
-    const switchOutTarget = pokemon;
+  public switchOutLogic(switchOutTarget: Pokemon): boolean {
+    const { currentBattle } = globalScene;
+    const { battleType, trainer, waveIndex } = currentBattle;
```
MokaStitcher:
```suggestion
    const { battleType, double, trainer, waveIndex } = globalScene.currentBattle;
```

src/data/abilities/ab-attrs/post-damage-force-switch-ab-attr.ts
```diff
@@ -190,13 +191,12 @@ class ForceSwitchOutHelper {
           500,
         );
 
-        if (globalScene.currentBattle.double) {
-          const allyPokemon = switchOutTarget.getAlly();
+        if (currentBattle.double && allyPokemon) {
```
MokaStitcher:
```suggestion
        if (double && allyPokemon) {
```

src/data/moves/move-attrs/force-switch-out-attr.ts
```diff
@@ -36,6 +36,8 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
   }
 
   override applyEffect(user: Pokemon, target: Pokemon, move: Move): boolean {
+    const { currentBattle } = globalScene;
+    const { battleType, double, trainer, waveIndex } = currentBattle;
```
MokaStitcher:
```suggestion
    const { battleType, double, trainer, waveIndex } = globalScene.currentBattle;
```

MokaStitcher (commented):
few nits, not playtested yet

DayKev (dismissed):

src/data/moves/move-attrs/await-combined-pledge-attr.ts
```diff
@@ -36,7 +36,7 @@ export class AwaitCombinedPledgeAttr extends OverrideMoveEffectAttr {
     });
 
     if (ret) {
-      const ally = user.getAlly();
+      const ally = user.getAlly()!;
```
torranx: shouldnt this be checked instead of adding a bang?

src/data/moves/move.ts
```diff
@@ -772,7 +772,7 @@ export abstract class Move implements Localizable {
     applyAbAttrs(AbAttrFlag.VARIABLE_MOVE_POWER, source, simulated, this, target, power);
 
     if (source.getAlly()) {
-      applyAbAttrs(AbAttrFlag.ALLY_MOVE_CATEGORY_POWER_BOOST, source.getAlly(), simulated, this, target, power);
+      applyAbAttrs(AbAttrFlag.ALLY_MOVE_CATEGORY_POWER_BOOST, source.getAlly()!, simulated, this, target, power);
```
torranx: is the bang needed here? the condition already checks for `source.getAlly()`

torranx (commented):
few questions, otherwise lgtm

src/data/moves/move.ts
DayKev: Because it's a function TS can't know that the `if` check will return the same value as here. That's why I used a `const` in most of the other spots in my review/suggestion.

src/data/moves/move-attrs/ability-copy-attr.ts
```diff
@@ -34,13 +34,14 @@ export class AbilityCopyAttr extends MoveEffectAttr {
       }),
     );
 
-    if (this.copyToPartner && globalScene.currentBattle?.double && user.getAlly().hp) {
-      user.getAlly().summonData.ability = target.getAbility().id;
+    const allyPokemon = user.getAlly();
+    if (this.copyToPartner && globalScene.currentBattle?.double && allyPokemon?.hp) {
```
innerthunder:
```suggestion
    if (this.copyToPartner && globalScene.currentBattle?.double && allyPokemon?.isActive(true)) {
```

src/field/pokemon.ts
```diff
@@ -3019,10 +3024,11 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     applyAbFunc(AbAttrFlag.RECEIVED_MOVE_DAMAGE_MULTIPLIER, this, simulated, source, move, receivedDamageMultiplier);
 
     /** Additionally apply friend guard damage reduction if ally has it. */
-    if (globalScene.currentBattle.double && this.getAlly()?.isActive(true)) {
+    const allyPokemon = this.getAlly();
+    if (globalScene.currentBattle.double && allyPokemon && allyPokemon.isActive(true)) {
```
```suggestion
    if (globalScene.currentBattle.double && allyPokemon?.isActive(true)) {
```
I think this can be condensed?

src/data/moves/move.ts
torranx: why are we not doing that here? I'd prefer the compiler checking instead of telling it it is not undefined

DayKev: I missed a couple when doing my review. 🤷

torranx (approved):

DayKev (approved):

PR #863, opened 2025-03-21
[Misc] Re-add type safety to `applyAbAttrs()`

DayKev:
N/A

Automatic type safety was lost with the circular dependency removal.

All `applyAbAttrs()` calls are now `applyAbAttrs<...>()` in order to manually apply type safety.
`applyAbAttrs()` now enforces developers to add type safety (see `src/data/abilities/apply-ab-attrs.ts`).

## Checklist

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -8,8 +8,7 @@ import type { PokemonSpeciesFilter } from "#app/@types/PokemonSpeciesFilter";
 import type { AnySettingKey, SettingsUpdateEventArgs } from "#app/@types/Settings";
 import { Animation } from "#app/animations";
 import { AudioManager } from "#app/audio-manager";
-import type { FixedBattleConfig } from "#app/battle";
-import Battle from "#app/battle";
+import Battle, { type FixedBattleConfig } from "#app/battle";
 import {
   IV_MAX,
   IV_MIN,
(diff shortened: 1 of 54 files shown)


MokaStitcher (dismissed):

PigeonBar:
image

Shown above, I found a hacky way to enforce type safety by forcing the programmer to pass a type into `TAttr`. If no type gets passed into `TAttr`, then it defaults to `NonsenseAbAttr`, and the param with type `never` causes a type error:

image

Would this be worth doing?

DayKev:
> Shown above, I found a hacky way to enforce type safety by forcing the programmer to pass a type into `TAttr`. If no type gets passed into `TAttr`, then it defaults to `NonsenseAbAttr`, and the param with type `never` causes a type error:
> Would this be worth doing?

Huh, interesting. It might be.

innerthunder:
> > Shown above, I found a hacky way to enforce type safety by forcing the programmer to pass a type into `TAttr`. If no type gets passed into `TAttr`, then it defaults to `NonsenseAbAttr`, and the param with type `never` causes a type error:
> > Would this be worth doing?
> 
> Huh, interesting. It might be.

There's an easier way to do this:

1. Change `applyAbAttrsInternal`'s signature to
```ts
function applyAbAttrsInternal<TAttr extends AbAttr = never>(
  abFilterOptions: AbilityFilterOptions,
  abAttrFlag: AbAttrFlag,
): string[] {
```
2. Make sure `applyAbAttrs` and `applyRevealedAbAttrs` call the internal function under the same type as specified, e.g.
```ts
export function applyAbAttrs<TAttr extends AbAttr = never>(
  abAttrFlag: AbAttrFlag,
): string[] {
  return applyAbAttrsInternal<TAttr>({ canApplyOnly: true }, abAttrFlag, ...params);
}
```

Then applying without a set type causes the error here:

image

innerthunder (approved):

MokaStitcher (approved):

PR #867, opened 2025-03-23
[Misc] Damage function refactor follow-up

DayKev:
- The number pop-up for healing will no longer sometimes be incorrect.
- Percentage based damage will respect Dynamax damage reduction.

Some issues were noticed after #753 was merged.

- Variable usage in `Pokemon.heal()` was fixed so that the correct value will be shown when healing is capped.
- The `ignoreDynamaxReduction` parameter was removed from `Pokemon.damageAndHeal()` and it is now automatically determined based on whether the `result` parameter is `HitResult.ONE_HIT_KO` or the new `HitResult.SELF_KO`.
- The enum value `HitResult.SELF_KO` was added and used for moves that cause the user to self-KO (such as Explosion or Lunar Dance).

## Checklist

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -3316,7 +3316,6 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
    * @param ignoreSegments - If `true`, boss bars are ignored. Only applies to {@linkcode EnemyPokemon}. Default `false`
    * @param preventEndure - If `true`, bypasses the effects of Endure, Sturdy, etc. Default `false`
    * @param ignoreFaintPhase - If `true`, doesn't push a {@linkcode FaintPhase}. Default `false`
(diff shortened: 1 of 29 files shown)


src/field/pokemon.ts
```diff
@@ -3338,6 +3336,8 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
       amount = 0;
     }
 
+    const ignoreDynamaxReduction = [HitResult.ONE_HIT_KO, HitResult.SELF_KO].includes(result) ? true : false;
```
torranx: array.includes() already returns a boolean value
```suggestion
    const ignoreDynamaxReduction = [HitResult.ONE_HIT_KO, HitResult.SELF_KO].includes(result);
```

DayKev: Whoops, you're right.

torranx (approved):

MokaStitcher (approved):

PR #869, opened 2025-03-23
[P0] Fix memory leak when Save & Quitting

### LINKED ISSUE #693: [Bug] Memory Leak when Saving and Quitting
### Describe the bug

When Saving and Quitting Pokerogue, a memory leak occurs.
-Memory used is increased by 0.2 to 0.3 GB at a consistent rate.
-Occurs independently of how the run was loaded. (Continue/Load Game)
-Occurs with any Run
-Occurs with any Browser or Device
-Occurs both Online and with Localhost.

** Fixed by Pokerogue #5128**

When the Browser is refreshed, memory used decreases drastically.

Link to Discord's Bug Report (memory leak)
https://discord.com/channels/1125469663833370665/1320050193567186984

### Reproduction

Load a run, then Save and Quit. Keep doing this until your RAM usage is extremely high.

### Expected behavior

There should not be a memory leak.

### Screenshots / Videos

Memory Leak Save and Quitting1

Memory Leak Save and Quitting2

### Session export file

_No response_

### User data export file

_No response_

### Additional context

> i've looked at this a bit, it looks to be due to ui elements not getting cleared properly on reset so they aren't garbage collected and new ones are added on top of them (biggest offender being `StarterContainer`s)

> just a small change to ensure that `StarterContainer`s get removed properly makes the graph look a lot nicer, though it's probably more than just those. two graphs here are a before/after of heap usage during 3 save and quits

**- mdeann**

<details><summary>
Before
</summary>

g1

</details>
<details><summary>
After
</summary>

g2

</details>

MokaStitcher:

The game will no longer take more and more memory every time the player saves & quits
<sub>and all other instances of calls to `battleScene.reset(true)`

Fixes #693 
The game is using a lot of memory as it is, no need for more

As explained in the Issue, when save & quitting BattleScene.reset() is called, with the `clearScene` parameter set to `true`. And when this happens the scene removes all its children elements, before recreating them and instantiating a new UI object. However before this the objects contained in the children containers never get destroyed, which results in the game keeping a lot of graphical elements of the previous instance of the UI in memory.

- In BattleScene.reset(), call `removeAll(true)` on all containers, which recursively destroys the Phaser elements contained within
- Add `clearMenus` function to the `NavigationManager` to delete existing references to menus it has
- override `ui.destroy` to do call `clearMenus` before the menus get destroyed
- Noticed a call to `BattleScene.initVariantData()` which was needed when changing the sprite type between consistent and experimental. That setting was removed in #101 so this call is no longer needed

Note:
This is kind of a "bruteforce" method and there are possibly some UI elements that get missed with this if they were created but not added to a container, but it is already a big improvement.
I also don't believe the approach of destroying the whole screen and UI is good in the first place. At the very least it could be very much optimized, but that is a much larger and different task

<sub>Another factor that could have minimized this issue would be for the `UI` to not keep a copy of each handler in memory at all times, which will be for another PR.

## Screenshots/Videos

Memory usage of the game after sae & quitting, before and after:
**BEFORE:** image **AFTER:** image

JS heap size, each new snapshot is taken after save & quitting, before and after:
**BEFORE:** image **AFTER:** image

## How to test the changes?

Run the game, start a run, save & quit.
Go through various UIs and make sure things work properly

## Checklist

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -1243,9 +1243,9 @@ export default class BattleScene extends SceneBase {
           this.clearPhaseQueue();
 
           // destroying those containers will call 'container.removeAll(true)', destroying all their children as well
-          this.uiContainer.destroy(true);
-          this.field.destroy(true);
-          this.fieldUI.destroy(true);
+          this.uiContainer.destroy();
+          this.field.destroy();
+          this.fieldUI.destroy();
 
           this.children.removeAll(true);
           this.game.domContainer.innerHTML = "";
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -1248,6 +1248,11 @@ export default class BattleScene extends SceneBase {
         onComplete: () => {
           this.clearPhaseQueue();
 
+          this.ui.destroy();
+          this.uiContainer.removeAll(true);
+          this.field.removeAll(true);
+          this.fieldUI.removeAll(true);
+
           this.children.removeAll(true);
           this.game.domContainer.innerHTML = "";
           this.launchBattle();
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -1236,9 +1236,6 @@ export default class BattleScene extends SceneBase {
     }
 
     if (clearScene) {
-      // Reload variant data in case sprite set has changed
-      this.initVariantData();
-
       this.audioManager.fadeOutBgm(250, false);
       this.tweens.add({
         targets: [this.uiContainer],
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -1242,9 +1242,10 @@ export default class BattleScene extends SceneBase {
         onComplete: () => {
           this.clearPhaseQueue();
 
-          this.uiContainer.removeAll(true);
-          this.field.removeAll(true);
-          this.fieldUI.removeAll(true);
+          // destroying those containers will call 'container.removeAll(true)', destroying all their children as well
+          this.uiContainer.destroy(true);
+          this.field.destroy(true);
+          this.fieldUI.destroy(true);
 
           this.children.removeAll(true);
           this.game.domContainer.innerHTML = "";
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -1252,6 +1252,8 @@ export default class BattleScene extends SceneBase {
 
           this.children.removeAll(true);
           this.game.domContainer.innerHTML = "";
+
+          // TODO: launchBattle will call reset(false, false, true) too, this can probably be done better
           this.launchBattle();
         },
       });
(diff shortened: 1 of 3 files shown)


innerthunder (dismissed):

src/ui/settings/navigation-menu.ts
```diff
@@ -44,6 +44,10 @@ export class NavigationManager {
     ];
   }
 
+  public clearMenus() {
+    this.navigationMenus.splice(0, this.navigationMenus.length);
```
Tempo-anon: Splicing with 0 to length is the same as setting it to an empty array so isn't `this.navigationMenus = [];` clearer?

src/battle-scene.ts
```diff
@@ -1245,8 +1242,14 @@ export default class BattleScene extends SceneBase {
         onComplete: () => {
           this.clearPhaseQueue();
 
+          this.uiContainer.removeAll(true);
+          this.field.removeAll(true);
+          this.fieldUI.removeAll(true);
```
DayKev, line 1247: These all call `.removeAll()` but what about `.destroy()`? Is there some reason we shouldn't be [also] calling that?

MokaStitcher:
Updated to directly destroy the containers instead of just removing their children, following our discussion
Results seems to be the same
image

Tempo-anon (dismissed):

DayKev (dismissed):

src/battle-scene.ts
```diff
@@ -1245,8 +1242,15 @@ export default class BattleScene extends SceneBase {
         onComplete: () => {
           this.clearPhaseQueue();
 
+          // destroying those containers will call 'container.removeAll(true)', destroying all their children as well
+          this.uiContainer.destroy(true);
+          this.field.destroy(true);
+          this.fieldUI.destroy(true);
```
DayKev: Got a [reply](https://phaser.discourse.group/t/what-does-the-fromscene-parameter-of-gameobject-destroy-do/15248/2?u=nightkev) from a Phaser dev, the `fromScene` parameter shouldn't be set to `true` when `.destroy()` is called outside of the internal Phaser functions.
```suggestion
          this.uiContainer.destroy();
          this.field.destroy();
          this.fieldUI.destroy();
```

PigeonBar:
As further evidence that certain objects now get garbage collected properly, it is possible to assign a 200 MB array to an object, which makes it much more obvious whether or not it has been garbage collected:

```
  constructor() {
    super(UiMode.STARTER_SELECT);
    Object.assign(this, {filler: Array(50000000).fill(123)});
  }
```

In addition, the heap snapshot now says that there is exactly one SSUI handler stored in memory, which means that the previous instances of SSUI handlers were garbage collected properly:

<details><summary>Screenshot</summary>
<p>

20250324 Heap Snapshot

</p>
</details>

src/battle-scene.ts
MokaStitcher: sweet, that makes things easier, thank you!

PigeonBar (approved):

Tempo-anon (approved):

PR #882, opened 2025-03-28
[Refactor] Add dedicated destroy function to UiHandlers

### LINKED ISSUE #905: [Bug] Enemy Pokemon do not get garbage collected when advancing a wave
### Describe the bug

#757 attempted to reduce memory usage by destroying enemy Pokemon objects when advancing to the next wave. However, according to the heap snapshot, this was not actually enough to make the enemy Pokemon get garbage collected.

### Reproduction

Play a few waves, then take a heap snapshot.

### Expected behavior

The enemy Pokemon should get garbage collected in order to free up memory.

### Screenshots / Videos

Image

After advancing a few waves, there are now 7 enemy Pokemon in the heap snapshot. According to the "Retainers" section, the enemy Pokemon object is still being referenced by something in `battle-anims.ts`. The given line of code points to some undocumented anims code that adds a listener to the enemy Pokemon's sprite and then never removes the listener:

https://github.com/Despair-Games/poketernity/blob/b2579311fb9d4aa5466c7d90541a9bff2303f74d/src/data/animations/battle-anims.ts#L271

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

### LINKED ISSUE #934: [Refactor] Dedicated destroy function to UiHandlers
Add a dedicated destroy function to UiHandler classes

MokaStitcher:

- Hopefully no more errors due to leftover listeners after save & quitting
- Save & quit memory leak reduced even further
- Properly clear up remaining reference to destroyed EnemyPokemon
- Destroy PlayerPokemon objects resulting from egg hatches

- Follow up to #869, some destroyed elements had registered listeners or infinitely looping tweens that were not being removed properly, causing issues
- What was done in #869 was destroying all contents of the UI container indiscriminately since the UI is fully rebuild upon save & quitting. This approach didn't make it possible for individual handlers to clear no longer needed listeners or tweens. So this PR adds dedicated `destroy` function in UiHandlers, and a few other updates to the UiHandler lifecycle. This will also be useful in the future where we no longer keep all handlers in memory all the time
- The EnemyPokemon and PlayerPokemon issues are not directly related to save & quitting since they happen during normal play, but they are still a memory leak cause that should be addressed (fixes #905 )

<details><summary>UI Handler updates ⚠️ :</summary>

> ℹ️ _This relates to **AbstractUiHandler**_

### Previous UiHandler lifecycle:
- `setup`, called once at creation, initializing some things
- `show`, called every time the ui mode is changed to the one corresponding to this handler, creating the needed UI elements
- `clear` called once the ui mode is switched back, destroying elements initialized in `show`

### Now:
This PR adds a new `destroy` function, that should destroy/clear all elements that were created during `setup`, similar to how `clear` removes all elements created in `show`.

This `destroy` function allows to properly destroy listeners, tweens or other elements that got missed before.
This will also allow each handler to be destroyed individually in the future, so that we don't have to keep all handlers in memory at all times.

To keep track of the 'setup/destroyed' status, a `ready` attribute was added in the UiHandler class, on top of the existing `active` attribute.

Both attributes were made private, with a public getter. Children classes should not have to modify them.
To ensure this doesn't happen, each step in the lifecycle has a public function that updates the ready or active status, and calls an internal abstract function that children should override and use to create/destroy its UI elements.
- `public init` is called once at creation. Calls `this.setup()` and sets `ready` to `true`
- `public start` is called when the mode is changed to this handler. Calls `this.show()` and sets `active` to `true`
- `public stop` is called when the mode is changed back. Calls `this.clear()` and sets `active` to `false`
- `public destroy` is called once at deleting. Calls `this.teardown()` and sets `ready` to `false`

 All these abstract functions are protected, expect for `show`, since its parameters are needed for argument type validation when changing the Ui Mode.
While right now `ready` is not being used outside of the AbstractUiHandler since all UIHandlers are always initialized, it will be needed in future PRs for the UI to keep track of handler status when it is no longer the case.

The UI class was updated to use these new functions, and the signature of all overridden functions in UiHandlers got updated and uniformized

> ℹ️ Calling `destroy()` on a Phaser Container:
> - removes it from its parent container
> - removes all its children from itself, and calls `destroy()` on them.
>
> This means that to destroy most UI Handlers we only need to call `destroy` the root container(s) where all children elements were added, and they will all be destroyed as well
> Another note, since destroyed handlers aren't intended to get reused and any reference to them is being deleted right after, we don't need to clear all references the handler hold to destroyed GameObjects for them to get garbage collected properly.

</details> 

With the above changes, the following fixes were able to be implemented:

<details><summary>List of listeners now being removed as needed</summary>

- EncounterPhase event listener in TimeOfDayWidget
- EggCountChanged event listener in EggCounterContainer
- Key down listeners in Keyboard and Gamepad Binding handlers
- Key down listeners in KeybordSettingsUiHandler
- Language Change listener from DisplaySettingsUiHandler
- Orientation Change listeners from GeneralSettingsUiHandler
- Remove destroyed InfoToggle references in the battle scene
- Remove listeners on sprites used for battle animation once the animation is done (fixes part of #905)

</details> 
<details><summary>List of tweens now being removed as needed</summary>

- Tera sparkle animation
- IconAnimHandler animations (used to make pokemon icons move up and down in various places)

</details> 
<details><summary>Other fixes and updates</summary>

- Removed logging of loaded SystemData and SessionData, as it was preventing the data from getting garbage collected, causing a big part of the remaining post save & quit memory leak
- TargetSelectUiHandler frees Pokemon object references once it's done (fixes the other part of #905)
- EggSummaryPhase destroys no longer PlayerPokemon objects once it's done
- EggHatchPhase destroys no longer PlayerPokemon objects once it's done
- Fix Keyboard binding deletion not working
- Renamed the various XXXSettingsUiHandlers for consistency with their associated UiMode, and each other

</details>

**Tests:**
The SelectModifierPhase test was manually altering the player's team, causing an error in a Ui Handler. This was already the case before this PR, but the test still passed. However with the ui handler changes the error now made the test time out. This resulted in subsequent tests failing because the UI Handler was left in an improper state.
- The test was fixed by using overrides instead of altering the team manually
- The GameManager now resets all UiHandlers between tests, ensuring such a bug would not impact on other tests in the future

## Known issues

> [!WARNING]
> InputTexts for the login and registration handlers do not get cleared properly even after the UI gets destroyed, but it seems like fixing it would require deeper investigation and its own PR. More details here: #1049 

## Screenshots/Videos

JS heap size, each snapshot taken after a save & quit. Still some increase over time, but much less substantial than before

| Before | After |
|--------|--------|
| before  | after | 

> [!NOTE]
> For the screenshots below I added console logs in the listeners and filtered the console messages to only show those, so that we can see the difference in number of times the listeners are called before and after save & quits, before and after the fix.
> In the early screenshots I did not write in the console the moments I saved & quit, so I added a red line to mark it

### Listeners:

<details><summary>Time of Day Widget</summary>

| Before | After |
|--------|--------|
| before  | after | 

</details> 

<details><summary>Hatching eggs </summary>

| Before | After |
|--------|--------|
| before  | after | 

</details> 

<details><summary>Info Toggle - Before / After </summary>

| Before | After |
|--------|--------|
| before  | after | 

</details> 

<details><summary>Opening the language menu after a save & quit - Before / After</summary>

image
image
image
</details> 

<details><summary>Changing control bindings - Before / After </summary>

image
image
image
</details> 

<details><summary>Mobile Orientation Change - Before / After </summary>

| Before | After |
|--------|--------|
| before  | after | 

</details> 

### Tweens / Animations:

<details><summary>IconAnimHelper - Before / After </summary>

| Before | After |
|--------|--------|
| before  | after | 

</details> 

<details><summary>Tera Sparkles Handler - Before / After </summary>

| Before | After |
|--------|--------|
| before  | after | 

</details> 

### Other:
<details><summary>EnemyPokemon references</summary>

#### Before, 7 EnemyPokemon in memory after having gone through a few waves:
Image

#### After, 2 snapshots taken on successive waves:
Image
</details> 

<details><summary>PlayerPokemon references</summary>

Before, 1st Snapshot: reference. 2nd snapshot: after hatching eggs (with summary). 3rd snapshot: after hatching eggs (no summary):
image

</details> 

## How to test the changes?

- Start the game
- Launch a run
- Save and quit
- Play through some waves / go through as many UIs as you can to make sure things aren't broken

## Checklist

- [ ] The PR is self-contained and cannot be split into smaller PRs? 💀 


src/battle-scene.ts
flx-sta, line 1: Alternatively we could set up our own `PktyAnimationmanager` that provides a `createIfNotExists()` method:

<details><summary>[Code Example]</summary>
<p>

### PktyAnimationManager | `src/gameobjects/pkty-animation-manager.ts`

```ts
import Phaser from "phaser";

export class PktyAnimationManager extends Phaser.Animations.AnimationManager {
  /**
   * Checks if an animation exists before calling {@linkcode PktyAnimationManager.create}
   *
   * If you wish to re-use an existing key, call `AnimationManager.remove` first, then this method.
   * @param config The configuration settings for the Animation.
   */
  public createIfNotExists(config: Phaser.Types.Animations.Animation): Phaser.Animations.Animation | undefined {
    if (!config.key || !this.exists(config.key)) {
      return this.create(config);
    } else {
      return undefined;
    }
  }
}
```

### SceneBase | `src/scene-base.ts`
```ts
export class SceneBase extends Phaser.Scene {

  public override anims: PktyAnimationManager;
  //...
}
```


### BattleScene | `src/battle-scene.ts`
```ts
//...
    this.anims.createIfNotExists({
      key: "prompt",
      frames: this.anims.generateFrameNumbers("prompt", { start: 1, end: 4 }),
      frameRate: 6,
      repeat: -1,
      showOnStart: true,
    });

    this.anims.createIfNotExists({
      key: "tera_sparkle",
      frames: this.anims.generateFrameNumbers("tera_sparkle", { start: 0, end: 12 }),
      frameRate: 18,
      repeat: 0,
      showOnStart: true,
      hideOnComplete: true,
    });
 //...
 ```

</p>
</details>

File: src/battle-scene.ts [hunk 1]
```diff
@@ -889,6 +894,13 @@ export default class BattleScene extends SceneBase {
     this.infoToggles.push(infoToggle);
   }
 
+  removeInfoToggle(infoToggle: InfoToggle): void {
+    const infoToggleIndex = this.infoToggles.indexOf(infoToggle);
+    if (infoToggleIndex !== -1) {
+      this.infoToggles.splice(infoToggleIndex, 1);
+    }
+  }
+
```
flx-sta: NIT: 
1. Missing scope `public`
2. Missing tsdoc

-----

```suggestion
  /**
   * Removes a specified {@linkcode InfoToggle} from the {@linkcode infoToggles} array.
   *
   * @param infoToggle - The {@linkcode InfoToggle} instance to be removed. If the toggle
   *                     is not found in the array, no action is taken.
   */
  public removeInfoToggle(infoToggle: InfoToggle): void {
    const infoToggleIndex = this.infoToggles.indexOf(infoToggle);
    if (infoToggleIndex !== -1) {
      this.infoToggles.splice(infoToggleIndex, 1);
    }
  }

```

src/battle-scene.ts
```diff
@@ -1172,6 +1184,13 @@ export default class BattleScene extends SceneBase {
         onComplete: () => {
           this.phaseManager.clearPhaseQueue();
 
+          // stop the tera sparkle handler
+          this.spriteSparkleHandler.destroy();
+
+          // destroy elements placed directly on the scene
+          this.arenaBg.destroy();
+          this.arenaBgTransition.destroy();
+
```
flx-sta: Q: I'm not expert on tweens, but is there also some kind of "catch" callback? If the tween fails to complete?

src/field/pokemon-sprite-sparkle-handler.ts
```diff
@@ -4,10 +4,12 @@ import { fixedNumber, randInt } from "#app/utils";
 export default class PokemonSpriteSparkleHandler {
   private sprites: Set<Phaser.GameObjects.Sprite>;
 
+  private counterTween: Phaser.Tweens.Tween | undefined;
```
flx-sta: NIT: Adding a `?` at the end of the definition will automatically tell the compiler that `undefined` is part of it's types
```suggestion
  private counterTween?: Phaser.Tweens.Tween;
```

src/field/pokemon-sprite-sparkle-handler.ts
flx-sta, line 1: Should this file import `Phaser` then, since it's using `Phaser.<...>` types

src/inputs-controller.ts
```diff
@@ -178,6 +178,8 @@ export class InputsController {
 
     this.setGamepadSupport(settings.gamepad.enabled);
 
+    // Note: at the moment we don't need to remove this listener because only a single instance
+    // Of InputController is ever created, during 'BattleScene.create'.
```
flx-sta, line 182: Maybe we should tag the class as `@singleton` or something in the comment
<sub> Out of scope possiblty </sub>

src/inputs-controller.ts
```diff
@@ -251,7 +253,7 @@ export class InputsController {
     if (gamepadName) {
       this.selectedDevice[Device.GAMEPAD] = gamepadName.toLowerCase();
     }
-    const handler = globalScene.ui?.handlers[UiMode.SETTINGS_GAMEPAD] as SettingsGamepadUiHandler;
+    const handler = globalScene.ui?.handlers[UiMode.SETTINGS_GAMEPAD] as GamepadSettingsUiHandler;
```
flx-sta, line 256: Q: We should possibly update the `Ui.getHandler<T>()` method to also allow a parameter. Thus we could call
```suggestion
    const handler = globalScene.ui?.getHandler<GamepadSettingsUiHandler>(UiMode.SETTINGS_GAMEPAD);
```
then

src/inputs-controller.ts
```diff
@@ -264,7 +266,7 @@ export class InputsController {
     if (layoutKeyboard) {
       this.selectedDevice[Device.KEYBOARD] = layoutKeyboard.toLowerCase();
     }
-    const handler = globalScene.ui?.handlers[UiMode.SETTINGS_KEYBOARD] as SettingsKeyboardUiHandler;
+    const handler = globalScene.ui?.handlers[UiMode.SETTINGS_KEYBOARD] as KeyboardSettingsUiHandler;
```
flx-sta, line 269: Q: We should possibly update the `Ui.getHandler<T>()` method to also allow a parameter. Thus we could call
```suggestion
    const handler = globalScene.ui?.getHandler<GamepadSettingsUiHandler>(UiMode.SETTINGS_GAMEPAD);
```
then

src/inputs-controller.ts
```diff
@@ -312,7 +314,7 @@ export class InputsController {
       globalScene.gameData?.saveMappingConfigs(gamepadID, this.configs[gamepadID]);
     }
     this.lastSource = "gamepad";
-    const handler = globalScene.ui?.handlers[UiMode.SETTINGS_GAMEPAD] as SettingsGamepadUiHandler;
+    const handler = globalScene.ui?.handlers[UiMode.SETTINGS_GAMEPAD] as GamepadSettingsUiHandler;
```
flx-sta, line 317: Q: We should possibly update the `Ui.getHandler<T>()` method to also allow a parameter. Thus we could call
```suggestion
    const handler = globalScene.ui?.getHandler<GamepadSettingsUiHandler>(UiMode.SETTINGS_GAMEPAD);
```
then

src/phases/egg-hatch-phase.ts
```diff
@@ -98,6 +98,8 @@ export class EggHatchPhase extends Phase {
 
       globalScene.audioManager.fadeOutBgm(undefined, false);
 
+      // TODO: the hatch phase and ui handler should not be intertwined in this way;
+      // the phase also should not be the one creating the graphical objects
       this.eggHatchHandler = globalScene.ui.getHandler() as EggHatchSceneUiHandler;
```
flx-sta, line 104: NIT: Possibly out of scope
```suggestion
      this.eggHatchHandler = globalScene.ui.getHandler<EggHatchSceneUiHandler;>();
```

src/system/settings/settings-gamepad.ts
```diff
@@ -86,7 +86,7 @@ export function setSettingGamepad(setting: SettingGamepad, value: number): boole
         if (globalScene.ui) {
           const cancelHandler = (success: boolean = false): boolean => {
             globalScene.ui.revertMode();
-            (globalScene.ui.getHandler() as SettingsGamepadUiHandler).updateBindings();
+            (globalScene.ui.getHandler() as GamepadSettingsUiHandler).updateBindings();
```
flx-sta, line 89: NIT: 
```suggestion
            globalScene.ui.getHandler<GamepadSettingsUiHandler>().updateBindings();
```

src/system/settings/settings-gamepad.ts
```diff
@@ -99,12 +99,12 @@ export function setSettingGamepad(setting: SettingGamepad, value: number): boole
         if (globalScene.ui && gp) {
           const cancelHandler = () => {
             globalScene.ui.revertMode();
-            (globalScene.ui.getHandler() as SettingsGamepadUiHandler).setOptionCursor(
+            (globalScene.ui.getHandler() as GamepadSettingsUiHandler).setOptionCursor(
```
flx-sta, line 102: NIT: 
```suggestion
            globalScene.ui.getHandler<GamepadSettingsUiHandler>().setOptionCursor(
```

src/system/settings/settings-gamepad.ts
```diff
@@ -99,12 +99,12 @@ export function setSettingGamepad(setting: SettingGamepad, value: number): boole
...
               Object.values(SettingGamepad).indexOf(SettingGamepad.Controller),
               0,
               true,
             );
-            (globalScene.ui.getHandler() as SettingsGamepadUiHandler).updateBindings();
+            (globalScene.ui.getHandler() as GamepadSettingsUiHandler).updateBindings();
```
flx-sta, line 107: NIT: 
```suggestion
            globalScene.ui.getHandler<GamepadSettingsUiHandler>().updateBindings();
```

src/system/settings/settings-keyboard.ts
```diff
@@ -135,7 +135,7 @@ export function setSettingKeyboard(setting: SettingKeyboard, value: number): boo
         if (globalScene.ui) {
           const cancelHandler = (success: boolean = false): boolean => {
             globalScene.ui.revertMode();
-            (globalScene.ui.getHandler() as SettingsKeyboardUiHandler).updateBindings();
+            (globalScene.ui.getHandler() as KeyboardSettingsUiHandler).updateBindings();
```
flx-sta, line 138: NIT: 
```suggestion
            globalScene.ui.getHandler<KeyboardSettingsUiHandler>().updateBindings();
```

src/ui/handlers/abstract-ui-handler.ts
```diff
@@ -18,25 +28,140 @@ export abstract class UiHandler {
...
 
-  show(..._args: unknown[]): boolean {
-    this.active = true;
+  /**
+   * Whether the handler is setup and ready to be displayed through calling {@linkcode start}.
+   * note: unused for now since all handlers are ready at all times.
+   */
+  public get ready(): boolean {
+    return this._ready;
+  }
```
flx-sta: NIT: I think @DayKev said that the docs for get/set should be on the field definition as they are "inherited" by the getter/setter

src/ui/handlers/autocomplete-ui-handler.ts
```diff
@@ -3,15 +3,19 @@ import { Button } from "#enums/buttons";
 import { OptionSelectUiHandler } from "./option-select-ui-handler";
 
 export class AutoCompleteUiHandler extends OptionSelectUiHandler {
-  private modalContainer: Phaser.GameObjects.Container;
+  private modalContainer: Phaser.GameObjects.Container | null;
```
flx-sta: NIT: It was decided to prefer the use of `undefined` instead of null for consistency

```suggestion
  private modalContainer?: Phaser.GameObjects.Container;
```

src/ui/handlers/battle-message-ui-handler.ts
```diff
@@ -143,17 +144,25 @@ export class BattleMessageUiHandler extends MessageUiHandler {
...
   }
 
-  processInput(button: Button): boolean {
+  protected override tearDown(): void {
+    this.bg.destroy();
+    this.commandWindow.destroy();
+    this.movesWindowContainer.destroy();
+    this.levelUpStatsContainer.destroy();
+    this.messageContainer.destroy();
+  }
```
flx-sta, line 161: Q: Might be worth it to set up some kind of getter that returns all the `GameObjects` (or Containers) that can call `.destroy()`

e.g.
```ts
protected get containers(): Phaser.GameObjects.Container {
  return [this.x, this.y, this.z, this.foo, this.bar];
}
```

then in `super.tearDown()` call: `this.containers.forEach((container) => container.destroy())`

flx-sta (commented):
22/76 Files Reviewed
image

src/inputs-controller.ts
MokaStitcher: Definitely if we want to keep things this way yes.
But ideally we should get rid of all those places in the code that access the handlers like this, I don't think they should be accessed publicly like this. I suspect most of the places where this is done could instead be replaced by events
Right now this whole thing only works because all handlers are always in memory, and hopefully the next big UI PR after this one will change that, so this is likely when this issue will be addressed.

src/field/pokemon-sprite-sparkle-handler.ts
MokaStitcher: Maybe? Phaser is used directly like this in many places in the codebase, esp in UI code so maybe that would be for a separate refactor

src/battle-scene.ts
MokaStitcher: Not sure I follow which tween you are referring to in this case, do you mean the one encompassing this code?
I wouldn't know the answer to this off hand though

src/inputs-controller.ts
MokaStitcher: It's definitely not being enforced as a singleton right now, it's only used that way, so I'm not sure how to reflect that in the documentation

src/ui/handlers/battle-message-ui-handler.ts
MokaStitcher: maybe, I'm a bit wary about adding more overhead for creating new handlers right now and adding more things that could get misused. Not all containers created by the handler should be returned here, only those added to the root UI container during setup. Every other container created differently or at another point shouldn't be part of it and I worry this could cause confusion.
Most handlers also only use a single container, so this would be overkill for many of them. It would also obfuscate the container's destruction to another place, while right now I quite like the "simplicity" of checking that `teardown` destroys everything that is added during `setup`, by looking at both functions in the same class, rather than having to refer back to the parent class.
Maybe if we also add a specific method for creating those containers in the parent class then this would even out.

I wonder what other people think.

src/battle-scene.ts
MokaStitcher: That's a good idea!

Trying to figure out if this should be part of this PR though, because to be honest that change wasn't part of the scope at all in the first place. This issue has been here forever I took the opportunity to fix it because the warning was annoying me :')

flx-sta: I'll pour into an EPIC. I always planned an epic to provide our own versions of GameObjects, that provide the property type-hints. It should probably be part of that

flx-sta: https://github.com/Despair-Games/poketernity/issues/1055?issue=Despair-Games%7Cpoketernity%7C1056

src/field/pokemon-sprite-sparkle-handler.ts
flx-sta: I would rather include it for any changes. In this case it's even just the type import, no harm in that.
I've had cases in the past where the missing phaser import caused issues.

src/ui/handlers/abstract-ui-handler.ts
```diff
@@ -18,25 +36,133 @@ export abstract class UiHandler {
...
+
+    this.tearDown();
+    this._ready = false;
+  }
+
+  /**
+   * Displays the handler.
+   * Should not be overridden. Calls {@linkcode show}, which subclasses should override.
+   *
+   * @param args the arguments needed by this handler for display.
```
flx-sta: NIT:  Recently learned that the dash `-` is supposed to be put manually.
```suggestion
   * @param args - The arguments needed by this handler for display.
```

src/ui/handlers/abstract-ui-handler.ts
```diff
@@ -18,25 +36,133 @@ export abstract class UiHandler {
...
+   */
+  protected abstract tearDown(): void;
+
+  /**
+   * Called when the mode corresponding this handler is set, and it should
+   * prepare the handler for display, based on the arguments provided.
+   *
+   * Elements (containers, gameObjects, listeners, ...) created here should be destroyed in {@linkcode clear}.
+   *
+   * @param _args the arguments needed by this handler for display.
```
flx-sta: NIT:  Recently learned that the dash `-` is supposed to be put manually.
```suggestion
   * @param _args - The arguments needed by this handler for display.
```

src/ui/handlers/egg-hatch-scene-ui-handler.ts
```diff
@@ -34,17 +34,25 @@ export class EggHatchSceneUiHandler extends UiHandler {
     }
   }
 
-  override show(): boolean {
-    super.show();
+  protected override tearDown(): void {
+    this.eggHatchContainer.destroy();
+    globalScene.anims.remove("egg_lightrays");
```
flx-sta, line 39: Q: Since you check if they exists for create, does `remove()` already check it by itself ? Or could this cause issues?

src/ui/handlers/egg-list-ui-handler.ts
```diff
@@ -106,9 +106,12 @@ export class EggListUiHandler extends MessageUiHandler {
     this.cursor = -1;
   }
 
-  override show(): boolean {
-    super.show();
+  protected override tearDown(): void {
+    this.iconAnimHandler.destroy();
+    this.eggListContainer.destroy();
+  }
```
flx-sta, line 112: Q: What about all the other game objects?
```ts
  private eggIcons: Phaser.GameObjects.Sprite[];
  private eggSprite: Phaser.GameObjects.Sprite;
  private eggNameText: Phaser.GameObjects.Text;
  private eggDateText: Phaser.GameObjects.Text;
  private eggHatchWavesText: Phaser.GameObjects.Text;
  private eggGachaInfoText: Phaser.GameObjects.Text;
  private eggListMessageBoxContainer: Phaser.GameObjects.Container;
```

flx-sta: If it's only affecting containers there is the `eggListMessageBoxContainer`

src/ui/handlers/game-stats-ui-handler.ts
```diff
@@ -295,9 +295,11 @@ export class GameStatsUiHandler extends UiHandler {
     this.gameStatsContainer.setVisible(false);
   }
 
-  override show(): boolean {
-    super.show();
+  protected override tearDown(): void {
+    this.gameStatsContainer.destroy();
+  }
```
flx-sta, line 300: Q: What about `statsContainer`?

src/ui/handlers/login-form-ui-handler.ts
```diff
@@ -64,8 +65,18 @@ export class LoginFormUiHandler extends FormModalUiHandler {
     this.infoContainer.setVisible(false);
   }
 
+  protected override tearDown(): void {
+    this.infoContainer.destroy();
+    this.externalPartyContainer.destroy();
+
+    // TODO: it seems like the password and username input remains in a detached state even after destruction
```
flx-sta, line 72: I did a little experiment of calling `this.inputs.forEach((input) => input.destroy())` and that seemed to get rid of the game-object & dom-element

src/ui/handlers/message-ui-handler.ts
```diff
@@ -36,7 +36,7 @@ export abstract class MessageUiHandler extends AwaitableUiHandler {
     }
   }
 
-  showText(
+  public showText(
```
flx-sta, line 49: NIT: Missing doc
```suggestion
  /**
   * Displays a dialogue message on the UI with optional delay and speaker name.
   *
   * @param text - The dialogue message to display.
   * @param _name - (Optional) The name of the speaker for the dialogue. Defaults to an empty string if not provided.
   * @param delay - (Optional) The delay in milliseconds before the dialogue is displayed. Defaults to `null` for no delay.
   * @param callback - (Optional) A callback function to execute after the dialogue is displayed. Defaults to `null` for no callback.
   * @param callbackDelay - (Optional) The delay in milliseconds before executing the callback. Defaults to `null` for no delay.
   * @param prompt - (Optional) Whether to show the prompt after displaying the dialogue.
   * @param promptDelay - (Optional) The delay in milliseconds before showing the prompt. Defaults to `null` for no delay.
   */
  public showText(
```
<sup>Generated by Copilot</sup>

src/ui/handlers/message-ui-handler.ts
```diff
@@ -50,7 +50,7 @@ export abstract class MessageUiHandler extends AwaitableUiHandler {
     this.showTextInternal(text, delay, callback, callbackDelay, prompt, promptDelay);
   }
 
-  showDialogue(
+  public showDialogue(
```
flx-sta, line 74: NIT: Missing doc

```suggestion
  /**
   * Displays a dialogue with the given text and optional parameters.
   *
   * @param text - The dialogue text to display.
   * @param _name - (Optional) The name of the character speaking the dialogue.
   * @param delay - (Optional) The delay in milliseconds before the dialogue is displayed.
   * @param callback - (Optional) A function to execute after the dialogue is displayed.
   * @param callbackDelay - (Optional) The delay in milliseconds before the callback is executed.
   * @param prompt - (Optional) Whether to display a prompt at the end of the dialogue.
   * @param promptDelay - (Optional) The delay in milliseconds before the prompt is displayed.
   */
  public showDialogue(
```
<sup> Generated by Copilot </sup>

src/ui/handlers/message-ui-handler.ts
```diff
@@ -249,20 +249,18 @@ export abstract class MessageUiHandler extends AwaitableUiHandler {
     };
   }
 
-  isTextAnimationInProgress() {
+  public isTextAnimationInProgress() {
```
flx-sta: NIT: Missing doc

```suggestion
  /**
   * Checks if the text animation is currently in progress.
   *
   * @returns `true` if the text animation is active and the timer has remaining repetitions; otherwise, `false`.
   *
   * The method evaluates the state of the `textTimer` property to determine whether the text animation is still ongoing.
   * If `textTimer` is defined and its `repeatCount` is less than its `repeat` value, the animation is considered in progress.
   * Otherwise, it returns `false`.
   */
  public isTextAnimationInProgress(): boolean {
```
<sup> Generated by Copilot </sup>

src/ui/handlers/message-ui-handler.ts
```diff
@@ -249,20 +249,18 @@ export abstract class MessageUiHandler extends AwaitableUiHandler {
...
     if (this.textTimer) {
       return this.textTimer.repeatCount < this.textTimer.repeat;
     }
 
     return false;
   }
 
-  clearText() {
-    this.message.setText("");
+  public clearText() {
```
flx-sta: NIT: MIssing docs

```suggestion
  /**
   * Clears the currently displayed text from the UI.
   *
   * This method sets the message text to an empty string and resets the `pendingPrompt` state.
   * It ensures that the text is only cleared if the `ready` state is `true` and the `message` object is active.
   */
  public clearText(): void {
```
<sup> Generated by Copilot </sup>

src/ui/handlers/modal-ui-handler.ts
```diff
@@ -69,6 +69,10 @@ export abstract class ModalUiHandler extends UiHandler {
     this.modalContainer.setVisible(false);
   }
 
+  protected override tearDown(): void {
+    this.modalContainer.destroy();
```
flx-sta, line 73: Q: What about the `buttonContainers`?

src/ui/handlers/modifier-select-ui-handler.ts
```diff
@@ -157,7 +157,19 @@ export class ModifierSelectUiHandler extends AwaitableUiHandler {
...
+  protected override tearDown(): void {
+    globalScene.removeInfoToggle(this.moveInfoOverlay);
+
+    this.modifierContainer.destroy();
+    this.continueButtonContainer.destroy();
+    this.transferButtonContainer.destroy();
+    this.rerollButtonContainer.destroy();
+    this.lockRarityButtonContainer.destroy();
+    this.checkButtonContainer.destroy();
+    this.moveInfoOverlay.destroy();
```
flx-sta, line 170: Q: Since MoveInfoOverlay is a custom class too, shouldn't it call `tearDown()` here?

src/ui/handlers/party-ui-handler.ts
```diff
@@ -206,7 +206,13 @@ export class PartyUiHandler extends MessageUiHandler {
     this.partySlots = [];
   }
 
-  override show(
+  protected override tearDown(): void {
+    this.partyContainer.destroy();
+    this.moveInfoOverlay.destroy();
+    this.iconAnimHandler.destroy();
+  }
```
flx-sta, line 214: Q: What about `partySlotsContainer`, `optionsContainer` & `partyCancelButton`?

src/ui/handlers/run-history-ui-handler.ts
```diff
@@ -72,9 +72,11 @@ export class RunHistoryUiHandler extends MessageUiHandler {
     globalScene.loadAtlas("rival_m", ImagesFolder.TRAINER);
   }
 
-  override show(): boolean {
-    super.show();
+  protected override tearDown(): void {
+    this.runSelectContainer.destroy();
+  }
```
flx-sta, line 78: Q: What about `runsContainer` &  `runs`

src/ui/handlers/run-info-ui-handler.ts
```diff
@@ -79,24 +79,28 @@ export class RunInfoUiHandler extends UiHandler {
...
+
     // The import of the modifiersModule is loaded here to sidestep async/await issues.
     this.modifiersModule = Modifier;
     this.runContainer.setVisible(false);
     globalScene.loadImage("encounter_exclaim", ImagesFolder.ME);
   }
 
+  protected override tearDown(): void {
+    this.runContainer.destroy();
+  }
```
flx-sta, line 94: Q: What about 
```ts
  private runResultContainer: Phaser.GameObjects.Container;
  private runInfoContainer: Phaser.GameObjects.Container;
  private partyContainer: Phaser.GameObjects.Container;
  private statsBgWidth: number;
  private hallofFameContainer: Phaser.GameObjects.Container;
  private endCardContainer: Phaser.GameObjects.Container;
```

flx-sta (commented):
45/76 Files viewed

src/ui/handlers/save-slot-select-ui-handler.ts
```diff
@@ -74,9 +74,11 @@ export class SaveSlotSelectUiHandler extends MessageUiHandler {
     this.sessionSlots = [];
   }
 
-  override show(mode: SaveSlotUiMode, slotSelectCallback: SaveSlotSelectCallback): boolean {
-    super.show();
+  protected override tearDown(): void {
+    this.saveSlotSelectContainer.destroy();
```
flx-sta, line 79: Q: What about:
```ts
  private sessionSlotsContainer: Phaser.GameObjects.Container;
  private saveSlotSelectMessageBoxContainer: Phaser.GameObjects.Container;
  private sessionSlots: SessionSlot[];
  private cursorObj: Phaser.GameObjects.Container | null;
```

src/ui/handlers/starter-select-ui-handler.ts
```diff
@@ -1017,8 +1017,14 @@ export class StarterSelectUiHandler extends MessageUiHandler {
     this.updateInstructions();
   }
 
-  override show(selectedStarterCallback?: StarterSelectCallback): boolean {
-    if (!this.starterPreferences) {
+  protected override tearDown(): void {
+    this.starterSelectContainer.destroy();
+    this.iconAnimHandler.destroy();
+    this.clearStarterPreferences();
```
flx-sta, line 1049: Q: What about all the other Containers?

flx-sta (commented):
49/76

src/ui/handlers/egg-hatch-scene-ui-handler.ts
MokaStitcher: Phaser does [seem to check its existence yes](https://github.com/phaserjs/phaser/blob/c62baca677399e098ce21bb49757974c4742895b/src/animations/AnimationManager.js#L979), so it shouldn't cause any issues if it's not defined.

<sub>I am also working under the assumption that teardown is always called after setup, hence the animation will be defined, although I supposed it's always possible that it failed to load for a reason or another

src/ui/handlers/egg-list-ui-handler.ts
MokaStitcher: what container.destroy() does is it will remove the elements contained within it, and call destroy on them recursively. Since all these elements are added to the main `eggListContainer`, we don't need to call `destroy` on them individually.

Of note, if we wanted to be able to reuse the handler after teardown we should set all these attributes to `undefined` or `null` here so that the destroyed objects can get garbage collected. But in our case after `teardown` any reference to the handler itself gets removed, so everything can be garbage collected without having to worry about doing that

src/ui/handlers/game-stats-ui-handler.ts
MokaStitcher: Same thing, statsContainer is being added to the gameStatsContainer, so it will be destroyed when gameStatsContainer is destroyed
https://github.com/Despair-Games/poketernity/blob/424514bc69361ffc374e15877d8f4a5936ffe026/src/ui/handlers/game-stats-ui-handler.ts#L282

Basically, the containers we have to destroy manually here are those who were added directly to the parent UI container, like here
https://github.com/Despair-Games/poketernity/blob/424514bc69361ffc374e15877d8f4a5936ffe026/src/ui/handlers/game-stats-ui-handler.ts#L291

(`container.destroy` will also remove them from their parent container)

src/ui/handlers/login-form-ui-handler.ts
MokaStitcher: are you sure? Did you compare with and without doing that, do you have screenshots of the heap? Also what browser are you using?

Because first, they should already be destroyed by the parent FormModalUiHandler's teardown
And secondly, I am certain I have tested doing the same just in case for some reason they weren't, and it didn't work (they get removed from the dom, but remain tied by some reference somewhere, hence the "detached" state).

src/ui/handlers/modal-ui-handler.ts
MokaStitcher: same thing, `buttonContainers` are children of the `modalContainer`, so this will destroy them
https://github.com/Despair-Games/poketernity/blob/424514bc69361ffc374e15877d8f4a5936ffe026/src/ui/handlers/modal-ui-handler.ts#L99

src/ui/handlers/modifier-select-ui-handler.ts
MokaStitcher: MoveInfoOverlay extends Container, not UiHandler, so it doesn't have a teardown function

src/ui/handlers/party-ui-handler.ts
MokaStitcher: same, reference to them are being held in the object to be able to modify them on the fly, but they're all added to the main `partyContainer`
https://github.com/Despair-Games/poketernity/blob/424514bc69361ffc374e15877d8f4a5936ffe026/src/ui/handlers/party-ui-handler.ts#L159
https://github.com/Despair-Games/poketernity/blob/424514bc69361ffc374e15877d8f4a5936ffe026/src/ui/handlers/party-ui-handler.ts#L183
https://github.com/Despair-Games/poketernity/blob/424514bc69361ffc374e15877d8f4a5936ffe026/src/ui/handlers/party-ui-handler.ts#L188

src/ui/handlers/run-history-ui-handler.ts
MokaStitcher: `runContainers` is part of `runSelectContainer`
https://github.com/Despair-Games/poketernity/blob/424514bc69361ffc374e15877d8f4a5936ffe026/src/ui/handlers/run-history-ui-handler.ts#L63
Contents of the `runs` array are being added to `runContainers` 
https://github.com/Despair-Games/poketernity/blob/424514bc69361ffc374e15877d8f4a5936ffe026/src/ui/handlers/run-history-ui-handler.ts#L184-L185

<sub> and like before we don't need to empty the array to clear the references, since the handler will be dropped altogether after teardown. But I may still do it in this case, for good measure

src/ui/handlers/run-info-ui-handler.ts
MokaStitcher: Not going to link them all this time but I double checked and they are all added to `runContainer`

src/ui/handlers/save-slot-select-ui-handler.ts
MokaStitcher: same

src/ui/handlers/starter-select-ui-handler.ts
MokaStitcher: same

src/ui/handlers/login-form-ui-handler.ts
flx-sta: Oh, I thought you were referencing "detached" as the DOM element *floating* around without the reference.
I'm confident that you did your research thus I'll trust you about the reference/heap stuff as I don't want to go down that rabbit hole :D

src/ui/handlers/message-ui-handler.ts
```diff
@@ -36,7 +36,18 @@ export abstract class MessageUiHandler extends AwaitableUiHandler {
...
+  /**
+   * Displays a dialogue message on the UI with optional delay and speaker name.
+   *
+   * @param text - The dialogue message to display.
+   * @param _name - (Optional) The name of the speaker for the dialogue. Defaults to an empty string if not provided.
+   * @param delay - (Optional) The delay in milliseconds before the dialogue is displayed. Defaults to `null` for no delay.
+   * @param callback - (Optional) A callback function to execute after the dialogue is displayed. Defaults to `null` for no callback.
+   * @param callbackDelay - (Optional) The delay in milliseconds before executing the callback. Defaults to `null` for no delay.
+   * @param prompt - (Optional) Whether to show the prompt after displaying the dialogue.
+   * @param promptDelay - (Optional) The delay in milliseconds before showing the prompt. Defaults to `null` for no delay.
```
flx-sta: I wonder why Co-Pilot added the `_name` thing.
Can't trust these AI things.
Sorry for the inconvenience, should have checked more
```suggestion
   * Displays a dialogue message on the UI with optional delay
   *
   * @param text - The dialogue message to display.
   * @param delay - (Optional) The delay in milliseconds before the dialogue is displayed. Defaults to `null` for no delay.
   * @param callback - (Optional) A callback function to execute after the dialogue is displayed. Defaults to `null` for no callback.
   * @param callbackDelay - (Optional) The delay in milliseconds before executing the callback. Defaults to `null` for no delay.
   * @param prompt - (Optional) Whether to show the prompt after displaying the dialogue.
   * @param promptDelay - (Optional) The delay in milliseconds before showing the prompt. Defaults to `null` for no delay.
```

src/utils.ts
```diff
@@ -662,6 +647,7 @@ export function hasTouchscreen(): boolean {
  * @returns `true` if the device is in landscape mode, otherwise `false` which means it is in portrait mode.
  */
 export function isLandscapeMode(): boolean {
+  // TODO: we should user Phaser's scale 'orientation' field instead
```
flx-sta: Created an issue for it: #1068

src/ui/settings/general-settings-ui-handler.ts
```diff
@@ -3,18 +3,41 @@ import { hasTouchscreen, isLandscapeMode } from "#app/utils";
 import { t } from "i18next";
 import { AbstractSettingsUiHandler } from "./abstract-settings-ui-handler";
 
-export class SettingsUiHandler extends AbstractSettingsUiHandler {
+export class GeneralSettingsUiHandler extends AbstractSettingsUiHandler {
+  private onWindowResizeEvent = () => this.updateMoveTouchControlsSettingsLabel();
```
flx-sta, line 7: NIT: alternatively use:
```suggestion
  private onWindowResizeEvent = this.updateMoveTouchControlsSettingsLabel.bind(this);
```

No difference, just giving insight

src/ui/settings/display-settings-ui-handler.ts
```diff
@@ -0,0 +1,81 @@
...
+import { supportedLanguages } from "#app/system/settings/supported-languages";
+import type { OptionSelectUiHandler } from "#app/ui/handlers/option-select-ui-handler";
+import type { OptionSelectItem } from "#app/ui/interfaces/option-select-config";
+import { displaySettingUiItems } from "#app/ui/settings/settings-ui-items";
+import { UiMode } from "#enums/ui-mode";
+import i18next from "i18next";
+import { AbstractSettingsUiHandler } from "./abstract-settings-ui-handler";
+
+export class DisplaySettingsUiHandler extends AbstractSettingsUiHandler {
+  private onLanguageChange = () => this.showLanguageOptions();
```
flx-sta: NIT: Alternatively use:
```suggestion
  private onLanguageChange = this.showLanguageOptions.bind(this);
```

NO difference, just giving insight

src/ui/helpers/pokemon-icon-anim-helper.ts
```diff
@@ -84,4 +87,12 @@ export class PokemonIconAnimHelper {
       this.icons.delete(i);
     }
   }
+
+  public destroy(): void {
```
flx-sta, line 123: NIT: Missing tsdoc

src/ui/helpers/pokemon-icon-anim-helper.ts
```diff
@@ -74,7 +77,7 @@ export class PokemonIconAnimHelper {
     }
   }
 
-  removeAll(): void {
+  public removeAll(): void {
```
flx-sta, line 108: NIT: Missing tsdoc

src/ui/helpers/pokemon-icon-anim-helper.ts
```diff
@@ -60,7 +63,7 @@ export class PokemonIconAnimHelper {
     }
   }
 
-  remove(icons: PokemonIcon | PokemonIcon[]): void {
+  public remove(icons: PokemonIcon | PokemonIcon[]): void {
```
flx-sta, line 91: NIT: Missing tsdoc

src/ui/helpers/pokemon-icon-anim-helper.ts
```diff
@@ -43,7 +46,7 @@ export class PokemonIconAnimHelper {
     }
   }
 
-  addOrUpdate(icons: PokemonIcon | PokemonIcon[], mode: PokemonIconAnimMode): void {
+  public addOrUpdate(icons: PokemonIcon | PokemonIcon[], mode: PokemonIconAnimMode): void {
```
flx-sta, line 69: NIT: Missing tsdoc

src/ui/helpers/pokemon-icon-anim-helper.ts
```diff
@@ -7,11 +7,14 @@ type PokemonIcon = Phaser.GameObjects.Container | Phaser.GameObjects.Sprite;
...
   private toggled: boolean;
+  private counterTween: Phaser.Tweens.Tween | undefined;
 
-  setup(): void {
+  constructor() {
     this.icons = new Map();
     this.toggled = false;
+  }
 
+  public setup(): void {
```
flx-sta, line 24: NIT: Missing tsdoc

flx-sta (commented):
> [!NOTE]
>  ### 76/76 Files Reviewed

_This concludes the code-review_

src/inputs-controller.ts
MokaStitcher: marking these ones as resolved as well, this should all be addressed in my next PR once this one makes it in

src/ui/settings/general-settings-ui-handler.ts
MokaStitcher: interesting, thank you 📝 

This was done this way initially as I was following a few other components/handlers already doing the same. So for consistency right now i'll probably leave it as is

src/ui/handlers/message-ui-handler.ts
MokaStitcher: I should've too, got a bit lazy on that one :')

src/battle-scene.ts
```diff
@@ -626,22 +626,26 @@ export default class BattleScene extends SceneBase {
...
     this.trainer = trainer;
 
-    this.anims.create({
-      key: "prompt",
-      frames: this.anims.generateFrameNumbers("prompt", { start: 1, end: 4 }),
-      frameRate: 6,
-      repeat: -1,
-      showOnStart: true,
-    });
+    if (!this.anims.exists("prompt")) {
```
SirzBenjie, line 626: It's worth pointing out that checking if it exists already isn't necessary. Phaser's animation manager will not create a new one if the animation already exists:
https://docs.phaser.io/api-documentation/class/animations-animationmanager#create

MokaStitcher: it does not indeed, but there's a warning in the console which annoys me :')

src/ui/handlers/abstract-ui-handler.ts
```diff
@@ -1,42 +1,168 @@
...
-import type { UiMode } from "#enums/ui-mode";
-import type { Button } from "#enums/buttons";
 import type { AwaitableUiHandler } from "#app/ui/handlers/awaitable-ui-handler";
+import type { Button } from "#enums/buttons";
+import type { UiMode } from "#enums/ui-mode";
 
 /**
  * A basic abstract class to act as a holder and processor for UI elements.
+ *
+ * Subclasses should override the following functions of a handler's lifecycle:
```
DayKev, line 9: This says these functions should be overridden, but the TSDocs of `destroy` says you should be overriding `tearDown` instead.

src/ui/handlers/abstract-ui-handler.ts
```diff
@@ -1,42 +1,168 @@
...
  */
 export abstract class UiHandler {
   protected mode: number | null;
   protected cursor: number = 0;
-  public active: boolean = false;
 
   /**
-   * @param mode The mode of the UI element. These should be unique.
+   * Whether the handler is setup and ready to be displayed through calling {@linkcode start}.
+   * note: unused for now since all handlers are ready at all times.
```
DayKev: This says it's unused but `initialize` and `destroy` both check it.

src/ui/handlers/abstract-ui-handler.ts
```diff
@@ -1,42 +1,168 @@
...
+  public get ready(): boolean {
+    return this._ready;
+  }
+
+  /**
+   * Prepares the handler for display.
+   * Should not be overridden. Calls {@linkcode setup}, which subclasses should override.
+   */
+  public initialize(): void {
+    if (this._ready) {
```
DayKev:
```suggestion
    if (this.ready) {
```

src/ui/handlers/autocomplete-ui-handler.ts
```diff
@@ -3,15 +3,19 @@ import { Button } from "#enums/buttons";
...
-  override show(config: OptionSelectModeConfig, container: Phaser.GameObjects.Container): boolean {
+  public override show(config: OptionSelectModeConfig, container: Phaser.GameObjects.Container): boolean {
     this.modalContainer = container;
 
     return super.show(config);
   }
 
-  override updateSizeForOptions(options: any): void {
+  protected override clear(): void {
+    this.modalContainer = undefined;
```
DayKev, line 16: Shouldn't this also call `.destroy` or something first?

src/ui/handlers/form-modal-ui-handler.ts
DayKev, line 1: Todo: Standardize file naming and add `abstract-` to the files that are missing it.

src/ui/handlers/message-ui-handler.ts
```diff
@@ -249,20 +270,33 @@ export abstract class MessageUiHandler extends AwaitableUiHandler {
...
+  public clearText(): void {
+    if (this.ready && this.message?.active) {
+      this.message.setText("");
+    }
     this.pendingPrompt = false;
   }
-
-  override clear() {
-    super.clear();
-  }
```
DayKev, line 267: lol

src/ui/handlers/title-ui-handler.ts
```diff
@@ -74,6 +75,17 @@ export class TitleUiHandler extends OptionSelectUiHandler {
     this.titleContainer.add(this.appVersionText);
   }
 
+  protected override tearDown(): void {
+    this.titleContainer.destroy();
+
+    if (this.splashTextTween) {
+      this.splashTextTween.remove();
```
DayKev: Does it not need to be `.destroy()`ed too?

src/ui/settings/move-touch-controls-handler.ts
```diff
@@ -109,11 +109,12 @@ export class MoveTouchControlsHandler {
 
   //#endregion
 
-  public initListeners() {
+  // Note: at the moment we don't need to remove those listeners because a single instance
```
DayKev:
```suggestion
  // Note: at the moment we don't need to remove those listeners because only a single instance
```

src/ui/settings/gamepad-settings-ui-handler.ts
```diff
@@ -24,15 +23,14 @@ import { MAPPING_CONFIG_LS_KEY } from "#app/constants/app";
...
  * @extends AbstractControlSettingsUiHandler
  */
-
-export class SettingsGamepadUiHandler extends AbstractControlSettingsUiHandler {
+export class GamepadSettingsUiHandler extends AbstractControlSettingsUiHandler {
   /**
-   * Creates an instance of SettingsGamepadUiHandler.
+   * Creates an instance of GamepadSettingsUiHandler.
    *
    * @param mode - The UI mode, optional.
```
DayKev:
```suggestion
```

src/ui/settings/gamepad-settings-ui-handler.ts
```diff
@@ -24,15 +23,14 @@ import { MAPPING_CONFIG_LS_KEY } from "#app/constants/app";
...
    */
-  constructor(mode: UiMode | null = null) {
-    super(mode);
+  constructor() {
```
DayKev, line 28: There's no `UiMode` for this handler?

src/ui/settings/general-settings-ui-handler.ts
```diff
@@ -3,18 +3,41 @@ import { hasTouchscreen, isLandscapeMode } from "#app/utils";
...
+
   /**
    * Creates an instance of SettingsGamepadUiHandler.
    *
    * @param mode - The UI mode, optional.
```
DayKev:
```suggestion
```

src/ui/settings/display-settings-ui-handler.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  constructor() {
+    super("display", displaySettingUiItems);
+  }
+
+  protected override setup(): void {
+    super.setup();
+
+    eventBus.on("language/change", this.onLanguageChange);
```
DayKev: Why not just directly use `this.showLanguageOptions`?

src/ui/ui.ts
```diff
@@ -201,6 +203,38 @@ export class UI extends Phaser.GameObjects.Container {
...
+    (this.handlers[UiMode.STARTER_SELECT] as StarterSelectUiHandler).clearStarterPreferences();
+    this.resetModeChain();
+  }
+
+  public resetModeChain(): void {
+    this.modeChain = [];
+    globalScene.updateGameInfo();
+  }
+
+  override destroy(fromScene?: boolean): void {
```
DayKev:
```suggestion
  public override destroy(fromScene?: boolean): void {
```

src/battle-scene.ts
```diff
@@ -889,6 +894,19 @@ export default class BattleScene extends SceneBase {
     this.infoToggles.push(infoToggle);
   }
 
+  /**
+   * Removes a specified {@linkcode InfoToggle} from the {@linkcode infoToggles} array.
+   *
+   * @param infoToggle - The {@linkcode InfoToggle} instance to be removed. If the toggle
+   *                     is not found in the array, no action is taken.
+   */
+  removeInfoToggle(infoToggle: InfoToggle): void {
```
```suggestion
  public removeInfoToggle(infoToggle: InfoToggle): void {
```

DayKev (commented):
We move ever closer to re-creating Phaser scenes within our UI code lmao

src/battle-scene.ts
DayKev: Ah, so that's why these appeared:
image

src/ui/handlers/abstract-ui-handler.ts
MokaStitcher: it's unused in the sense that if it didn't exist right now the game would behave the same.
Right now all handlers are initialized and thus always in a ready state. As I attempt to explain in the PR description it is more here to prepare for the future where this is no longer the case

src/ui/handlers/abstract-ui-handler.ts
MokaStitcher: good catch, I changed things at some point and missed that

src/ui/handlers/abstract-ui-handler.ts
DayKev: Hm, alright. It's still used in the sense that it protects from a handler being initialized multiple times (ie: it protects against [future] bugs).

src/ui/handlers/autocomplete-ui-handler.ts
MokaStitcher: <sub>this whole autoComplete handler is a a bit confusing unfortunately. It should probably be removed or at least reworked, but it's not a priority at the moment since it's only being used by a non player facing UI

Since this handler is not the one creating or populating the container it is not the one responsible for destroying it, no.
The container is being passed to it in the show function, and the handler only keeps a reference to it. So when it gets cleared all it does is free that reference so that it can get garbage collected if destroyed by its actual "parent handler"

src/ui/handlers/title-ui-handler.ts
MokaStitcher: I had one tween that I tried to use `destroy` directly on, and it broke things. I don't remember if it was this one and I found the doc to be a little unclear on whether remove was enough or not, I may need to give it another look...

src/ui/settings/gamepad-settings-ui-handler.ts
MokaStitcher: good catch

<sub>fun fact. That mode attribute of ui handlers is I think never actually being accessed/used in any meaningful way

src/ui/settings/display-settings-ui-handler.ts
MokaStitcher: i was following the template used by other handlers who did similar things.

I think if we wanted to use `this.showLanguageOptions` we'd have to do `eventBus.on("language/change", this.showLanguageOptions, this);` to give it the proper context otherwise it wouldn't be able to resolve the proper `this`, which is solved by using this extra step of `onLanguageChange`.

I don't know if there's any real difference but it would probably be cleaner to change it.
But it would mean we should update all handlers doing this and that means going through allll handlers and components to double check for this pattern..

src/ui/handlers/form-modal-ui-handler.ts
MokaStitcher: Should this be done in this PR? (probably not but i already renamed some handlers so... maybe...)

DayKev: If you want, this was just a note for the future.

src/ui/handlers/abstract-ui-handler.ts
MokaStitcher: this line will go once this is no longer the case 🤷

src/ui/settings/display-settings-ui-handler.ts
MokaStitcher: updated for this file. Left the others as is for now. The main ones using this pattern are the arena and battle flyouts

src/ui/handlers/form-modal-ui-handler.ts
MokaStitcher: hmm.. honestly I'm not sure about adding Abstract to the abstract class names. For the very very base UiHandler I can see it, but for the others I don't know, seems a bit weird. May need to discuss this with the others and leave it for another time

src/ui/handlers/title-ui-handler.ts
MokaStitcher: Seems like it is fine to destroy this one.
There are a few other tweens that are being removed and not destroyed in the code base, most of them relating to scrolling text. For now I have left them but will create an issue to have a look at the scrolling text

flx-sta (dismissed):

src/data/animations/battle-anims.ts
```diff
@@ -237,6 +237,10 @@ export abstract class BattleAnim {
       targetSprite.pipelineData["tone"] = [0.0, 0.0, 0.0, 0.0];
       targetSprite.setAngle(0);
 
+      // Remove animation event listeners
+      userSprite.off("animationupdate");
+      targetSprite.off("animationupdate");
+
```
flx-sta, line 243: Q: Where are those added?

MokaStitcher: https://github.com/Despair-Games/poketernity/blob/726f13b8bec71448ddd5a855b0d795820dfb1fb9/src/data/animations/battle-anims.ts#L343-L349

spriteSource is either userSprite or targetSprite, depending
This is a bit of a brute force approach to removing them but from my testing things were ok

flx-sta: Thanks for the details

src/ui/handlers/egg-gacha-ui-handler.ts
```diff
@@ -349,6 +351,11 @@ export class EggGachaUiHandler extends MessageUiHandler {
     return true;
   }
 
+  protected override clear(): void {
+    this.setGachaCursor(-1);
```
flx-sta, line 355: Q: Why `-1`?

src/ui/handlers/menu-ui-handler.ts
```diff
@@ -20,7 +20,7 @@ import { TextStyle } from "#enums/text-style";
 import { Tutorial } from "#enums/tutorial";
 import { UiMode } from "#enums/ui-mode";
 import i18next from "i18next";
-import type { SettingsUiHandler } from "../settings/settings-ui-handler";
+import type { GeneralSettingsUiHandler } from "../settings/general-settings-ui-handler";
```
flx-sta: NIT: 
```suggestion
import type { GeneralSettingsUiHandler } from "#app/ui/settings/general-settings-ui-handler";
```

src/ui/handlers/title-ui-handler.ts
```diff
@@ -26,12 +26,13 @@ export class TitleUiHandler extends OptionSelectUiHandler {
   private eventDisplay?: TimedEventDisplay;
 
   private titleStatsTimer: NodeJS.Timeout | null;
```
flx-sta, line 28: Q: I think I added that type, but I'm not sure it's correct

src/ui/handlers/egg-gacha-ui-handler.ts
MokaStitcher: to reset it so that it doesn't keep in mind the last machine used the next time it's showed (this is not a change of this PR, the clear function was just moved up from another part of the file)

src/ui/handlers/title-ui-handler.ts
MokaStitcher: that's probably because it's being set to `null` in `clear()`. Which is done to free it from memory while the handler itself remains in memory at all times.
I don't know if it's the best approach but I don't think it's necessarily wrong either

src/utils/app-utils.ts
```diff
@@ -60,6 +60,7 @@ export function hasTouchscreen(): boolean {
  * @returns `true` if the device window's `width > height`, and `false` otherwise
  */
 export function isLandscapeMode(): boolean {
+  // TODO: we should user Phaser's scale 'orientation' field instead
```
flx-sta, line 63: You can probably remove that, as I've created a follow up task for that
#1068

<sub>Alternative you can also link the issue-number ofc</sub>

flx-sta (dismissed):

DayKev (dismissed):
Got this error when I caused the game window to be resized by opening the browser devtools.
image

It only happened once though. Not sure if it's related to this PR or just a coincidence. Regardless, the game continued to run normally afterwards so... 🤷 

Other than that, I tried a few things and didn't find anything broken. 👍

MokaStitcher:
> Got this error when I caused the game window to be resized by opening the browser devtools. image
> 
> It only happened once though. Not sure if it's related to this PR or just a coincidence. Regardless, the game continued to run normally afterwards so... 🤷
> 
> Other than that, I tried a few things and didn't find anything broken. 👍

I've seen it a few times when playing with the dev tools as well I think, so I'm not fully sure. I'll try and get a second look at it

src/ui/handlers/title-ui-handler.ts
```diff
@@ -75,6 +76,18 @@ export class TitleUiHandler extends OptionSelectUiHandler {
     this.titleContainer.add(this.appVersionText);
   }
 
+  protected override tearDown(): void {
+    this.titleContainer.destroy();
+
+    if (this.splashTextTween) {
+      //this.splashTextTween.remove();
```
innerthunder: remove comment

innerthunder (dismissed):

MokaStitcher:
> > Got this error when I caused the game window to be resized by opening the browser devtools. image
> > It only happened once though. Not sure if it's related to this PR or just a coincidence. Regardless, the game continued to run normally afterwards so... 🤷
> > Other than that, I tried a few things and didn't find anything broken. 👍
> 
> I've seen it a few times when playing with the dev tools as well I think, so I'm not fully sure. I'll try and get a second look at it

Investigated a little and I think it's caused by somehow the first input registered by the game being a key up one, rather than a key down one. The keyboard is initialized only on key down events, and so no mapping is found if the game encounters a key up event without having gone through a key down one first. So if that's correct, it should only happen when launching the game and somehow pressing a button with perfect timing during loading to make this happen.
Does this fit your experience @DayKev or do you remenber it happening in the middle of gameplay?
For now considering it as unrelated to this PR and made an issue for it because the file could really use an overhaul #1094

Tempo-anon (approved):

flx-sta (approved):

DayKev:
> Investigated a little and I think it's caused by somehow the first input registered by the game being a key up one, rather than a key down one. The keyboard is initialized only on key down events, and so no mapping is found if the game encounters a key up event without having gone through a key down one first. So if that's correct, it should only happen when launching the game and somehow pressing a button with perfect timing during loading to make this happen. Does this fit your experience @DayKev or do you remenber it happening in the middle of gameplay?

Ah, could be. I have one of the buttons on my mouse set up as a macro to open the browser devtools and it only happened when I pushed the button as the page started to load. I haven't seen the error before now iirc so it requiring a very specific timing makes sense.
image

flx-sta:
> > Investigated a little and I think it's caused by somehow the first input registered by the game being a key up one, rather than a key down one. The keyboard is initialized only on key down events, and so no mapping is found if the game encounters a key up event without having gone through a key down one first. So if that's correct, it should only happen when launching the game and somehow pressing a button with perfect timing during loading to make this happen. Does this fit your experience @DayKev or do you remenber it happening in the middle of gameplay?
> 
> Ah, could be. I have one of the buttons on my mouse set up as a macro to open the browser devtools and it only happened when I pushed the button as the page started to load. I haven't seen the error before now iirc so it requiring a very specific timing makes sense. image

Guess you’ll have to fix that bug yourself then 😁

PR #885, opened 2025-03-30
[Feature] Add uniform dynamic turn command scheduling

### LINKED ISSUE #940: [Feature] TurnCommand management
See #885

innerthunder:

Dynamic turn order now applies to commands of *all* types, rather than just `FIGHT` commands. The main difference is that multiple `POKEMON` commands (i.e. switching) is now handled dynamically instead of statically based on speed order at the start of the turn. Consider the following example:
- You're in a double battle against a Trainer, and the battle starts with Rain on the field.
- You want to switch out a Buizel (100 Spd) with Swift Swim, while the enemy Trainer wants to switch out both of its Pokemon: a Buizel (102 Spd + Swift Swim) and a Hoppip (80 Spd + Chlorophyll). The first enemy Pokemon to switch in has Drought.
- Under the current logic, the switch order would be Enemy Buizel > Player Buizel > Enemy Hoppip. The change in weather as the Pokemon switch does not affect order.
- With the changes here, the switch order would be Enemy Buizel > Enemy Hoppip > Player Buizel. Because the change in weather after the first switch deactivates the player Buizel's Swift Swim and activates the enemy Hoppip's Chlorophyll, the overall switch order changes.

A few reasons:
1. Dynamic order currently only applies to `FIGHT` commands, so overall turn order may not be accurate to mainline in some cases.
2. A uniform approach for command handling makes it easier to introduce an event-based approach to command handling in the future.
3. These changes will make it easier to implement [Pursuit](https://bulbapedia.bulbagarden.net/wiki/Pursuit_(move)) and other complex effects on turn order.

- `MoveEndPhase` has been repurposed into the more general `PostActionPhase`. This is queued after every command to
  1. Search for commands in the turn command manager's queue.
  2. If valid commands are found, schedule the next valid command in turn order.
  3. If no valid commands are found, start the end-of-turn phase sequence. 
- `TurnCommandManager#startTurn` now only schedules the first command in turn order to start the turn instead of scheduling all non-`FIGHT` commands in static order. Some other scheduling functions have been rewritten in support of this change.

## How to test the changes?

`npm run test`

## Checklist

----

> [!CAUTION]
> Blocks #988


DayKev:
For Terastallization, actions will need to go in this order (just tested on Showdown to confirm):
1. All switches happen in speed order, then
2. All Terastallization happens in speed order, then
3. All moves happen in priority+speed order

Will the changes make this easier to handle?

innerthunder:
> For Terastallization, actions will need to go in this order (just tested on Showdown to confirm):
> 
> 1. All switches happen in speed order, then
> 2. All Terastallization happens in speed order, then
> 3. All moves happen in priority+speed order
> 
> Will the changes make this easier to handle?

It's not complete atm, but it sounds like this applies alongside the logic to apply move header attributes. It should be easy to insert Tera before those attributes apply once the logic is fully worked out.

src/phases/weather-effect-phase.ts
```diff
@@ -38,60 +42,88 @@ export class WeatherEffectPhase extends CommonAnimPhase {
...
-    this.executeForAll((pokemon: Pokemon) => {
-      const immune: boolean =
-        !pokemon
-        || pokemon.getTypes(true, true).filter((t) => weather?.isTypeDamageImmune(t)).length > 0
-        || pokemon.switchOutStatus;
-      if (!immune) {
-        inflictDamage(pokemon);
-      }
-    });
+    const damage = toDmgValue(pokemon.getMaxHp() / 16);
```
flx-sta: Every weather damage is always 1/16 of max-life? 
Can we put this value into a constant ?

DayKev: Looks like end of turn weather damage is always 1/16. https://bulbapedia.bulbagarden.net/wiki/Weather

flx-sta (dismissed):

DayKev:
Updated the Focus Punch test so it'll output more descriptive errors if any of the `expect`s fail (since that test file was already being updated a bunch by this PR).

DayKev (approved):

PR #893, opened 2025-03-31
[Refactor] Add `Pokemon#hasTag()` function

DayKev:
N/A

Less `!!`.

New `Pokemon#hasTag()` function added which returns `true`/`false` for if a Pokémon has one of the listed `BattlerTag`s.
Existing uses of `!!*.getTag()` were replaced with `*.hasTag()`.

## Checklist

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -1716,7 +1716,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     // Note: This code is also copied in `GroundedTag.onAdd()`, to check whether or not the Pokemon
     // was grounded before receiving the `GroundedTag`.
     return (
-      !!this.getTag(BattlerTagType.IGNORE_FLYING)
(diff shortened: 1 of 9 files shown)


src/field/pokemon.ts
```diff
@@ -3451,6 +3451,13 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     return this.summonData.tags.find((t) => tagTypes.includes(t.tagType)) as T | undefined;
   }
 
+  public hasTag(...tagTypes: BattlerTagType[]): boolean {
```
Tempo-anon, line 3459: Public function should have documentation.

```suggestion

  /**
   * Helper function to check if a Pokemon has any of the input tag types
   * @param tagTypes the battler tag types to search for
   * returns true if the Pokemon has at least one of the battler tags, false otherwise
   */
  public hasTag(...tagTypes: BattlerTagType[]): boolean {
```

Tempo-anon (approved):

torranx (approved):

PR #900, opened 2025-04-02
[P1] `WeatherEffectPhase` no longer requires an active Pokemon

innerthunder:
> [!NOTE]
> **This is an [`Addendum`](https://github.com/Despair-Games/poketernity/labels/Addendum) for the [`post-action-phase`](https://github.com/Despair-Games/poketernity/tree/post-action-phase) branch** 
> Meaning the base branch is **not** ~~`beta`~~ 

The game should no longer crash (as seen in #885) when the field is empty at the end of a turn.

Fixing a bug in #885. `WeatherEffectPhase` extending `PokemonPhase` caused issues after its scheduling was moved from the start of the turn (where at least one Pokemon is guaranteed to be on the field)

`WeatherEffectPhase` has been rewritten as follows:
- It now extends `FieldPhase` instead of `CommonAnimPhase`
- The `CommonBattleAnims` used for weather in this phase should now play (to the best of their ability) without needing a reference to a Pokemon on the field
- Some of the internal `const` functions have been rewritten to protected methods
- Reorganized some logic to better reflect the timing of the phase's effects.

## Screenshots/Videos

<details><summary>Sun animation now plays on an empty field</summary>
<p>

https://github.com/user-attachments/assets/6636d0ed-6d07-40ad-8c9f-c2549e84e2a0

Overrides:
```ts
const overrides = {
  WEATHER_OVERRIDE: WeatherType.SUNNY,
  MOVESET_OVERRIDE: MoveId.EXPLOSION
```

</p>
</details> 

## How to test the changes?

`npm run test [aftermath | battle | destiny-bond]`: These fail in #885 due to `WeatherEffectPhase`'s implementation

## Checklist

- [x] ⚠️ If this is a PR for `main` (such as a hotfix), has the game version been updated (`npm run update-version:patch` / `npm run update version:minor`?

### DIFF
--- a/src/phases/weather-effect-phase.ts
+++ b/src/phases/weather-effect-phase.ts
@@ -38,7 +38,8 @@ export class WeatherEffectPhase extends FieldPhase {
     const weatherAnim = new CommonBattleAnim(weatherAnimType, globalScene.getPlayerPokemon()!, undefined, true);
 
     globalScene.ui.showText(getWeatherLapseMessage(weather.weatherType) ?? "", null, () => {
-      weatherAnim.play(false, () => {
+      // (36, 80) is the player's center-field position
+      weatherAnim.playWithoutTargets(36, 80, 2, 5, () => {
         this.applyWeatherEffects(weather);
         this.end();
       });
--- a/src/phases/weather-effect-phase.ts
+++ b/src/phases/weather-effect-phase.ts
@@ -16,6 +16,14 @@ import { HitResult } from "#enums/hit-result";
 import { PhaseId } from "#enums/phase-id";
 import { WeatherType } from "#enums/weather-type";
 
+/**
+ * Applies the end-of-turn effects from active {@linkcode Weather}, including
+ * - the weather's post-turn animation and message
+ * - the damaging effects of Hail and Sandstorm
+ * - all post-turn ability triggers dependent on the current weather
+ * (e.g. Rain Dish, Dry Skin)
+ * @extends FieldPhase
+ */
 export class WeatherEffectPhase extends FieldPhase {
   override readonly id = PhaseId.WEATHER_EFFECT;
 
@@ -35,6 +43,7 @@ export class WeatherEffectPhase extends FieldPhase {
     }
 
     const weatherAnimType: CommonAnim = CommonAnim.SUNNY + (weather.weatherType - 1);
+    /** @todo Rework animation params so that the placeholder "user" can be removed */
     const weatherAnim = new CommonBattleAnim(weatherAnimType, globalScene.getPlayerPokemon()!, undefined, true);
 
     globalScene.ui.showText(getWeatherLapseMessage(weather.weatherType) ?? "", null, () => {
@@ -50,9 +59,9 @@ export class WeatherEffectPhase extends FieldPhase {
    * Applies the end-of-turn effects from the current weather, including
    * - Damage from Sandstorm and Hail
    * - Ability triggers, e.g. healing from {@link https://bulbapedia.bulbagarden.net/wiki/Rain_Dish_(Ability) | Rain Dish}
-   * @param weather the current {@linkcode Weather} on the field
+   * @param weather - The current {@linkcode Weather} on the field
    */
-  protected applyWeatherEffects(weather: Weather): void {
+  private applyWeatherEffects(weather: Weather): void {
     if (this.tryCancelWeatherEffects(weather)) {
       return;
     }
@@ -68,14 +77,14 @@ export class WeatherEffectPhase extends FieldPhase {
 
   /**
    * Checks the field for effects that nullify the current weather's end-of-turn effects
-   * @param weather the current {@linkcode Weather} on the field
+   * @param weather - The current {@linkcode Weather} on the field
    * @returns `true` if the weather is suppressed
    *
    * @todo Merge this logic with {@linkcode Weather.isEffectSuppressed}. This method applies the suppressing attributes directly
    * while `isEffectSuppressed` uses a `hasAbility` check to avoid flyouts appearing at the wrong time. These may
    * be combined by rewriting `isEffectSuppressed` to support simulated application.
    */
-  protected tryCancelWeatherEffects(weather: Weather): boolean {
+  private tryCancelWeatherEffects(weather: Weather): boolean {
     const cancelled = new BooleanHolder(false);
 
     this.executeForAll((pokemon: Pokemon) =>
@@ -90,10 +99,10 @@ export class WeatherEffectPhase extends FieldPhase {
    * of specific {@linkcode WeatherType | types of weather}. This accounts for
    * immunity to weather-based damage from the Pokemon's typing, abilities, and
    * other effects.
-   * @param weather the current {@linkcode Weather} on the field
-   * @param pokemon the {@linkcode Pokemon} to which damage may apply
+   * @param weather - The current {@linkcode Weather} on the field
+   * @param pokemon - The {@linkcode Pokemon} to which damage may apply
    */
-  protected tryInflictWeatherDamage(weather: Weather, pokemon: Pokemon): void {
+  private tryInflictWeatherDamage(weather: Weather, pokemon: Pokemon): void {
     if (!weather.isDamaging()) {
       return;
     }
--- a/src/phases/weather-effect-phase.ts
+++ b/src/phases/weather-effect-phase.ts
@@ -3,10 +3,11 @@ import type { PostWeatherLapseAbAttr } from "#app/data/abilities/ab-attrs/post-w
 import type { PreWeatherDamageAbAttr } from "#app/data/abilities/ab-attrs/pre-weather-damage-ab-attr";
 import type { SuppressWeatherEffectAbAttr } from "#app/data/abilities/ab-attrs/suppress-weather-effect-ab-attr";
 import { applyAbAttrs } from "#app/data/abilities/apply-ab-attrs";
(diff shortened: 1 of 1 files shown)


src/phases/weather-effect-phase.ts
```diff
@@ -33,60 +34,87 @@ export class WeatherEffectPhase extends CommonAnimPhase {
...
+    });
+  }
 
-    if (!weather.isDamaging()) {
-      return end();
+  /**
+   * Applies the end-of-turn effects from the current weather, including
+   * - Damage from Sandstorm and Hail
+   * - Ability triggers, e.g. healing from {@link https://bulbapedia.bulbagarden.net/wiki/Rain_Dish_(Ability) | Rain Dish}
+   * @param weather the current {@linkcode Weather} on the field
```
DayKev:
```suggestion
   * @param weather - The current {@linkcode Weather} on the field
```

src/phases/weather-effect-phase.ts
```diff
@@ -33,60 +34,87 @@ export class WeatherEffectPhase extends CommonAnimPhase {
       return this.end();
     }
 
-    this.setAnimation(CommonAnim.SUNNY + (weather.weatherType - 1));
+    const weatherAnimType: CommonAnim = CommonAnim.SUNNY + (weather.weatherType - 1);
+    const weatherAnim = new CommonBattleAnim(weatherAnimType, globalScene.getPlayerPokemon()!, undefined, true);
```
flx-sta, line 47: Can we please avoid adding more `!` to the codebase?

DayKev, line 47: That `!` reminds me that the animation code is jank af.

src/phases/weather-effect-phase.ts
```diff
@@ -33,60 +34,87 @@ export class WeatherEffectPhase extends CommonAnimPhase {
...
+    this.executeForAll((pokemon: Pokemon) => {
+      if (!pokemon.switchOutStatus) {
+        applyAbAttrs<PostWeatherLapseAbAttr>(AbAttrFlag.POST_WEATHER_LAPSE, pokemon, false, weather);
+      }
+    });
+  }
+
+  /**
+   * Checks the field for effects that nullify the current weather's end-of-turn effects
+   * @param weather the current {@linkcode Weather} on the field
```
DayKev:
```suggestion
   * @param weather - The current {@linkcode Weather} on the field
```

src/phases/weather-effect-phase.ts
```diff
@@ -33,60 +34,87 @@ export class WeatherEffectPhase extends CommonAnimPhase {
...
 
-      applyAbAttrs<PreWeatherDamageAbAttr>(AbAttrFlag.PRE_WEATHER_DAMAGE, pokemon, false, weather, cancelled);
-      applyAbAttrs<BlockNonDirectDamageAbAttr>(AbAttrFlag.BLOCK_NON_DIRECT_DAMAGE, pokemon, false, cancelled);
+  /**
+   * Inflicts damage on the given Pokemon from the end-of-turn effects
+   * of specific {@linkcode WeatherType | types of weather}. This accounts for
+   * immunity to weather-based damage from the Pokemon's typing, abilities, and
+   * other effects.
+   * @param weather the current {@linkcode Weather} on the field
+   * @param pokemon the {@linkcode Pokemon} to which damage may apply
```
DayKev:
```suggestion
   * @param weather - The current {@linkcode Weather} on the field
   * @param pokemon - The {@linkcode Pokemon} to which damage may apply
```

src/phases/weather-effect-phase.ts
DayKev: The problem is that the params for `CommonBattleAnim` and `BattleAnim` are not in sync; the animation code needs to be refactored.

src/phases/weather-effect-phase.ts
```diff
@@ -33,60 +34,87 @@ export class WeatherEffectPhase extends CommonAnimPhase {
...
 
-    if (!weather.isDamaging()) {
-      return end();
+  /**
+   * Applies the end-of-turn effects from the current weather, including
+   * - Damage from Sandstorm and Hail
+   * - Ability triggers, e.g. healing from {@link https://bulbapedia.bulbagarden.net/wiki/Rain_Dish_(Ability) | Rain Dish}
+   * @param weather the current {@linkcode Weather} on the field
+   */
+  protected applyWeatherEffects(weather: Weather): void {
```
torranx: i think the methods should be `private`?

torranx (commented):
wouldnt tag this as a refactor but a bug fix (p1/p2) instead

src/phases/weather-effect-phase.ts
```diff
@@ -33,60 +34,87 @@ export class WeatherEffectPhase extends CommonAnimPhase {
       return this.end();
     }
 
-    this.setAnimation(CommonAnim.SUNNY + (weather.weatherType - 1));
+    const weatherAnimType: CommonAnim = CommonAnim.SUNNY + (weather.weatherType - 1);
```
Tempo-anon, line 45: nit: unrelated to this PR but seeing `blahblah.SUNNY + (weather.weatherType -1)` in multiple places in the codebase makes me feel like there should be a helper function that gets the correct value for a given weatherType

src/phases/weather-effect-phase.ts
```diff
@@ -15,7 +16,7 @@ import { HitResult } from "#enums/hit-result";
 import { PhaseId } from "#enums/phase-id";
 import { WeatherType } from "#enums/weather-type";
 
-export class WeatherEffectPhase extends CommonAnimPhase {
+export class WeatherEffectPhase extends FieldPhase {
```
Tempo-anon, line 27: This class should have some documentation describing what it's doing and how it's being used

src/phases/weather-effect-phase.ts
```diff
@@ -33,60 +34,87 @@ export class WeatherEffectPhase extends CommonAnimPhase {
...
+    applyAbAttrs<PreWeatherDamageAbAttr>(AbAttrFlag.PRE_WEATHER_DAMAGE, pokemon, false, weather, cancelled);
+    applyAbAttrs<BlockNonDirectDamageAbAttr>(AbAttrFlag.BLOCK_NON_DIRECT_DAMAGE, pokemon, false, cancelled);
 
-      globalScene.queueMessage(getWeatherDamageMessage(weather.weatherType, pokemon) ?? "");
-      pokemon.damageAndUpdate(damage, { result: HitResult.EFFECTIVE, preventEndure: true });
-    };
+    if (
+      cancelled.value
+      || pokemon.getTypes(true, true).some((t) => weather.isTypeDamageImmune(t))
+      || pokemon.getTag(BattlerTagType.UNDERGROUND, BattlerTagType.UNDERWATER)
```
Tempo-anon, line 118: Just to double check, is this only underground and underwater but not any of the other semi invulnerable tags?

src/phases/weather-effect-phase.ts
innerthunder: Not sure if this is in the PR's scope

Tempo-anon: agree, just pointing it out since I also saw it a lot while refactoring other weather stuff

src/phases/weather-effect-phase.ts
innerthunder: Added a `todo` comment. It's an underlying issue with `BattleAnim.playWithoutTargets` that can be seen with `EncounterAnims` in other parts of the codebase. I'm just copying how those are initialized (e.g. in `fiery-fallout-encounter.ts`) for now.

src/phases/weather-effect-phase.ts
innerthunder: Bulba confirms this for [Dig](https://bulbapedia.bulbagarden.net/wiki/Dig_(move)) and [Dive](https://bulbapedia.bulbagarden.net/wiki/Dive_(move)), and omits that info from [Fly](https://bulbapedia.bulbagarden.net/wiki/Fly_(move)) and [Phantom Force](https://bulbapedia.bulbagarden.net/wiki/Phantom_Force_(move)). I'm assuming this doesn't need to be changed according to that.

DayKev (approved):

src/phases/weather-effect-phase.ts
```diff
@@ -33,60 +42,88 @@ export class WeatherEffectPhase extends CommonAnimPhase {
...
-        this.executeForAll((pokemon: Pokemon) => {
-          if (!pokemon.switchOutStatus) {
-            applyAbAttrs<PostWeatherLapseAbAttr>(AbAttrFlag.POST_WEATHER_LAPSE, pokemon, false, weather);
-          }
-        });
-
-        super.start();
+    globalScene.ui.showText(getWeatherLapseMessage(weather.weatherType) ?? "", null, () => {
+      // (36, 80) is the player's center-field position
+      weatherAnim.playWithoutTargets(36, 80, 2, 5, () => {
```
PigeonBar, line 51: Unfortunately, this same positioning does not seem to work for other weathers, such as snow:

https://github.com/user-attachments/assets/8adb9c49-9c5b-4aaa-ae2b-650bec86e515

Tempo-anon: Is this the only blocker for this PR? If it isn't too egregious this is just a new p3 that can be fixed later. Worst case scenario we can even have hardcoded values for all the weather types

innerthunder: Will be addressed in the battle anim refactor. I think we can just leave this as a P3 for now

PR #909, opened 2025-04-04
[Refactor][Items] Potions/Berries

flx-sta:

A first idea for the item refactor with the example of `potion` and `super-potion`

TBA

## Checklist


src/data/items/consumable/healing/potions.ts
```diff
@@ -0,0 +1,25 @@
...
+export class SuperPotion extends HealingItem {
+  constructor() {
+    super({
+      id: ItemId.SUPER_POTION,
+      rarity: ItemRarity.GREAT,
+      healAmount: 50,
+      healPercentage: 0.25,
+    });
+  }
+}
```
Tempo-anon: I think putting all the `Potion` classes in this file is fine but I also know there's the opinion that every class deserves its own file

src/data/items/consumable/consumable-item.ts
```diff
@@ -0,0 +1,13 @@
+import type { Pokemon } from "#app/field/pokemon";
+import { BaseItem } from "../base-item";
```
Tempo-anon: nit: `..` import

src/data/items/consumable/healing/potions.ts
flx-sta: I agree. I'd prefer to put them into separate files in the long run. 

_I just cut corners now for the simplicity of it being a proposition_

src/data/items/berry-items/berry-item.ts
```diff
@@ -0,0 +1,11 @@
+import type { Consumable, ConsumeContext } from "#app/@types/item/Consumable";
+import type { FlingContext, Holdable } from "#app/@types/item/Holdable";
+import { BaseItem } from "#app/data/items/base-item";
+
+export abstract class BerryItem extends BaseItem implements Holdable, Consumable {
```
innerthunder: I don't think berries match the definition of a `Consumable` item. Presumably the `onConsume` event is triggered by the player actively using the item on a Pokemon. Berries, on the other hand, passively apply under certain battle conditions (i.e. in a different, in-battle event). They should apply like other held items, but remove themselves after applying.

src/@types/item/Consumable.ts
```diff
@@ -0,0 +1,15 @@
+import type { Pokemon } from "#app/field/pokemon";
+
+export interface Consumable {
+  beforeConsume?(): void;
+  onConsume(context: ConsumeContext): void;
+  afterConsume?(): void;
+}
```
innerthunder: I don't think an event-driven approach works as well for consumables as it does for held items. It makes sense for held items since they respond to specific events and conditions in battle, but consumables are individually selected and applied by the player.

Tempo-anon: I think that's sort of on me for the ambiguous naming of Consumable. I originally intended for Consumable to refer to all the things you saw in the reward/shop screen where an item is picked and used immediately to apply its effects, without having to consider about anything afterwards

flx-sta: @innerthunder I would disagree [partially]. The user simply emits the "use" (or "consume") event thus the even handlers will be called. It's the same with any element behaviors in html. The element has it's even handler for "click" and as soon as the user clicks on the element it executes it's click-handler

src/@types/item/Eatable.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * An item that can be eaten by a {@linkcode Pokemon}. (e.g. a {@linkcode BerryItem})
+ */
+export interface Eatable {
```
Tempo-anon: nit: should be Edible rather than Eatable

src/@types/item/Transferable.ts
```diff
@@ -0,0 +1,20 @@
+import type { Pokemon } from "#app/field/pokemon";
+
+/**
+ * An item that can be transferred from one {@linkcode Pokemon} to another.
+ */
+export interface Transferable {
```
Tempo-anon: Is this needed within the current context? Can't this be a boolean field on the `Holdable` interface?

src/data/items/berry-items/cheri-berry.ts
```diff
@@ -0,0 +1,31 @@
...
+import { ItemId } from "#enums/item-id";
+import { ItemRarity } from "#enums/item-rarity";
+import { StatusEffect } from "#enums/status-effect";
+
+export class SitrusBerry extends BerryItem {
+  public static readonly HEAL_PERCENTAGE = 0.25;
+
+  constructor(holder: Pokemon) {
+    super({
+      id: ItemId.SITRUS_BERRY,
```
Tempo-anon: Are these filler values for now since there is no cheri-berry right now? You can probably use the existing lum berry that cures all status effects (and confusion)

src/data/items/berry-items/cheri-berry.ts
```diff
@@ -0,0 +1,31 @@
...
+    });
+  }
+
+  public override eat(): void {
+    if (this.holder.status?.statusEffect === StatusEffect.PARALYSIS) {
+      this.holder.resetStatus();
+    }
+  }
+
+  public override onStatusEffectApplied({ statusEffect }: StatusEffectAppliedContext): void {
```
Tempo-anon: berries should be applying their effect whenever it's applicable (the holder has a status, or the holder is at low hp) rather than only checking for when a status is applied or damage is done

src/@types/item/Eatable.ts
flx-sta: I had that initially but thought it might be confused with some 420 bs :D

src/@types/item/Transferable.ts
flx-sta: My intention was to include conditional listeners or something in that interface in the future.
But with the current given it's possible that this should only be a simple boolean on Holdable

src/data/items/berry-items/cheri-berry.ts
flx-sta: Whops. No there is an `ItemId` for it I simply forgot it.

So the current game has no cheri-berry? I didn't check on any I simply went through bulbapedia and chose the few I liked

src/data/items/berry-items/cheri-berry.ts
flx-sta: What is your suggestion? 
I would simply include the check in another listener like "onAfterBattle" or "onSendOut" etc.
That was my intention

Tempo-anon: For the healing berries and stat boosting berries I believe they both trigger at the end of the turn if their conditions are met. The only other type of berry that aren't implemented are Custap (that increases user's priority by +1 if low hp, and the type damage reduction berries)

flx-sta: Oh so any current implemented berries simply need to listen to `onEndOfTurn` (or something)?

Tempo-anon: I believe so, though there's also a few moves/abilities that might trigger a berry's eat effect

src/@types/item/Holdable.ts
```diff
@@ -0,0 +1,38 @@
+import type { Pokemon } from "#app/field/pokemon";
+
+/**
+ * An item that can be held by a {@linkcode Pokemon}.
+ */
+export interface Holdable {
```
innerthunder, line 6:
```suggestion
export interface Holdable {
  /** The {@linkcode Pokemon} holding the item */
  holder: Pokemon;
```

I feel like this would help simplify the event contexts.

flx-sta: That makes perfect sense :D

PR #910, opened 2025-04-05
[Refactor] Decouple phase system from battle-scene

torranx:
none

`BattleScene` is one hell of a singleton, this is another effort to break it down
`PhaseManager` currently does not, and should not know the scene instance by passing it directly to the constructor.
- prevents tight coupling between the phase system and scene
- dependencies should be passed through method parameters instead.

- moved all phase-related logic from `battle-scene.ts` to its own new file `src/phase-manager.ts`
- renamed some functions to keep consistency - `queue{x}Phase`
- added some @\todo

## How to test the changes?
playtest and/or run `npm run test`

## Checklist


DayKev (dismissed):
Seems good? It's basically just moving some code around without any real changes. The game still runs, all the tests pass.

src/phase-manager.ts
```diff
@@ -0,0 +1,503 @@
...
+   * Adds a phase to the conditional queue and ensures it is executed only when the specified condition is met.
+   *
+   * This method allows deferring the execution of a phase until certain conditions are met, which is useful for handling
+   * situations like abilities and entry hazards that depend on specific game states.
+   *
+   * @param phase The {@linkcode Phase} to be added to the conditional queue.
+   * @param condition A function that returns a boolean indicating whether the phase should be executed.
+   *
+   */
+  public pushConditionalPhase(phase: Phase, condition: () => boolean): void {
```
innerthunder, line 106: This is probably out of the PR's scope, but imo this (and the conditional queue in general) should be deprecated. The way the conditional phase queue is handled in `shiftPhase` is unintuitive, and if we want to translate the Phase sequence to state-machine-esque logic, this makes doing so much more difficult. I'm also pretty sure its only current usage in `encounter-phase.ts` can be removed by changing the order in which the Phases scheduled there are pushed/unshifted

src/phase-manager.ts
```diff
@@ -0,0 +1,503 @@
...
+    this.conditionalQueue.push([condition, phase]);
+  }
+
+  /**
+   * Adds a phase to nextCommandPhaseQueue, as long as boolean passed in is false
+   * @param phase {@linkcode Phase} the phase to add
+   * @param defer boolean on which queue to add to, defaults to false, and adds to phaseQueue
+   */
+  public pushPhase(phase: Phase, defer: boolean = false): void {
```
innerthunder, line 116: Another thing to consider down the line: a factory function for Phases based on `PhaseId`. If this and other scheduling functions had a signature like
```ts
public pushPhase<P extends Phase>(phase: PhaseId, ...params: ConstructorParameters<P>)
```
then we wouldn't have to worry about circular dependencies, and we wouldn't need the bespoke helper functions (e.g. `useMove`)

src/phase-manager.ts
```diff
@@ -0,0 +1,503 @@
...
+      }
+    }
+
+    if (this.currentPhase) {
+      console.log(`%cStart Phase ${this.currentPhase.constructor.name}`, "color:green;");
+      this.currentPhase.start();
+    }
+  }
+
+  public overridePhase(phase: Phase): boolean {
```
innerthunder, line 229: This should also probably be deprecated. Logic that requires this is probably better off as an `async` function

src/phase-manager.ts
```diff
@@ -0,0 +1,503 @@
...
+  source?: Pokemon;
+}
+
+/**
+ * This is responsible for managing the game's {@linkcode Phase | phases}.
+ */
+export class PhaseManager {
+  /** PhaseQueue: dequeue/remove the first element to get the next phase */
+  public phaseQueue: Phase[] = [];
+  /** phaseManager.phaseQueuePrepend: is a temp storage of what will be added to PhaseQueue */
```
PigeonBar:
```suggestion
  /** phaseManager.phaseQueuePrepend: is a temp storage of what will be added to the front of PhaseQueue */
```

src/phase-manager.ts
```diff
@@ -0,0 +1,503 @@
...
+  public phaseQueuePrepend: Phase[] = [];
```
PigeonBar: Also, these should probably be `private` instead of `public`.

src/phase-manager.ts
```diff
@@ -0,0 +1,503 @@
...
+      this.unshiftPhase(phase);
+      return false;
+    }
+  }
+
+  /**
+   * @todo this is unused, may be removed?
+   * Moves everything from nextCommandPhaseQueue to phaseQueue (keeping order)
+   */
+  public populatePhaseQueue(): void {
```
PigeonBar, line 340: The entire `nextCommandPhaseQueue` system seems to only be used for moves like Healing Wish, where it is used to delay a `PokemonHealPhase` until after a replacement Pokemon has been sent out. This should become unnecessary if Healing Wish gets reimplemented to use an arena tag (which is also needed to store the heal if the next Pokemon sent out is fully healthy).

src/phase-manager.ts
```diff
@@ -0,0 +1,503 @@
...
+   *
+   */
+  public pushConditionalPhase(phase: Phase, condition: () => boolean): void {
+    this.conditionalQueue.push([condition, phase]);
+  }
+
+  /**
+   * Adds a phase to nextCommandPhaseQueue, as long as boolean passed in is false
+   * @param phase {@linkcode Phase} the phase to add
+   * @param defer boolean on which queue to add to, defaults to false, and adds to phaseQueue
```
PigeonBar:
```suggestion
   * Queues a phase to be run at a future point in time.
   * @param phase - The {@linkcode Phase} to add
   * @param defer - If `false`, adds the phase to `phaseQueue`. If `true`, adds the phase to `nextCommandPhaseQueue`. Default `false`.
```

src/phase-manager.ts
PigeonBar: This is currently used to activate `FormChangePhase` when using form change items in the "Check Team" menu, so it would be hard to change it into an `async` function in a reasonable way.

src/phase-manager.ts
```diff
@@ -0,0 +1,503 @@
...
+  public populatePhaseQueue(): void {
+    if (this.nextCommandPhaseQueue.length) {
+      this.phaseQueue.push(...this.nextCommandPhaseQueue);
+      this.nextCommandPhaseQueue.splice(0, this.nextCommandPhaseQueue.length);
+    }
+    this.phaseQueue.push(new TurnInitPhase());
+  }
+
+  /**
+   * Adds a MessagePhase, either to PhaseQueuePrepend or nextCommandPhaseQueue
```
PigeonBar:
```suggestion
   * Adds a MessagePhase, either to PhaseQueuePrepend or phaseQueue
```

src/phase-manager.ts
```diff
@@ -0,0 +1,503 @@
...
+   * @param message string for MessagePhase
+   * @param callbackDelay optional param for MessagePhase constructor
+   * @param prompt optional param for MessagePhase constructor
+   * @param promptDelay optional param for MessagePhase constructor
+   * @param defer boolean for which queue to add it to, false -> add to PhaseQueuePrepend, true -> nextCommandPhaseQueue
```
PigeonBar:
```suggestion
   * @param defer boolean for which queue to add it to, false -> add to PhaseQueuePrepend, true -> phaseQueue
```

src/phase-manager.ts
```diff
@@ -0,0 +1,503 @@
...
+    prompt?: boolean | null,
+    promptDelay?: number | null,
+    defer?: boolean | null,
+  ) {
+    const phase = new MessagePhase(message, callbackDelay, prompt, promptDelay);
+    if (!defer) {
+      // adds to the end of PhaseQueuePrepend
+      this.unshiftPhase(phase);
+    } else {
+      //remember that pushPhase adds it to nextCommandPhaseQueue
```
PigeonBar:
```suggestion
      // adds to the end of phaseQueue
```

src/phase-manager.ts
PigeonBar: Also, this function is not unused, and its main purpose is to start a new `TurnInitPhase`.

test/battler-tags/substitute.test.ts
PigeonBar, line 1: (TODO: This file needs to be updated, since scene has been removed from Pokemon objects (#5).)

src/phase-manager.ts
DayKev: Modified the suggestion slightly (added the `- ` and changed the `phase` param doc wording).

DayKev (commented):
#915 
Also included Pigeon's suggestions (the ones that used GitHub's "suggestion" feature).

src/battle-scene.ts
```diff
@@ -262,18 +202,8 @@ export default class BattleScene extends SceneBase {
   public gameData: GameData;
   public sessionSlotId: number;
 
-  /** PhaseQueue: dequeue/remove the first element to get the next phase */
-  public phaseQueue: Phase[];
-  public conditionalQueue: Array<[() => boolean, Phase]>;
-  /** PhaseQueuePrepend: is a temp storage of what will be added to PhaseQueue */
-  private phaseQueuePrepend: Phase[];
+  public phaseManager: PhaseManager;
```
flx-sta: NIT: I assume this doesn't change after init anymore? Should be `readonly` then to prevent modification from outside

src/phase-manager.ts
```diff
@@ -0,0 +1,528 @@
...
+  /** A temporary storage of what will be added to the front of {@linkcode phaseQueue} */
+  private phaseQueuePrepend: Phase[] = [];
+
+  /** overrides default of inserting phases to end of phaseQueuePrepend array, useful for inserting Phases "out of order" */
+  public phaseQueuePrependSpliceIndex: number = -1;
+  public conditionalQueue: Array<[() => boolean, Phase]> = [];
+  public nextCommandPhaseQueue: Phase[] = [];
+
+  public currentPhase: Phase | null = null;
+  public standbyPhase: Phase | null = null;
```
flx-sta: Since they are `public`, unless they are supposed to be modifiable from "outside", they should probably be protected via `get/set` pattern

flx-sta (changes requested):

src/phase-manager.ts
torranx: ~~this is out of the PR's scope; should be in another PR~~
Edit: nvm since I isolated them in a class this is within scope

flx-sta (dismissed):
Guess being able to tag them as `private`  made it much easier instead of having to integrate the whole get/set pattern :D

src/phase-manager.ts
```diff
@@ -0,0 +1,527 @@
...
+
+  private currentPhase: Phase | null = null;
+  private standbyPhase: Phase | null = null;
+
+  public getCurrentPhase<P extends Phase = Phase>(): P | null {
+    return this.currentPhase as P;
+  }
+
+  public getStandbyPhase(): Phase | null {
+    return this.standbyPhase;
```
DayKev: To match `getCurrentPhase()`:
```suggestion
  public getStandbyPhase<P extends Phase = Phase>(): P | null {
    return this.standbyPhase as P;
```

DayKev (dismissed):

flx-sta (approved):

src/phase-manager.ts
DayKev: Something like this?
```ts
  public createPhase<P extends AbstractConstructor<Phase>>(phase: PhaseId, ...params: ConstructorParameters<P>) {
    switch (phase) {
      case PhaseId.ATTEMPT_CAPTURE:
        // not sure how to get the params out though; `params[0]`/etc didn't work either
        return new AttemptCapturePhase(...params);
      // ...
    }
  }
```
image
image

PR #911, opened 2025-04-06
[Docs] Update comments in version migration examples

DayKev:
N/A

Brought up in a [comment](https://github.com/Despair-Games/poketernity/pull/776/files#r2017733819) on the previous PR after it was merged.

Updated documentation.

## Checklist

### DIFF
--- a/src/system/version_migration/version_converter.ts
+++ b/src/system/version_migration/version_converter.ts
@@ -12,6 +12,7 @@ import { version } from "../../../package.json";
 // versions/vA_B_C.ts
 
 // The version for each migrator should match the filename, ie: `vA_B_C.ts` -> `version: "A.B.C"
+// This is the target version (aka the version we're ending up on after the migrators are run)
 
 // The name for each migrator should match its purpose. For example, if you're fixing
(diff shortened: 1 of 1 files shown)


Tempo-anon (approved):
This is not purely documentation right? Looks like there's also a renaming

DayKev:
I only changed the stuff that was in comments. If there were existing save migrators it would be a refactor.

PigeonBar:
Just noticed this as well:
https://github.com/Despair-Games/poketernity/blob/dc53bd895aa72bcf903b4069be4688e9310b10b1/src/system/version_migration/version_converter.ts#L124-L126
I'm pretty sure the safeguard is supposed to run every time, not just when upgrading versions.

DayKev:
> Just noticed this as well:
> 
> https://github.com/Despair-Games/poketernity/blob/dc53bd895aa72bcf903b4069be4688e9310b10b1/src/system/version_migration/version_converter.ts#L124-L126
> 
> 
> I'm pretty sure the safeguard is supposed to run every time, not just when upgrading versions.

It does: https://github.com/pagefaultgames/pokerogue/pull/4485/files

PigeonBar (approved):

PR #919, opened 2025-04-08
[Balance] Gym leaders now occur every 20 waves from 20-160

DayKev:
Gym leaders will appear every 20 waves in classic from waves 20-160, instead of every 30 waves from 20-170 or 30-180.

General consensus was to change it to 8 gym leaders per classic run.

- `BattleScene#offsetGym` and `BattleScene#getGeneratedOffsetGym()` were removed.
- `GameMode#isGymWave()` created that returns `true` if it's classic mode, the current wave is a multiple of 20 and the current wave is less than or equal to 160.
- Various wave checks were replaced with `.isGymWave()`.

## How to test the changes?
Go to a wave that's a multiple of 20 in classic mode.

## Checklist

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -254,8 +254,8 @@ export default class BattleScene extends SceneBase {
 
   public seed: string;
   public waveSeed: string;
+  /** @todo what is this? */
   public waveCycleOffset: number;
-  public offsetGym: boolean;
 
   public damageNumberHandler: DamageNumberHandler;
   private spriteSparkleHandler: PokemonSpriteSparkleHandler;
@@ -1039,7 +1039,6 @@ export default class BattleScene extends SceneBase {
     this.seed = seed;
     this.rngCounter = 0;
     this.waveCycleOffset = this.getGeneratedWaveCycleOffset();
-    this.offsetGym = this.gameMode.isClassic && this.getGeneratedOffsetGym();
   }
 
   /**
@@ -1555,18 +1554,7 @@ export default class BattleScene extends SceneBase {
     return this.arena.getSpeciesFormIndex(species);
   }
 
-  private getGeneratedOffsetGym(): boolean {
-    let ret = false;
-    this.executeWithSeedOffset(
-      () => {
-        ret = !randSeedInt(2);
-      },
-      0,
-      this.seed.toString(),
-    );
-    return ret;
-  }
-
+  /** @todo what is this? */
   private getGeneratedWaveCycleOffset(): number {
     let ret = 0;
     this.executeWithSeedOffset(
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -254,7 +254,7 @@ export default class BattleScene extends SceneBase {
 
   public seed: string;
   public waveSeed: string;
-  /** @todo what is this? */
+  /** Random multiple of `5` between `0-35` used to start at a random time of day */
   public waveCycleOffset: number;
 
   public damageNumberHandler: DamageNumberHandler;
@@ -1554,7 +1554,7 @@ export default class BattleScene extends SceneBase {
     return this.arena.getSpeciesFormIndex(species);
   }
 
-  /** @todo what is this? */
+  /** @returns A random multiple of 5 between 0-35, used to start at a random time of day */
   private getGeneratedWaveCycleOffset(): number {
     let ret = 0;
     this.executeWithSeedOffset(
--- a/src/game-mode.ts
+++ b/src/game-mode.ts
@@ -194,7 +194,7 @@ export class GameMode implements GameModeConfig {
     return this.isClassic && gymWaves.includes(waveIndex);
   }
 
-  isTrainerBoss(waveIndex: number, biomeId: BiomeId): boolean {
+  public isTrainerBoss(waveIndex: number, biomeId: BiomeId): boolean {
     switch (this.modeId) {
       case GameModes.DAILY:
         return waveIndex > 10 && waveIndex < 50 && !(waveIndex % 10);
--- a/src/game-mode.ts
+++ b/src/game-mode.ts
@@ -142,19 +142,18 @@ export class GameMode implements GameModeConfig {
    * @returns `true` if a trainer should be generated, `false` otherwise
    */
   isWaveTrainer(waveIndex: number, arena: Arena): boolean {
-    /**
-     * Daily spawns trainers on floors 5, 15, 20, 25, 30, 35, 40, and 45
-     */
+    // Daily spawns trainers on floors 5, 15, 20, 25, 30, 35, 40, and 45
     if (this.isDaily) {
       return waveIndex % 10 === 5 || (!(waveIndex % 10) && waveIndex > 10 && !this.isWaveFinal(waveIndex));
     }
-    if (waveIndex % 30 === (globalScene.offsetGym ? 0 : 20) && !this.isWaveFinal(waveIndex)) {
+
+    if (this.isGymWave(waveIndex) && !this.isWaveFinal(waveIndex)) {
       return true;
-    } else if (waveIndex % 10 !== 1 && waveIndex % 10) {
-      /**
-       * Do not check X1 floors since there's a bug that stops trainer sprites from appearing
-       * after a X0 full party heal
-       */
+    }
+
+    // Trainers must not appear on X1 waves due to a bug that prevents trainer sprites from appearing
+    // after the full party heal that happens between X0 and X1 waves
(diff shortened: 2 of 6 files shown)


src/battle-scene.ts
```diff
@@ -254,8 +254,8 @@ export default class BattleScene extends SceneBase {
 
   public seed: string;
   public waveSeed: string;
+  /** @todo what is this? */
   public waveCycleOffset: number;
```
Tempo-anon, line 258: This is a random number that's between 0-35 inclusive and is a multiple of 5. This is used for starting the game at a random time of day

src/game-mode.ts
```diff
@@ -182,18 +184,24 @@ export class GameMode implements GameModeConfig {
       }
       return Boolean(allowTrainerBattle && trainerChance && !randSeedInt(trainerChance));
     }
+
     return false;
   }
 
-  isTrainerBoss(waveIndex: number, biomeId: BiomeId, offsetGym: boolean): boolean {
+  /** @returns `true` if the wave is a multiple of `20` between `20-160` in classic mode */
```
Tempo-anon, line 191: The main reason there's a random offset is that players don't meet the same gym leaders every run. Will have to look into biome connections to ensure that there's enough to not have even parity

src/battle-scene.ts
DayKev: Docs added. 👍

src/game-mode.ts
DayKev: Biomes should be more loosely linked together once we implement the new Biome layout.

src/phases/victory-phase.ts
```diff
@@ -77,15 +77,14 @@ export class VictoryPhase extends PokemonPhase {
             globalScene.phaseManager.pushPhase(new ModifierRewardPhase(modifierTypes.GOLDEN_POKEBALL));
           }
         } else {
-          const superExpWave = !isEndless ? (offsetGym ? 0 : 20) : 10;
           if (isEndless && waveIndex === 10) {
             globalScene.phaseManager.pushPhase(new ModifierRewardPhase(modifierTypes.EXP_SHARE));
           }
 
-          if (waveIndex <= 750 && (waveIndex <= 500 || waveIndex % 30 === superExpWave)) {
+          if (waveIndex <= 750 && (waveIndex <= 500 || gameMode.isGymWave(waveIndex))) {
```
MokaStitcher: The superExpWave had a specific value for endless so the check was true every 30 waves for endless even past 500, now it'll always be false so I think endless loses on super exp charms altogether, and on all exp charms past floor 500
Probably doesn't matter much since we don't want to keep endless in this form, but I don't think functionality is being maintained here
If we keep it like this though we could remove the isGymWave check here since the waveIndex <= 500 will always be true when isGymWave would be lol

src/game-mode.ts
```diff
@@ -182,18 +184,24 @@ export class GameMode implements GameModeConfig {
...
+  public isGymWave(waveIndex: number): boolean {
+    return this.isClassic && waveIndex % 20 === 0 && waveIndex <= 160;
```
Tempo-anon: If there are only 8 intended values I'm also wondering if it would just be better if this was a hardcoded array.

```ts
const gymWaves = [20, 40, 60, 80, 100, 120, 140, 160]
return this.isClassic && gymWaves.includes(waveIndex);
```

Tempo-anon (commented):
Another thing to consider in a follow up is to look at moving the evil bosses and/or rivals onto the X0 floors that aren't multiplies of 20 like 30, 50, 70, 90, 110, 130, 150, 170. If we were to do that though, I think it would also make sense to move the guaranteed boss Pokemon encounter to X9 floors. That also comes with the benefit of guaranteeing more boss spawns during a playthrough

src/phases/victory-phase.ts
DayKev: Oh true, this was kind of a mess of checks...

DayKev:
> Another thing to consider in a follow up is to look at moving the evil bosses and/or rivals onto the X0 floors that aren't multiplies of 20 like 30, 50, 70, 90, 110, 130, 150, 170. If we were to do that though, I think it would also make sense to move the guaranteed boss Pokemon encounter to X9 floors. That also comes with the benefit of guaranteeing more boss spawns during a playthrough

Yeah, I was going to suggest moving around some of the other fixed battles after this PR.

src/game-mode.ts
```diff
@@ -182,18 +184,25 @@ export class GameMode implements GameModeConfig {
...
       return Boolean(allowTrainerBattle && trainerChance && !randSeedInt(trainerChance));
     }
+
     return false;
   }
 
-  isTrainerBoss(waveIndex: number, biomeId: BiomeId, offsetGym: boolean): boolean {
+  /** @returns `true` if the wave is a multiple of `20` between `20-160` in classic mode */
+  public isGymWave(waveIndex: number): boolean {
+    const gymWaves = [20, 40, 60, 80, 100, 120, 140, 160];
```
Tempo-anon, line 193: Just another random thing that comes to mind, but I wonder if it's a bit underwhelming for floor 100 to be just another gym. I know right now everything is evenly spaced by 20 but `gymWaves` could also be something like [20, 40, 60, 80, 110, 130, 150, 170] with something more climactic or story related happening in the second half. Just food for thought

src/phases/victory-phase.ts
```diff
@@ -52,63 +52,64 @@ export class VictoryPhase extends PokemonPhase {
...
+            phaseManager.pushPhase(new ModifierRewardPhase(modifierTypes.EXP_CHARM));
+          }
+
+          if (isEndless && waveIndex <= 750 && (waveIndex <= 500 || waveIndex % 30 === 10)) {
+            phaseManager.pushPhase(
               new ModifierRewardPhase(
-                waveIndex % 30 !== superExpWave || waveIndex > 250
-                  ? modifierTypes.EXP_CHARM
-                  : modifierTypes.SUPER_EXP_CHARM,
+                !(waveIndex % 30 === 10) || waveIndex > 250 ? modifierTypes.EXP_CHARM : modifierTypes.SUPER_EXP_CHARM,
```
Tempo-anon, line 94: Not necessarily for this PR but wow victory-phase is not well documented for how it rewards special items on special floors

Tempo-anon (dismissed):
Should be fine for this PR for now but I assume there will be some follow up stuff

src/phases/victory-phase.ts
DayKev: Yeah... I've got a refactor for this file ready to go after this PR is merged that hopefully makes it easier to parse what it's doing (and make changes if necessary).

DayKev:
Ugh I don't know why but the git history is messed up and it's seeing an extra commit. Well, it probably has something to do with GitHub being broken earlier today.
Whatever, just ignore the extra commit I guess.

Edit: managed to fix the git history issue.

src/phases/victory-phase.ts
```diff
@@ -52,63 +52,64 @@ export class VictoryPhase extends PokemonPhase {
...
-          const superExpWave = !isEndless ? (offsetGym ? 0 : 20) : 10;
           if (isEndless && waveIndex === 10) {
-            globalScene.phaseManager.pushPhase(new ModifierRewardPhase(modifierTypes.EXP_SHARE));
+            phaseManager.pushPhase(new ModifierRewardPhase(modifierTypes.EXP_SHARE));
           }
 
-          if (waveIndex <= 750 && (waveIndex <= 500 || waveIndex % 30 === superExpWave)) {
-            globalScene.phaseManager.pushPhase(
+          if (!isEndless && gameMode.isGymWave(waveIndex)) {
+            phaseManager.pushPhase(new ModifierRewardPhase(modifierTypes.EXP_CHARM));
```
PigeonBar: The previous formula was to give a Super Exp. Charm after gym battles, and an Exp. Charm after wild boss battles, so this new formula will give out much fewer Exp. Charms in classic mode. Is this intended?

DayKev: Oh, I think I see what you mean. Lemme fix that.

Tempo-anon: I also wouldn't worry too much about exp charms and when they're given out too much at the moment, though it does make sense that a gym would give better rewards than a non gym. The whole exp charms and enemy level curve system needs to be reexamined anyway

DayKev: Oh yeah that's also going to mean more S Exp Charms are going to be given out because there's 2 more gyms... yeah I guess it doesn't really matter.

flx-sta (approved):

PR #913, opened 2025-04-06
[Doc] Add `BattleAnim` comments and minor cleanup

innerthunder:
## What are the changes the user will see?

N/A

## Why am I making these changes?

Prepping for a larger refactor. I made these comments while researching.

## What are the changes from a developer perspective?

Added comments to the following:
- `BattleAnim` and its subclasses
- `BattleAnim` methods: `getGraphicFrameData`, `play`, `playWithoutTargets`
- `AnimConfig` fields
- Related enums: `AnimFocus`, `AnimFrameTarget`
- Added `@todo`s for issues caught during documentation

Some changes have also been made to simplify `BattleAnim` and its subclasses. The most notable change is that `CommonAnimPhase`'s constructor parameters have been reorganized.

## How to test the changes?

The test suite mocks out animations, so... just play and hope nothing catches fire 😅 

## Checklist

- [ ] ⚠️ If this is a PR for `main` (such as a hotfix), has the game version been updated (`npm run update-version:patch` / `npm run update version:minor`?
- [x] Otherwise: **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [?] Have I tested the changes manually?

- [x] Are all unit tests still passing? (`npm run test:silent`)
  - [n/a] Have I created new automated tests (`npm run test:create`) or updated existing tests related to the PR's changes?
- [n/a] Have I provided screenshots/videos of the changes (if applicable)?
  - [n/a] Have I made sure that any UI change works for both UI themes (dark and light)?

PR #920, opened 2025-04-08
[Refactor] Create `AnimConfig` JSON schema w/ new JSON format

### LINKED ISSUE #948: [EPIC] BattleAnim Rework
Rework/Refactor BattleAnim

## Tasks

- [x] Create schema for new animation interface
  - [x] #920
  - [x] #973 
  - [x] Set up validation tests for anims against schema
- [ ] Rewrite `anim-config.ts` adhering to schema
- [ ] Rewrite animation loading functions
- [ ] Rewrite `BattleAnim` methods
  - [ ] `getGraphicFrameData`
  - [ ] `play` (using [TweenChain](https://docs.phaser.io/api-documentation/class/tweens-tweenchain))
  - [ ] Remove `XWithoutTargets` methods
- [ ] Consolidate phases related to animations
  - [ ] Optional: Update `PokemonAnimPhase` to make use of battle anim format
- [ ] Rewrite battle animations in `public/battle-anims` under new schema

innerthunder:

N/A

This is me laying out my plans for reformatting and refactoring Battle Anims 😄 
- The new JSON format should require less data and be easier to work with
- Before this, Battle Anim JSONs were never properly validated before being plugged into animations

- `anim-config-schema`: Created `animConfigSchema` with sub-schemas `animPropSchema`, `keyFrameSchema`, and `animTimedEventSchema`. These are meant to enforce a new JSON format with curve-based motion (instead of frame-based)
  - This requires [ajv](https://www.npmjs.com/package/ajv) as a new dependency, which:
  - > "Supports JSON Schema draft-04/06/07/2019-09/2020-12 ([draft-04 support](https://ajv.js.org/json-schema.html#draft-04) requires ajv-draft-04 package) and JSON Type Definition [RFC8927](https://datatracker.ietf.org/doc/rfc8927/)."
- `anim-frame-target`: `AnimFrameTarget` is now a string enum, and some enum keys have changed:
  - `USER` -> `SOURCE`
  - `GRAPHIC` -> `IMAGE`
  - `TARGET` is unchanged
- `validate-battle-anims.test`: New unit test to validate all JSONs in `public/battle-anims` against the new schema. This is currently disabled, and can be enabled once battle anims are replaced.

## Screenshots/Videos

After `npm run test validate-battle-anims` with the new test enabled, the test fails when validating `absorb.json`:

image

The error message is due to the JSON being an array of two (duplicate?) configs instead of the expected single object:

image

## Checklist

- [?] Have I tested the changes manually?

    - Yes, but deactivated for now... 

### DIFF
--- a/src/data/animations/anim-config-schema.ts
+++ b/src/data/animations/anim-config-schema.ts
@@ -1,7 +1,6 @@
 // -- start tsdoc imports --
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import type { AnimConfig } from "#app/data/animations/anim-config";
-import type { AnimTimedSoundEvent, AnimTimedAddBgEvent } from "#app/data/animations/anim-config";
+import type { AnimConfig, AnimTimedSoundEvent, AnimTimedAddBgEvent } from "#app/data/animations/anim-config";
 import { easeFunctions } from "#app/data/animations/ease-functions";
 import type { MoveAnim } from "#app/data/animations/move-anim";
 /* eslint-enable @typescript-eslint/no-unused-vars */
--- a/src/data/animations/anim-config-schema.ts
+++ b/src/data/animations/anim-config-schema.ts
@@ -9,7 +9,7 @@ import type { MoveAnim } from "#app/data/animations/move-anim";
 import { AnimBlendType } from "#enums/anim-blend-type";
 import { AnimFrameTargets } from "#enums/anim-frame-target";
 import { MoveId } from "#enums/move-id";
-import { type Schema } from "jsonschema";
+import type { Schema } from "ajv";
 
 /**
  * Schema for a single keyframe of an animated object.
@@ -128,8 +128,8 @@ const keyFrameSchema: Schema = {
         minimum: 0,
         maximum: 255,
       },
-      minLength: 3,
-      maxLength: 4,
+      minItems: 3,
+      maxItems: 4,
     },
 
     /**
@@ -287,7 +287,7 @@ export const animConfigSchema: Schema = {
     props: {
       type: "array",
       items: animPropSchema,
-      minLength: 1,
+      minItems: 1,
     },
 
     /**
@@ -297,7 +297,7 @@ export const animConfigSchema: Schema = {
     timedEvents: {
       type: "array",
       items: animTimedEventSchema,
-      minLength: 1,
+      minItems: 1,
     },
   },
   required: ["props"],
--- a/src/data/animations/anim-config-schema.ts
+++ b/src/data/animations/anim-config-schema.ts
@@ -2,6 +2,7 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
 import type { AnimConfig } from "#app/data/animations/anim-config";
 import type { AnimTimedSoundEvent, AnimTimedAddBgEvent } from "#app/data/animations/anim-config";
+import { easeFunctions } from "#app/data/animations/ease-functions";
 import type { MoveAnim } from "#app/data/animations/move-anim";
 /* eslint-enable @typescript-eslint/no-unused-vars */
 // -- end tsdoc imports --
@@ -47,7 +48,7 @@ const keyFrameSchema: Schema = {
       type: ["number", "object"],
       properties: {
         value: { type: "number" },
-        ease: { type: "string" }, // TODO: can this be more specific to Phaser's ease functions?
+        ease: { enum: easeFunctions },
       },
     },
 
@@ -66,7 +67,7 @@ const keyFrameSchema: Schema = {
       type: ["number", "object"],
       properties: {
         value: { type: "number" },
-        ease: { type: "string" }, // TODO: can this be more specific to Phaser's ease functions?
+        ease: { enum: easeFunctions },
       },
     },
 
@@ -154,7 +155,7 @@ const keyFrameSchema: Schema = {
      * @see {@link https://rexrainbow.github.io/phaser3-rex-notes/docs/site/ease-function/#get-ease-function-via-string | Ease Functions}
      */
     ease: {
-      type: "string",
+      enum: easeFunctions,
       default: "Linear",
     },
   },
--- a/src/data/animations/anim-config-schema.ts
+++ b/src/data/animations/anim-config-schema.ts
@@ -7,7 +7,7 @@ import type { MoveAnim } from "#app/data/animations/move-anim";
 // -- end tsdoc imports --
 
 import { AnimBlendType } from "#enums/anim-blend-type";
-import { AnimFrameTarget } from "#enums/anim-frame-target";
+import { AnimFrameTargets } from "#enums/anim-frame-target";
 import { MoveId } from "#enums/move-id";
 import { type Schema } from "jsonschema";
 
@@ -162,7 +162,7 @@ const keyFrameSchema: Schema = {
 
 /**
  * Schema for the animation properties of an object.
- * This specifies the property's {@linkcode AnimFrameTarget | focus}
+ * This specifies the property's {@linkcode AnimFrameTargets | focus}
  * and contains the {@linkcode keyFrameSchema | keyframes} applied
  * to the focus.
  */
@@ -171,9 +171,9 @@ const animPropSchema: Schema = {
   properties: {
     /**
      * The type of sprite affected by the animation
-     * @see {@linkcode AnimFrameTarget}
+     * @see {@linkcode AnimFrameTargets}
      */
-    focus: { enum: Object.values(AnimFrameTarget) },
+    focus: { enum: Object.values(AnimFrameTargets) },
 
     /**
      * The keyframes used to animate the sprite.
--- a/src/data/animations/anim-config-schema.ts
+++ b/src/data/animations/anim-config-schema.ts
@@ -1,0 +1,186 @@
+// -- start tsdoc imports --
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import type { AnimConfig } from "#app/data/animations/anim-config";
+// -- end tsdoc imports --
+
+import { AnimBlendType } from "#enums/anim-blend-type";
+import { AnimFrameTarget } from "#enums/anim-frame-target";
+import { MoveId } from "#enums/move-id";
+import { type Schema } from "jsonschema";
+
+const keyFrameSchema: Schema = {
+  type: "object",
+  properties: {
+    /**
+     * For battle animations, the origin point for keyframes is defined
+     * along the line connecting the start point ("source") and end point ("target").
+     * The `u`-value is the fraction of the distance between the start and end point
+     * the origin point is away from the source. If `u = 0`, then the origin point is
+     * the source; if `u = 1`, then the origin point is the target.
+     */
+    u: {
+      type: "number",
+      minimum: 0,
+      maximum: 1,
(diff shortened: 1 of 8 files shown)


src/data/animations/anim-config-schema.ts
```diff
@@ -0,0 +1,304 @@
...
+ * to the focus.
+ */
+const animPropSchema: Schema = {
+  type: "object",
+  properties: {
+    /**
+     * The type of sprite affected by the animation
+     * @see {@linkcode AnimFrameTarget}
+     */
+    focus: { enum: Object.values(AnimFrameTarget) },
```
innerthunder: Just noting here that this is equivalent to the current `"target"` field. The old `"focus"` field and `AnimFocus` are to be deprecated.

src/enums/anim-frame-target.ts
DayKev, line 1: Based on what Benjie's said about enums vs const objects ("accesses are faster and the size of the data is less than half") we should probably not use enums for anything that doesn't need the reverse mapping (e.g.: `EnumName[EnumValue]`, which I checked and isn't used by this). So this might be a good time to convert this one (I think it's supposed to be like this?):
```ts
/** Specifies the type of sprite affected by an animation */
export const AnimFrameTargets = {
  /**
   * Affects the animation's source or start point, e.g.
   * a Pokemon using a move.
   */
  SOURCE: "source",
  /**
   * Affects the animation's end point, e.g. a Pokemon
   * hit by a move.
   */
  TARGET: "target",
  /** Affects a component of the animation's visual effects */
  IMAGE: "image",
} as const;

/** The sprite types affected by an animation */
export type AnimFrameTarget = typeof AnimFrameTargets[keyof typeof AnimFrameTargets];
```

test/animations/validate-battle-anims.test.ts
```diff
@@ -0,0 +1,19 @@
+import { animConfigSchema } from "#app/data/animations/anim-config-schema";
+import { readdirSync, readFileSync } from "fs";
```
DayKev, line 3: They really called it `readdirSync` and not `readDirSync` image

src/data/animations/anim-config-schema.ts
```diff
@@ -0,0 +1,304 @@
...
+     *  value: 100,
+     *  ease: "Sine.easeIn"
+     * }
+     * ```
+     */
+    x: {
+      type: ["number", "object"],
+      properties: {
+        value: { type: "number" },
+        ease: { type: "string" }, // TODO: can this be more specific to Phaser's ease functions?
```
DayKev: Based on the [enum](https://json-schema.org/understanding-json-schema/reference/enum) reference, you can probably make an array containing strings of all the Phaser ease functions (`const easeFuncs = ["Sine.easeIn", "Sine.easeOut", ...] as const;`) and then do `enum: easeFuncs` here instead of `type: "string"`.

src/enums/anim-frame-target.ts
DayKev: [0002-Convert-enum-to-const-object.patch](https://github.com/user-attachments/files/19663347/0002-Convert-enum-to-const-object.patch)

src/data/animations/anim-config-schema.ts
innerthunder: `as const` seems to cause issues, but otherwise this works. I guess the `enum` array is expected to be mutable...

DayKev: Huh, weird. Oh well.

src/data/animations/anim-config-schema.ts
```diff
@@ -0,0 +1,305 @@
+// -- start tsdoc imports --
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import type { AnimConfig } from "#app/data/animations/anim-config";
+import type { AnimTimedSoundEvent, AnimTimedAddBgEvent } from "#app/data/animations/anim-config";
```
DayKev:
```suggestion
import type { AnimConfig, AnimTimedSoundEvent, AnimTimedAddBgEvent } from "#app/data/animations/anim-config";
```

DayKev (dismissed):

src/enums/anim-frame-target.ts
```diff
@@ -1,15 +1,18 @@
 /** Specifies the type of sprite affected by an animation */
-export enum AnimFrameTarget {
+export const AnimFrameTargets = {
```
Tempo-anon, line 2: Should this still be in the enum folder if it's not an enum?

src/data/animations/battle-anims.ts
```diff
@@ -499,11 +499,11 @@ export abstract class BattleAnim {
     frames: AnimFrame[],
     targetInitialX: number,
     targetInitialY: number,
-  ): Map<number, Map<AnimFrameTarget, GraphicFrameData>> {
-    const ret: Map<number, Map<AnimFrameTarget, GraphicFrameData>> = new Map([
-      [AnimFrameTarget.GRAPHIC, new Map<AnimFrameTarget, GraphicFrameData>()],
-      [AnimFrameTarget.USER, new Map<AnimFrameTarget, GraphicFrameData>()],
-      [AnimFrameTarget.TARGET, new Map<AnimFrameTarget, GraphicFrameData>()],
+  ): Map<AnimFrameTarget, Map<number, GraphicFrameData>> {
```
Tempo-anon, line 502: Not sure if this was in the PR desc, but can you explain why this was changed from a `Map<number, Map,AnimFrameTarget, GraphicFrameData>>` to a `Map<AnimFrameTarget, Map<number, GraphicFrameData>>` and what that means?

flx-sta (commented):
I'm surprised to see JSON schema again. I've worked with it quite a lot in the past. For validation we always used [ajv](https://github.com/ajv-validator/ajv).

It's probably worth mentioning which version of the of JSON-schema is being used here

src/enums/anim-frame-target.ts
flx-sta: Why the change to `const` from `enum` to begin with?

SirzBenjie: Imo if you're using something as an enum, regardless of the _mechanism_ for it, it's still an "enum". It's just implemented differently.

src/data/animations/battle-anims.ts
innerthunder: This was changed for `getGraphicFrameData` in #913; I just missed it until now. The types for the map keys were mixed up, and that actually causes issues now that `AnimFrameTarget` is a string.

DayKev (dismissed):

DayKev (approved):

flx-sta (approved):

src/data/animations/anim-config-schema.ts
```diff
@@ -0,0 +1,304 @@
...
+     * @see {@link https://docs.phaser.io/api-documentation/constant/blendmodes}
+     */
+    blendType: { enum: Object.values(AnimBlendType) },
+
+    /**
+     * If this keyframe is for a graphic, specifies the tile index used
+     * for the graphic during the tween
+     * @todo Should this be decoupled from keyframes?
+     */
+    graphicFrame: { type: "number" },
```
pom-eranian, line 114: Changing the target type from GRAPHIC to IMAGE means other params like graphicFrame are less clear.

src/data/animations/anim-config-schema.ts
```diff
@@ -0,0 +1,304 @@
...
+    /**
+     * The {@linkcode MoveId | identifier} for the move
+     * associated with the anim (if applicable).
+     * This is required for all {@linkcode MoveAnim | MoveAnims}.
+     */
+    id: { enum: Object.values(MoveId) },
+
+    /**
+     * The name of the tileset used for the animation.
+     * @todo Should this be required?
```
pom-eranian, line 279: Should not be required. It is possible for a move anim to not have any.

src/data/animations/anim-config-schema.ts
```diff
@@ -0,0 +1,304 @@
...
+    /**
+     * The pitch of the played sound effect
+     * @default 100
+     */
+    pitch: {
+      type: "number",
+      default: 100,
+    },
+
+    // Background Image Event fields
```
pom-eranian, line 236: Please add a scale property
`x = y = value`

src/data/animations/anim-config-schema.ts
```diff
@@ -0,0 +1,304 @@
...
+     * for the affected sprite.
+     * @see {@linkcode keyFrameSchema}
+     */
+    keyFrames: {
+      type: "array",
+      items: keyFrameSchema,
+      minItems: 1,
+    },
+  },
+  required: ["focus", "keyframes"],
```
pom-eranian, line 190: I think that `focus` is used too loosely to refer to different things. Position coordinates, and the sprite object at that location for example. Because multiple individual sprites could have the exact same focus, but be a different category.

Another term that suffers from this is `target`.

src/data/animations/anim-config-schema.ts
```diff
@@ -0,0 +1,304 @@
...
+     * - 0 is behind all other sprites (except BG)
+     * - 1 is on top of player field
+     * - 3 is on top of both fields
+     * - 5 is on top of player sprite
+     */
+    priority: {
+      enum: [0, 1, 3, 5],
+    },
+
+    /** The duration of the tween for this keyframe (ms) */
```
pom-eranian, line 146: `The duration of the tween for this keyframe (ms)`
This one could be a bit more specific. Are the values defined on this frame considered the start or end values?

```
frame 0  {x: 5},
frame 3  {x: 10, duration: 4},
frame 8  {x: 15}
```
on frame 3, what is the value of x? Will it be 5 or 10?
and once the duration is over will the value be 10 or 15?

innerthunder: The defined values are the end values. `duration` defines the time it takes from the end of the previous keyframe to reach the end values. Frames aren't really relevant in this model. The example could be defined instead like
```
1 frame = 60 ms
in:  kf:
[0]  {x: 5}
[1]  {x: 10, duration: 180} // 3 * 60 ms
[2]  {x: 15, duration: 300} // 5 * 60 ms
```

src/data/animations/anim-config-schema.ts
innerthunder: The idea was for what's currently `target` to be renamed to `focus` to avoid confusion with the `TARGET` enum. What's currently `focus` will be removed completely since it can be derived completely in code.

src/data/animations/battle-anims.ts
```diff
@@ -156,15 +156,15 @@ export abstract class BattleAnim {
...
             && isReversed(this.srcLine[0], this.srcLine[2], this.dstLine[0], this.dstLine[2])
           ) {
             zoomX = zoomX * -1;
           }
           break;
       }
       const angle = -frame.angle;
-      const key = frame.target === AnimFrameTarget.GRAPHIC ? g++ : frame.target === AnimFrameTarget.USER ? u++ : t++;
+      const key = frame.target === AnimFrameTargets.IMAGE ? g++ : frame.target === AnimFrameTargets.SOURCE ? u++ : t++;
       ret.get(frame.target)!.set(key, { x, y, zoomX, zoomY, angle }); // TODO: is the bang correct?
```
DayKev, line 168: For some reason `frame.target` is a number here and not a string matching the `AnimFrameTargets` const.

PR #922, opened 2025-04-09
[Refactor] Remove a bunch of `!!`

DayKev:
N/A

General code cleanup.

- Replaced `!!*.find()` with `*.some()`
- Replaced `!!*.filter().length` with `*.some()`
- Added some default values to function parameters which allowed the removal of some `!!` and `??`
- Replaced `!!number` with `number > 0`
- Replaced `!!randSeedInt(2)` with `!randSeedInt(2)`
- Removed some leftover `!!` that were already unnecessary
- Condensed a couple of `if` conditions
- Fixed the return type hint of `Move#findAttr`

## Checklist

### DIFF
--- a/src/data/arena-tag.ts
+++ b/src/data/arena-tag.ts
@@ -69,7 +69,7 @@ export abstract class ArenaTag {
   public onOverlap(_arena: Arena): void {}
 
   public lapse(_arena: Arena): boolean {
-    return this.turnCount < 1 || !!--this.turnCount;
+    return this.turnCount < 1 || --this.turnCount !== 0;
   }
 
   public getMoveName(): string | null {
--- a/src/data/moves/move-attrs/heal-attr.ts
+++ b/src/data/moves/move-attrs/heal-attr.ts
@@ -19,11 +19,11 @@ export class HealAttr extends MoveEffectAttr {
   /** Should an animation be shown? */
   private showAnim: boolean;
 
-  constructor(healRatio?: number, showAnim?: boolean, selfTarget?: boolean) {
-    super(selfTarget === undefined || selfTarget);
+  constructor(healRatio: number = 1, showAnim: boolean = false, selfTarget: boolean = true) {
+    super(selfTarget);
 
-    this.healRatio = healRatio || 1;
-    this.showAnim = !!showAnim;
+    this.healRatio = healRatio;
+    this.showAnim = showAnim;
   }
 
   override applyEffect(user: Pokemon, target: Pokemon, move: Move): boolean {
--- a/src/data/mystery-encounters/encounters/mysterious-challengers-encounter.ts
+++ b/src/data/mystery-encounters/encounters/mysterious-challengers-encounter.ts
@@ -43,7 +43,7 @@ export const MysteriousChallengersEncounter: MysteryEncounter = MysteryEncounter
     const normalConfig = allTrainerConfigs[normalTrainerType].clone();
     let female = false;
     if (normalConfig.hasGenders) {
-      female = !!randSeedInt(2);
+      female = !randSeedInt(2);
     }
     const normalSpriteKey = normalConfig.getSpriteKey(female, normalConfig.doubleOnly);
     encounter.enemyPartyConfigs.push({
@@ -73,7 +73,7 @@ export const MysteriousChallengersEncounter: MysteryEncounter = MysteryEncounter
     hardConfig.setPartyTemplates(hardTemplate);
     female = false;
     if (hardConfig.hasGenders) {
-      female = !!randSeedInt(2);
+      female = !randSeedInt(2);
     }
     const hardSpriteKey = hardConfig.getSpriteKey(female, hardConfig.doubleOnly);
     encounter.enemyPartyConfigs.push({
@@ -93,7 +93,7 @@ export const MysteriousChallengersEncounter: MysteryEncounter = MysteryEncounter
     brutalConfig.partyTemplateFunc = null; // Overrides gym leader party template func
     female = false;
     if (brutalConfig.hasGenders) {
-      female = !!randSeedInt(2);
+      female = !randSeedInt(2);
     }
     const brutalSpriteKey = brutalConfig.getSpriteKey(female, brutalConfig.doubleOnly);
     encounter.enemyPartyConfigs.push({
--- a/src/phases/evolution-phase.ts
+++ b/src/phases/evolution-phase.ts
@@ -34,6 +34,7 @@ export class EvolutionPhase extends FormChangeBasePhase {
 
   private preEvolvedPokemonName: string;
 
+  /** @todo why is this able to be `null`? */
   private readonly evolution: SpeciesFormEvolution | null;
   private evolutionBgm: AnySound;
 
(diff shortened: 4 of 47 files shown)


src/data/arena-tag.ts
```diff
@@ -69,7 +69,7 @@ export abstract class ArenaTag {
   public onOverlap(_arena: Arena): void {}
 
   public lapse(_arena: Arena): boolean {
-    return this.turnCount < 1 || !!--this.turnCount;
+    return this.turnCount < 1 || --this.turnCount !== 0;
```
Tempo-anon, line 72: nit: Is it more clear if it were `--this.turnCount > 0` ?

src/data/moves/move-attrs/heal-attr.ts
```diff
@@ -19,11 +19,11 @@ export class HealAttr extends MoveEffectAttr {
   /** Should an animation be shown? */
   private showAnim: boolean;
 
-  constructor(healRatio?: number, showAnim?: boolean, selfTarget?: boolean) {
-    super(selfTarget === undefined || selfTarget);
+  constructor(healRatio: number = 1, showAnim: boolean = false, selfTarget: boolean = true) {
```
Tempo-anon, line 22: nit: I think healRatio with a default value of 0.5 makes more sense since that's more common

src/data/mystery-encounters/encounters/mysterious-challengers-encounter.ts
```diff
@@ -43,7 +43,7 @@ export const MysteriousChallengersEncounter: MysteryEncounter = MysteryEncounter
     const normalConfig = allTrainerConfigs[normalTrainerType].clone();
     let female = false;
     if (normalConfig.hasGenders) {
-      female = !!randSeedInt(2);
+      female = !randSeedInt(2);
```
Tempo-anon, line 46: nit: `!randSeedInt(2)` is equivalent to `randSeedInt(2)`

src/phases/form-change-phase.ts
```diff
@@ -48,7 +48,7 @@ export class FormChangePhase extends FormChangeBasePhase {
   }
 
   public override validate(): boolean {
-    return !!this.formChange;
+    return true;
```
Tempo-anon, line 51: form change phase validate is always true?

src/system/pokemon-data.ts
```diff
@@ -56,6 +56,7 @@ export default class PokemonData {
   /** Data that can customize a Pokemon in non-standard ways from its Species */
   public customPokemonData: CustomPokemonData;
 
+  // TODO: this shouldn't be `| any`
```
Tempo-anon, line 59: nit: this PR uses both TODO as well as @todo. We should probably decide on which one to use for consistency

src/system/pokemon-data.ts
```diff
@@ -56,6 +56,7 @@ export default class PokemonData {
...
   constructor(source: Pokemon | any, forHistory: boolean = false) {
     const sourcePokemon = source.type === "Pokemon" ? source : null;
```
Tempo-anon, line 61: Is there really no better way in TS to check something's type?

Tempo-anon (dismissed):
Refactors look okay and tests still pass. Just some nits

src/phases/form-change-phase.ts
DayKev: That seems like the current behavior. `this.formChange` should always evaluate to `true` from what I can see.

src/phases/evolution-phase.ts
```diff
@@ -34,6 +34,7 @@ export class EvolutionPhase extends FormChangeBasePhase {
 
   private preEvolvedPokemonName: string;
 
+  /** @todo why is this able to be `null`? */
```
PigeonBar, line 37: Most likely a relic from when `FormChangePhase` was a subclass of `EvolutionPhase`

src/system/pokemon-data.ts
DayKev: That used to be `instanceof Pokemon` but flx changed it in the circular imports removal PR.
There might be a better way of doing things here (this is a file I want to refactor because that `| any` is just... ugh) but I haven't looked into it yet.

src/system/pokemon-data.ts
DayKev: `@todo` is a TSDoc tag, but single-line `//` comments are not TSDocs.

src/data/moves/move-attrs/heal-attr.ts
DayKev: That would change the existing behavior (I replaced the `?? 1` with a default value for the parameter).

src/data/arena-tag.ts
DayKev: That might change the existing behavior if it's possible for `turnCount` to be negative (perhaps accidentally?). I wanted to preserve existing behavior as much as possible in this PR and just make what are effectively style changes.

src/ui/handlers/starter-select-ui-handler.ts
```diff
@@ -2851,10 +2851,10 @@ export class StarterSelectUiHandler extends MessageUiHandler {
           }
         }
 
-        container.starterPassiveBgs.setVisible(!!globalScene.gameData.starterData[speciesId].passiveAttr);
+        container.starterPassiveBgs.setVisible(globalScene.gameData.starterData[speciesId].passiveAttr > 0);
         container.hiddenAbilityIcon.setVisible(
-          !!globalScene.gameData.dexData[speciesId].caughtAttr
-            && !!(globalScene.gameData.starterData[speciesId].abilityAttr & 4),
+          globalScene.gameData.dexData[speciesId].caughtAttr > 0
+            && (globalScene.gameData.starterData[speciesId].abilityAttr & 4) > 0,
```
PigeonBar:
```suggestion
            && (globalScene.gameData.starterData[speciesId].abilityAttr & AbilityAttr.ABILITY_HIDDEN) > 0,
```

innerthunder (dismissed):

PigeonBar (dismissed):

src/data/mystery-encounters/encounters/mysterious-challengers-encounter.ts
PigeonBar: It's not equivalent, since `!randSeedInt(2)` is a `boolean` and `randSeedInt(2)` is a `number`

DayKev: That would make it be `0 | 1` but we want it to stay a boolean.
Edit: Missed Pigeon's reply lol.

src/data/arena-tag.ts
DayKev: I could still change it anyway if you think that's better though.

flx-sta: Wouldn’t this change the behavior? if the previous condition is met the decrement won’t be executed. I guess the outcome is the same.. but is keep it out of this scope of the PR

DayKev: So change it later, ok. 👍

PR #925, opened 2025-04-10
[Refactor] Simplify/clarify `VictoryPhase` and add TSDocs

DayKev:
N/A

The phase was a bit confusing and undocumented.

- `if/else` blocks moved around and early returns added to reduce indentation and improve readability / code flow.
- TSDocs added.

## How to test the changes?
Defeat some pokemon on various waves.

## Checklist

### DIFF
--- a/src/phases/victory-phase.ts
+++ b/src/phases/victory-phase.ts
@@ -1,23 +1,46 @@
-import { type BattlerIndex } from "#enums/battler-index";
-import { BattleType } from "#enums/battle-type";
+// -- start tsdoc imports --
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import type { GameOverPhase } from "#app/phases/game-over-phase";
+// -- end tsdoc imports--
+
 import { handleMysteryEncounterVictory } from "#app/data/mystery-encounters/utils/encounter-phase-utils";
+import { EVIL_BOSS_2_WAVE } from "#app/data/special-waves";
 import { globalScene } from "#app/global-scene";
 import { type CustomModifierSettings } from "#app/modifier/modifier-type";
 import { modifierTypes } from "#app/modifier/modifier-types";
-import { PokemonPhase } from "./abstract-pokemon-phase";
-import { BattleEndPhase } from "./battle-end-phase";
-import { EggLapsePhase } from "./egg-lapse-phase";
-import { ModifierRewardPhase } from "./modifier-reward-phase";
-import { NewBattlePhase } from "./new-battle-phase";
-import { SelectModifierPhase } from "./select-modifier-phase";
-import { TrainerVictoryPhase } from "./trainer-victory-phase";
+import { PokemonPhase } from "#app/phases/abstract-pokemon-phase";
+import { BattleEndPhase } from "#app/phases/battle-end-phase";
+import { EggLapsePhase } from "#app/phases/egg-lapse-phase";
+import { ModifierRewardPhase } from "#app/phases/modifier-reward-phase";
+import { NewBattlePhase } from "#app/phases/new-battle-phase";
+import { SelectModifierPhase } from "#app/phases/select-modifier-phase";
+import { TrainerVictoryPhase } from "#app/phases/trainer-victory-phase";
 import { ArenaTagType } from "#enums/arena-tag-type";
-import { EVIL_BOSS_2_WAVE } from "#app/data/special-waves";
+import { BattleType } from "#enums/battle-type";
+import { type BattlerIndex } from "#enums/battler-index";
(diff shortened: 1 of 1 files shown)


src/phases/victory-phase.ts
```diff
@@ -45,91 +68,95 @@ export class VictoryPhase extends PokemonPhase {
...
+          new ModifierRewardPhase(!(waveIndex % 250) ? modifierTypes.VOUCHER_PREMIUM : modifierTypes.VOUCHER_PLUS),
+        );
       }
+    } else {
+      const modifierType = gameMode.isGymWave(waveIndex) ? modifierTypes.SUPER_EXP_CHARM : modifierTypes.EXP_CHARM;
+      phaseManager.pushPhase(new ModifierRewardPhase(modifierType));
+    }
+
+    if ([50, 100, 150].includes(waveIndex)) {
+      phaseManager.pushPhase(new ModifierRewardPhase(modifierTypes.GOLDEN_POKEBALL));
```
Tempo-anon, line 141: All these formulas for different modes is hmmm. I wonder if it would be helpful to create 3 helper functions to handle the classic/endless/daily cases. Each of those functions can then be individually documented with a table or something so the rewards are understandable at a glance

src/phases/victory-phase.ts
```diff
@@ -45,91 +68,95 @@ export class VictoryPhase extends PokemonPhase {
...
    * If this wave is a fixed battle with special custom modifier rewards,
-   * will pass those settings to the upcoming {@linkcode SelectModifierPhase}`.
+   * pass those settings to the upcoming {@linkcode SelectModifierPhase}.
    */
   protected getFixedBattleCustomModifiers(): CustomModifierSettings | undefined {
     const gameMode = globalScene.gameMode;
     const waveIndex = globalScene.currentBattle.waveIndex;
     if (gameMode.isFixedBattle(waveIndex)) {
       return gameMode.getFixedBattle(waveIndex).customModifierRewardSettings;
     }
```
Tempo-anon, line 161: Will this `if` ever be false? Seems a bit weird to have the only `return` statement in an `if`

src/phases/victory-phase.ts
```diff
@@ -45,91 +68,95 @@ export class VictoryPhase extends PokemonPhase {
...
+
+    if ([50, 100, 150].includes(waveIndex)) {
+      phaseManager.pushPhase(new ModifierRewardPhase(modifierTypes.GOLDEN_POKEBALL));
     }
 
     this.end();
   }
 
+  public override end(): void {
+    globalScene.phaseManager.pushPhase(new NewBattlePhase());
```
Tempo-anon, line 148: Is this always guaranteed to be a `NewBattlePhase`?

src/phases/victory-phase.ts
```diff
@@ -1,23 +1,46 @@
-import { type BattlerIndex } from "#enums/battler-index";
-import { BattleType } from "#enums/battle-type";
+// -- start tsdoc imports --
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
flx-sta: Please use the multi-line disable to make it easier for future developers to add additional tsdoc imports

src/phases/victory-phase.ts
flx-sta: Might wanna go for whole victory phases per mode but Idk if that would be easy to achieve or cause cluster-f

src/phases/victory-phase.ts
DayKev: The `if` is `false` when it's not a fixed battle (aka it's not wave 5, 8, ...).

src/phases/victory-phase.ts
DayKev: Yeah I want to break up `VictoryPhase` into `PostKnockoutPhase` and... I guess one would still be called `VictoryPhase`? I'm not sure, but this phase is doing a few too many things. For now I want it to be easier to understand though, then a follow-up can break it apart.

src/phases/victory-phase.ts
DayKev: Yes, I used `this.end()` in all the places there was formerly a `pushPhase(new NewBattlePhase())`.

PigeonBar: Yes (if the `NewBattlePhase` is unwanted, `super.end()` gets called instead)

src/phases/victory-phase.ts
DayKev: Also if there's no `return` then it implicitly returns `undefined`, which is why I removed the `return undefined;` statement.

PigeonBar (approved):

torranx (approved):

PR #918, opened 2025-04-08
[Documentation] Add documentation for how to add a new biome

Tempo-anon:
## What are the changes the user will see?

None

## Why am I making these changes?

Mentioned in #916. More documentation is also good

## What are the changes from a developer perspective?

Created a readme

## Screenshots/Videos

n/a

## How to test the changes?

n/a

## Checklist

- [ ] ⚠️ If this is a PR for `main` (such as a hotfix), has the game version been updated (`npm run update-version:patch` / `npm run update version:minor`?
- [x] Otherwise: **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [ ] Have I tested the changes manually?

- [ ] Are all unit tests still passing? (`npm run test:silent`)
  - [ ] Have I created new automated tests (`npm run test:create`) or updated existing tests related to the PR's changes?
- [ ] Have I provided screenshots/videos of the changes (if applicable)?
  - [ ] Have I made sure that any UI change works for both UI themes (dark and light)?

#### Are there any localization additions or changes? If so:

- [ ] Has a locales PR been created on the [locales](https://github.com/despair-games/poketernity-locales) repo?
  - [ ] If so, please leave a link to it here:
- [ ] Have I added the `Localization` tag to this PR?

#### If there are no locale changes:
- [ ] Have I made sure **not** to commit any changes to the locale repo on this branch?

PR #926, opened 2025-04-11
[Balance] Add chargestone cave biome

### LINKED ISSUE #930: [Biome] Add Chargestone Cave
Add a new biome called "Chargestone Cave"

Tempo-anon:

A new biome is added

* There was only 1 electric biome previously so this biome will be another option for electric types
* Trying to challenge some Sam traditions (regional forms being rarer) while also exploring some different ideas

From a design perspective, this is a combination of Unova's [Chargestone Cave](https://bulbapedia.bulbagarden.net/wiki/Chargestone_Cave) and [Chargestone Cavern](https://bulbapedia.bulbagarden.net/wiki/Chargestone_Cavern)

For a Pokemon pool perspective here's the main basis I used for determining Pokemon
Common: These should be the most iconic Pokemon of the biome. Ones that you should expect to see a lot of.
Uncommon: Pokemon that would also make sense in the biome but not as much as often as the common pool.
Rare: This should be for Pokemon that are a step up from uncommon, be they rarer, stronger, or just generally more desirable.
Super rare: This should be reserved for Pokemon that are really rare but not quite to the level of sublegends. Psuedolegends would belong here.
Ultra rare: This should be where all sublegends belong. This includes sublegends, some mythics, ultra beasts, paradoxes

Boss: Pokemon from the common/uncommon/rare pool but focusing on the common pool for iconic Pokemon
Boss Rare: Similar to Boss, but more focusing on the uncommon/rare pool
Boss Super Rare: This should be composed of Pokemon from the Super Rare and Ultra Rare pool so everything in here should be quite desirable.
Boss Ultra Rare: Reserved for legendaries (like the 600+bst box legends) and really rare mythics

Assets (all temp values, art stolen from cave, audio stolen from power plant)
* chargestone_cave_a.png
* chargestone_cave_b.png
* chargestone_cave_bg.png
* chargestone_cave.mp3

biome-link.ts (temp value)
Made Chargestone Cave an alternate version of Cave where it has all the same inward and outward paths

copy-biome-type-attr.ts
* Chargestone cave makes user electric

nature-power-attr.ts
* Chargestone cave becomes charge beam
> [!IMPORTANT]
>  * Changed Beach to Surf instead of scald since I forgot to do that in the steam vent PR

secret-power-attr.ts
* Chance to paralyze

chargestone-cave.ts
* Created a new biome

## Screenshots/Videos

image

## How to test the changes?

```ts
const overrides = {
  STARTING_BIOME_OVERRIDE:BiomeId.CHARGESTONE_CAVE
}
```

## Checklist

   - NOPE! Pay attention to the nature-power fix (see IMPORTANT)


src/data/biomes/chargestone-cave.ts
```diff
@@ -0,0 +1,146 @@
...
+    [TimeOfDay.DUSK]: [],
+    [TimeOfDay.NIGHT]: [],
+    [TimeOfDay.ALL]: [SpeciesId.BELDUM, SpeciesId.ROTOM],
+  },
+  [BiomePoolTier.ULTRA_RARE]: {
+    [TimeOfDay.DAWN]: [],
+    [TimeOfDay.DAY]: [],
+    [TimeOfDay.DUSK]: [],
+    [TimeOfDay.NIGHT]: [],
+    [TimeOfDay.ALL]: [SpeciesId.SANDY_SHOCKS, SpeciesId.IRON_THORNS],
```
PigeonBar, line 66: Did we ever decide to make Paradox Pokemon encounterable and catchable outside of the End biome?

Tempo-anon: I don't see any reason why they should all be packed into a single story related location, especially since there's not even a connection between them and Eternatus. I think it makes more sense to treat them like UBs or other sublegendaries and have them more accessible, but I'm also interested in how others feel

flx-sta: Wouldn't that make it possible to catch them? Since catching is only (initially) prohibited in the `END` biome?

src/data/biomes/chargestone-cave.ts
```diff
@@ -0,0 +1,146 @@
...
+import { TerrainType } from "#enums/terrain-type";
+import { TimeOfDay } from "#enums/time-of-day";
+import { TrainerType } from "#enums/trainer-type";
+import { WeatherType } from "#enums/weather-type";
+
+/**
+ * TODO: Add original art and audio assets
+ */
+
+const pokemonPool = {
```
flx-sta, line 14: Don't we have interfaces at the ready for those? So the compiler would catch it if something was missing?

src/data/biomes/chargestone-cave.ts
```diff
@@ -0,0 +1,146 @@
...
+  [BiomePoolTier.BOSS_ULTRA_RARE]: {
+    [TimeOfDay.DAWN]: [],
+    [TimeOfDay.DAY]: [],
+    [TimeOfDay.DUSK]: [],
+    [TimeOfDay.NIGHT]: [],
+    [TimeOfDay.ALL]: [SpeciesId.ZEKROM],
+  },
+};
+
+const trainerPool = {
```
flx-sta, line 113: Don't we have interfaces at the ready for those? So the compiler would catch it if something was missing?

src/data/biomes/chargestone-cave.ts
```diff
@@ -0,0 +1,146 @@
...
+  [BiomePoolTier.RARE]: [],
+  [BiomePoolTier.SUPER_RARE]: [],
+  [BiomePoolTier.ULTRA_RARE]: [],
+  [BiomePoolTier.BOSS]: [TrainerType.CLAY, TrainerType.ELESA],
+  [BiomePoolTier.BOSS_RARE]: [],
+  [BiomePoolTier.BOSS_SUPER_RARE]: [],
+  [BiomePoolTier.BOSS_ULTRA_RARE]: [],
+};
+
+const weatherPool = {
```
flx-sta, line 125: Don't we have interfaces at the ready for those? So the compiler would catch it if something was missing?

src/data/biomes/chargestone-cave.ts
```diff
@@ -0,0 +1,146 @@
...
+  [WeatherType.NONE]: 1,
+};
+
+/**
+ * 1/2 of Electric terrain
+ */
+const terrainPool: Partial<Record<TerrainType, number>> = {
```
flx-sta, line 132: We should create a dedicated *alias* type for this as there is probably multiple instances of this being used:

`@types/TerrainPool.ts`
```ts
type FullTerrainPool = Record<TerrainType, number>>;

export type TerrainPool = Partial<FullTerrainPool>;
```
and use it via:
```ts
const terrainPool: TerrainPool = {
```

src/data/moves/move-attrs/copy-biome-type-attr.ts
```diff
@@ -102,6 +102,7 @@ export class CopyBiomeTypeAttr extends MoveEffectAttr {
       case BiomeId.ISLAND:
         return ElementalType.FAIRY;
       case BiomeId.POWER_PLANT:
+      case BiomeId.CHARGESTONE_CAVE:
         return ElementalType.ELECTRIC;
```
flx-sta, line 106: NIT (Out of scope):
I'm wondering if a biome should define it's elemental-type equivalent. 
Might be counterintuitive. <sub>Thinking out loud</sub>

flx-sta (dismissed):
Looks fine. Some comments

src/data/biomes/chargestone-cave.ts
Tempo-anon: Yes. Keeping 22 Pokemon completely locked behind an arbitrary gacha was one of the decisions Sam made that mainly led to artificial longevity for 100% dex completion.

src/data/moves/move-attrs/copy-biome-type-attr.ts
Tempo-anon: Since it's only ever used for this one attr, I think it makes more sense to have it on the attr rather than the biome class

src/data/biomes/chargestone-cave.ts
PigeonBar: In that case, we could definitely start allowing players to catch Paradox Pokemon in the End biome before hatching them. However, I still think that the End biome having its own set of signature species is a cool idea, and it matches how Paradox Pokemon are only encountered at the very end of Scarlet/Violet.

Tempo-anon: I think it'd make more sense to just swap out the Pokemon that appear in the end biome then. Maybe something like the UBs? They seem to fit more as extra dimensional beings and there's only 11 of them.

PigeonBar: If we do that, it probably requires a separate PR that also handles some other stuff, including:
- Giving Paradox Pokemon boss/pseudolegendary status
- Some ME Pokemon pools that specifically exclude Paradox Pokemon

src/data/biomes/chargestone-cave.ts
```diff
@@ -0,0 +1,146 @@
...
+    ],
+  },
+  [BiomePoolTier.SUPER_RARE]: {
+    [TimeOfDay.DAWN]: [],
+    [TimeOfDay.DAY]: [],
+    [TimeOfDay.DUSK]: [],
+    [TimeOfDay.NIGHT]: [],
+    [TimeOfDay.ALL]: [SpeciesId.BELDUM, SpeciesId.ROTOM],
+  },
+  [BiomePoolTier.ULTRA_RARE]: {
```
PigeonBar, line 61: It turns out that Power Plant has 5 possible Ultra Rare spawns, which is a lot, so some of them should also be moved to Chargestone Cave.

PigeonBar:
Also remember to set the biome to be indoors in `arena.ts`, and also the tasks mentioned in the comment https://github.com/Despair-Games/poketernity/pull/918/files#r2038637699.

src/data/biomes/chargestone-cave.ts
Tempo-anon: That's another thing I wanted to bring up. Should the cover legends and/or gym leaders be exclusive to a single biome? Or is it okay for some to appear in multiple? 🤔 I know in the steam vents PR I also did some cleanup in the volcano one but I did not do the power plant clean up in this one.

src/data/biomes/chargestone-cave.ts
flx-sta: I feel like some of those thoughts might actually unravel as soon as we start steering towards the custom storyline. Right now everything feels quite blurry and hard to asses

src/data/biomes/chargestone-cave.ts
Tempo-anon: I think Raikou and Thunurus don't really belong in a power plant but I don't have any particularly good places to place them so.....

I'll leave touching other biomes in a separate PR

PR #927, opened 2025-04-11
[Refactor] Breakup `pokemon.ts` file

### LINKED ISSUE #801: [Refactor] Move PlayerPokemon and EnemyPokemon into their own files
### Is your feature request related to a problem? Please describe.

They shouldn't all be in the Pokemon.ts file

### Describe the Feature

.

### Additional context

_No response_

DayKev:
N/A

The `pokemon.ts` file is way too massive.
Closes #801

- `DamageResult` type moved to `src/@types/DamageResult.ts`
- `DamageCalculationResult` interface moved to `src/@types/DamageCalculationResult.ts`
- `EnemyPokemon` class moved to `src/field/enemy-pokemon.ts`
- `PlayerPokemon` class moved to `src/field/player-pokemon.ts`
- `PokemonBattleData` renamed to `PokemonWaveData`, changed from class to interface, moved to `src/@types/PokemonWaveData.ts`; `Pokemon#battleData` renamed to `Pokemon#waveData`
- `PokemonTurnData` changed from class to interface, moved to `src/@types/PokemonTurnData.ts`
- Created `PhaseManager#queueStatStageChangePhase` to fix newly introduced circular import issue

## Checklist


src/phase-manager.ts
```diff
@@ -524,4 +526,20 @@ export class PhaseManager {
...
     }
   }
+
+  public queueStatStageChangePhase(
+    battlerIndex: BattlerIndex,
+    source: Pokemon | null,
+    stats: BattleStat[],
+    stages: number,
+    options: SSCPhaseOptions = {},
+    eager: boolean = true,
```
Tempo-anon, line 536: What is this `eager` field? A quick ctrl+f of this PR only has it showing up in `berry.ts` where it's `true` in both cases

DayKev: This is ultimately supposed to replace all `[push|unshift]Phase(new StatStageChangePhase(...))` calls so I copied the way the other replacement functions do it. For the purposes of this PR I didn't replace any of them other than the ones that were directly causing the new circular import issue; the rest would be done in a follow-up PR.

src/field/pokemon.ts
```diff
@@ -4035,7 +4048,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
   }
 
   resetBattleData(): void {
```
PigeonBar: This should probably also be renamed to `resetWaveData`.

PigeonBar (dismissed):

src/field/pokemon.ts
DayKev: Oh good catch, looks like I missed this one.

PigeonBar (approved):

PR #960, opened 2025-04-12
[P2][Balance] Thief and Covet always proc (no more global RNG)

### LINKED ISSUE #531: [P2 Bug] Reused RNG calls causing unintended RNG correlations
Copied over from PKR 4527

### Describe the bug

[Original report (Discord)](https://discord.com/channels/1125469663833370665/1126338176659763210/threads/1289920989106274335)

Some in-battle RNG mechanics (e.g., Covet/Thief, Endure Tokens) actually call global RNG instead of battle RNG. Due to #3952 resetting RNG at the start of each battle, this means that the aforementioned RNG mechanics will reuse RNG calls that were used to generated the wild Pokemon.

One significant way that this bug shows up is: If the opponents have 10 Endure Tokens, and if you naturally encounter a wild double battle with 0 lures, then the Endure Tokens will generally be guaranteed to activate on Turn 1.

Here is the RNG call for rolling a double battle:

20240930 Double Battle Endure 1

And here is the RNG call for rolling the Endure Tokens:

20240930 Double Battle Endure 2

### FAQ's:

**Q**: If lures are used to force double battles, why does this bug no longer occur?

**A**: Lures increase the allowed RNG threshold for double battles, which makes it possible to get a double battle without an Endure Token activation.

**Q**: Why was this bug not reported before the ME's release?

**A**: Before the ME's release, there was a separate bug that prevented Endure Tokens from working on full HP Pokemon.

### Reproduction

Play an endless run naturally with 10 Endure Tokens and 0 lures, and observe correlations between double battles and Endure Token activations.

### Expected behavior

RNG calls should not be reused. Ideally, all in-battle mechanics should use battle RNG.

### Screenshots / Videos

_No response_

### Session export file

For reference, the above screenshots used the following session (which will have a double battle on wave 172):
[20240930 Double Battle Endure.prsv.txt](https://github.com/user-attachments/files/17194535/20240930.Double.Battle.Endure.prsv.txt)

### User data export file

_No response_

### Additional context

_No response_

Tempo-anon:

Thief and Covet will always proc their effects instead of having a 30% to

* Closes #531
* More mainline accuracy

Deleted the portions of code that used `Phaser.Math.RND.realInRange(0, 1);`

## Screenshots/Videos

n/a

## How to test the changes?

Override with thief/covet

## Checklist


src/data/moves/move-attrs/steal-held-item-chance-attr.ts
```diff
@@ -13,22 +13,15 @@ import { MoveEffectAttr } from "#app/data/moves/move-attrs/move-effect-attr";
  * - "If Knock Off causes a Pokémon with the Sticky Hold Ability to faint, it can now remove that Pokémon's held item."
  */
 export class StealHeldItemChanceAttr extends MoveEffectAttr {
```
flx-sta: Should probably remove the `Chance` word from the Attr then

```suggestion
export class StealHeldItemAttr extends MoveEffectAttr {
```

flx-sta (changes requested):

PigeonBar:
This might be too overpowered, since Magician also has a 100% activation rate and is also sometimes considered overpowered. Unlike Magician, Thief/Covet also always steals the most valuable item from the opponent, which makes Thief/Covet even stronger. (The reason why a 100% activation rate works in mainline is because in mainline, Magician/Thief/Covet has a downside of requiring the user to start with no held item.)

Tempo-anon:
The goal is to move to one held item per Pokemon eventually so there will be less itemspam and juggernauting but if current balance is a concern then I can add on the clause of it requiring the user to have no item

flx-sta:
I can see how it could be considered overpowered right now. Since there is no release in sight right now I can see it as not being an issue unless we release. We should definitely document that somewhere so it’s not forgotten

flx-sta (dismissed):

src/data/moves/move-attrs/steal-held-item-attr.ts
```diff
@@ -12,23 +12,16 @@ import { MoveEffectAttr } from "#app/data/moves/move-attrs/move-effect-attr";
...
   override applyEffect(user: Pokemon, target: Pokemon, move: Move): boolean {
     if (move.hitsSubstitute(user, target)) {
       return false;
     }
-
-    const rand = Phaser.Math.RND.realInRange(0, 1);
-    if (rand >= this.chance) {
-      return false;
-    }
+    // TODO: Refactor this mess after items are properly implemented
```
innerthunder:
```suggestion
    /** @todo Refactor this mess after items are properly implemented */
```

innerthunder (commented):
1 nit, otherwise lgtm

PR #968, opened 2025-04-12
[Refactor] Break part of `VictoryPhase` into `PostKnockoutPhase`

### LINKED ISSUE #966: [Refactor] Investigate breaking up/refactoring `VictoryPhase`

DayKev:
N/A

Keep phase responsibility limited. Resolves #966 

The code that only ran when a wave was completed was left in `VictoryPhase` and the rest was broken out into `PostKnockoutPhase`.

## How to test the changes?
Defeat some pokemon, clear a wave.

## Checklist


MokaStitcher:
If I follow this phase is only called post enemy knockout, not player, right?
If yes the name should probably be clarified
My main issue with the name though Is that it specifically references KO, but is also called when an enemy is caught, so it's a bit misleading. PostEnemyDefeatPhase maybe?

Is VictoryPhase only ever called from PosrKOPhase now? I really feel something should be done with renaming both those Phases if yes, to clarify their individual purpose and difference. WaveVictoryPhase could help for the VictoryPhase maybe? or something like that?

DayKev:
Yeah the naming is difficult because of where/when these phases are called... I wasn't sure what exactly to do about it.

Tempo-anon:
`PostEnemyKoOrCapturePhase`? It's a bit of a mouthful but at least then it would capture the exact usecases

PigeonBar (approved):

flx-sta (approved):

src/phases/post-knockout-phase.ts
```diff
@@ -0,0 +1,55 @@
...
+import { PhaseId } from "#enums/phase-id";
+
+/**
+ * Handles the actions after the player KOs a pokemon:
+ * - Increases {@linkcode globalScene.gameData.gameStats.pokemonDefeated | pokemon defeated} count in game stats
+ * - Applies EXP gain
+ * - If this is a Mystery Encounter: hand off to a ME function and end the phase
+ * - If there are no more unfainted pokemon on the enemy team, unshift a {@linkcode VictoryPhase}
+ */
+export class PostKnockoutPhase extends PokemonPhase {
```
Tempo-anon, line 16: Moka's comment says that this is also called when catching a Pokemon. If that's the case the name/tsdoc should also reflect that

PR #971, opened 2025-04-13
[Test] Add test for effectiveness of Ghost moves vs Normal pkmn

DayKev:
N/A

Improve testing coverage.

Adds regression tests for Ghost moves vs primary and secondary Normal type Pokémon. Also updated the other test descriptions to specify whether the target Pokémon's typing was primary or secondary.

## How to test the changes?
`npm run test:silent effectiveness`

## Checklist

### DIFF
--- a/test/moves/effectiveness.test.ts
+++ b/test/moves/effectiveness.test.ts
@@ -53,22 +53,28 @@ describe("Moves - Type Effectiveness", () => {
     game.phaseInterceptor.restoreOg();
   });
 
-  it("Normal-type attacks are neutrally effective against Normal-type Pokemon", () =>
+  it("Normal-type attacks are neutrally effective against primary Normal-type Pokemon", () =>
     testMoveEffectiveness(game, MoveId.TACKLE, SpeciesId.SNORLAX, 1));
 
-  it("Normal-type attacks are not very effective against Steel-type Pokemon", () =>
+  it("Normal-type attacks are not very effective against primary Steel-type Pokemon", () =>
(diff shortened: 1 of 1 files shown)


PigeonBar (approved):

test/moves/effectiveness.test.ts
```diff
@@ -80,13 +86,13 @@ describe("Moves - Type Effectiveness", () => {
   it("Electric-type attacks are super-effective against Tera-Water Pokemon", () =>
     testMoveEffectiveness(game, MoveId.THUNDERBOLT, SpeciesId.EXCADRILL, 2, AbilityId.BALL_FETCH, ElementalType.WATER));
 
-  it("Powder moves have no effect on Grass-type Pokemon", () =>
+  it("Powder moves have no effect on primary Grass-type Pokemon", () =>
```
Tempo-anon, line 89: Is this clarification needed if there also isn't a test for secondary grass type?

Tempo-anon (approved):
Not sure if there's any need to clarify primary/secondary typing when they're mostly identical for a lot of tests

DayKev:
I figured it would be a good idea to specify even if it isn't necessarily relevant (but who knows, it might become relevant at some point).

The reason for adding the new tests was to prevent the same regression that happened at PKR where there was a check for type2 being falsy instead of specifically `null`, which meant if type2 was Normal (enum value 0) the check would make it as if the pokemon had no secondary typing.

innerthunder (approved):

PR #972, opened 2025-04-13
[Misc] Remove unused overworld forms of [Mir|Kor]aidon

DayKev:
N/A

These forms are completely unused and can potentially break things if they're accidentally used for something.

The extra forms of Miraidon and Koraidon that were only used for the overworld in SV are removed.

## Checklist


Tempo-anon (approved):
Do these two need the apex build still? Showdown just has the base form

DayKev:
I can't remove the form that they actually use (ie: making them "formless") because it requires repacking the icon textures and I can't do that.

Tempo-anon:
Maybe someone with texture packer can help? Feels like this is taking the right step and just 99% of the way there

torranx (approved):

PR #977, opened 2025-04-14
[Refactor] Use `Move.checkFlag()` instead of `.hasFlag()`

### LINKED ISSUE #949: [Refactor] Use `Move.checkFlag` instead of `hasFlag`
The use of `Move.hasFlag()` is outdated and should be replaced with using `Move.checkFlag()`.

> [!WARNING]
>  This doesn't mean that `hasFlag()` should be removed. But it should possibly be marked as "private"

## Check Flag
https://github.com/Despair-Games/poketernity/blob/8d9145598dbdd49aaa890c6b0a49ab5167e9c6fe/src/data/moves/move.ts#L617-L658

## Has Flag

https://github.com/Despair-Games/poketernity/blob/8d9145598dbdd49aaa890c6b0a49ab5167e9c6fe/src/data/moves/move.ts#L204-L213

flx-sta:

Closes #949

- Replaced `Move.checkFlag()` with `.hasFlag()`
- Use `describe.each` in gooey test
- change `checkFlag()` parameter `target` from `target: Pokemon | null` to `target?: Pokemon`

## Screenshots/Videos

> [!NOTE]
> Is it required here? 

## How to test the changes?

Run the tests

## Checklist


src/data/animations/battle-anims.ts
```diff
@@ -35,7 +35,7 @@ const targetFocusY = 84 - 32;
 
 export abstract class BattleAnim {
   /** The {@linkcode Pokemon} from which the effect of this animation originated */
-  public user?: Pokemon;
+  public user: Pokemon;
```
DayKev, line 38: It's interesting that this was marked optional but was always passed in anyways. This'll allow for the removal of a bunch of `!` now.

DayKev (approved):

src/data/animations/battle-anims.ts
flx-sta: I totally forgot to mention that in the changes. Originally I just removed it and follow along and found out that nothing seemed to have it NOT defined. But sometimes those class definitions are a bit hidden.

src/data/init/init-abilities.ts
```diff
@@ -1041,9 +1041,9 @@ export function initAbilities() {
     new Ability(AbilityId.LONG_REACH, 7)
       .attr(IgnoreContactAbAttr),
     new Ability(AbilityId.LIQUID_VOICE, 7)
-      .attr(MoveTypeChangeAbAttr, ElementalType.WATER, 1, (_user, _target, move) => !!move?.hasFlag(MoveFlags.SOUND_MOVE)),
+      .attr(MoveTypeChangeAbAttr, ElementalType.WATER, 1, (user, target, move) => !!user && !!move?.checkFlag(MoveFlags.SOUND_MOVE, user, target)),
```
Tempo-anon, line 1044: Can this just be `user && blahblah` instead of `!!user && blahblah`

src/data/animations/battle-anims.ts
innerthunder: This was changed recently in #913 and may be relevant for #885 with the changes to `WeatherEffectPhase`

innerthunder (approved):

src/data/init/init-abilities.ts
innerthunder: No, the return type would then be `boolean | undefined`

PR #978, opened 2025-04-14
[Test] Fix `EvolutionPhase` tests and add `LEVEL_CAP_OVERRIDE`

DayKev:
N/A

The `EvolutionPhase` tests were flaky because sometimes the test would encounter a trainer battle instead of a wild pokemon.

Added a new `LEVEL_CAP_OVERRIDE` (negative numbers disable the cap, positive numbers set the cap to that value) and changed the tests to use the new override instead of setting the wave to 100.

## How to test the changes?
`npm run test evolution-phase`

## Checklist

### DIFF
--- a/test/phases/evolution-phase.test.ts
+++ b/test/phases/evolution-phase.test.ts
@@ -25,7 +25,7 @@ describe("Evolution Phase", () => {
   beforeEach(() => {
     game = new GameManager(phaserGame);
     game.override
-      .startingWave(100) // Make sure level cap is high enough for evolution
+      .levelCap(-1)
       .ability(AbilityId.BALL_FETCH)
       .battleType("single")
       .disableCrits()
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -1914,10 +1914,14 @@ export default class BattleScene extends SceneBase {
     this.currentBattle.battleScore += Math.ceil(scoreIncrease);
   }
 
-  getMaxExpLevel(ignoreLevelCap?: boolean): number {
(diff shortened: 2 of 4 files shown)


flx-sta (approved):

test/phases/evolution-phase.test.ts
```diff
@@ -25,7 +25,7 @@ describe("Evolution Phase", () => {
   beforeEach(() => {
     game = new GameManager(phaserGame);
     game.override
-      .startingWave(100) // Make sure level cap is high enough for evolution
+      .levelCap(-1)
```
Tempo-anon, line 28: I'm confused, I thought the test was flaky because a trainer showed up instead of a wild battle. In that case, wouldn't it be better to force the test to be a wild battle? What does the level cap have to do with anything?

PigeonBar: We are moving the starting wave back to 1 in order to force wild battles. This would set the level cap to be too low for Bulbasaur to evolve, so the level cap needs to be disabled.

PigeonBar (approved):

test/phases/evolution-phase.test.ts
Tempo-anon: Ah ok that makes sense. My first thought would have just been to change the startingWave from 100 to 101 because trainers are already not allowed to appear on X1 floors

DayKev: That's true for now, but if we ever fix the bug that prevents us from allowing trainers on X1 waves and then enable them, that could also break the test again.

PR #982, opened 2025-04-14
[Refactor] Replace `is*Tag` in battler-tags

### LINKED ISSUE #950: [Refactor] Replace `.is*Tag()` with checks against the `*TagTypes` arrays
Instead of using the very static `is*Tag()` method there should be predefined arrays that can be checked against instead to simplify the code and reduce redundancy

flx-sta:

Closes #950 

- add `BattlerTag.isType<>()` method
- add `TypeBoostTagType` array (frozen)
- replaced `BattlerTag.isMysteryEncounterPostSummonTag()` calls with `isType<>()`
- replaced `BattlerTag.isTypeBoostTag()` calls with `isType<>()`

## Checklist


src/data/battler-tags/battler-tag.ts
```diff
@@ -1,11 +1,9 @@
-import type { MysteryEncounterPostSummonTag } from "./mystery-encounter-post-summon-tag";
-import type { TypeBoostTag } from "./type-boost-tag";
 import { allMoves } from "#app/data/data-lists";
-import type { Pokemon } from "#app/field/pokemon";
+import { type Pokemon } from "#app/field/pokemon";
```
DayKev:
```suggestion
import type { Pokemon } from "#app/field/pokemon";
```

src/data/battler-tags/battler-tag.ts
```diff
@@ -1,11 +1,9 @@
...
 import { globalScene } from "#app/global-scene";
-import type { BattlerTagLapseType } from "#enums/battler-tag-lapse-type";
-import type { BattlerTagType } from "#enums/battler-tag-type";
-import type { MoveId } from "#enums/move-id";
+import { type BattlerTagLapseType } from "#enums/battler-tag-lapse-type";
+import { type BattlerTagType } from "#enums/battler-tag-type";
+import { type MoveId } from "#enums/move-id";
```
DayKev:
```suggestion
import type { BattlerTagLapseType } from "#enums/battler-tag-lapse-type";
import type { BattlerTagType } from "#enums/battler-tag-type";
import type { MoveId } from "#enums/move-id";
```

src/phases/post-summon-phase.ts
```diff
@@ -1,11 +1,12 @@
-import type { CommanderAbAttr } from "#app/data/abilities/ab-attrs/commander-ab-attr";
-import type { PostSummonAbAttr } from "#app/data/abilities/ab-attrs/post-summon-ab-attr";
+import { type CommanderAbAttr } from "#app/data/abilities/ab-attrs/commander-ab-attr";
+import { type PostSummonAbAttr } from "#app/data/abilities/ab-attrs/post-summon-ab-attr";
```
DayKev:
```suggestion
import type { CommanderAbAttr } from "#app/data/abilities/ab-attrs/commander-ab-attr";
import type { PostSummonAbAttr } from "#app/data/abilities/ab-attrs/post-summon-ab-attr";
```

src/phases/post-summon-phase.ts
```diff
@@ -1,11 +1,12 @@
...
 import { applyAbAttrs } from "#app/data/abilities/apply-ab-attrs";
+import { type MysteryEncounterPostSummonTag } from "#app/data/battler-tags/mystery-encounter-post-summon-tag";
```
DayKev:
```suggestion
import type { MysteryEncounterPostSummonTag } from "#app/data/battler-tags/mystery-encounter-post-summon-tag";
```

src/phases/post-summon-phase.ts
```diff
@@ -1,11 +1,12 @@
...
 import { globalScene } from "#app/global-scene";
 import { PokemonPhase } from "#app/phases/abstract-pokemon-phase";
 import { EntryHazardArenaTagTypes } from "#app/utils/arena-tag-type-utils";
 import { AbAttrFlag } from "#enums/ab-attr-flag";
-import type { BattlerIndex } from "#enums/battler-index";
+import { type BattlerIndex } from "#enums/battler-index";
```
DayKev:
```suggestion
import type { BattlerIndex } from "#enums/battler-index";
```

PigeonBar (dismissed):

DayKev (dismissed):
I've recently learned that TS will potentially not fully erase imports that are `import { type X } from Y` but will for `import type { X } from Y`. So for imports that aren't mixed (like `import { X, type Y } from Z`) we should keep the `type` declaration on the outside of the brackets.
image
https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax
We aren't currently using this specific option, but it's likely that TS exhibits approximately the same behavior without it, since this option just disables the TS compiler's "try to autodetect undeclared type imports" behavior.

src/data/battler-tags/battler-tag.ts
flx-sta: Despair. And here I naively was going for the less redundant approach

DayKev: Yeah, it's... like, I kinda get it, but it is annoying.
For mixed type/non-type imports it's still fine to combine them though.

PR #1003, opened 2025-04-18
[Refactor] Consolidate constants files into `src/constants/`

### LINKED ISSUE #998: Move all constants files into `src/constants/`
# See #1003

DayKev:
N/A

Resolves #998

Moved scattered constants files to `src/constants/` and split apart `constants.ts` into multiple files.
Some TSDocs were added/updated as well.

## Checklist


src/constants/app.ts
```diff
@@ -0,0 +1,44 @@
...
+export const bypassLogin = import.meta.env.VITE_BYPASS_LOGIN === "1";
+
+/** Key for the local storage item storing the settings. */
+export const SETTINGS_LS_KEY = `${LS_PREFIX}/settings`;
+
+/** Key for the local storage item storing the input mapping configs. */
+export const MAPPING_CONFIG_LS_KEY = `${LS_PREFIX}/mapping/configs`;
+
+/** The ratio at which PRSFX sound volumes are played is adjusted since they are sigificantly louder. */
+export const PRSFX_SOUND_ADJUSTMENT_RATIO = 0.5;
```
flx-sta, line 35: Should be a `game` constant (?)

src/constants/app.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/** The maximum number of language options to display simultaneously. */
+export const LANGUAGE_MAX_OPTIONS = 7;
```
flx-sta: Should be a `ui` constant (?)

src/constants/app.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/** All available game speeds. */
+export const GAME_SPEEDS = [1, 1.25, 1.5, 2, 2.5, 3, 4, 5] as const;
```
flx-sta: Should be a `game` constant (?)

flx-sta: And the array should probably be frozen, instead of `as const` (if that's possible)

src/constants/game.ts
```diff
@@ -0,0 +1,63 @@
...
+/** The damage multiplier applied by Reflect, Light Screen, and Aurora Veil in single battles.*/
+export const SCREEN_SINGLES_DMG_FACTOR = 0.5;
+
+/** The damage multiplier applied by Reflect, Light Screen, and Aurora Veil in double battles.*/
+export const SCREEN_DOUBLES_DMG_FACTOR = 2732 / 4096;
+
+/**
+ * A list of all Gigantamax form keys, excluding Eternamax.
+ */
+export const G_MAX_FORM_KEYS: readonly string[] = Object.freeze([
```
flx-sta: NIT: 
```suggestion
export const G_MAX_FORM_KEYS  = Object.freeze<string[]>([
```

src/constants/mystery-encounters.ts
```diff
@@ -0,0 +1,50 @@
+/** Min - Max waves for mystery encounter in classic mode. */
+export const CLASSIC_MODE_MYSTERY_ENCOUNTER_WAVES: [min: number, max: number] = [10, 180];
+
+/** Min - Max waves for mystery encounter in challenge mode. */
+export const CHALLENGE_MODE_MYSTERY_ENCOUNTER_WAVES: [min: number, max: number] = [10, 180];
```
flx-sta: We should freeze all arrays in constants

```suggestion
export const CHALLENGE_MODE_MYSTERY_ENCOUNTER_WAVES = Object.freeze<[min: number, max: number]>([10, 180]);
```

src/constants/mystery-encounters.ts
```diff
@@ -0,0 +1,50 @@
+/** Min - Max waves for mystery encounter in classic mode. */
+export const CLASSIC_MODE_MYSTERY_ENCOUNTER_WAVES: [min: number, max: number] = [10, 180];
```
flx-sta: We should freeze all arrays in constants
```suggestion
export const CLASSIC_MODE_MYSTERY_ENCOUNTER_WAVES = Object.freeze<[min: number, max: number]>([10, 180]);
```

flx-sta (changes requested):
I feel like the constants files should still have the `-const` or `-constant` suffix. This is especially helpful when searching for files as the path is not always super prominent immediately.
e.g. I use the `Cmd + P` to search for my files by their name and it would help me a lot. But I'm open for other opinions.

flx-sta:
> [!NOTE]
> @nightkev could you also move `PRIMAL_WEATHER` into some kind of constant or utils file in your constant refactor PR?
and rename it to `PRIMAL_WEATHER_TYPES`?
(currently it's in `src/data/weaather.ts`)

src/constants/game.ts
```diff
@@ -0,0 +1,70 @@
...
+]);
+
+/** Default amount of money the player starts with. Same for all game modes. */
+export const DEFAULT_STARTING_MONEY = 1000;
+
+/**
+ * Weather types that are associated with the primal forms of the Generation III cover legendaries
+ * and cannot be overwritten by weaker weather types.
+ */
+export const PRIMAL_WEATHER = Object.freeze([WeatherType.HARSH_SUN, WeatherType.HEAVY_RAIN, WeatherType.STRONG_WINDS]);
```
flx-sta: NIT: Use `PRIMAL_WEATHER_TYPES` to be more precisely named. And add explicit typing
```suggestion
export const PRIMAL_WEATHER_TYPES = Object.freeze<WeatherType[]>([WeatherType.HARSH_SUN, WeatherType.HEAVY_RAIN, WeatherType.STRONG_WINDS]);
```

flx-sta (dismissed):

src/constants/app.ts
Tempo-anon: Can also rename PRSFX

innerthunder: I feel like this is a fine place for this. `game` constants are more about in-game properties (i.e. Pokemon mechanics)

innerthunder (commented):
Seems good, pending flx's comments.

src/constants/game.ts
DayKev: Renamed. The typing isn't necessary since it's already properly inferred:
image

PigeonBar (approved):

DayKev:
@flx-sta Should we make a poll about the filenames or is it okay to leave them without `-constants` appended?

flx-sta:
> @flx-sta Should we make a poll about the filenames or is it okay to leave them without `-constants` appended?

We can make a poll but that shouldn’t stop this PR

innerthunder (approved):

Issue #946, opened 2025-04-11
[EPIC] Move and Ability Implementations


flx-sta:
EPIC for finishing up Move/Ability Implementations

[GDocs -> Move/Ability implementation notes](https://docs.google.com/spreadsheets/d/1TpQ2YhKYOgeYeEi2_a6EyE60pcr3yhqxjYPgSQwB7ko/edit?usp=sharing)

> [!NOTE]
> The list below only includes implementations that are marked unimplemented (N) or incomplete (P). Moves and abilities with documented edge cases are omitted.

-------

### Ready
- [x] Bide (N) #1011 
- [ ] Ally Switch (N) #1023 
- [x] Healing Wish / Lunar Dance (P) #1050 

### Dependent on Item Refactor #928 
- [ ] Trick / Switcheroo (N) #1024 
- [ ] Recycle (N) #1025 
- [ ] Natural Gift (N) #1026 
- [ ] Embargo (N) #1027 
- [ ] Magic Room (N) #1028 
- [ ] Fling (N) #1029 
- [ ] Corrosive Gas (N) #1030 
- [ ] G-max Replenish (P) #1031 

### Dependent on `TurnCommandManager` changes #940 
- [ ] Pursuit (P) #1032 

### Requires changes to ability application methods
- [ ] Future Sight / Doom Desire (P) #1033 

### Dependent on **verdict** on #252 
- [ ] Rage Fist (P) #1034 
- [ ] Last Respects (P) #1035 

-------

### Ready
- [ ] Flower Veil (N) #1036 
- [ ] Flower Gift (P) #1037 
- [ ] Victory Star (P) #1038 
- [ ] Sweet Veil (P) #1039 
- [ ] Shields Down (P) #1040 
- [ ] Protosynthesis / Quark Drive (P) #1041 

### Dependent on Item Refactor #928 
- [ ] Klutz (N) #1042 
- [ ] Symbiosis (N) #1043 
- [ ] Cud Chew (N) #1044 

### Requires changes to ability application methods
- [ ] Neutralizing Gas (P) #1045 

### Dependent on Form Change refactor
- [x] Embody Aspect (P) #1046 

### Dependent on AI Rework #945 
- [ ] Illusion (N) #1047 

### Dependent on verdict on #252
- [ ] Supreme Overlord (P) #1048

PR #1002, opened 2025-04-18
[Check][Ability] Add Solar Power Ability Checkup

flx-sta:
## What are the changes the user will see?

n/a

## Why am I making these changes?

Closes #686 

## What are the changes from a developer perspective?

- Add test-coverage for Solar Power Ability
- Add documentation in `WeatherType`
- add `weather-utils`
- Add documentation in `PostWeatherLapseDamageAbAttr`
- Add new `NEW_WEATHER_DURATION_OVERRIDE` override to control the duration of weathers (not primal!)
- Add custom matchers:
    - `toHaveStat`,
    - `toHaveEffectiveStat`
    - `toHaveTakenDamage`
    - `toHaveWeather`

## Screenshots/Videos

n/a

## How to test the changes?

Run the automated tests

## Checklist

- [x] Otherwise: **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test:silent`)
  - [x] Have I created new automated tests (`npm run test:create`) or updated existing tests related to the PR's changes?

PR #1011, opened 2025-04-20
[Move] Partially Implement Bide

innerthunder:
> [!IMPORTANT]
> **This PR updates the locales** 

Bide is partially implemented. It is functionally different from mainline in that
- Its execution is completely cancelled when any of its turns are interrupted by any effect (e.g. Paralysis, flinching). In mainline, only Sleep has this effect on the move.
- Its priority is calculated for each individual turn instead of preserving the calculated priority from the first turn.

Part of #946 

📈 

Added:
- A new battler tag `BideTag` which forces the Pokemon to use Bide for 3 turns and stores received attack damage during Bide's execution
- 4 new attributes to implement Bide's characteristics
  - `BideEffectAttr`: controls which effects are applied (animations, damage, etc) for each turn of Bide's execution
  - `BideDamageAttr`: resolves the damage dealt by Bide on its final turn of execution
  - `BideMessageAttr` (and parent `VariableMoveMessageAttr`): may override the default message `"{pokemon} used {move}!"` with a custom message (in Bide's case, the messages depend on the turn of the move's execution).
- Resolved TODO in Mimic and Sketch's condition (WIP)

## How to test the changes?

`npm run test bide` (WIP)

## Checklist

  - [x] If so, please leave a link to it here: https://github.com/Despair-Games/poketernity-locales/pull/45
- [x] Have I added the `Localization` tag to this PR?


src/data/battler-tags/bide-tag.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * On the last turn of execution, Bide attacks the last Pokemon that attacked the user,
+   * or a random near enemy if no such Pokemon exists.
+   * @param pokemon the {@linkcode Pokemon} with this tag
+   * @param move the {@linkcode Move} to be used (in this case, Bide)
+   * @returns an array containing the target's {@linkcode BattlerIndex}
+   */
+  protected override getNextTargets(pokemon: Pokemon, move: Move): BattlerIndex[] {
+    if (this.turnCount > 1) {
```
Tempo-anon, line 50: Out of scope for this PR, but this makes me think this variable should probably be renamed `turnsRemaining` or something since most cases have this as a counter that ticks down

src/data/battler-tags/bide-tag.ts
```diff
@@ -0,0 +1,62 @@
...
+    this.lastAttackerId = attacker.id;
+    this.damageReceived += damage;
+  }
+
+  /**
+   * On the last turn of execution, Bide attacks the last Pokemon that attacked the user,
+   * or a random near enemy if no such Pokemon exists.
+   * @param pokemon the {@linkcode Pokemon} with this tag
+   * @param move the {@linkcode Move} to be used (in this case, Bide)
+   * @returns an array containing the target's {@linkcode BattlerIndex}
```
Tempo-anon: Should mention that returns the user's index if the move is still charging

src/data/init/init-moves.ts
```diff
@@ -621,7 +624,14 @@ export function initMoves() {
...
       .target(MoveTarget.USER)
-      .unimplemented(),
+      .attr(BideEffectAttr)
+      .attr(BideDamageAttr)
+      .attr(BideMessageAttr)
+      /**
+       * - Does not preserve original priority throughout execution
+       * - Is cancelled completely when interrupted by any effect (not just Sleep)
+       */
+      .edgeCase(),
```
Tempo-anon: In my mind the difference between a `partial` and `edgeCase` is if a player will likely see the anomaly. On one hand, interruption effects seem to be a common occurrence. On the other hand, nobody is realistically using bide so I think this can go either way.

src/data/moves/move-attrs/bide-effect-attr.ts
```diff
@@ -0,0 +1,48 @@
...
+   * Bide fails on its last turn of execution if the user
+   * has not received any attack damage since it was first used.
+   */
+  public override getCondition(): MoveConditionFunc {
+    return (user: Pokemon) => {
+      const bideTag = user.getTag<BideTag>(BattlerTagType.BIDE);
+      if (bideTag) {
+        return bideTag.turnCount > 1 || bideTag.attackDamage > 0;
+      }
+      return true;
```
Tempo-anon, line 46: The default move condition is true if the user doesn't have the tag?

src/data/moves/move-attrs/variable-move-message-attr.ts
```diff
@@ -0,0 +1,34 @@
...
+import type { Move } from "#app/data/moves/move";
+import { MoveAttr } from "#app/data/moves/move-attrs/move-attr";
+import type { Pokemon } from "#app/field/pokemon";
+
+/**
+ * Attribute to override the default move usage message (e.g. "Pikachu used Thunderbolt!")
+ * with a custom message which can change based on game state.
+ * @extends MoveAttr
+ */
+export abstract class VariableMoveMessageAttr extends MoveAttr {
```
Tempo-anon, line 11: Is this only for bide? Or are there other moves that will also be using this?

src/data/moves/move-attrs/bide-effect-attr.ts
```diff
@@ -0,0 +1,48 @@
...
+ * - Cancelling Bide's damage on all but the last turn of execution
+ * @extends OverrideMoveEffectAttr
+ */
+export class BideEffectAttr extends OverrideMoveEffectAttr {
+  public override apply(user: Pokemon, _target: Pokemon | null, _move: Move, overridden: BooleanHolder): boolean {
+    const bideTag = user.getTag<BideTag>(BattlerTagType.BIDE);
+
+    if (!bideTag || bideTag.turnCount > 1) {
+      // Add Bide's tag to the user if it doesn't exist yet
+      user.addTag(BattlerTagType.BIDE);
```
Tempo-anon: The `if` seems to attempt to add the tag even if the tag already exists so that doesn't match the comment

src/data/moves/move-attrs/bide-effect-attr.ts
innerthunder: Yes, the user doesn't have the tag on the move's first turn

src/data/battler-tags/bide-tag.ts
innerthunder: `turnsLeft`? Maybe not the best fit since the counter doesn't always correspond with turns (see a lot of the `CUSTOM` lapsed tags)

src/data/moves/move-attrs/variable-move-message-attr.ts
innerthunder: Seems like it'll only be for Bide, looking at the implementations left to do. I just wanted to maintain flexibility in case some new move happens to require this in the future.

src/data/battler-tags/bide-tag.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  /**
+   * Updates Bide's future target and records damage after receiving an attack.
+   * @param attacker - The {@linkcode Pokemon} that attacked the tag owner
+   * @param damage - The damage dealt to the tag owner
+   */
+  public updateAttackData(attacker: Pokemon, damage: number): void {
+    this.lastAttackerId = attacker.id;
+    this.damageReceived += damage;
```
flx-sta, line 38: Is damage guaranteed to be a damage value ?

innerthunder: Yes, it's obtained from the results of `MoveEffectPhase#applyMove` and is updated alongside other data related to damage.

test/moves/bide.test.ts
```diff
@@ -0,0 +1,154 @@
...
+          move: allMoves.get(MoveId.BIDE),
+          ignorePP: true,
+        }),
+      );
+      expect(enemy.isFullHp()).toBeTruthy();
+    }
+
+    const receivedDamage = player.getInverseHp();
+    // For test stability; we don't want the enemy to faint from Bide's attack
+    expect(receivedDamage).toBeLessThan(enemy.getMaxHp() / 2);
```
Tempo-anon, line 64: Is there a way to spy on how much damage is dealt? Then it wouldn't matter if the enemy faints or not (though in this case of a Chansey doing 2 tackles in terms of damage I don't think that will happen)

test/moves/bide.test.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    await game.classicMode.startBattle([SpeciesId.AGGRON]);
+
+    const player = game.field.getPlayerPokemon();
+    const enemy = game.field.getEnemyPokemon();
+
+    game.move.use(MoveId.BIDE);
+
+    for (let i = 0; i < 3; i++) {
+      // Attacking more than once can cause the player to nearly faint
```
Tempo-anon: Can use moves like dragon rage or seismic toss which do guaranteed damage values

test/moves/bide.test.ts
flx-sta: There will be with my `Solar Power Checkup` PR #1002 

https://github.com/Despair-Games/poketernity/pull/1002/files#diff-0e6d6285d5324ee0f5b8c7d14f52e3a4f6afd0167c76fdcbc4b6033b6b8d41f3

test/moves/bide.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    for (let i = 0; i < 2; i++) {
+      await game.toEndOfTurn();
+      expect(player).toHaveMoveResult(MoveResult.SUCCESS);
+    }
+
+    await game.toEndOfTurn();
+    expect(player).toHaveMoveResult(MoveResult.FAIL);
+  });
+
+  it("should stop execution if the user falls asleep", async () => {
```
Tempo-anon, line 90: Should there be some similar tests with todo for the complicated cases where the user is interrupted in other ways?

flx-sta (approved):

Tempo-anon (approved):

src/data/moves/move-conditions/target-move-copiable-condition.ts
```diff
@@ -20,6 +22,13 @@ export const targetMoveCopiableCondition: MoveConditionFunc = (_user, target, _m
     return false;
   }
 
-  // TODO: Add last turn of Bide
+  /**
+   * Bide can only be copied after it fully executes
+   * @todo Verify this interaction (on Showdown?)
```
PigeonBar, line 27: According to Showdown, it is possible to copy Bide on any turn, including before it fully executes.

20250426 Copiable Bide

PR #1013, opened 2025-04-21
[Refactor] Remove `any` from `PokemonData` constructor

DayKev:
> [!TIP]
> It's best to view the changes with the "[ignore whitespace](https://github.com/Despair-Games/poketernity/pull/1013/files?diff=unified&w=1)" option enabled

N/A

The `PokemonData` constructor could use some cleaning up.

- The constructor of `PokemonData` now has the `source` parameter typed as `Pokemon | PokemonData`.
- Remove `sourcePokemon` variable and replace with new utility function `isPokemon` in the couple of places it's necessary.
- Replace `x !== undefined ? y : z` with `??`.

## How to test the changes?
Load some session save data.

## Checklist

### DIFF
--- a/src/system/game-data.ts
+++ b/src/system/game-data.ts
@@ -1107,6 +1107,10 @@ export class GameData {
           v = [];
         }
         for (const pd of v) {
+          // TODO: remove this later
+          if (pd.hasOwnProperty("species")) {
+            pd.speciesId = pd.species;
+          }
           ret.push(new PokemonData(pd));
         }
         return ret;
--- a/src/system/game-data.ts
+++ b/src/system/game-data.ts
@@ -1107,7 +1107,8 @@ export class GameData {
           v = [];
         }
         for (const pd of v) {
-          // TODO: remove this later
+          // TODO: remove later, temporary to prevent devs from needing to wipe their local storage
+          // due to the field in `PokemonData` being renamed from `species` to `speciesId`
           if (pd.hasOwnProperty("species")) {
             pd.speciesId = pd.species;
           }
--- a/src/system/pokemon-data.ts
+++ b/src/system/pokemon-data.ts
@@ -2,17 +2,17 @@ import { loadBattlerTag } from "#app/data/battler-tags/utils/load-battler-tag";
 import { CustomPokemonData } from "#app/data/custom-pokemon-data";
 import { Status } from "#app/data/status-effect";
 import type { Variant } from "#app/data/variant";
-import { EnemyPokemon } from "#app/field/enemy-pokemon";
 import type { Pokemon } from "#app/field/pokemon";
 import { PokemonMove } from "#app/field/pokemon-move";
 import { PokemonSummonData } from "#app/field/pokemon-summon-data";
 import { globalScene } from "#app/global-scene";
+import { isPokemon } from "#app/utils";
 import { getPokemonSpecies } from "#app/utils/pokemon-species-utils";
 import { BattleType } from "#enums/battle-type";
 import type { BiomeId } from "#enums/biome-id";
 import type { Gender } from "#enums/gender";
 import { MoveId } from "#enums/move-id";
-import type { Nature } from "#enums/nature";
+import { Nature } from "#enums/nature";
 import type { PokeballType } from "#enums/pokeball-type";
 import type { SpeciesId } from "#enums/species-id";
 import { TrainerSlot } from "#enums/trainer-slot";
@@ -20,7 +20,7 @@ import { TrainerSlot } from "#enums/trainer-slot";
 export default class PokemonData {
   public id: number;
   public player: boolean;
-  public species: SpeciesId;
+  public speciesId: SpeciesId;
   public nickname: string;
   public formIndex: number;
   public abilityIndex: number;
@@ -37,7 +37,7 @@ export default class PokemonData {
   public ivs: number[];
   public nature: Nature;
   public moveset: PokemonMove[];
-  public status: Status | null;
+  public status: Status | null = null;
   public friendship: number;
   public metLevel: number;
   public metBiome: BiomeId | -1; // -1 for starters
@@ -49,7 +49,7 @@ export default class PokemonData {
   public usedTMs: MoveId[];
   public evoCounter: number;
 
-  public boss: boolean;
+  public boss: boolean = false;
   public bossSegments?: number;
 
   public summonData: PokemonSummonData;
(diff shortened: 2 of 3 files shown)


src/system/pokemon-data.ts
```diff
@@ -49,22 +49,20 @@ export default class PokemonData {
...
 
   public summonData: PokemonSummonData;
 
   /** Data that can customize a Pokemon in non-standard ways from its Species */
   public customPokemonData: CustomPokemonData;
 
-  // TODO: this shouldn't be `| any`
-  constructor(source: Pokemon | any) {
-    const sourcePokemon = source.type === "Pokemon" ? source : null;
+  constructor(source: Pokemon | PokemonData) {
```
innerthunder, line 64: This seems weird. Why would we want to construct a `PokemonData` out of a `PokemonData`? It seems like the compatibility between the source and the new `PokemonData` isn't 1:1, so shouldn't this use a separate interface?

DayKev: The source `PokemonData` is from the session save data. Handling of save data could use a major overhaul but this is a minor refactor until then.

src/system/pokemon-data.ts
```diff
@@ -77,61 +75,66 @@ export default class PokemonData {
...
-    this.pokerus = !!source.pokerus;
+    this.pokerus = source.pokerus;
     this.usedTMs = source.usedTMs ?? [];
 
     this.customPokemonData = new CustomPokemonData(source.customPokemonData);
 
-    this.boss = (source instanceof EnemyPokemon && source.bossSegments > 0) || (!this.player && !!source.boss);
-    this.bossSegments = source.bossSegments;
+    if (source.hasOwnProperty("bossSegments")) {
+      // @ts-expect-error
```
flx-sta: NIT: `@ts-expect-error` should have a description behind it why it's expected

src/system/pokemon-data.ts
```diff
@@ -77,61 +75,66 @@ export default class PokemonData {
...
+      this.boss = source.bossSegments > 0;
+      // @ts-expect-error
```
flx-sta: NIT: `@ts-expect-error` should have a description behind it why it's expected

src/system/pokemon-data.ts
```diff
@@ -77,61 +75,66 @@ export default class PokemonData {
...
 
-        if (source.summonData.tags) {
-          this.summonData.tags = source.summonData.tags?.map((t) => loadBattlerTag(t));
-        } else {
-          this.summonData.tags = [];
-        }
-      }
+    this.summonData = new PokemonSummonData();
+    if (!source.summonData) {
+      return;
```
flx-sta, line 124: Q: Is this an expected exit point?

src/utils.ts
```diff
@@ -684,3 +684,7 @@ export function deepFreeze<T>(obj: T): Readonly<T> {
   }
   return obj;
 }
+
+export function isPokemon(data: any): data is Pokemon {
```
flx-sta, line 689: NIT: ts-doc

src/system/game-data.ts
```diff
@@ -1107,6 +1107,10 @@ export class GameData {
           v = [];
         }
         for (const pd of v) {
+          // TODO: remove this later
```
Tempo-anon: Would also help if this todo had a bit more context. Why is this needed? Will it be done in a follow up PR soon? Is it waiting on some other change?

src/system/pokemon-data.ts
```diff
@@ -2,25 +2,25 @@ import { loadBattlerTag } from "#app/data/battler-tags/utils/load-battler-tag";
...
+import { Nature } from "#enums/nature";
 import type { PokeballType } from "#enums/pokeball-type";
 import type { SpeciesId } from "#enums/species-id";
 import { TrainerSlot } from "#enums/trainer-slot";
 
 export default class PokemonData {
   public id: number;
   public player: boolean;
-  public species: SpeciesId;
+  public speciesId: SpeciesId;
```
Tempo-anon, line 23: 👍

src/system/pokemon-data.ts
Tempo-anon: Would like a comment or something explaining that then. A class constructor taking in itself as special handling feels weird

src/system/pokemon-data.ts
DayKev: Yes, if there's no summon data to assign then we can skip the summon data assignment.

src/system/game-data.ts
DayKev: Added a note.

src/system/pokemon-data.ts
DayKev: Added a comment.

flx-sta (approved):

Tempo-anon (approved):

src/system/pokemon-data.ts
```diff
@@ -77,61 +79,66 @@ export default class PokemonData {
...
     this.evoCounter = source.evoCounter ?? 0;
-    this.pokerus = !!source.pokerus;
+    this.pokerus = source.pokerus;
     this.usedTMs = source.usedTMs ?? [];
 
     this.customPokemonData = new CustomPokemonData(source.customPokemonData);
 
-    this.boss = (source instanceof EnemyPokemon && source.bossSegments > 0) || (!this.player && !!source.boss);
-    this.bossSegments = source.bossSegments;
+    if (source.hasOwnProperty("bossSegments")) {
```
PigeonBar, line 96: `if (!isPokemon(source) || source.isEnemy()) {` can be used here to resolve the `@ts-expect-error` statements (it also assumes that player Pokemon are not allowed to have boss segments).

src/system/pokemon-data.ts
```diff
@@ -77,61 +79,66 @@ export default class PokemonData {
...
-        this.summonData.abilitySuppressed = source.summonData.abilitySuppressed;
-        this.summonData.abilitiesApplied = source.summonData.abilitiesApplied;
+      return;
+    }
 
-        this.summonData.ability = source.summonData.ability;
-        this.summonData.moveset = source.summonData.moveset?.map((m) => PokemonMove.loadMove(m));
-        this.summonData.types = source.summonData.types;
+    // TODO: is this `.map()` needed? why?
+    this.moveset = (source.moveset || [new PokemonMove(MoveId.TACKLE), new PokemonMove(MoveId.GROWL)]).map(
```
PigeonBar, line 113: Based on some testing, the `.map()` is needed because the `PokemonMove` data stored in save files are not full `PokemonMove` objects (in particular, they are missing `PokemonMove`'s functions, which causes a crash), so `new PokemonMove()` must be called to fix this.

src/system/pokemon-data.ts
```diff
@@ -77,61 +79,66 @@ export default class PokemonData {
...
+      (m) => new PokemonMove(m.moveId, m.ppUsed, m.ppUp, m.virtual, m.maxPpOverride),
+    );
+    // TODO: can this just be `this.status = source.status`?
```
PigeonBar, line 116: Similarly, `new Status()` must be called to convert the save data into a full `Status` object.

PR #1063, opened 2025-04-25
[Test] Fix test matchers' behaviour when receiving wrong object type

PigeonBar:

When using the newly introduced matchers (see `matchers.setup.ts`), if `expect(object).not` is used and the object is of the wrong type, the test now correctly fails.

I just realized that since `expect().not` inverts the `pass` result, returning `pass: false` when the object is of the wrong type causes the test to pass when it should not.

According to the [vitest docs](https://vitest.dev/guide/extending-matchers), `this.isNot` can be used to tell whether `expect().not` was used. So, if the test needs to fail due to the received object being the wrong type, it now returns `pass: false` if `this.isNot === false` and returns `pass: true` if `this.isNot === true`.

## How to test the changes?

`expect(undefined).not.toHaveFullHp(); // Test should fail`

## Checklist


test/test-utils/testUtils.ts
```diff
@@ -76,7 +76,7 @@ export function receivedStr(received: unknown, expectedType: TypeOfResult = "obj
     what = received.constructor.name;
   }
 
-  return `but got ${what}!`;
+  return `${what}`;
```
Tempo-anon: Removing the`but got`?

Also not a fan of the `what` name. Maybe something like `receivedObject` or `resultStr`?

Tempo-anon (dismissed):

test/test-utils/testUtils.ts
PigeonBar: The `but got` is removed here because it already appears in all of the matcher files.

test/test-utils/testUtils.ts
```diff
@@ -64,19 +64,17 @@ export function getAppRootDir() {
...
+    return "undefined";
   } else if (typeof received !== expectedType) {
-    what = typeof received;
+    return typeof received;
   } else if (expectedType === "object") {
-    what = received.constructor.name;
+    return received.constructor.name;
+  } else {
+    return "unknown";
   }
```
DayKev:
```suggestion
  }

  return "unknown";
```

test/test-utils/testUtils.ts
```diff
@@ -64,19 +64,17 @@ export function getAppRootDir() {
  * @returns A human readable string of the received object (type)
  */
 export function receivedStr(received: unknown, expectedType: TypeOfResult = "object"): string {
-  let what = "unknown";
-
   if (received === null) {
-    what = "null";
+    return "null";
   } else if (received === undefined) {
```
DayKev:
```suggestion
  }
  if (received === undefined) {
```

test/test-utils/testUtils.ts
```diff
@@ -64,19 +64,17 @@ export function getAppRootDir() {
...
-    what = "undefined";
+    return "undefined";
   } else if (typeof received !== expectedType) {
```
DayKev:
```suggestion
  }
  if (typeof received !== expectedType) {
```

test/test-utils/testUtils.ts
```diff
@@ -64,19 +64,17 @@ export function getAppRootDir() {
...
-    what = typeof received;
+    return typeof received;
   } else if (expectedType === "object") {
```
DayKev:
```suggestion
  }
  if (expectedType === "object") {
```

flx-sta:
@PigeonBar would you mind adding the missing response types to
The matchers where I forgot to add it ?

flx-sta:
EDIT: Provided those changes as addendum:
#1067 

-----

<details><summary>[Original Comments]</summary>
<p>

> @PigeonBar would you mind adding the missing response types to
> The matchers where I forgot to add it ?

> We should also add:
> ```ts
> import type { MatcherState, [...] } from "@vitest/expect";
> 
> export function <MATCHER_NAME>(
>   this: MatcherState,
> ```
>
> To every matcher.
> It makes sure to have proper type hints for `this`, but the `this` doesn't have to be provided as extra param when calling. That's just how `expect.extend()` works. Just gotta make sure to not include the `this` param in the vitest.d.ts definition 

</p>
</details>

flx-sta (approved):

Tempo-anon (approved):

PR #1064, opened 2025-04-25
[Refactor] Minor refactoring in exp

Tempo-anon:

None

Small refactor to prep for a larger change

* Made the 1.2 factor in exp a constant
* Replaced the hardcoded trainer `baseLevel` with the function that was created for it to reduce duplicated code

## Screenshots/Videos

n/a

## How to test the changes?

`npm run start`

## Checklist

### DIFF
--- a/src/field/trainer.ts
+++ b/src/field/trainer.ts
@@ -1,4 +1,6 @@
+import { LEVEL_CAP_SCALE_FACTOR } from "#app/constants/game";
 import { EntryHazardTag } from "#app/data/arena-tag";
+import { getLevelForWaveFunc } from "#app/data/exp";
 import { pokemonPreEvolutions } from "#app/data/pokemon-pre-evolutions";
 import type PokemonSpecies from "#app/data/pokemon-species";
 import { signatureSpecies } from "#app/data/signatureSpecies";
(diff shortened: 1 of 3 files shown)


src/field/trainer.ts
```diff
@@ -281,7 +284,7 @@ export default class Trainer extends Phaser.GameObjects.Container {
           multiplier = 1.1;
           break;
         case PartyMemberStrength.STRONG:
-          multiplier = 1.2;
+          multiplier = LEVEL_CAP_SCALE_FACTOR;
```
flx-sta: Is this as correct use-case for it? Seems a bit out of place unless I’m missing the context

DayKev (dismissed):

src/field/trainer.ts
DayKev: Huh, yeah, is this a mistake?

Tempo-anon: the STRONG modifier is usually reserved for important trainer Pokemon so they will be at 1.2x the level of wild encounters (the level cap). I can remove it from here if we want to later change the scale factor

flx-sta: I se. Yeah probably reasonable. Sorry for the Hickhack

flx-sta (approved):

src/field/trainer.ts
DayKev: Ah so it intentionally matches the level cap here. Is the `STRONGER` value used for anything?

DayKev (approved):

src/field/trainer.ts
Tempo-anon: It's used in certain Psychics (the trainer class), gym leaders, elite four (which evilboss2 also uses), champion, and rival 6. This is why those trainers will actually have Pokemon past the player level cap.

DayKev: Interesting. It might be good to make them all relative to the level cap constant then (for example `STRONGER` becomes `multiplier = LEVEL_CAP_SCALE_FACTOR + 0.05`) later.

PR #1065, opened 2025-04-25
[Dev] Add `ban-ts-comment` TypeScript-ESLint rule

DayKev:
N/A

`@ts-expect-error` should be used instead of `@ts-ignore`.

Added ESLint rule `@typescript-eslint/ban-ts-comment` to disallow usage of `@ts-ignore`.
Uses of `@ts-expect-error` are also required to have a comment explaining their purpose now (min length 3).

## Checklist


flx-sta:
Before anyone asks why we didn't use this before:
`ts-expect-error` is fairly new

flx-sta (approved):

DayKev:
It doesn't seem to be that new, it was in as early as [TS 3.9](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments).
image

flx-sta:
> It doesn't seem to be that new, it was in as early as [TS 3.9](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments). image

*Pushed for in a rather new manner

Tempo-anon (approved):

PR #1066, opened 2025-04-25
[Refactor] Phaser stuff cleanup

flx-sta:

n/a

- One unnecessary src level directory less
- cleaner structure

- move `typings/phaser/index.d.ts` into `@types` and rename it `phaser.d.ts`
- move phaser class prototype extension into `phaser-extension.ts`
    - Include file in `main.ts`
    - Include file in `test/test-utils/testFileInitialization.ts`
- Make `setRealativePosition` return `this` 

## Screenshots/Videos

n/a

## How to test the changes?

- Run the tests
- Run the game

Everything should feel the same

## Checklist


SirzBenjie (approved):
This is phenomenal. The QoL from `setPositionRelative` now returning `this` is very much appreciated.

test/test-utils/testFileInitialization.ts
```diff
@@ -79,24 +80,8 @@ export function initTestFile() {
...
-   * @param guideObject {@linkcode Phaser.GameObjects.GameObject} to base the position off of
-   * @param x The relative x position
-   * @param y The relative y position
-   */
-  const setPositionRelative = function (guideObject: any, x: number, y: number) {
-    const offsetX = guideObject.width * (-0.5 + (0.5 - guideObject.originX));
-    const offsetY = guideObject.height * (-0.5 + (0.5 - guideObject.originY));
-    this.setPosition(guideObject.x + offsetX + x, guideObject.y + offsetY + y);
-  };
+  // await import("#app/phaser-extensions");
```
DayKev:
```suggestion
```
Leftover from testing?

src/@types/phaser.d.ts
DayKev, line 1: Can we use this to override the type hints of unchanged Phaser functions/etc too?

src/phaser-extensions.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Positions this object relative to the {@linkcode guideObject}.
+ * @param guideObject - The object to base the position off of
+ * @param x - The relative x position
+ * @param y - The relative y position
+ * @returns The positioned instance of {@linkcode T}
+ */
+function setPositionRelative<T extends Phaser.GameObjects.GameObject>(
+  guideObject: Phaser.GameObjects.GameObject,
```
DayKev: Should this be `: T`?

test/test-utils/testFileInitialization.ts
Tempo-anon: Is this piece of commented code needed?

flx-sta: yup. Thanks

src/phaser-extensions.ts
flx-sta: You mean the `guideObjects`? No, this can be something else

src/@types/phaser.d.ts
flx-sta: @DayKev you have an example

flx-sta: Foner

flx-sta: Done. `Tweens.Tween.getValue()` now notes a possible `null` as return value

src/data/mystery-encounters/utils/encounter-pokemon-utils.ts
```diff
@@ -520,7 +520,7 @@ export function trainerThrowPokeball(
                   repeatDelay: 500,
                   onUpdate: (t) => {
                     if (shakeCount && shakeCount < 4) {
-                      const value = t.getValue();
+                      const value = t.getValue() ?? 1;
```
innerthunder:
```suggestion
                      const value = t.getValue() ?? 0;
```
Makes more sense imo, since this affects x-position relative to the poke ball's original position

src/phases/attempt-capture-phase.ts
```diff
@@ -137,7 +137,7 @@ export class AttemptCapturePhase extends PokemonPhase {
                 repeatDelay: 500,
                 onUpdate: (t) => {
                   if (shakeCount && shakeCount < (isCritical ? 2 : 4)) {
-                    const value = t.getValue();
+                    const value = t.getValue() ?? 1;
```
innerthunder:
```suggestion
                    const value = t.getValue() ?? 0;
```
Same here (is this just recycled code?)

DayKev: Yeah MEs copy/pasted the capture code.

src/@types/phaser.d.ts
```diff
@@ -0,0 +1,98 @@
...
+         * This is called automatically when a gamepad is connected or disconnected, and during the update loop.
+         */
+        refreshPads(): void;
+      }
+    }
+  }
+
+  namespace Tweens {
+    interface Tween {
+      getValue(index?: number): number | null;
```
MokaStitcher, line 99: Where is this coming from?

DayKev: Phaser's return type hint for this function is incorrect, so we're manually fixing it.

MokaStitcher: maybe we should add a comment to explain that, so that when/if they fix it we know we can remove it

src/phaser-extensions.ts
```diff
@@ -0,0 +1,34 @@
...
+  const offsetX = guideObject.width * (-0.5 + (0.5 - guideObject.originX));
+  const offsetY = guideObject.height * (-0.5 + (0.5 - guideObject.originY));
+  this.setPosition(guideObject.x + offsetX + x, guideObject.y + offsetY + y);
+
+  return this;
+}
+
+//#endregion
+
+export default {};
```
MokaStitcher: what does this do?

MokaStitcher (dismissed):
The game won't run on my machine with the following error:
image

otherwise this looks good. This gives a good template on how add `setPositionRelative` to the rex plugin's `BBCodeText` and maybe `InputText` in the future too!

src/phaser-extensions.ts
flx-sta: Nothing anymore.
When I was experimenting with using `await import("#app/phaser-extensions.ts")` it would make sure that the compiler doesn't complain about the file not being a module, as it wasn't exporting anything.
Removed!

flx-sta:
> The game won't run on my machine with the following error: image
> 
> otherwise this looks good. This gives a good template on how add `setPositionRelative` to the rex plugin's `BBCodeText` and maybe `InputText` in the future too!

Fixed the missing phaser import.

src/phaser-extensions.ts
```diff
@@ -0,0 +1,56 @@
+import Phaser from "phaser";
+
+//#region Types
+
+type GuideObject = Phaser.GameObjects.Components.Origin &
+  Phaser.GameObjects.Components.Origin &
+  Phaser.GameObjects.Components.Size &
+  Phaser.GameObjects.Components.Transform;
```
DayKev:
```suggestion
// prettier-ignore
type GuideObject =
  & Phaser.GameObjects.Components.Origin
  & Phaser.GameObjects.Components.Size
  & Phaser.GameObjects.Components.Transform;
```

src/phaser-extensions.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Positions this object relative to the {@linkcode guideObject}.
+ * @param guideObject - The {@linkcode GuideObject} to base the position off of.
+ * @param x - The relative x position
+ * @param y - The relative y position
+ * @returns The positioned instance of {@linkcode T}
+ */
+function setPositionRelative<T extends Phaser.GameObjects.Components.Transform>(
+  this: T,
```
DayKev, line 31: Using `this` as the name of a param seems dangerous.

flx-sta: No, this is intended.
This help to have proper type-hints for `this`.
it's *auto injected*

src/phaser-extensions.ts
flx-sta: This would require adjusting prettier somehow. Otherwise it would always be undone

DayKev: Wait, prettier put the `&` at the end of the line? It shouldn't be doing that.

DayKev: Okay I updated the suggestion to add `// prettier-ignore` to this for now... apparently there's a bug in the prettier fork we're using if it's doing this. Since the feature we're using the fork for (which is literally putting stuff like `&` and `||` etc at the start of the line instead of the end) was finally pulled into upstream I'll see if updating to the latest upstream version of prettier fixes this.

DayKev: Okay test successful, upstream prettier has a fixed version of this. Will make a PR to update the deps to upstream Prettier later.

src/@types/phaser.d.ts
```diff
@@ -0,0 +1,97 @@
...
+      getByType<T extends Phaser.GameObjects.GameObject>(type: string): T | null;
+    }
+
+    interface Container {
+      /**
+       * Positions this object relative to the {@linkcode guideObject}.
+       * @param guideObject - The object to base the position off of
+       * @param x - The relative x position
+       * @param y - The relative y position
+       * @returns The positioned instance of {@linkcode T}
```
MokaStitcher: Is this doc correct? I know this is copied from the setPositionRelative generic function doc but it seems weird in this context since there's no T I think?

<sub> same with all the others below

MokaStitcher (dismissed):
much better :p

src/@types/phaser.d.ts
flx-sta: Good point. Fixed https://github.com/Despair-Games/poketernity/pull/1066/commits/53a6ca9a494d7c4698fbc16077e443ba594c57d7

DayKev (dismissed):

MokaStitcher (dismissed):
should we wait for #1070 so that we can remove the prettier ignore?

PR #1067, opened 2025-04-25
[Addendum][Tests] Add Matchers `this` type and return type

flx-sta:
Addendum for #1063


test/matchers/to-have-full-hp.ts
innerthunder, line 1: This should be renamed to `to-have-full-hp-matcher.ts`

PigeonBar (approved):

PR #1074, opened 2025-04-26
[Refactor] Consolidate arena tag functions

DayKev:
N/A

Reduce code redundancy, reduce import pollution.

- `Arena#getTagOnSide` and `Arena#findTagsOnSide` consolidated into `Arena#findTag` and `Arena#getTags`.
- Added TSDocs to `Arena#hasTag`.
- Replaced some instances of `findTag` with `hasTag` where it was being used as a boolean.
- `Arena#findTag` no longer accepts class constructors as a parameter.

## Checklist

### DIFF
--- a/src/data/init/init-abilities.ts
+++ b/src/data/init/init-abilities.ts
@@ -303,7 +303,7 @@ export function initAbilities() {
         AttackTypeImmunityAbAttr,
         ElementalType.GROUND,
         (pokemon: Pokemon) =>
-          !pokemon.getTag(BattlerTagType.IGNORE_FLYING) && !globalScene.arena.getTag(ArenaTagType.GRAVITY),
+          !pokemon.getTag(BattlerTagType.IGNORE_FLYING) && !globalScene.arena.findTag(ArenaTagType.GRAVITY),
       )
       .ignorable(),
     new Ability(AbilityId.EFFECT_SPORE, 3)
--- a/src/data/init/init-abilities.ts
+++ b/src/data/init/init-abilities.ts
@@ -303,7 +303,7 @@ export function initAbilities() {
         AttackTypeImmunityAbAttr,
         ElementalType.GROUND,
         (pokemon: Pokemon) =>
-          !pokemon.getTag(BattlerTagType.IGNORE_FLYING) && !globalScene.arena.findTag(ArenaTagType.GRAVITY),
+          !pokemon.getTag(BattlerTagType.IGNORE_FLYING) && !globalScene.arena.hasTag(ArenaTagType.GRAVITY),
       )
       .ignorable(),
     new Ability(AbilityId.EFFECT_SPORE, 3)
--- a/src/data/moves/move-attrs/delayed-attack-attr.ts
+++ b/src/data/moves/move-attrs/delayed-attack-attr.ts
@@ -56,7 +56,7 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
       // Add a Delayed Attack tag to the arena if it doesn't already exist
       globalScene.arena.addTag(ArenaTagType.DELAYED_ATTACK, user.id);
       // Queue an attack on the added (or existing) tag
-      const tag = globalScene.arena.getTag(ArenaTagType.DELAYED_ATTACK) as DelayedAttackTag;
+      const tag = globalScene.arena.findTag(ArenaTagType.DELAYED_ATTACK) as DelayedAttackTag;
       if (tag) {
         tag.addAttack(user, move.id, target.getBattlerIndex());
       }
@@ -75,7 +75,7 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
   /** Delayed attacks fail if another delayed attack is already queued against the target */
   override getCondition(): MoveConditionFunc {
     return (_user, target, _move) => {
-      const delayedAttackTag = globalScene.arena.getTag(ArenaTagType.DELAYED_ATTACK) as DelayedAttackTag;
+      const delayedAttackTag = globalScene.arena.findTag(ArenaTagType.DELAYED_ATTACK) as DelayedAttackTag;
       return !delayedAttackTag?.delayedAttacks.some((attack) => attack.targetIndex === target.getBattlerIndex());
     };
   }
--- a/src/data/moves/move-attrs/delayed-attack-attr.ts
+++ b/src/data/moves/move-attrs/delayed-attack-attr.ts
@@ -29,7 +29,8 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
   }
 
   /**
-   * If used virtually, this queues a message and proceeds normally.
+   * If used virtually, this queues a message and proceeds normally. (TODO: change this)
+   *
    * Otherwise, this adds a delayed attack to the field and cancels other move effects
    * for the current attack.
    */
@@ -56,10 +57,9 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
       // Add a Delayed Attack tag to the arena if it doesn't already exist
       globalScene.arena.addTag(ArenaTagType.DELAYED_ATTACK, user.id);
       // Queue an attack on the added (or existing) tag
-      const tag = globalScene.arena.findTag(ArenaTagType.DELAYED_ATTACK) as DelayedAttackTag;
-      if (tag) {
-        tag.addAttack(user, move.id, target.getBattlerIndex());
-      }
+      const tag = globalScene.arena.findTag<DelayedAttackTag>(ArenaTagType.DELAYED_ATTACK);
+      tag?.addAttack(user, move.id, target.getBattlerIndex());
+
       return true;
     } else {
       globalScene.phaseManager.queueMessagePhase(
@@ -75,7 +75,7 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
   /** Delayed attacks fail if another delayed attack is already queued against the target */
   override getCondition(): MoveConditionFunc {
     return (_user, target, _move) => {
-      const delayedAttackTag = globalScene.arena.findTag(ArenaTagType.DELAYED_ATTACK) as DelayedAttackTag;
+      const delayedAttackTag = globalScene.arena.findTag<DelayedAttackTag>(ArenaTagType.DELAYED_ATTACK);
       return !delayedAttackTag?.delayedAttacks.some((attack) => attack.targetIndex === target.getBattlerIndex());
     };
   }
--- a/src/data/moves/move-attrs/delayed-attack-attr.ts
+++ b/src/data/moves/move-attrs/delayed-attack-attr.ts
@@ -29,10 +29,12 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
   }
 
   /**
-   * If used virtually, this queues a message and proceeds normally. (TODO: change this)
+   * If used virtually, this queues a message and proceeds normally.
    *
    * Otherwise, this adds a delayed attack to the field and cancels other move effects
    * for the current attack.
+   *
+   * @todo Remove the virtual check, this behavior is incorrect and doesn't match mainline
    */
   override apply(user: Pokemon, target: Pokemon, move: Move, overridden: BooleanHolder, virtual: boolean): boolean {
     // Edge case for the move applied on a pokemon that has fainted
--- a/src/field/arena.ts
+++ b/src/field/arena.ts
@@ -858,7 +858,7 @@ export class Arena {
   getTags<T extends ArenaTag = ArenaTag>(
     tagPredicate: (t: ArenaTag) => boolean,
     side: ArenaTagSide = ArenaTagSide.BOTH,
-  ): T[] {
+  ): T[] | undefined {
     const validSides = new Set<ArenaTagSide>([ArenaTagSide.BOTH, side]);
 
     return this.tags.filter((t) => tagPredicate(t) && (side === ArenaTagSide.BOTH || validSides.has(t.side))) as T[];
--- a/src/field/arena.ts
+++ b/src/field/arena.ts
@@ -795,7 +795,7 @@ export class Arena {
     side: ArenaTagSide = ArenaTagSide.BOTH,
     quiet: boolean = false,
   ): boolean {
-    const existingTag = this.getTag(tagType, side);
+    const existingTag = this.findTag(tagType, side);
     if (existingTag) {
       existingTag.onOverlap(this);
 
@@ -824,7 +824,7 @@ export class Arena {
   }
 
   hasTag(tagType: ArenaTagType): boolean {
-    return !!this.getTag(tagType);
+    return !!this.findTag(tagType);
   }
 
   /**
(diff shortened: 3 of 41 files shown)


src/data/init/init-abilities.ts
```diff
@@ -302,7 +302,7 @@ export function initAbilities() {
         AttackTypeImmunityAbAttr,
         ElementalType.GROUND,
         (pokemon: Pokemon) =>
-          !pokemon.getTag(BattlerTagType.IGNORE_FLYING) && !globalScene.arena.getTag(ArenaTagType.GRAVITY),
+          !pokemon.getTag(BattlerTagType.IGNORE_FLYING) && !globalScene.arena.hasTag(ArenaTagType.GRAVITY),
```
flx-sta, line 306: Q: I assume `pokemon` doens't have `hasTag()`?

src/data/moves/move-attrs/delayed-attack-attr.ts
```diff
@@ -29,7 +29,8 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
   }
 
   /**
-   * If used virtually, this queues a message and proceeds normally.
+   * If used virtually, this queues a message and proceeds normally. (TODO: change this)
```
flx-sta: NIT: Why not `@todo`?

src/field/arena.ts
```diff
@@ -794,57 +794,44 @@ export class Arena {
...
   }
 
   /**
    * Attempts to get a tag from the Arena from a specific side (the tag passed in has to either apply to both sides, or the specific side only)
    *
    * eg: `MIST` only applies to the user's side, while `MUD_SPORT` applies to both user and enemy side
-   * @param tagType The {@linkcode ArenaTagType} or {@linkcode ArenaTag} to get
-   * @param side The {@linkcode ArenaTagSide} to look at
+   * @param tagType - The {@linkcode ArenaTagType} to get
+   * @param side - (Default `ArenaTagSide.BOTH`) The {@linkcode ArenaTagSide} to look at
```
flx-sta, line 845: NIT:
```suggestion
   * @param side - The {@linkcode ArenaTagSide} to look at (Optional. Defaults to {@linkcode ArenaTagSide.BOTH})
```

src/field/arena.ts
```diff
@@ -794,57 +794,44 @@ export class Arena {
...
-   * Uses {@linkcode findTagsOnSide} to filter (using the parameter function) for specific tags that apply to both sides
-   * @param tagPredicate a function mapping {@linkcode ArenaTag}s to `boolean`s
-   * @returns array of {@linkcode ArenaTag}s from which the Arena's tags return true and apply to both sides
-   */
-  findTags(tagPredicate: (t: ArenaTag) => boolean): ArenaTag[] {
-    return this.findTagsOnSide(tagPredicate, ArenaTagSide.BOTH);
+  findTag<T extends ArenaTag = ArenaTag>(tagType: ArenaTagType, side: ArenaTagSide = ArenaTagSide.BOTH): T | undefined {
+    return this.tags.find(
+      (t) => tagType === t.tagType && (side === ArenaTagSide.BOTH || t.side === ArenaTagSide.BOTH || t.side === side),
+    ) as T;
```
flx-sta: Is a `as T` cast here correct? Shouldn't it be ` as T ?? undefined`. I'm never sure about these things

src/field/arena.ts
```diff
@@ -794,57 +794,44 @@ export class Arena {
...
   }
 
   /**
    * Returns specific tags from the arena that pass the `tagPredicate` function passed in as a parameter, and apply to the given side
-   * @param tagPredicate a function mapping {@linkcode ArenaTag}s to `boolean`s
-   * @param side The {@linkcode ArenaTagSide} to look at
+   * @param tagPredicate - A function mapping {@linkcode ArenaTag}s to `boolean`s
+   * @param side - (Default `ArenaTagSide.BOTH`) The {@linkcode ArenaTagSide} to look at
```
flx-sta, line 857: NIT:
```suggestion
   * @param side - The {@linkcode ArenaTagSide} to look at (Optional. Defaults to {@linkcode ArenaTagSide.BOTH})
```

src/ui/components/arena-flyout.ts
```diff
@@ -278,7 +278,8 @@ export class ArenaFlyout extends Phaser.GameObjects.Container {
           return;
         }
 
-        const isEntryHazardTag = globalScene.arena.getTag(tagAddedEvent.arenaTagType) instanceof EntryHazardTag;
+        // TODO: fix this
```
flx-sta: Q: Fix what?

src/data/init/init-abilities.ts
DayKev: I was going to take a look at battlertags after this; I wanted to keep this PR focused on the arena tag functions.

src/field/arena.ts
DayKev: That's handled by the function return type hint being `T | undefined`.

src/field/arena.ts
DayKev: I think we should standardize param docs to be "`@param paramName - (Optional | Default ...) Description here`". This way if we want to see what the default of a param is we can look at the same place every time instead of having to scan to the end of the description which could be any arbitrary length.

src/ui/components/arena-flyout.ts
DayKev: I want to remove the `instanceof`; forgot to go back and do that. Committed the change now.

src/data/moves/move-attrs/delayed-attack-attr.ts
DayKev: It was easier to add it that way than add a todo block, but I guess it's good to add why it should be changed so I'll move it to a todo block tag instead.

src/field/arena.ts
flx-sta: Sure. I think tsdoc itself doesn't have a specification for that, as there is technically no necessity to mark it in tsdoc, as ts has the typings for that

DayKev: Unfortunately for parameters with a default value it doesn't show what that value is, just that the parameter is optional (and hovering the param itself doesn't even show that it's optional, though that's less relevant).
image
image

src/ui/components/arena-flyout.ts
```diff
@@ -1,23 +1,24 @@
...
-import { fixedNumber, isNullOrUndefined } from "#app/utils/common-utils";
-import { toCamelCaseString, toTitleCase } from "#app/utils/string-utils";
+import { BattleSceneEventType } from "#enums/battle-scene-event-type";
+import { TerrainType } from "#enums/terrain-type";
+import { TextStyle } from "#enums/text-style";
+import { WeatherType } from "#enums/weather-type";
+import { WindowVariant } from "#enums/window-variant";
 import type { ParseKeys } from "i18next";
 import i18next from "i18next";
+import { isNullOrUndefined } from "util";
```
MokaStitcher:
```suggestion
import { isNullOrUndefined } from "#app/utils/common-utils";
```

test/abilities/screen-cleaner.test.ts
```diff
@@ -38,13 +38,13 @@ describe("Abilities - Screen Cleaner", () => {
     game.move.select(MoveId.HAIL);
     await game.phaseInterceptor.to(TurnEndPhase);
 
-    expect(game.scene.arena.getTag(ArenaTagType.AURORA_VEIL)).toBeDefined();
+    expect(game.scene.arena.findTag(ArenaTagType.AURORA_VEIL)).toBeDefined();
```
MokaStitcher: hm, all these tests really could check that `arena.hasTag()` is true instead, and also probably check the proper side.
But not sure it's the scope of this PR

test/moves/ceaseless-edge.test.ts
```diff
@@ -50,11 +50,11 @@ describe("Moves - Ceaseless Edge", () => {
 
     await game.phaseInterceptor.to(MoveEffectPhase, false);
     // Spikes should not have any layers before move effect is applied
-    const tagBefore = game.scene.arena.getTagOnSide(ArenaTagType.SPIKES, ArenaTagSide.ENEMY) as EntryHazardTag;
+    const tagBefore = game.scene.arena.findTag(ArenaTagType.SPIKES, ArenaTagSide.ENEMY) as EntryHazardTag;
     expect(tagBefore instanceof EntryHazardTag).toBeFalsy();
```
MokaStitcher: that's... weird :')

```suggestion
    const tagBefore = game.scene.arena.findTag<EntryHazardTag>(ArenaTagType.SPIKES, ArenaTagSide.ENEMY);
    expect(tagBefore).toBeUndefined();
```
is that the better way?

<sub>(the rest of the file also does type casting that could be changed but it probably requires using `!` to not bother with the `undefined` issue)

test/moves/toxic-spikes.test.ts
```diff
@@ -112,7 +112,7 @@ describe("Moves - Toxic Spikes", () => {
     game.move.select(MoveId.SPLASH, 1);
     await game.toNextTurn();
 
-    const arenaTags = game.scene.arena.getTagOnSide(ArenaTagType.TOXIC_SPIKES, ArenaTagSide.ENEMY) as EntryHazardTag;
+    const arenaTags = game.scene.arena.findTag(ArenaTagType.TOXIC_SPIKES, ArenaTagSide.ENEMY) as EntryHazardTag;
     expect(arenaTags.tagType).toBe(ArenaTagType.TOXIC_SPIKES);
     expect(arenaTags.layers).toBe(1);
```
MokaStitcher: not sure that's the best way to handle the undefined case
```suggestion
    const arenaTag = game.scene.arena.findTag<EntryHazardTag>(ArenaTagType.TOXIC_SPIKES, ArenaTagSide.ENEMY);
    expect(arenaTag).toBeDefined();
    expect(arenaTag!.tagType).toBe(ArenaTagType.TOXIC_SPIKES);
    expect(arenaTag!.layers).toBe(1);
```

src/ui/components/arena-flyout.ts
DayKev: Okay we need to rename this function so this stops happening. Maybe `isNil`?

test/abilities/screen-cleaner.test.ts
DayKev: Hm, since I already changed these files I guess it's not a big deal to change them to `hasTag`.

DayKev: Done.

MokaStitcher (dismissed):
sorry about having you go and replace all these `hasTag` in test, I didn't realize there were so many tests doing this when I made that comment... :')

src/field/arena.ts
```diff
@@ -1,3 +1,5 @@
+import { DEFAULT_NEW_TERRAIN_DURATION, PRIMAL_WEATHER_TYPES } from "#app/constants/game";
+import { DEFAULT_NEW_WEATHER_DURATION } from "#app/constants/weather";
```
flx-sta: I guess I forgot to move `PRIMAL_WEATHER_TYPES` into `#app/constants/weather`

src/field/arena.ts
```diff
@@ -824,57 +824,44 @@ export class Arena {
...
-  getTag(tagType: ArenaTagType | AbstractConstructor<ArenaTag>): ArenaTag | undefined {
-    return this.getTagOnSide(tagType, ArenaTagSide.BOTH);
-  }
-
-  hasTag(tagType: ArenaTagType): boolean {
-    return !!this.getTag(tagType);
+  hasTag(tagType: ArenaTagType, side: ArenaTagSide = ArenaTagSide.BOTH): boolean {
+    return this.tags.some(
+      (t) => t.tagType === tagType && (side === ArenaTagSide.BOTH || t.side === ArenaTagSide.BOTH || t.side === side),
+    );
```
flx-sta: NIT: I feel like the 2nd section of the `some()` condition is quite hard to understand if one doesn't know what's going on. It might be easier to pre-populate a `Set` of all the valid choices, or what do you think?

```suggestion
  hasTag(tagType: ArenaTagType, side: ArenaTagSide = ArenaTagSide.BOTH): boolean {
    const validSides= new Set<ArenaTagSide>([ArenaTagSide.BOTH, side]);
    
    return this.tags.some(
      (t) => t.tagType === tagType && validSides.has(t.side),
    );
```

Why a set? That way we don't have to worry about putting duplicates into it. The set will handle it for us

src/field/arena.ts
```diff
@@ -824,57 +824,44 @@ export class Arena {
...
-   * Uses {@linkcode findTagsOnSide} to filter (using the parameter function) for specific tags that apply to both sides
-   * @param tagPredicate a function mapping {@linkcode ArenaTag}s to `boolean`s
-   * @returns array of {@linkcode ArenaTag}s from which the Arena's tags return true and apply to both sides
-   */
-  findTags(tagPredicate: (t: ArenaTag) => boolean): ArenaTag[] {
-    return this.findTagsOnSide(tagPredicate, ArenaTagSide.BOTH);
+  findTag<T extends ArenaTag = ArenaTag>(tagType: ArenaTagType, side: ArenaTagSide = ArenaTagSide.BOTH): T | undefined {
+    return this.tags.find(
+      (t) => tagType === t.tagType && (side === ArenaTagSide.BOTH || t.side === ArenaTagSide.BOTH || t.side === side),
+    ) as T;
```
flx-sta: NIT: I feel like the 2nd section of the `find()` condition is quite hard to understand if one doesn't know what's going on. It might be easier to pre-populate a `Set` of all the valid choices, or what do you think?

```suggestion
  findTag<T extends ArenaTag = ArenaTag>(tagType: ArenaTagType, side: ArenaTagSide = ArenaTagSide.BOTH): T | undefined {
    const validSides= new Set<ArenaTagSide>([ArenaTagSide.BOTH, side]);
    return this.tags.find(
      (t) => tagType === t.tagType && validSides.has(t.side),
    ) as T;
```

Why a set? That way we don't have to worry about putting duplicates into it. The set will handle it for us

src/field/arena.ts
```diff
@@ -824,57 +824,44 @@ export class Arena {
...
+   * @param side - (Default `ArenaTagSide.BOTH`) The {@linkcode ArenaTagSide} to look at
    * @returns array of {@linkcode ArenaTag}s from which the Arena's tags return `true` and apply to the given side
    */
-  findTagsOnSide(tagPredicate: (t: ArenaTag) => boolean, side: ArenaTagSide): ArenaTag[] {
+  getTags<T extends ArenaTag = ArenaTag>(
+    tagPredicate: (t: ArenaTag) => boolean,
+    side: ArenaTagSide = ArenaTagSide.BOTH,
+  ): T[] {
     return this.tags.filter(
       (t) => tagPredicate(t) && (side === ArenaTagSide.BOTH || t.side === ArenaTagSide.BOTH || t.side === side),
```
flx-sta: NIT: I feel like the 2nd section of the `filter()` condition is quite hard to understand if one doesn't know what's going on. It might be easier to pre-populate a `Set` of all the valid choices, or what do you think?

```suggestion
    const validSides= new Set<ArenaTagSide>([ArenaTagSide.BOTH, side]);
    return this.tags.filter(
      (t) => tagPredicate(t) && validSides.has(t.side),
```

Why a set? That way we don't have to worry about putting duplicates into it. The set will handle it for us

flx-sta (dismissed):

flx-sta (approved):

Tempo-anon (approved):

PR #1088, opened 2025-04-29
Add `Battle#isTrainerBattle`, reformat functions, update comments

DayKev:
Added some `TODO`s around the item-based multipliers as well.

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -449,6 +449,16 @@ export default class Battle {
   isBattleMysteryEncounter(): boolean {
     return this.battleType === BattleType.MYSTERY_ENCOUNTER;
   }
+
+  /**
+   * @param includeMEs - Whether to count Mystery Encounter trainer battles
+   * @returns `true` if the current battle is a trainer battle
+   */
+  public isTrainerBattle(includeMEs: boolean = false): boolean {
+    const { battleType, mysteryEncounter } = this;
+    const trainerME = includeMEs ? mysteryEncounter?.encounterMode === MysteryEncounterMode.TRAINER_BATTLE : false;
+    return battleType === BattleType.TRAINER || trainerME;
+  }
 }
 
 export class FixedBattle extends Battle {
--- a/src/data/exp.ts
+++ b/src/data/exp.ts
@@ -1,10 +1,8 @@
 import type { EnemyPokemon } from "#app/field/enemy-pokemon";
 import type { PlayerPokemon } from "#app/field/player-pokemon";
-import type { Pokemon } from "#app/field/pokemon";
-import { BattleType } from "#enums/battle-type";
+import { globalScene } from "#app/global-scene";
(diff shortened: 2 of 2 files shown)


src/data/exp.ts
```diff
@@ -195,50 +193,45 @@ export function getGrowthRateColor(growthRate: GrowthRate) {
...
-  // Trainer exp bonus
-  const trainerExpBonus =
-    this.currentBattle.battleType === BattleType.TRAINER
-    || this.currentBattle.mysteryEncounter?.encounterMode === MysteryEncounterMode.TRAINER_BATTLE
-      ? 1.5
-      : 1;
-  // This game does not use the outsider Pokemon bonus
-  const outsiderExpBonus = 1;
-  // Unused
-  const expPointPowerBonus = 1;
```
DayKev, line 241: Since we never plan to use these I removed them and mentioned them in the main function TSDoc instead.

src/data/exp.ts
```diff
@@ -195,50 +193,45 @@ export function getGrowthRateColor(growthRate: GrowthRate) {
...
+  const trainerExpBonus = globalScene.currentBattle.isTrainerBattle(true) ? 1.5 : 1;
+
+  const baseExpGain = (baseExp * enemyLevel) / 5;
```
DayKev, line 234: Extracted these out to reduce parentheses nesting.

src/battle.ts
```diff
@@ -449,6 +449,16 @@ export default class Battle {
   isBattleMysteryEncounter(): boolean {
     return this.battleType === BattleType.MYSTERY_ENCOUNTER;
   }
+
+  /**
+   * @param includeMEs - Whether to count Mystery Encounter trainer battles
+   * @returns `true` if the current battle is a trainer battle
+   */
+  public isTrainerBattle(includeMEs: boolean = false): boolean {
```
DayKev, line 457: In a follow-up PR we should replace existing `battleType === TRAINER` code with uses of this function.

src/data/exp.ts
```diff
@@ -195,50 +193,45 @@ export function getGrowthRateColor(growthRate: GrowthRate) {
 
 /**
  * Function to get the exp given from a defeated Pokemon from gen 1-4
+ *
+ * Not included from the mainline formula:
+ * - Original Trainer bonus
```
Tempo-anon, line 198: nit: the bonus is actually for pokemon that are not of the original trainer but that doesn't really apply in a single player game. Maybe any NPC trades the player can get will just come with a lucky egg and that is good enough

Tempo-anon (approved):

src/data/exp.ts
DayKev: Oh yeah I forgot about the possibility of NPC trading. Well, if we add that in the future we can put the EXP bonus back in.

PR #1090, opened 2025-04-29
[Refactor] Add and implement `coerceArray` utility function

DayKev:
N/A

Code simplification.

A new utility function `makeArray(input)` is added.
```ts
// old
if (!Array.isArray(input)) {
  input = [input];
}

// new
input = coerceArray(input);
```

## Checklist


src/data/battler-tags/utils/apply-battler-tags.ts
```diff
@@ -17,7 +18,7 @@ export function applyBattlerTags<T extends BattlerTag = BattlerTag>(
   ...params: Parameters<T["apply"]>
 ): boolean {
   const [pokemon, simulated, ...args] = params;
-  const tagTypeArr = Array.isArray(tagTypes) ? tagTypes : [tagTypes];
+  const tagTypeArr = coerceArray(tagTypes) as BattlerTagType[];
```
DayKev: `as BattlerTagType[]` is required:
<details><summary>No manual type hints</summary>
<p>

image

</p>
</details> 
<details><summary>Using "var: BattlerTagType[]"</summary>
<p>

image

</p>
</details> 
<details><summary>Using "func&lt;BattlerTagType[]>()"</summary>
<p>

image

</p>
</details> 
<details><summary>Original behavior (was typed as "any[]")</summary>
<p>

image

</p>
</details>

src/data/abilities/ab-attrs/flinch-stat-stage-change-ab-attr.ts
```diff
@@ -11,7 +12,7 @@ export class FlinchStatStageChangeAbAttr extends FlinchEffectAbAttr {
   constructor(stats: BattleStat[], stages: number) {
     super();
 
-    this.stats = Array.isArray(stats) ? stats : [stats];
+    this.stats = coerceArray(stats);
```
MokaStitcher: according to the constructor's signature stats is already an array, so either the constructor is wrong or that `coerceArray` shouldn't be needed

Tempo-anon (dismissed):

src/data/battler-tags/utils/apply-battler-tags.ts
MokaStitcher: I may completely be underthinking this but by doing this:
```ts
export function coerceArray<T>(input: T | readonly T[]): T[];
export function coerceArray<T>(input: T | T[]): T[] {
  return Array.isArray(input) ? input : [input];
}
```
The typing seems to get resolved properly without needing to cast anything. I just had to add the overload with `readonly` because the compiler wasn't happy about apply battler tags otherwise. Not entirely sure it's good but at least the compiler doesn't complain and the resolved typing seemed correct everywhere I checked

Can you explain a bit more what the original signature and overload does?
```ts
export function coerceArray<T>(input: T): T extends any[] ? T : [T];
export function coerceArray<T>(input: T): T | [T] {
```

It seems the option I proposed above also allows T to be an array, for example, with the current version we can't get it to return an array of arrays, even with telling it T is an array:
image

With the version above, by default:
image
And if we tell it that T is an array of strings:
image

src/data/abilities/ab-attrs/pre-apply-battler-tag-immunnity-ab-attr.ts
```diff
@@ -1,7 +1,7 @@
 import type { BattlerTag } from "#app/data/battler-tags/battler-tag";
 import type { Pokemon } from "#app/field/pokemon";
 import { getPokemonNameWithAffix } from "#app/messages";
-import type { BooleanHolder } from "#app/utils/common-utils";
+import { coerceArray, type BooleanHolder } from "#app/utils/common-utils";
```
flx-sta, line 4: Q: I remember you mentioning to avoid combining import & type import

src/field/arena.ts
```diff
@@ -717,19 +717,19 @@ export class Arena {
...
   applyTagsForSide(
-    tagType: ArenaTagType | ArenaTagType[],
+    tagTypes: ArenaTagType | ArenaTagType[],
     side: ArenaTagSide,
     simulated: boolean,
     ...args: unknown[]
   ): void {
-    const tagTypeArr = Array.isArray(tagType) ? tagType : [tagType];
-    let tags = this.tags.filter((t) => tagTypeArr.includes(t.tagType));
+    tagTypes = coerceArray(tagTypes);
```
flx-sta: NIT: Generally it's discouraged to alter params, that's probably why the other variable had the `arr` suffix.

src/utils/common-utils.ts
```diff
@@ -170,3 +170,12 @@ export function deepFreeze<T>(obj: T): Readonly<T> {
...
 }
+
+/**
+ * If the input isn't already an array, turns it into one.
+ * @returns An array with the same type as the type of the input
+ */
+export function coerceArray<T>(input: T): T extends any[] ? T : [T];
+export function coerceArray<T>(input: T): T | [T] {
+  return Array.isArray(input) ? input : [input];
+}
```
flx-sta: Q: Why define it twice?

src/data/abilities/ab-attrs/flinch-stat-stage-change-ab-attr.ts
DayKev: Ah good catch, I just search and replaced stuff using this pattern.

src/data/abilities/ab-attrs/pre-apply-battler-tag-immunnity-ab-attr.ts
DayKev: Combining them is okay, it's only when all the imports are `type` imports that the `type` identifier needs to be moved outside the `{}`.
```ts
// good
import { BooleanHolder, type NumberHolder } from "#app/utils/common-utils";
// good
import type { NumberHolder, nil } from "#app/utils/common-utils";
// bad
import { type NumberHolder, type nil } from "#app/utils/common-utils";
```
We don't have to worry about it though because I added an ESLint rule to check for us.

src/data/battler-tags/utils/apply-battler-tags.ts
DayKev: > Can you explain a bit more what the original signature and overload does?

From @SirzBenjie:
> This way gives you a bit more power with typescript inference, using type predicates. It's a lot nicer, as it forces the return type of the method to be exactly the type that was put in, rather than a generic array. This preserves tuples, if the input was known to be a tuple. It's great.

MokaStitcher: hmm, we don't use tuples though, so I don't know if it outweighs the downside of needing to type cast
I may experiment with it a bit more

src/utils/common-utils.ts
DayKev: Discussion in this comment chain: https://github.com/Despair-Games/poketernity/pull/1090#discussion_r2066039955

src/field/arena.ts
DayKev: I changed it because of this error:
image

DayKev: I can change it to be like the other one?
```ts
const tagTypeArr = coerceArray(tagTypes) as ArenaTagType[];
```

src/data/battler-tags/utils/apply-battler-tags.ts
DayKev: I'm not overly attached to the more precise version, given the use case for the function in the existing codebase means it doesn't need to be more precise.

src/field/arena.ts
MokaStitcher: would the solution proposed here also fix this? https://github.com/Despair-Games/poketernity/pull/1090#discussion_r2066675010

DayKev: It does.

src/data/battler-tags/utils/apply-battler-tags.ts
DayKev: Swapped to the less precise version.
I also changed the return from `array ? input : [input]` to `array ? [...input] : [input]` to account for `readonly` arrays.

flx-sta (dismissed):

flx-sta (dismissed):

PigeonBar (dismissed):
A couple more places where `coerceArray` can be used:

https://github.com/Despair-Games/poketernity/blob/e334a2dee6abee607e74b14a42220e1613fa9253/src/data/mystery-encounters/utils/encounter-phase-utils.ts#L476

https://github.com/Despair-Games/poketernity/blob/e334a2dee6abee607e74b14a42220e1613fa9253/src/data/mystery-encounters/utils/encounter-phase-utils.ts#L826

flx-sta (approved):

Tempo-anon (approved):

PR #1092, opened 2025-04-30
[Refactor] Rename `isNullOrUndefined` to `isNil` and add TSDoc

DayKev:
N/A

This prevents an issue where VSCode would incorrectly import a same-name function from `util` instead of `src/utils/common-utils.ts` which would break things. As a bonus this makes the function name way shorter, reducing line length.

`isNullOrUndefined` renamed to `isNil` and TSDocs added.

## Checklist

### DIFF
--- a/src/data/init/init-encounter-anims.ts
+++ b/src/data/init/init-encounter-anims.ts
@@ -1,8 +1,8 @@
 import { AnimConfig } from "#app/data/animations/anim-config";
-import { encounterAnims } from "../animations/encounter-anims";
 import { globalScene } from "#app/global-scene";
-import { getEnumKeys, isNullOrUndefined } from "#app/utils/common-utils";
+import { getEnumKeys, isNil } from "#app/utils/common-utils";
 import { EncounterAnim } from "#enums/encounter-anims";
+import { encounterAnims } from "../animations/encounter-anims";
 
 /**
  * Fetches animation configs to be used in a Mystery Encounter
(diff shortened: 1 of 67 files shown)


innerthunder (dismissed):

src/data/init/init-encounter-anims.ts
```diff
@@ -1,8 +1,8 @@
 import { AnimConfig } from "#app/data/animations/anim-config";
-import { encounterAnims } from "../animations/encounter-anims";
 import { globalScene } from "#app/global-scene";
-import { getEnumKeys, isNullOrUndefined } from "#app/utils/common-utils";
+import { getEnumKeys, isNil } from "#app/utils/common-utils";
 import { EncounterAnim } from "#enums/encounter-anims";
+import { encounterAnims } from "../animations/encounter-anims";
```
MokaStitcher:
```suggestion
```

src/data/init/init-encounter-anims.ts
```diff
@@ -1,8 +1,8 @@
 import { AnimConfig } from "#app/data/animations/anim-config";
-import { encounterAnims } from "../animations/encounter-anims";
 import { globalScene } from "#app/global-scene";
```
MokaStitcher, line 3:
```suggestion
import { encounterAnims } from "#app/data/animations/encounter-anims";
import { globalScene } from "#app/global-scene";
```

MokaStitcher (dismissed):

PR #1093, opened 2025-04-30
[Refactor] Move `BattleSummonData` fields into `SummonData`

DayKev:
N/A

`BattleSummonData` was used in exactly the same way as `SummonData`, so having two separate data structures for the same purpose is redundant.

All fields in `BattleSummonData` were moved to `SummonData` and references to `Pokemon#battleSummonData` were changed to `Pokemon#summonData`.

## Checklist


innerthunder (dismissed):

MokaStitcher:
So this seems to mean that the turn count, wave turn count and move history will now be saved as part of the session data
https://github.com/Despair-Games/poketernity/blob/fa9cfb288c37602107944978889ab660d21e213a/src/system/pokemon-data.ts#L59
But those things are only relevant to the ongoing battle and shouldn't be saved, if only because they will use storage space. (I am also a bit worried about potential bugs caused by it)
I have to assume this is why summonData and battleSummonData were handled separately. I agree about the confusing aspect of it, but  we probably should try to find a solution that does not compromise on the storage issue

src/field/pokemon.ts
```diff
@@ -4151,7 +4151,13 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     if (!this.waveData) {
       this.resetWaveData();
     }
-    this.resetBattleSummonData();
+    // TODO: why is this here?
+    if (this.getTag(BattlerTagType.SEEDED)) {
+      this.lapseTag(BattlerTagType.SEEDED);
+    }
```
Tempo-anon, line 4104: I assume this is trying to proc leech seed one more time when a battle ends?

Tempo-anon:
> I have to assume this is why summonData and battleSummonData were handled separately. I agree about the confusing aspect of it, but we probably should try to find a solution that does not compromise on the storage issue

Would be good to have this documented in the `PokemonSummonData` and `PokemonBattleSummonData` classes (as well as documentation for them in general and what they're used for). I think it makes sense to keep them separate if one of them will only be used for a battle because right now the game does not save information relating only to a battle

DayKev:
> So this seems to mean that the turn count, wave turn count and move history will now be saved as part of the session data
> 
> https://github.com/Despair-Games/poketernity/blob/fa9cfb288c37602107944978889ab660d21e213a/src/system/pokemon-data.ts#L59
> 
> 
> But those things are only relevant to the ongoing battle and shouldn't be saved, if only because they will use storage space. (I am also a bit worried about potential bugs caused by it)
> I have to assume this is why summonData and battleSummonData were handled separately. I agree about the confusing aspect of it, but we probably should try to find a solution that does not compromise on the storage issue

If only this had been documented anywhere...
Well, we need to re-do save data anyway, so maybe for now it doesn't matter?

flx-sta:
> > So this seems to mean that the turn count, wave turn count and move history will now be saved as part of the session data
> > https://github.com/Despair-Games/poketernity/blob/fa9cfb288c37602107944978889ab660d21e213a/src/system/pokemon-data.ts#L59
> > 
> > But those things are only relevant to the ongoing battle and shouldn't be saved, if only because they will use storage space. (I am also a bit worried about potential bugs caused by it)
> > I have to assume this is why summonData and battleSummonData were handled separately. I agree about the confusing aspect of it, but we probably should try to find a solution that does not compromise on the storage issue
> 
> If only this had been documented anywhere... Well, we need to re-do save data anyway, so maybe for now it doesn't matter?

I'd say it's fine for now. For the future we should probably have some kind of `export()` method inside classes that get saved anyway to purge and unnecessary data. 

------

@DayKev Please resolve the conflicts

flx-sta (dismissed):
Please make sure to set up the follow-up issue (task) for the long-term solution for values that shouldn't be in the session-data

DayKev:
> Please make sure to set up the follow-up issue (task) for the long-term solution for values that shouldn't be in the session-data

#1131

DayKev:
> So this seems to mean that the turn count, wave turn count and move history will now be saved as part of the session data
>
> But those things are only relevant to the ongoing battle and shouldn't be saved

Maybe they should be moved into wave data instead then, since we wanted to have every wave be its own battle.

flx-sta:
Sorry, hit `Update Branch` in the wrong tab @DayKev

flx-sta (approved):

innerthunder (approved):

PR #1096, opened 2025-04-30
[P2] Transistor is 1.3x instead of 1.5x

Tempo-anon:

Transistor has a 1.3x multiplier instead of 1.5x

This ability was changed from gen 8 to gen 9

Changed the `powerMultiplier` field to 1.3 from the default 1.5x

Technically the ability is supposed to boost the user's attacking stats by 1.3 instead of the damage, but functionally it's the same.

## Screenshots/Videos

n/a

## How to test the changes?

Override with transistor

## Checklist

### DIFF
--- a/src/data/init/init-abilities.ts
+++ b/src/data/init/init-abilities.ts
@@ -1388,7 +1388,7 @@ export function initAbilities() {
     new Ability(AbilityId.CURIOUS_MEDICINE, 8)
       .attr(PostSummonClearAllyStatStagesAbAttr),
     new Ability(AbilityId.TRANSISTOR, 8)
-      .attr(MoveTypePowerBoostAbAttr, ElementalType.ELECTRIC),
+      .attr(MoveTypePowerBoostAbAttr, ElementalType.ELECTRIC, 1.3),
     new Ability(AbilityId.DRAGONS_MAW, 8)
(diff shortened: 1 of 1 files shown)


innerthunder (approved):

flx-sta:
1. Assigned PR to @Tempo-anon .
2. Added PR to `PKTY Board`
3. Update board status to `In Review`

src/data/init/init-abilities.ts
```diff
@@ -1388,7 +1388,7 @@ export function initAbilities() {
     new Ability(AbilityId.CURIOUS_MEDICINE, 8)
       .attr(PostSummonClearAllyStatStagesAbAttr),
     new Ability(AbilityId.TRANSISTOR, 8)
```
flx-sta, line 1390: Does the Gen stay `8` still, since it was added then?

flx-sta:
To save you from requesting to write test-coverage for this I've added an ability-checkup ticket:
#1099

flx-sta (approved):
lgtm. 
~~Just 1 question regarding the Gen assign~~

src/data/init/init-abilities.ts
innerthunder: Yes, this always refers to the Gen the ability was introduced.

PR #1113, opened 2025-05-02
[Refactor] Consolidate `try/catch` for sprites in one place

### LINKED ISSUE #1071: [Refactor] Replace `Sprite.play()` with a `try/catch` version of it
Almost all our `play` calls are wrapped into a `try/catch`.
We want to replace the original `Phaser.GameObjects.Sprite.prototype.play` method with a custom one that simply wraps the original in a `try/catch`

DayKev:
Third time's the charm.

Potentially less crashes due to animation issues, though most were already surrounded with `try/catch` so likely no change.

Reduce duplicated code.
Resolves #1071 

All `try/catch` surrounding `sprite.play()` are removed and instead `Phaser.GameObjects.Sprite.prototype.play` is overridden to have a `try/catch` inside it.

## Checklist

### DIFF
--- a/src/phaser-extensions.ts
+++ b/src/phaser-extensions.ts
@@ -1,16 +1,16 @@
 import Phaser from "phaser";
 
-// #region Original Functions
-
-const originalPlay = Phaser.GameObjects.Sprite.prototype.play;
-const originalStop = Phaser.GameObjects.Sprite.prototype.stop;
-
-// #endregion
 // #region Types
 
 type GuideObject = Phaser.GameObjects.Components.Origin
   & Phaser.GameObjects.Components.Size
   & Phaser.GameObjects.Components.Transform;
+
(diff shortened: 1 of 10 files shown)


src/phaser-extensions.ts
flx-sta, line 1: Please keep the `Types` region right below the imports of the file.
I've always tried to have it in this order:
1. tsdoc imports
2. imports
3. Types
4. constants (or in this case `Original Functions`
5. whatever order works

flx-sta (dismissed):
lgtm, just 1 wannabeOCD complaint

src/phaser-extensions.ts
DayKev: Moved original functions below types.

Tempo-anon (approved):

flx-sta (approved):

PR #1118, opened 2025-05-02
[Misc] Fix ME `build()` functions

DayKev:
Nothing, probably?

I was reminded of this when writing PR 1113.

Restored the `this` "param" to the two ME `build()` methods.

## How to test the changes?
Check some MEs I guess.

## Checklist

### DIFF
--- a/src/data/mystery-encounters/mystery-encounter-option.ts
+++ b/src/data/mystery-encounters/mystery-encounter-option.ts
@@ -333,7 +333,7 @@ export class MysteryEncounterOptionBuilder implements Partial<IMysteryEncounterO
     return this;
   }
 
-  build() {
-    return new MysteryEncounterOption(this as unknown as IMysteryEncounterOption);
+  build(this: IMysteryEncounterOption) {
+    return new MysteryEncounterOption(this);
   }
 }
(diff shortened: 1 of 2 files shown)


DayKev:
Basically, this PR reverts it to how it was originally. It got messed up by an ESLint rule being misconfigured (?) and then changed to the current version shortly thereafter to fix the issue.

MokaStitcher (approved):

flx-sta (approved):
~~And that works? I didn't know you can put `this` as first param anywhere to add a type to it. Is there and sources for this?~~

https://www.typescriptlang.org/docs/handbook/2/functions.html#declaring-this-in-a-function

> ### Declaring `this` in a Function
>
> TypeScript will infer what the this should be in a function via code flow analysis, for example in the following:
> ```ts
> const user = {
>   id: 123,
> 
>   admin: false,
>   becomeAdmin: function () {
>       this.admin = true;
>    },
> };
> ```
> 
> TypeScript understands that the function user.becomeAdmin has a corresponding this which is the outer object user. this, heh, can be enough for a lot of cases, but there are a lot of cases where you need more control over what object this represents. The JavaScript specification states that you cannot have a parameter called this, and so TypeScript uses that syntax space to let you declare the type for this in the function body.

Issue #1117, opened 2025-05-02
[Refactor] Initialize UI handlers dynamically


MokaStitcher:
Currently all UI Handlers are initialized on startup and always stay in memory, even if most handlers are rarely if ever used.

The most basic system would be to have a list of handlers that should stay in memory because they are often used (message, option select, confirm, menu, fight, command, settings?), and have others handlers initialized only when needed, then destroyed

PR #1119, opened 2025-05-02
[Refactor] Make UI's handler list private

### LINKED ISSUE #1114: [Refactor] Make UI's handlers list private
Currently the list of handlers managed by the UI is available publicly, and accessed directly by some parts of the code.
This is a bad pattern and is a problem for moving away from having all handlers in memory at all times

The following entities call directly `ui.handlers`, which should be replaced:
- [x] SummaryUiHandler
- [x] InputsController
- [x] CommandPhase
- [x] Tests:
   - [x] ME Test utils
   - [x] Each Individual ME Test
   - [x] SelectModifierPhase Test

MokaStitcher:

N/A

More preparation for #1117
Fixes #1114 tldr. The ui's handlers list shouldn't be public

- Command Phase: Moved the cursor reset to the CommandUiHandler
- InputsController: Replaced the direct calls to the gamepad and keyboard handlers with events
- SummaryUiHandler: Add a `fromPartyMenu` parameter to replace fetching the handlers list

Tests:
- ME test utils:
   - Removed unnecessary `onNextPrompt`s from `runSelectMysteryEncounterOption` which were clogging the prompt handler queue, preventing subsequent prompt handlers from running
   - Replaced accessing the uiHandlers directly by `onNextPrompt`s in `handleSecondaryOptionSelect`
- Individual tests:
   - Replaced accessing the uiHandlers directly by `onNextPrompt`s in Clowning Around test. This required clearing out the current list of prompt handlers to ensure those new ones run. This is not ideal, but was already done in another test (Bug Type Superfan) and refactoring the test prompt handlers is beyond the scope of this PR.
   - Other modified tests were searching through the handlers list for the `ModifierSelectUiHandler` instead of simply calling `ui.get(Current)Handler()`

<sub> will make a separate PR to rename `getHandler` to `getCurrentHandler`, to clarify its meaning

## How to test the changes?

Run the game
- Play with a controller, then open the settings, the controller tab should be initialized
- Play a few waves, making sure the cursor still resets to `Fight` each wave, but not each turn
- Access the Pokemon Summary from various places (party screen, catching a mon, learning a move, ...)

## Checklist


test/mystery-encounter/encounter-test-utils.ts
```diff
@@ -150,40 +123,48 @@ export async function runSelectMysteryEncounterOption(
   }
 }
 
+/**
+ * @param game - The {@linkcode GameManager}.
+ * @param pokemonNo - The Pokemon to select, from 1 to 6.
+ * @param optionNo - (Optional) If the Pokemon selection is followed by another menu,
```
Tempo-anon: nit: not a fan of `No` as an abbreviation for number as that can be confusing. Maybe `pokemonNum` or `pokemonSlot` and `defaultOptionNum`?

Tempo-anon (dismissed):

test/mystery-encounter/encounter-test-utils.ts
MokaStitcher: I initially changed it to Index before realizing it wasn't starting at 0, so just rolled it back since I couldn't find a great name. Agree that "No" is confusing :')
I like slot

flx-sta: I think `No` is fine, as it's the official abbreviation. But I'd also go with `Slot`. `Num` is probably my least favorite, still fine though

src/@types/Settings.ts
```diff
@@ -119,3 +119,5 @@ export interface SettingsUpdateEventArgs {
 }
 
 export type SettingsEvent = "settings/updated" | "settings/update/failed" | "settings/saved";
+
+export type ControlsEvent = "keyboard/init" | "gamepad/init";
```
flx-sta: This type should be moved into it's own file.
Probably called `Controls.ts` as it has strictly nothing  to do with Settings.

src/inputs-controller.ts
flx-sta, line 1: Instead of having to add two new `emitInitEvent` parameters we should consider restructuring the calls to better fit the event-driven approach.
e.g. `setupGamePad()` should probably not call `setChosenGamepad`, but instead call all the involved methods separately
```ts
    this.deactivatePressedKey();
    this.initChosenGamepad(gamepad);
    eventBus.emit("gamepad/init");
```
Although I'm wondering if `deactivatePressedKey` is even necessary to call there.

----

An alternative approach would be to have some kind of logic that knows if the gamepad was already initialized and only emit the event if not yet.

MokaStitcher: This whole controls things is a whole can of spaghetti, definitely not the scope of this PR to fix
It needs a full refactor

src/@types/Settings.ts
MokaStitcher: Both of these are called by and used for settings but sure

flx-sta: I know but technically it's not bound to settings. The controls can also "live" without settings, right?

test/mystery-encounter/encounter-test-utils.ts
MokaStitcher: updated `pokemonNo` to `partySlot` (I'm pretty sure we're using that in other places) and `optionNo` to `optionNumber` (no need for an abbreviation, especially since optionNo can be interpreted at a glance as the no option in a menu or something)

src/@types/Settings.ts
MokaStitcher: moved it and renamed it InputsEvent

src/inputs-controller.ts
flx-sta: Follow up EPIC created #1121

flx-sta (approved):

Tempo-anon (approved):

PR #1123, opened 2025-05-02
[Bug] Fix gym leader party generation

Tempo-anon:
## What are the changes the user will see?

Gym leaders will actually use their appropriate teams instead of always using the 8th gym

## Why am I making these changes?

Bug found while playtesting

## What are the changes from a developer perspective?

The game technically generates the party for a gym leader on the PREVIOUS wave so I updated the switch statement
image

## Screenshots/Videos

image
image

## How to test the changes?

Override to a wave and then play to a gym leader or override straight to a gym leader

## Checklist

- [ ] ⚠️ If this is a PR for `main` (such as a hotfix), has the game version been updated (`npm run update-version:patch` / `npm run update version:minor`?
- [x] Otherwise: **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?

- [x] Are all unit tests still passing? (`npm run test:silent`)
  - [ ] Have I created new automated tests (`npm run test:create`) or updated existing tests related to the PR's changes?
- [ ] Have I provided screenshots/videos of the changes (if applicable)?
  - [ ] Have I made sure that any UI change works for both UI themes (dark and light)?

#### Are there any localization additions or changes? If so:

- [ ] Has a locales PR been created on the [locales](https://github.com/despair-games/poketernity-locales) repo?
  - [ ] If so, please leave a link to it here:
- [ ] Have I added the `Localization` tag to this PR?

#### If there are no locale changes:
- [ ] Have I made sure **not** to commit any changes to the locale repo on this branch?

PR #1126, opened 2025-05-03
[Balance] Change exp formula to gen 1-4

Tempo-anon:

Exp formula has been changed

Moving away from custom legacy code and to make exp scale better with the new level caps

* ~~Disabled the `useWaveIndexMultiplier` in `battle-scene.ts`~~ Unfortunately all ME code uses this so it will be a very large change
* Changed `EnemyPokemon` to `Pokemon` in `genOneThroughFourExpFormula` because the PostKnockoutPhase only holds a `Pokemon` instead of `EnemyPokemon`
* Changed the exp formula to use `genOneThroughFourExpFormula` instead of `getExpValue` in `post-knockout-phase.ts`

## Screenshots/Videos

I tested with the changes in #1123 

Manually tested the first 110 waves and noted my findings. I specifically made sure to only use MEDIUM_FAST Pokemon so the exp smoothing wasn't a concern. I used two Pokemon so I could get an idea of what level the player's primary Pokemon would be compared to the rest of their party. The level scaling looks pretty good for the first 50 or so waves but the player hits the level cap earlier and earlier as the game goes on so the game is definitely handing out too many exp charms

```
8: Got Super exp charm (1) , got exp share (1)
10: get exp charm (1)
11 start: 11/6 grabbed a lure on floor 14
20: get super exp charm (2)
21 start: 16/11
23: got exp all (2)
25: hit level cap, got exp all (3)
30: got exp charm (2)
31 start: 22/20
38: hit level cap
41: 26/26
45: hit level cap
47: hit level cap
50: get exp charm (3)
51 start: 29/29
54 got exp charm (4)
55 hit level cap (all)
60: got super exp charm (4)
61 start: 33/33
62: hit level cap
63: get exp share (4)
70: get exp charm (5)
71 start: 36/36
73: hit level cap
74: hit level cap
80: got super exp charm (5)
81 start: 40/40
84: hit level cap (all)
90: get exp charm (6)
91 start: 45/45
93: hit level cap (all)
100: got super exp charm (6)
101 start: 48/48
102: hit level cap (all)
111: 52/52
```

## How to test the changes?

`npm run start`

## Checklist

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -2744,7 +2744,7 @@ export default class BattleScene extends SceneBase {
   applyPartyExp(
     expValue: number,
     pokemonDefeated: boolean,
-    _useWaveIndexMultiplier?: boolean,
+    useWaveIndexMultiplier?: boolean,
     pokemonParticipantIds?: Set<number>,
   ): void {
     const participantIds = pokemonParticipantIds ?? this.currentBattle.playerParticipantIds;
@@ -2757,10 +2757,13 @@ export default class BattleScene extends SceneBase {
     const nonFaintedPartyMembers = party.filter((p) => p.hp);
     const expPartyMembers = nonFaintedPartyMembers.filter((p) => p.level < this.getMaxExpLevel());
     const partyMemberExp: number[] = [];
-    // EXP value calculation is based off Pokemon.getExpValue
-    // if (useWaveIndexMultiplier) {
-    //   expValue = Math.floor((expValue * this.currentBattle.waveIndex) / 5 + 1);
-    // }
+    /**
+     * Lots of ME code still use this
+     * TODO: Change/remove this
+     */
+    if (useWaveIndexMultiplier) {
+      expValue = Math.floor((expValue * this.currentBattle.waveIndex) / 5 + 1);
+    }
 
     if (participantIds.size > 0) {
       if (
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -2744,7 +2744,7 @@ export default class BattleScene extends SceneBase {
   applyPartyExp(
     expValue: number,
     pokemonDefeated: boolean,
-    useWaveIndexMultiplier?: boolean,
+    _useWaveIndexMultiplier?: boolean,
     pokemonParticipantIds?: Set<number>,
   ): void {
     const participantIds = pokemonParticipantIds ?? this.currentBattle.playerParticipantIds;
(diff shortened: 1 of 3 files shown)


flx-sta (approved):
LGTM.

I'm wondering if we should use a generic `getExpFormula()` function instead and have a `game-constants` that's called `EXP_FORMULAR_GEN` with a value of `1` (or `2`,`3`,`4` or `I-IV`<sup>make your choice</sup>) and will select the correct one. That way it's easier to switch (if we ever have to).

```ts
type FormulaGen = "I-IV" | "V" /*...*/;

export function getExpFormula(gen: "I-IV" | "V", /*...*/) {
  switch (gen) {
    case "I-IV":
       return genOneThroughFourExpFormula(/*...*/);
    //...
    default:
      throw new Error(`Unknown formula ${gen}!`);
  }
}
```

Tempo-anon:
I considered that, but if any changes are going to be made with a PR then I don't see too much difference between changing a `I-IV` -> `V` vs changing something like a `genOneThroughFourExpFormula` -> `genFiveExpFormula` since it's functionally the same just with less characters being changed

flx-sta:
> I considered that, but if any changes are going to be made with a PR then I don't see too much difference between changing a `I-IV` -> `V` vs changing something like a `genOneThroughFourExpFormula` -> `genFiveExpFormula` since it's functionally the same just with less characters being changed

There is still 2 other benefits:
1. It's easier to spot which formula is being used by simply looking at those constant values
2. It's easier to *read* the code as it would state `getExpFormula()` instead of `getAbcdefghgtrenjgFormula()` (minor benefit)

src/battle-scene.ts
```diff
@@ -2757,7 +2757,10 @@ export default class BattleScene extends SceneBase {
     const nonFaintedPartyMembers = party.filter((p) => p.hp);
     const expPartyMembers = nonFaintedPartyMembers.filter((p) => p.level < this.getMaxExpLevel());
     const partyMemberExp: number[] = [];
-    // EXP value calculation is based off Pokemon.getExpValue
+    /**
+     * Lots of ME code still use this
+     * TODO: Change/remove this
```
innerthunder, line 2762:
```suggestion
     * @todo Change/remove this
```

innerthunder (approved):

src/battle-scene.ts
flx-sta: In this case it's fine to use `TODO:` since it's not actually a comment for a variable, function or class

PR #1125, opened 2025-05-03
[Addendum][AI-Rework][Test] Add matchers + Improve test utils

innerthunder:
## What are the changes the user will see?

N/A

## Why am I making these changes?

Existing test utils were a bit clunky, and I wanted to improve on them in support of testing specific attribute effect score bonuses.

## What are the changes from a developer perspective?

- Updated `getEnemyMoveChoices` util function to no longer require `Scene` and `MoveChoiceSet` arguments
  - The `BattleScene` is instead extrapolated from the `pokemon` input
  - `MoveChoiceSet` is now more explicitly typed as a `Record<MoveId, number>`
  - The function now generates a `MoveChoiceSet` internally and returns it.
- Created two new matchers on `EnemyPokemon` for common AI testing assertions. Both use `getEnemyMoveChoices` internally to evaluate move selection.
  - `toPreferSelectingMove`: to check whether the given enemy selects a specific move a plurality of the time.
  - `toNeverSelectMove`: to check whether the given enemy never selects a move or set of moves in any trial.

Existing tests have been updated to use the `toNeverSelectMove` matcher. `toPreferSelectingMove` is currently unused, but will be used for future tests on effect score bonuses.

## Screenshots/Videos

N/A

## How to test the changes?

`npm run test [enemy-command | low-accuracy | priority]`

## Checklist

- [n/a] Otherwise: **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?

- [x] Are all unit tests still passing? (`npm run test:silent`)
  - [x] Have I created new automated tests (`npm run test:create`) or updated existing tests related to the PR's changes?
- [n/a] Have I provided screenshots/videos of the changes (if applicable)?
  - [n/a] Have I made sure that any UI change works for both UI themes (dark and light)?

PR #1130, opened 2025-05-03
[Refactor] Initialize `waveData` and `turnData` in constructor

DayKev:
Nothing, ideally.

`Pokemon#turnData` and `Pokemon#waveData` could be `undefined` at times, which causes many issues (and was not properly communicated by the type system).

Both `turnData` and `waveData` are now initialized in the `Pokemon` constructor, and all the guards for them being `undefined` are removed.

## Checklist

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -156,8 +156,6 @@ import {
   BooleanHolder,
   NumberHolder,
   coerceArray,
-  deepCopy,
-  deepFreeze,
   fixedNumber,
   getEnumValues,
   isNil,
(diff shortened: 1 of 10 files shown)


src/field/pokemon.ts
```diff
@@ -4161,7 +4137,12 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
   }
 
   resetWaveData(): void {
-    this.waveData = deepCopy<PokemonWaveData>(defaultWaveData);
```
DayKev, line 4152: Changed due to breaking things: https://github.com/Despair-Games/poketernity/pull/1125#discussion_r2072342279

DayKev: Since the default values object wasn't used in more than one place any way there's no need to overcomplicate it.

Tempo-anon (approved):

innerthunder (approved):

PR #1132, opened 2025-05-03
[Refactor] Change `Status` from class to interface

DayKev:
N/A

General refactoring.

- `Pokemon#status` is now `protected`
- Added `Pokemon#setStatus`, `Pokemon#advanceStatusCounter` and `Pokemon#resetToxicTurnCounter`
- Renamed `Status#statusEffect` to `Status#effect`
- Added getters for `toxicTurnCount` and `sleepTurnsRemaining` in `Pokemon`
- Moved `Status` interface to `src/@types/Status.ts`

## How to test the changes?
`npm run test:silent`
Do various things involving status effects.

## Checklist

### DIFF
--- a/src/@types/Status.ts
+++ b/src/@types/Status.ts
@@ -1,0 +1,7 @@
+import type { StatusEffect } from "#enums/status-effect";
+
+export interface Status {
+  effect: StatusEffect;
+  toxicTurnCount: number;
+  sleepTurnsRemaining: number;
+}
--- a/src/@types/Status.ts
+++ b/src/@types/Status.ts
@@ -1,7 +1,19 @@
 import type { StatusEffect } from "#enums/status-effect";
 
+/** Contains fields related to {@linkcode StatusEffect}s */
 export interface Status {
-  effect: StatusEffect;
+  /** @see {@linkcode StatusEffect} */
+  readonly effect: StatusEffect;
+  /**
+   * Toxic damage is `1/16 max HP * toxicTurnCount`; increases by 1 per turn.
+   * Ignored if the effect is not {@linkcode StatusEffect.TOXIC}
+   * @defaultValue 0
+   */
   toxicTurnCount: number;
+  /**
+   * The pokemon wakes up when this is `0` and the {@linkcode effect} is {@linkcode StatusEffect.SLEEP}.
+   * Ignored if the effect is not sleep.
+   * @defaultValue 0
+   */
   sleepTurnsRemaining: number;
 }
--- a/src/data/abilities/ab-attrs/reduce-sleep-duration-ab-attr.ts
+++ b/src/data/abilities/ab-attrs/reduce-sleep-duration-ab-attr.ts
@@ -1,5 +1,4 @@
 import type { Pokemon } from "#app/field/pokemon";
-import type { NumberHolder } from "#app/utils/common-utils";
 import { AbAttrFlag } from "#enums/ab-attr-flag";
 import { StatusEffect } from "#enums/status-effect";
 import { AbAttr } from "./ab-attr";
@@ -27,14 +26,9 @@ export class ReduceSleepDurationAbAttr extends AbAttr {
    * - `[1]` - The number of turns remaining until the status is healed
    * @returns `true` if the ability was applied
    */
-  override apply(
-    _pokemon: Pokemon,
-    _simulated: boolean,
-    statusEffect: StatusEffect,
-    turnsRemaining: NumberHolder,
-  ): boolean {
+  override apply(pokemon: Pokemon, _simulated: boolean, statusEffect: StatusEffect): boolean {
     if (statusEffect === this.statusEffect) {
-      turnsRemaining.value -= 1;
+      pokemon.advanceStatusCounter();
       return true;
     }
 
--- a/src/data/mystery-encounters/utils/encounter-phase-utils.ts
+++ b/src/data/mystery-encounters/utils/encounter-phase-utils.ts
@@ -13,7 +13,6 @@ import type MysteryEncounterOption from "#app/data/mystery-encounters/mystery-en
 import { showEncounterText } from "#app/data/mystery-encounters/utils/encounter-dialogue-utils";
 import { getNatureName } from "#app/data/nature";
 import type PokemonSpecies from "#app/data/pokemon-species";
-import { Status } from "#app/data/status-effect";
 import type { TrainerConfig } from "#app/data/trainer-config";
 import { allTrainerConfigs } from "#app/data/trainer-configs/all-trainer-configs";
 import type { Variant } from "#app/data/variant";
@@ -269,7 +268,7 @@ export async function initBattleWithEnemyConfig(partyConfig: EnemyPartyConfig): 
 
     // Make sure basic data is clean
     enemyPokemon.hp = enemyPokemon.getMaxHp();
-    enemyPokemon.status = null;
+    enemyPokemon.resetStatus();
     enemyPokemon.passive = false;
 
     if (e < (doubleBattle ? 2 : 1)) {
@@ -350,7 +349,8 @@ export async function initBattleWithEnemyConfig(partyConfig: EnemyPartyConfig): 
           : statusEffects === StatusEffect.SLEEP
             ? 3
             : undefined;
-        enemyPokemon.status = new Status(status, 0, cureTurn);
+        // @ts-expect-error - `Pokemon#setStatus` is protected; TODO: change this?
+        enemyPokemon.setStatus(status, { sleepTurnsRemaining: cureTurn });
       }
 
       // Set summon data fields
--- a/src/data/status-effect.ts
+++ b/src/data/status-effect.ts
@@ -2,30 +2,6 @@ import { StatusEffect } from "#enums/status-effect";
 import type { ParseKeys } from "i18next";
 import i18next from "i18next";
 
-export class Status {
-  protected effect: StatusEffect;
-  /** Toxic damage is `1/16 max HP * toxicTurnCount` */
-  public toxicTurnCount: number = 0;
-  public sleepTurnsRemaining?: number;
-
-  constructor(effect: StatusEffect, toxicTurnCount: number = 0, sleepTurnsRemaining?: number) {
-    this.effect = effect;
-    this.toxicTurnCount = toxicTurnCount;
-    this.sleepTurnsRemaining = sleepTurnsRemaining;
-  }
-
-  get statusEffect(): StatusEffect {
-    return this.effect;
-  }
-
-  incrementTurn(): void {
-    this.toxicTurnCount++;
-    if (this.sleepTurnsRemaining) {
-      this.sleepTurnsRemaining--;
-    }
-  }
-}
-
 function getStatusEffectMessageKey(statusEffect: StatusEffect | undefined): string {
   switch (statusEffect) {
     case StatusEffect.POISON:
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -4066,7 +4066,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
 
   protected setStatus(
     effect: StatusEffect,
-    { toxicTurnCount = 0, sleepTurnsRemaining = 0 }: { toxicTurnCount?: number; sleepTurnsRemaining?: number },
+    { toxicTurnCount = 0, sleepTurnsRemaining = 0 }: Partial<Omit<Status, "effect">>,
   ): void {
     this.status = {
       effect,
@@ -4093,6 +4093,9 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
           break;
         }
         this.status.sleepTurnsRemaining--;
+        break;
+      default:
+        // intentionally left blank
         break;
     }
   }
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -21,7 +21,11 @@ import {
   SEMI_INVULNERABLE_BATTLER_TAG_TYPES,
   TRAPPED_BATTLER_TAG_TYPES,
 } from "#app/constants/battler-tag-constants";
-import { DYNAMAX_DAMAGE_TAKEN_FACTOR } from "#app/constants/game-constants";
+import {
+  DEFAULT_MAX_SLEEP_DURATION,
+  DEFAULT_MIN_SLEEP_DURATION,
+  DYNAMAX_DAMAGE_TAKEN_FACTOR,
+} from "#app/constants/game-constants";
 import type { AbAttr } from "#app/data/abilities/ab-attrs/ab-attr";
 import type { AddSecondStrikeAbAttr } from "#app/data/abilities/ab-attrs/add-second-strike-ab-attr";
 import type { AlliedFieldDamageReductionAbAttr } from "#app/data/abilities/ab-attrs/allied-field-damage-reduction-ab-attr";
@@ -4042,7 +4046,10 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     const sleepTurnsRemaining: NumberHolder = new NumberHolder(0);
 
     if (effect === StatusEffect.SLEEP) {
-      sleepTurnsRemaining.value = turnsRemaining !== 0 ? turnsRemaining : this.randSeedIntRange(2, 4);
+      sleepTurnsRemaining.value =
+        turnsRemaining !== 0
+          ? turnsRemaining
+          : this.randSeedIntRange(DEFAULT_MIN_SLEEP_DURATION, DEFAULT_MAX_SLEEP_DURATION);
 
       this.setFrameRate(4);
 
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -4006,6 +4006,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     return true;
   }
 
+  /** @todo Refactor this */
   trySetStatus(
     effect: StatusEffect,
     asPhase: boolean = false,
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -9,10 +9,18 @@ import type { FaintPhase } from "#app/phases/faint-phase";
 import type { AbilityFilterOptions } from "#app/@types/AbilityFilterOptions";
 import type { DamageCalculationResult } from "#app/@types/DamageCalculationResult";
 import type { DamageFunctionOptions } from "#app/@types/DamageFunctionOptions";
+import type { nil } from "#app/@types/nil";
 import type { PokemonTurnData } from "#app/@types/PokemonTurnData";
 import type { PokemonWaveData } from "#app/@types/PokemonWaveData";
+import type { Status } from "#app/@types/Status";
 import type { TurnMove } from "#app/@types/TurnMove";
 import type { AnySound } from "#app/audio-manager";
+import { WEAKEN_MOVE_SCREEN_ARENA_TAG_TYPES } from "#app/constants/arena-tag-constants";
+import {
+  CRIT_BOOST_BATTLER_TAG_TYPES,
+  SEMI_INVULNERABLE_BATTLER_TAG_TYPES,
+  TRAPPED_BATTLER_TAG_TYPES,
+} from "#app/constants/battler-tag-constants";
 import { DYNAMAX_DAMAGE_TAKEN_FACTOR } from "#app/constants/game-constants";
 import type { AbAttr } from "#app/data/abilities/ab-attrs/ab-attr";
 import type { AddSecondStrikeAbAttr } from "#app/data/abilities/ab-attrs/add-second-strike-ab-attr";
@@ -114,7 +122,7 @@ import { SpeciesFormChangeActiveTrigger } from "#app/data/species-form-change-tr
 import { SpeciesFormChangeMoveLearnedTrigger } from "#app/data/species-form-change-triggers/species-form-change-move-learned-trigger";
 import { SpeciesFormChangePostMoveTrigger } from "#app/data/species-form-change-triggers/species-form-change-post-move-trigger";
 import { SpeciesFormChangeStatusEffectTrigger } from "#app/data/species-form-change-triggers/species-form-change-status-effect-trigger";
-import { Status, getNonVolatileStatusEffects } from "#app/data/status-effect";
+import { getNonVolatileStatusEffects } from "#app/data/status-effect";
 import { tmPoolTiers, tmSpecies } from "#app/data/tms";
 import { getTypeDamageMultiplier, getTypeRgb, type TypeDamageMultiplier } from "#app/data/type";
 import { variantData, type Variant } from "#app/data/variant";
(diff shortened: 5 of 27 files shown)


src/field/pokemon.ts
```diff
@@ -4068,6 +4077,46 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    { toxicTurnCount = 0, sleepTurnsRemaining = 0 }: { toxicTurnCount?: number; sleepTurnsRemaining?: number },
+  ): void {
+    this.status = {
+      effect,
+      toxicTurnCount,
+      sleepTurnsRemaining,
+    };
+  }
+
+  public advanceStatusCounter(): void {
```
flx-sta, line 4085: We should stick to already used terms in the project, which would be `lapse` in this case

```suggestion
  public lapseStatusCounter(): void {
```

flx-sta: Also - NIT: Missing tsdocs

DayKev: Oops, forgot about the docs.

src/@types/Status.ts
```diff
@@ -0,0 +1,7 @@
+import type { StatusEffect } from "#enums/status-effect";
+
+export interface Status {
```
flx-sta, line 4: NIT: docs (Best case even for every property/key)

src/data/status-effect.ts
flx-sta, line 1: The rest of the file should probably be moved into `status-constants` and `status-utils` but possibly in another PR

src/field/pokemon.ts
```diff
@@ -447,6 +449,14 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     this._teraType = value;
   }
 
+  public get toxicTurnCount(): number {
```
flx-sta, line 439: NIT: doc (?)

src/field/pokemon.ts
```diff
@@ -447,6 +449,14 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    return this.status?.toxicTurnCount ?? 0;
+  }
+
+  public get sleepTurnsRemaining(): number {
```
flx-sta, line 448: NIT: doc (?)

src/field/pokemon.ts
```diff
@@ -4068,6 +4077,46 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+        if (Overrides.STATUS_ACTIVATION_OVERRIDE === false) {
+          this.status.sleepTurnsRemaining = 0;
+          break;
+        }
+        this.status.sleepTurnsRemaining--;
+        break;
+    }
+  }
+
+  public resetToxicTurnCounter(): void {
```
flx-sta, line 4111: NIT: docs

src/field/pokemon.ts
DayKev: This doesn't do the same thing though, because it also ticks up the toxic counter which is different from `[Battler|Arena]Tag`s.

flx-sta: I see. Keep it then

src/@types/Status.ts
```diff
@@ -0,0 +1,19 @@
...
+   * Ignored if the effect is not {@linkcode StatusEffect.TOXIC}
+   * @defaultValue 0
+   */
+  toxicTurnCount: number;
+  /**
+   * The pokemon wakes up when this is `0` and the {@linkcode effect} is {@linkcode StatusEffect.SLEEP}.
+   * Ignored if the effect is not sleep.
+   * @defaultValue 0
+   */
+  sleepTurnsRemaining: number;
```
innerthunder, line 18: Shouldn't these be optional?

src/phases/obtain-status-effect-phase.ts
```diff
@@ -38,9 +38,10 @@ export class ObtainStatusEffectPhase extends PokemonPhase {
 
   public override start(): void {
     const pokemon = this.getPokemon();
-    if (pokemon && !pokemon.status) {
+    if (pokemon && !pokemon.hasNonVolatileStatusEffect(false, true)) {
       if (pokemon.trySetStatus(this.statusEffect, false, this.sourcePokemon)) {
         if (this.turnsRemaining) {
+          // @ts-expect-error - `Pokemon#status` is protected; TODO: change this
           pokemon.status!.sleepTurnsRemaining = this.turnsRemaining;
```
innerthunder:
```suggestion
          /** @todo Change the accessibility of `Pokemon#status` */
          pokemon["status"]!.sleepTurnsRemaining = this.turnsRemaining;
```

Alternatively, you could make a `Pokemon` method to set the Pokemon's `sleepTurnsRemaining`

test/abilities/magic-bounce.test.ts
```diff
@@ -310,7 +310,7 @@ describe("Abilities - Magic Bounce", () => {
     // Turn 1 - thunder wave immunity test
     game.move.use(MoveId.THUNDER_WAVE);
     await game.toEndOfTurn();
-    expect(game.field.getPlayerPokemon().status).toBeUndefined();
+    expect(game.field.getPlayerPokemon().hasNonVolatileStatusEffect(false, true)).toBeFalsy();
```
innerthunder:
```suggestion
    expect(game.field.getPlayerPokemon()).toHaveStatusEffect(StatusEffect.NONE);
```

test/abilities/magic-bounce.test.ts
```diff
@@ -336,7 +336,7 @@ describe("Abilities - Magic Bounce", () => {
     vi.spyOn(opponent, "getAccuracyMultiplier").mockReturnValue(0);
     game.move.use(MoveId.SPORE);
     await game.toEndOfTurn();
-    expect(game.field.getPlayerPokemon().status).toBeUndefined();
+    expect(game.field.getPlayerPokemon().hasNonVolatileStatusEffect(false, true)).toBeFalsy();
```
innerthunder:
```suggestion
    expect(game.field.getPlayerPokemon()).toHaveStatusEffect(StatusEffect.NONE);
```

test/abilities/magic-bounce.test.ts
```diff
@@ -371,7 +371,7 @@ describe("Abilities - Magic Bounce", () => {
     game.setTurnOrder([BattlerIndex.ENEMY, BattlerIndex.PLAYER]);
     await game.toEndOfTurn();
     expect(game.field.getEnemyPokemon().getStatusEffect()).toBe(StatusEffect.TOXIC);
-    expect(game.field.getPlayerPokemon().status).toBeUndefined();
+    expect(game.field.getPlayerPokemon().hasNonVolatileStatusEffect(false, true)).toBeFalsy();
```
innerthunder:
```suggestion
    expect(game.field.getPlayerPokemon()).toHaveStatusEffect(StatusEffect.NONE);
```

test/moves/magic-coat.test.ts
```diff
@@ -289,7 +289,7 @@ describe("Moves - Magic Coat", () => {
     // Turn 1 - thunder wave immunity test
     game.move.use(MoveId.THUNDER_WAVE);
     await game.toEndOfTurn();
-    expect(game.field.getPlayerPokemon().status).toBeUndefined();
+    expect(game.field.getPlayerPokemon().hasNonVolatileStatusEffect(false, true)).toBeFalsy();
```
innerthunder:
```suggestion
    expect(game.field.getPlayerPokemon()).toHaveStatusEffect(StatusEffect.NONE);
```

test/moves/magic-coat.test.ts
```diff
@@ -315,7 +315,7 @@ describe("Moves - Magic Coat", () => {
     vi.spyOn(opponent, "getAccuracyMultiplier").mockReturnValue(0);
     game.move.use(MoveId.SPORE);
     await game.toEndOfTurn();
-    expect(game.field.getPlayerPokemon().status).toBeUndefined();
+    expect(game.field.getPlayerPokemon().hasNonVolatileStatusEffect(false, true)).toBeFalsy();
```
innerthunder:
```suggestion
    expect(game.field.getPlayerPokemon()).toHaveStatusEffect(StatusEffect.NONE);
```

test/moves/purify.test.ts
```diff
@@ -43,7 +42,8 @@ describe("Moves - Purify", () => {
     const playerPokemon: PlayerPokemon = game.scene.getPlayerPokemon()!;
 
     playerPokemon.hp = playerPokemon.getMaxHp() - 1;
-    enemyPokemon.status = new Status(StatusEffect.BURN);
+    enemyPokemon.trySetStatus(StatusEffect.BURN);
+    expect(enemyPokemon.hasStatusEffect(StatusEffect.BURN)).toBeTruthy();
```
innerthunder:
```suggestion
    expect(enemyPokemon).toHaveStatusEffect(StatusEffect.BURN);
```

test/moves/toxic.test.ts
```diff
@@ -47,7 +47,7 @@ describe("Moves - Toxic", () => {
     game.move.select(MoveId.TOXIC);
     await game.toEndOfTurn();
 
-    expect(game.scene.getEnemyPokemon()!.status).toBeUndefined();
+    expect(game.field.getEnemyPokemon().hasNonVolatileStatusEffect(false, true)).toBeFalsy();
```
innerthunder:
```suggestion
    expect(game.field.getEnemyPokemon()).toHaveStatusEffect(StatusEffect.NONE);
```

test/moves/toxic.test.ts
```diff
@@ -71,7 +71,7 @@ describe("Moves - Toxic", () => {
     game.setTurnOrder([BattlerIndex.ENEMY, BattlerIndex.PLAYER]);
     await game.toEndOfTurn();
 
-    expect(game.scene.getEnemyPokemon()!.status).toBeUndefined();
+    expect(game.field.getEnemyPokemon().hasNonVolatileStatusEffect(false, true)).toBeFalsy();
```
innerthunder:
```suggestion
    expect(game.field.getEnemyPokemon()).toHaveStatusEffect(StatusEffect.NONE);
```

test/mystery-encounter/encounters/fiery-fallout-encounter.test.ts
```diff
@@ -214,7 +213,8 @@ describe("Fiery Fallout - Mystery Encounter", () => {
 
       const party = scene.getPlayerParty();
       const lapras = party.find((pkm) => pkm.species.speciesId === SpeciesId.LAPRAS)!;
-      lapras.status = new Status(StatusEffect.POISON);
+      lapras.trySetStatus(StatusEffect.POISON);
+      expect(lapras.hasStatusEffect(StatusEffect.POISON)).toBeTruthy();
```
innerthunder:
```suggestion
      expect(lapras).toHaveStatusEffect(StatusEffect.POISON);
```

src/@types/Status.ts
DayKev: Since these fields are only checked if the appropriate status is set, it's easier to keep them mandatory and default to 0 to save on unnecessary null checks.

src/field/pokemon.ts
```diff
@@ -229,7 +230,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
   public ivs: number[];
   public nature: Nature;
   public moveset: PokemonMove[];
-  public status: Status | null;
+  protected status: Status | null = null;
```
flx-sta, line 237: Q: Why `null` and not `Status.NONE`? Has this always been that way?

src/field/pokemon.ts
```diff
@@ -425,6 +427,24 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     this._teraType = value;
   }
 
+  /**
+   * Toxic damage is `1/16 max HP * toxicTurnCount`; increases by 1 per turn.
+   * Ignored if the effect is not {@linkcode StatusEffect.TOXIC}
+   * @defaultValue 0
```
flx-sta, line 437: `@defaultValue` is new to me. 
https://tsdoc.org/pages/tags/defaultvalue/
_I feel like a newbie about tsdoc with all these new tags_

begs the question if we wanna stick to using `@linkcode` in the long run as it's not a tsdoc standard

src/field/pokemon.ts
```diff
@@ -425,6 +427,24 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   */
+  public get toxicTurnCount(): number {
+    return this.status?.toxicTurnCount ?? 0;
+  }
+
+  /**
+   * The pokemon wakes up when this is `0` and the {@linkcode effect} is {@linkcode StatusEffect.SLEEP}.
+   * Ignored if the effect is not sleep.
```
flx-sta, line 445: NIT: Match `@linkcode` from `toxicTurnCount`:
```suggestion
   * Ignored if the effect is not {@linkcode StatusEffect.SLEEP}.
```

src/field/pokemon.ts
```diff
@@ -4018,10 +4039,10 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
       return true;
     }
 
-    let sleepTurnsRemaining: NumberHolder;
+    const sleepTurnsRemaining: NumberHolder = new NumberHolder(0);
 
     if (effect === StatusEffect.SLEEP) {
-      sleepTurnsRemaining = new NumberHolder(turnsRemaining !== 0 ? turnsRemaining : this.randSeedIntRange(2, 4));
+      sleepTurnsRemaining.value = turnsRemaining !== 0 ? turnsRemaining : this.randSeedIntRange(2, 4);
```
flx-sta: Q: Where is the `2-4` range coming from? Seems a bit like *magic numbers*.

src/field/pokemon.ts
```diff
@@ -4044,6 +4064,47 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     return true;
   }
 
+  protected setStatus(
+    effect: StatusEffect,
+    { toxicTurnCount = 0, sleepTurnsRemaining = 0 }: { toxicTurnCount?: number; sleepTurnsRemaining?: number },
```
flx-sta: A possible alternative for the type would be `Partial<Omit<Status, "effect">>`.
- `Omit` is a helper type to remove 1-n keys from the references type/interface.
- `Partial` to make all keys optional.

```suggestion
    { toxicTurnCount = 0, sleepTurnsRemaining = 0 }: Partial<Omit<Status, "effect">>,
```

src/field/pokemon.ts
```diff
@@ -4044,6 +4064,47 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+  ): void {
+    this.status = {
+      effect,
+      toxicTurnCount,
+      sleepTurnsRemaining,
+    };
+  }
+
+  public advanceStatusCounter(): void {
+    if (!this.status) {
```
flx-sta, line 4086: When is status not set, it's a required param, right?
I'd prefer to use a more descriptive condition here:
```suggestion
    if (this.status === StatusEffect.NONE) {
```

src/field/pokemon.ts
```diff
@@ -4044,6 +4064,47 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+          break;
+        }
+        if (Overrides.STATUS_ACTIVATION_OVERRIDE === false) {
+          this.status.sleepTurnsRemaining = 0;
+          break;
+        }
+        this.status.sleepTurnsRemaining--;
+        break;
+    }
+  }
```
flx-sta: NIT: It's generally recommended to always have a `default` block in a switch-case.
```suggestion
    switch (this.status.effect) {
      case StatusEffect.TOXIC:
        this.status.toxicTurnCount++;
        break;
      case StatusEffect.SLEEP:
        if (Overrides.STATUS_ACTIVATION_OVERRIDE === true) {
          this.status.sleepTurnsRemaining = Math.max(this.status.sleepTurnsRemaining, 1);
          break;
        }
        if (Overrides.STATUS_ACTIVATION_OVERRIDE === false) {
          this.status.sleepTurnsRemaining = 0;
          break;
        }
        this.status.sleepTurnsRemaining--;
        break;
      default:
        break;
    }
  }
```

src/field/pokemon.ts
DayKev: Yeah, I only added the `= null` default there (looks like GitHub's change display isn't the most accurate lol).

src/field/pokemon.ts
DayKev: I think for now we should stick with `@linkcode` since it changes how it displays in VSCode and the docs generator supports it (typedoc is fairly permissive).

src/field/pokemon.ts
DayKev: That's already in the previous line which is why I didn't add it again there.

src/field/pokemon.ts
DayKev: It's the mainline numbers for Gen 5+. I'll add constants for them.

src/field/pokemon.ts
DayKev: Oh that's useful.

src/field/pokemon.ts
DayKev: Currently `Pokemon#status` is set to `null` if the pokemon doesn't have a status effect, so that wouldn't work.

src/field/pokemon.ts
flx-sta: Anything speaking against to add the other default ?

DayKev: Dunno, would be a separate refactor though.

src/data/abilities/ab-attrs/reduce-sleep-duration-ab-attr.ts
```diff
@@ -27,14 +26,9 @@ export class ReduceSleepDurationAbAttr extends AbAttr {
...
-  override apply(
-    _pokemon: Pokemon,
-    _simulated: boolean,
-    statusEffect: StatusEffect,
-    turnsRemaining: NumberHolder,
-  ): boolean {
+  override apply(pokemon: Pokemon, _simulated: boolean, statusEffect: StatusEffect): boolean {
     if (statusEffect === this.statusEffect) {
-      turnsRemaining.value -= 1;
+      pokemon.advanceStatusCounter();
```
MokaStitcher, line 31: so before this used a NumberHolder and now it's a simple nmber. Was the NumberHolder not needed?

src/data/mystery-encounters/utils/encounter-phase-utils.ts
```diff
@@ -350,7 +349,8 @@ export async function initBattleWithEnemyConfig(partyConfig: EnemyPartyConfig):
           : statusEffects === StatusEffect.SLEEP
             ? 3
             : undefined;
-        enemyPokemon.status = new Status(status, 0, cureTurn);
+        // @ts-expect-error - `Pokemon#setStatus` is protected; TODO: change this?
```
MokaStitcher, line 352: hmm... Is there a way to call trySetStatus and force it to ignore any immunities?
(assuming wanting to ignore immunities is why setStatus is being called here rather than trySetStatus)

src/data/status-effect.ts
MokaStitcher, line 1: where do we draw the line as what `utils` are? And would the new content of this file count as just status utils?

src/phases/move-phase.ts
```diff
@@ -268,21 +268,8 @@ export class MovePhase extends BattlePhase {
...
           applyMoveAttrs(BypassSleepAttr, this.pokemon, null, this.move.getMove());
-          const turnsRemaining = new NumberHolder(this.pokemon.status!.sleepTurnsRemaining ?? 0);
-          applyAbAttrs<ReduceSleepDurationAbAttr>(
-            AbAttrFlag.REDUCE_SLEEP_DURATION,
-            this.pokemon,
-            false,
-            statusEffect,
-            turnsRemaining,
-          );
-          if (Overrides.STATUS_ACTIVATION_OVERRIDE === true) {
```
MokaStitcher, line 279: why are we removing the status activation override code here? Is that why the NumberHolder was used before?

Not sure I agree with removing it, but if others do, the documentation for that override should be updated to reflect it

test/data/status_effect.test.ts
```diff
@@ -369,7 +368,8 @@ describe("Status Effects", () => {
       await game.classicMode.startBattle([SpeciesId.FEEBAS]);
 
       const player = game.scene.getPlayerPokemon()!;
-      player.status = new Status(StatusEffect.SLEEP, 0, 4);
+      // @ts-expect-error - `Pokemon#setStatus` is protected
+      player.setStatus(StatusEffect.SLEEP, { sleepTurnsRemaining: 4 });
```
MokaStitcher: why not use `trySetStatus`?

MokaStitcher (commented):
some questions

src/data/abilities/ab-attrs/reduce-sleep-duration-ab-attr.ts
DayKev: Previously a numberholder was created with the current value of the `status.sleepTurnsRemaining` field, then this decremented that by one and the value of the numberholder was passed back to that field. Now it calls `advanceStatusCounter` which directly updates the field without the need for the "middleman".

src/phases/move-phase.ts
DayKev: The override is handled in a different place now (in `advanceStatusCounter`).

test/data/status_effect.test.ts
DayKev: True, I guess that works. I was just replicating the existing behavior of the test.

src/data/mystery-encounters/utils/encounter-phase-utils.ts
DayKev: Hm, it might be reasonable to just use `trySetStatus` here... I'm not sure if it was necessary for the status to be manually set previously.

MokaStitcher: It may be... If we want an event that forces a status on mons that would normally be immune, so I'm not sure. I don't think any current me does that though 🤔 
As long as the documentation makes the limitation clear it should be fine to change it

src/phases/move-phase.ts
MokaStitcher: Ah sorry, guess I missed it, I'll take another look

flx-sta (dismissed):

src/modifier/modifier-type.ts
```diff
@@ -472,7 +473,10 @@ export class PokemonStatusHealModifierType extends PokemonModifierType {
       iconImage,
       (_type, args) => new PokemonStatusHealModifier(this, (args[0] as PlayerPokemon).id),
       (pokemon: PlayerPokemon) => {
-        if (!pokemon.hp || (!pokemon.status && !pokemon.getTag(BattlerTagType.CONFUSED))) {
+        if (
+          !pokemon.hp
+          || (!pokemon.hasNonVolatileStatusEffect(false, true) && !pokemon.getTag(BattlerTagType.CONFUSED))
+        ) {
```
PigeonBar:
```suggestion
        if (!pokemon.hp || !pokemon.hasNonVolatileStatusEffect(true, true)) {
```

src/modifier/modifier-type.ts
```diff
@@ -405,7 +405,8 @@ export class PokemonHpRestoreModifierType extends PokemonModifierType {
           if (
             !pokemon.hp
             || (pokemon.isFullHp()
-              && (!this.healStatus || (!pokemon.status && !pokemon.getTag(BattlerTagType.CONFUSED))))
+              && (!this.healStatus
+                || (!pokemon.hasNonVolatileStatusEffect(false, true) && !pokemon.getTag(BattlerTagType.CONFUSED))))
```
PigeonBar:
```suggestion
              && (!this.healStatus || !pokemon.hasNonVolatileStatusEffect(true, true)))
```

PigeonBar (dismissed):

src/data/status-effect.ts
flx-sta: Generally a class should be living alone in it's file (carrying the same name).
Any "utility-ish" functions only being used by the class can live in the same file but exported ones should be in a `utils` file. 

----

That's how I personally *draw the line*

PR #1134, opened 2025-05-04
[Refactor] Rename `getFieldPokemonByBattlerIndex`

DayKev:
Also make param required

N/A

The function name is a bit long, and the `battlerIndex` parameter should be required.

- `BattleScene#getFieldPokemonByBattlerIndex` renamed to `BattleScene#getPokemonByBattlerIndex` (only field pokemon can have a battler index).
- Made `battlerIndex` a required parameter.

## Checklist

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -1,3 +1,4 @@
+import type { AbstractConstructor } from "#app/@types/AbstractConstructor";
 import type { HeldModifierConfig } from "#app/@types/HeldModifierConfig";
 import type { Localizable } from "#app/@types/locales";
 import type { ModifierPredicate } from "#app/@types/ModifierPredicate";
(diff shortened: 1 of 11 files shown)


Tempo-anon (dismissed):

src/battle-scene.ts
```diff
@@ -831,9 +831,9 @@ export default class BattleScene extends SceneBase {
   /**
    * Returns the Pokemon currently on the field that has a certain battler index, or `undefined` if no such Pokemon exists.
    * This function is allowed to return non-active (e.g., fainted) Pokemon.
-   * @param battlerIndex The battler index to search for.
+   * @param battlerIndex - The battler index to search for.
```
innerthunder:
```suggestion
   * @param battlerIndex - The {@linkcode BattlerIndex} to search for.
```

innerthunder (dismissed):

PR #1138, opened 2025-05-06
[Bug] Fix parsing of local save data

DayKev:
Data saving/loading will no longer break if non-latin characters end up in the local save data somehow.

Crash fix.

The code path for saving and loading local save data now uses `encodeURIComponent`/`decodeURIComponent`.

## How to test the changes?
~I forgot to save the system/session data from before the fix, so I have no idea what caused it. This save is from after the fix, on the wave that was previously broken. I'm not sure if it's possible to observe the bug from it though, it was system data that was broken.~

~[sessionData1_Guest - local save encoding bug.pkty.txt](https://github.com/user-attachments/files/20053308/sessionData1_Guest.-.local.save.encoding.bug.pkty.txt)~

The following session save triggers the bug after you complete the wave (system data was freshly imported from the "everything" dev save in `test/test-utils/resources/saves/everything.pkty.txt`):

[sessionData1_Guest - local save encoding bug 2.pkty.txt](https://github.com/user-attachments/files/20053742/sessionData1_Guest.-.local.save.encoding.bug.2.pkty.txt)

## Checklist

### DIFF
--- a/src/system/game-data.ts
+++ b/src/system/game-data.ts
@@ -104,15 +104,15 @@ export function getDataTypeKey(dataType: GameDataType, slotId: number = 0): stri
 }
 
 export function encrypt(data: string, bypassLogin: boolean): string {
-  return (bypassLogin ? (data: string) => btoa(data) : (data: string) => AES.encrypt(data, saveKey))(
-    data,
-  ) as unknown as string; // TODO: is this correct?
+  const localFunc = (data: string): string => btoa(encodeURIComponent(data));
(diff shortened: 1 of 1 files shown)


innerthunder (approved):

Tempo-anon (commented):
I feel like a comment in these functions to explain what `localFunc` is doing (preventing crashes of non latin chars) would be good

flx-sta (approved):
LGTM but it might be worth adding a simple automated test to validate this

PR #1139, opened 2025-05-06
[Dev] Update `compilerOptions.target` to `ES2022` from `ES2020`

DayKev:
N/A?

~While apparently we're supposed to be using `nodenext`, that causes a billion errors so would potentially require extensive refactoring.~
~`npm run typecheck` output when using `nodenext`: "Found 3684 errors in 559 files."~
> ~`node16`, `node18`, and `nodenext` are the only correct `module` options for all apps and libraries that are intended to run in Node.js v12 or later, whether they use ES modules or not.~

Top level `await` now supported.
> [!NOTE]
> I tried also updating `compilerOptions.target` to `ES2022` but that caused a strange bug with session loading where movesets would get deleted for all pokemon. Need to investigate further to change that. 

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "@eslint/js": "^9.24.0",
     "@hpcc-js/wasm": "^2.22.4",
     "@stylistic/eslint-plugin-ts": "^4.2.0",
+    "@types/crypto-js": "^4.2.2",
     "@types/jsdom": "^21.1.7",
     "@types/node": "^22.14.1",
     "@typescript-eslint/eslint-plugin": "^8.29.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,6 +27,7 @@
         "@eslint/js": "^9.24.0",
         "@hpcc-js/wasm": "^2.22.4",
         "@stylistic/eslint-plugin-ts": "^4.2.0",
+        "@types/crypto-js": "^4.2.2",
         "@types/jsdom": "^21.1.7",
         "@types/node": "^22.14.1",
         "@typescript-eslint/eslint-plugin": "^8.29.1",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -31,6 +31,7 @@
     "@eslint/js": "^9.24.0",
     "@hpcc-js/wasm": "^2.22.4",
     "@stylistic/eslint-plugin-ts": "^4.2.0",
+    "@types/crypto-js": "^4.2.2",
```
MokaStitcher, line 34: what is this?
Can you reply here or maybe better, update the description to document why this dependency was added for future reference? (Although I realize now that it's.a dev dependency, so not that important)

Tempo-anon (approved):

flx-sta (approved):

flx-sta:
> While apparently we're supposed to be using nodenext

Why would we use use `nodenext`? This isn't a node application

DayKev:
> > While apparently we're supposed to be using nodenext
> 
> Why would we use use `nodenext`? This isn't a node application

Oh, we don't use node?

flx-sta:
> Oh, we don't use node?

We obviously use node for compilation, bundling and development but the app is targeted for web. Node isn’t for web but for standalone. We run in ECMA Script aka ES aka modern JS

DayKev:
Ah, I see.

MokaStitcher (commented):
oops

package.json
DayKev: We're using the package in `game-data.ts` and `gameManager.ts` but we didn't have the types installed so it meant we had no type data for it:
image

MokaStitcher: Oh 😂
So yeah, definitely good to document that especially since it's not directly related to the PR itself
Thank you

PR #1140, opened 2025-05-06
[Dev] Enable some `strict` options in `tsconfig.json`

DayKev:
N/A

Safer development.

The following options that are part of the [strict](https://www.typescriptlang.org/tsconfig/#strict) option set are enabled: `alwaysStrict`, `strictBindCallApply`, `strictBuiltinIteratorReturn`, `noImplicitThis`.
Additionally, `useUnknownInCatchVariables` could be enabled relatively easily if we want it.
The rest (`strictFunctionTypes`, `strictPropertyInitialization` and `noImplicitAny`) would require a lot of code changes (`npm run typecheck -> Found 3675 errors in 285 files.` when all 3 of them are enabled), and so should be done separately.

## Checklist

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,10 +5,20 @@
     "moduleResolution": "bundler",
     "resolveJsonModule": true,
     "esModuleInterop": true,
-    "strictNullChecks": true,
     "noImplicitOverride": true,
     "sourceMap": false,
+    // #region `strict` options
     "strict": false,
+    "alwaysStrict": true,
+    "strictBindCallApply": true,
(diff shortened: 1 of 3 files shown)


flx-sta (approved):

tsconfig.json
flx-sta, line 1: I'm surprised you can use `//` comments in a json file.
Guess internally it's a jsonc

flx-sta (approved):

tsconfig.json
DayKev: Seems so.

Tempo-anon (approved):

PR #1152, opened 2025-05-08
[Bug] Fix timing of Terastallization phase

### LINKED ISSUE #988: Fix Terastallization timing
- [x] Requires #885 
- [ ] Remove temporary code from `TerastallizationPhase`
- [ ] Related: #1085

### LINKED ISSUE #1085: [P2] Instant tera allows a slower Pokemon to move first
### Describe the bug

Discovered while testing #1084 

~~When an enemy trainer instant tera's, they will go first even if they are slower~~ Terastallization will always make a Pokemon go first

### Reproduction

Add a gen 9 gym leader to the town boss pool
```ts
const trainerPool: Record<BiomePoolTier, TrainerType[]> = {
  [BiomePoolTier.COMMON]: [TrainerType.YOUNGSTER],
  [BiomePoolTier.UNCOMMON]: [],
  [BiomePoolTier.RARE]: [],
  [BiomePoolTier.SUPER_RARE]: [],
  [BiomePoolTier.ULTRA_RARE]: [],
  [BiomePoolTier.BOSS]: [TrainerType.KATY],
  [BiomePoolTier.BOSS_RARE]: [],
  [BiomePoolTier.BOSS_SUPER_RARE]: [],
  [BiomePoolTier.BOSS_ULTRA_RARE]: [],
};
```

Jump to wave 20 with a high level Pokemon
```ts
const overrides = {
  STARTING_WAVE_OVERRIDE:20,
  STARTING_LEVEL_OVERRIDE:50,
}
```

Observe that they go first

Test 2:
```ts
const overrides = {
  STARTING_MODIFIER_OVERRIDE: [{name: "TERA_ORB"}],
  ENEMY_SPECIES_OVERRIDE:SpeciesId.REGIELEKI,
  ENEMY_LEVEL_OVERRIDE:1000,
}
```
Tera and use an attack and observe the player going first even if they are low level

### Expected behavior

Even if a Pokemon terastalizes they should obey the turn order

### Screenshots / Videos

Proof enemy speeds are lower
Image

My weedle's 70 speed
Image

A non tera Pokemon gets outsped and KO'd
Image

When a Pokemon tera's they can outspeed and damage me before I get the KO
Image

Image

Test 2:

Image

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

DayKev:
Terastallization will happen after switches and before moves.

Fixes #1085 
Resolves #988 

- Turn commands are sorted by priority via a mapping of command to priority value.
- When a Tera command is handled by the turn manager, a Terastallization Phase is unshifted and then a new Fight command is added to the turn queue for that Pokémon to be handled later. Previously both a Terastallization Phase and a Move Phase would be queued at the same time.
- The check for the Pokémon being fainted is removed from Terastallization Phase as it should no longer be necessary.

## How to test the changes?
Use the following overrides:
```ts
const overrides = {
  ENEMY_LEVEL_OVERRIDE: 20,
  ENEMY_SPECIES_OVERRIDE: SpeciesId.REGIELEKI,
  STARTING_MODIFIER_OVERRIDE: [{name: "TERA_ORB"}],
```
Start with 3 pokemon. Have one Terastallize and use a move, and the other switch to your third pokemon. First the switch should happen, then the other pokemon should Terastallize, then one of the enemies should use a move.

## Checklist

### DIFF
--- a/src/turn-command-manager.ts
+++ b/src/turn-command-manager.ts
@@ -36,6 +36,7 @@ import { PhaseId } from "#enums/phase-id";
 import { Stat } from "#enums/stat";
 import { SwitchType } from "#enums/switch-type";
 
+/** Lower number = lower priority */
 const COMMAND_PRIORITY_MAP = {
   [BattleCommand.FIGHT]: 0,
   [BattleCommand.TERA]: 1,
@@ -426,6 +427,13 @@ export class TurnCommandManager {
     }
   }
 
+  /**
+   * Validates a given {@linkcode BattleCommand.TERA | Tera} command and
+   * schedules a {@linkcode TerastallizationPhase} if valid.
+   * Then queues a new {@linkcode BattleCommand.FIGHT | Fight} command.
+   * @param turnCommand - The {@linkcode TurnCommand} to validate
+   * @returns Whether the command was successful
+   */
   private handleTeraCommand(turnCommand: TurnCommand): boolean {
     const { pokemon } = turnCommand;
     if (!pokemon.isActive(true)) {
--- a/src/turn-command-manager.ts
+++ b/src/turn-command-manager.ts
@@ -36,6 +36,14 @@ import { PhaseId } from "#enums/phase-id";
 import { Stat } from "#enums/stat";
 import { SwitchType } from "#enums/switch-type";
 
+const COMMAND_PRIORITY_MAP = {
+  [BattleCommand.FIGHT]: 0,
+  [BattleCommand.TERA]: 1,
+  [BattleCommand.POKEMON]: 2,
+  [BattleCommand.BALL]: 3,
+  [BattleCommand.RUN]: 4,
+} as const;
+
 /**
  * Interface representing an action taken by a Pokemon for the turn.
  * Encompasses both Player and Enemy commands.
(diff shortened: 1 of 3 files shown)


src/turn-command-manager.ts
```diff
@@ -36,6 +36,14 @@ import { PhaseId } from "#enums/phase-id";
 import { Stat } from "#enums/stat";
 import { SwitchType } from "#enums/switch-type";
 
+const COMMAND_PRIORITY_MAP = {
```
Tempo-anon, line 40: nit: Might want to toss a comment here about how higher number = more priority just to be clear

src/turn-command-manager.ts
```diff
@@ -418,14 +426,29 @@ export class TurnCommandManager {
     }
   }
 
+  private handleTeraCommand(turnCommand: TurnCommand): boolean {
```
Tempo-anon, line 437: nit: Should add ts doc

Tempo-anon (commented):
Another way to manually test would be to fight against a trainer that uses tera like Katy

src/turn-command-manager.ts
DayKev: Ah I forgot, got distracted trying to figure out something that turned out to not even be a bug lol.

flx-sta (approved):

innerthunder (approved):

PR #1153, opened 2025-05-09
[Dev] Add a `FORCE_ENEMY_TERA_OVERRIDE`

PigeonBar:

If the override is enabled, then every enemy Pokemon (including wild Pokemon) will Terastallize on the first turn that it chooses to use a move.

Making it easier to test things while working on Terastallization (e.g., testing scenarios when both sides Terastallize on the same turn).

- If `FORCE_ENEMY_TERA_OVERRIDE` is `true`, then the enemy Pokemon will select `BattleCommand.TERA` instead of `BattleCommand.FIGHT` during `EnemyCommandPhase`.
- As a proof of concept, one of the Tera Blast tests is rewritten to use this override.

## Screenshots/Videos

https://github.com/user-attachments/assets/0f16dfe5-751c-4c50-b729-3f1d7b92d40d

## How to test the changes?

```
  FORCE_ENEMY_TERA_OVERRIDE: true,
```

## Checklist

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -401,7 +401,8 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
   public get teraType(): ElementalType {
     if (this.isPlayer() && Overrides.TERA_TYPE_OVERRIDE !== ElementalType.UNKNOWN) {
       return Overrides.TERA_TYPE_OVERRIDE;
-    } else if (this.isEnemy() && Overrides.ENEMY_TERA_TYPE_OVERRIDE !== ElementalType.UNKNOWN) {
+    }
+    if (this.isEnemy() && Overrides.ENEMY_TERA_TYPE_OVERRIDE !== ElementalType.UNKNOWN) {
       return Overrides.ENEMY_TERA_TYPE_OVERRIDE;
     }
 
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -399,6 +399,12 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
    * - Shedinja is always Bug
    */
   public get teraType(): ElementalType {
+    if (this.isPlayer() && Overrides.TERA_TYPE_OVERRIDE !== ElementalType.UNKNOWN) {
+      return Overrides.TERA_TYPE_OVERRIDE;
+    } else if (this.isEnemy() && Overrides.ENEMY_TERA_TYPE_OVERRIDE !== ElementalType.UNKNOWN) {
+      return Overrides.ENEMY_TERA_TYPE_OVERRIDE;
+    }
+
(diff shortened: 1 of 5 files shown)


innerthunder (dismissed):

src/overrides.ts
```diff
@@ -141,6 +141,11 @@ class DefaultOverrides {
    * - Infatuated: set to `true` to force its activation, set to `false` to do the opposite
    */
   readonly STATUS_ACTIVATION_OVERRIDE: boolean | null = null;
+  /**
+   * If `true`, every enemy Pokemon Terastallizes on the first turn that it decides to use a move.
+   * If `false`, this override is ignored.
+   */
+  readonly FORCE_ENEMY_TERA_OVERRIDE: boolean = false;
```
Tempo-anon: Is there also an override for the enemy's tera type? I think right now it'll randomly pick one of their types

Tempo-anon (dismissed):

src/overrides.ts
PigeonBar: Overrides for Tera type are now also included in this PR.

src/field/pokemon.ts
```diff
@@ -399,6 +399,12 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
    * - Shedinja is always Bug
    */
   public get teraType(): ElementalType {
+    if (this.isPlayer() && Overrides.TERA_TYPE_OVERRIDE !== ElementalType.UNKNOWN) {
+      return Overrides.TERA_TYPE_OVERRIDE;
+    } else if (this.isEnemy() && Overrides.ENEMY_TERA_TYPE_OVERRIDE !== ElementalType.UNKNOWN) {
```
```suggestion
    }
    if (this.isEnemy() && Overrides.ENEMY_TERA_TYPE_OVERRIDE !== ElementalType.UNKNOWN) {
```

DayKev (dismissed):

src/field/pokemon.ts
Tempo-anon: I agree that not having the `else` makes more sense stylistically

Tempo-anon (dismissed):

MokaStitcher:
Could it be useful at some point to have it be 3 states
- `true` to force tera
- `false` to prevent tera
- null for default behavior?

Similar to some other overrides

flx-sta (dismissed):

PigeonBar:
> Could it be useful at some point to have it be 3 states
> 
>     * `true` to force tera
> 
>     * `false` to prevent tera
> 
>     * null for default behavior?
> 
> 
> Similar to some other overrides

I don't think we need an override to prevent Tera since enemy Tera is limited to just a few specific enemy trainers.

DayKev (approved):

innerthunder (approved):

PR #1156, opened 2025-05-09
[Dev] Replace ESLint with Biome (mostly)

### LINKED ISSUE #1135: [Dev] Replace ESLint with Biome
Reason: Biome is ~15x faster than base ESLint (ref: https://github.com/biomejs/biome/tree/main/benchmark#linting) 

Main site: https://biomejs.dev/

This issue for discussion/demonstration of potential rule additions/changes.

## "Safe" fixes

[noInferrableTypes](https://biomejs.dev/linter/rules/no-inferrable-types)

<details><summary>noInferrableTypes</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[useFlatMap](https://biomejs.dev/linter/rules/use-flat-map)

<details><summary>useFlatMap</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[useNumberNamespace](https://biomejs.dev/linter/rules/use-number-namespace)

<details><summary>useNumberNamespace</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[noBannedTypes](https://biomejs.dev/linter/rules/no-banned-types)

<details><summary>noBannedTypes</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[noMultipleSpacesInRegularExpressionLiterals](https://biomejs.dev/linter/rules/no-multiple-spaces-in-regular-expression-literals)

<details><summary>noMultipleSpacesInRegularExpressionLiterals</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[useArrowFunction](https://biomejs.dev/linter/rules/use-arrow-function)

<details><summary>useArrowFunction</summary>
<p>

Before:

Image

After:

Image

</p>
</details>

[noUselessThisAlias](https://biomejs.dev/linter/rules/no-useless-this-alias)

<details><summary>noUselessThisAlias</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[noUselessRename](https://biomejs.dev/linter/rules/no-useless-rename)

<details><summary>noUselessRename</summary>
<p>

Before:

Image

</p>
</details> 

## "Unsafe" fixes (requires manual application and double checking, though some of these seem pretty safe imo)

[useNodejsImportProtocol](https://biomejs.dev/linter/rules/use-nodejs-import-protocol)

<details><summary>useNodejsImportProtocol</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[noUselessElse](https://biomejs.dev/linter/rules/no-useless-else)

<details><summary>noUselessElse</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[noSwitchDeclarations](https://biomejs.dev/linter/rules/no-switch-declarations)

<details><summary>noSwitchDeclarations</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[useOptionalChain](https://biomejs.dev/linter/rules/use-optional-chain)

<details><summary>useOptionalChain</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[noUnusedTemplateLiteral](https://biomejs.dev/linter/rules/no-unused-template-literal)

<details><summary>noUnusedTemplateLiteral</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[noUselessConstructor](https://biomejs.dev/linter/rules/no-useless-constructor)

<details><summary>noUselessConstructor</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[noUselessTernary](https://biomejs.dev/linter/rules/no-useless-ternary)

Needs additional hand-holding (it adds useless `!!` sometimes)

<details><summary>noUselessTernary</summary>
<p>

Before:

Image

After:

Image

</p>
</details>

[noExtraBooleanCast](https://biomejs.dev/linter/rules/no-extra-boolean-cast)

<details><summary>noExtraBooleanCast</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[noGlobalIsNan](https://biomejs.dev/linter/rules/no-global-is-nan)

<details><summary>noGlobalIsNan</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[useIsArray](https://biomejs.dev/linter/rules/use-is-array)

<details><summary>useIsArray</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[noConfusingVoidType](https://biomejs.dev/linter/rules/no-confusing-void-type)

<details><summary>noConfusingVoidType</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[useDefaultParameterLast](https://biomejs.dev/linter/rules/use-default-parameter-last)

Note: Fix can cause parameters to become untyped if `noInferrableTypes` has already been applied first

<details><summary>useDefaultParameterLast</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[noDelete](https://biomejs.dev/linter/rules/no-delete)

<details><summary>noDelete</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[useBlockStatements](https://biomejs.dev/linter/rules/use-block-statements)

<details><summary>useBlockStatements</summary>
<p>

Before:

Image

After:

Image

</p>
</details> 

[noUnnecessaryContinue](https://biomejs.dev/linter/rules/no-unnecessary-continue)

Fix can leave behind useless code, needs manual intervention

<details><summary>noUnnecessaryContinue</summary>
<p>

Before:

Image

After:

Image

</p>
</details>

## Rules with no fixers

https://github.com/Despair-Games/poketernity/issues/1135#issuecomment-2858931463

DayKev:
N/A

Biome is a lot faster than ESLint.
Resolves #1135 (rule examples are in there)

ESLint is now only run as a GitHub workflow for the rules that don't exist [yet] in Biome. The VSCode extension will still work as well.
Current rules (updated): https://github.com/Despair-Games/poketernity/blob/710fe0ab3e2ceb0d8618293371e535a5271d004e/biome.jsonc
`noInferrableTypes` was not enabled from the list in the issue.
Some rules are set as "warn" or "info" for now.

## Checklist


Tempo-anon (dismissed):
trust

.github/workflows/biome.yml
```diff
@@ -0,0 +1,36 @@
...
+  pull_request:
+    branches:
+      - main
+      - beta
+  merge_group:
+    types: [checks_requested]
+
+jobs:
+  biome-lint:
+    name: Lint changes with Biome
```
MokaStitcher: not sure I fully understand that name

.github/workflows/eslint.yml
```diff
@@ -1,38 +1,36 @@
...
+      - main
+      - beta
   merge_group:
     types: [checks_requested]
 
 jobs:
-  run-linters: # Define a job named "run-linters"
-    name: Run linters # Human-readable name for the job
+  run-eslint:
+    name: Lint changes with ESLint
```
MokaStitcher: same

biome.jsonc
```diff
@@ -0,0 +1,119 @@
...
+        "noDoubleEquals": "error",
+        "noExplicitAny": "off", // TODO: Refactor and make this an error
+        "noAssignInExpressions": "warn",
+        "noPrototypeBuiltins": "off", // TODO: investigate adding `"lib": ["ES2022"]` to `tsconfig.json`
+        "noFallthroughSwitchClause": "error", // Prevents accidental automatic fallthroughs in switch cases (use disable comment if needed)
+        "noImplicitAnyLet": "warn", // TODO: Refactor and make this an error
+        "noRedeclare": "error",
+        "noGlobalIsNan": "error",
+        "noAsyncPromiseExecutor": "warn", // TODO: Refactor and make this an error
+        "useDefaultSwitchClauseLast": "info" // TODO: make error?
```
MokaStitcher: I wouldn't mind making this an error. Do we have any part of the codebase that doesn't do have `default` as last statement? oO

index.css
```diff
@@ -162,6 +162,7 @@ input:-internal-autofill-selected {
   --small-control-size: calc(var(--controls-size) / 3);
   --rect-control-size: calc(var(--controls-size) * 0.74);
 
+  /* biome-ignore lint/a11y/useGenericFontNames: does this matter? */
```
MokaStitcher: guess it wouldn't hurt to add a fallback font family. may do that in the font pr if i don't forget

src/data/abilities/ab-attrs/non-super-effective-immunity-ab-attr.ts
```diff
@@ -12,7 +12,7 @@ import { TypeImmunityAbAttr } from "./type-immunity-ab-attr";
  */
 export class NonSuperEffectiveImmunityAbAttr extends TypeImmunityAbAttr {
   constructor(condition?: AbAttrCondition) {
-    super(null, condition);
+    super(null!, condition);
```
MokaStitcher, line 17: what?

src/data/abilities/ab-attrs/move-type-change-ab-attr.ts
```diff
@@ -25,7 +25,7 @@ export class MoveTypeChangeAbAttr extends PreAttackAbAttr {
     moveType?: NumberHolder,
     power?: NumberHolder,
   ): boolean {
-    if (this.condition && this.condition(pokemon, defender, move)) {
+    if (this.condition?.(pokemon, defender, move)) {
```
MokaStitcher, line 28: i don't like that notation for function calls, I find it confusing, but will follow the opinion of others

src/data/animations/anim-config-schema.ts
```diff
@@ -342,6 +342,7 @@ const animTimedEventSchema: JSONSchemaType<AnimTimedEvent> = {
   additionalProperties: false,
   required: ["eventType", "time", "resourceName"],
   if: { properties: { eventType: { const: "AnimTimedSoundEvent" } } },
+  // biome-ignore lint/suspicious/noThenProperty: should be safe?
```
MokaStitcher: https://biomejs.dev/linter/rules/no-then-property/

what do you think @innerthunder ?

src/data/mystery-encounters/mystery-encounter-option.ts
```diff
@@ -16,6 +16,7 @@ import { MysteryEncounterOptionMode } from "#enums/mystery-encounter-option-mode
 import type { CanLearnMoveRequirementOptions } from "./requirements/can-learn-move-requirement";
 import { CanLearnMoveRequirement } from "./requirements/can-learn-move-requirement";
 
+// biome-ignore lint/suspicious/noConfusingVoidType: TODO: change this?
 export type OptionPhaseCallback = () => Promise<void | boolean>;
```
MokaStitcher, line 20: should be as easy to fix as changing `void` to `undefined` right? Or is there more to it? ([doc for reference](https://biomejs.dev/linter/rules/no-confusing-void-type/))

src/data/mystery-encounters/mystery-encounters.ts
```diff
@@ -202,15 +202,15 @@ export function initMysteryEncounters() {
 
   // Add ANY biome encounters to biome map
   // eslint-disable-next-line
-  let encounterBiomeTableLog = "";
+  let _encounterBiomeTableLog = "";
```
MokaStitcher: what's the issue with this? And do we still need to `eslint-disable` comment?

src/data/mystery-encounters/utils/encounter-phase-utils.ts
```diff
@@ -554,6 +554,7 @@ export function generateModifierTypeOption(
  * @param selectablePokemonFilter - A filter for which Pokemon are allowed
  */
 export function selectPokemonForOption(
+  // biome-ignore lint/suspicious/noConfusingVoidType: TODO: refactor this?
   onPokemonSelected: (pokemon: PlayerPokemon) => void | OptionSelectItem[],
```
MokaStitcher, line 558: same question
```suggestion
  onPokemonSelected: (pokemon: PlayerPokemon) => undefined | OptionSelectItem[],
```

src/data/pokemon-species-form.ts
```diff
@@ -312,6 +312,7 @@ export abstract class PokemonSpeciesForm {
         case SpeciesId.DUDUNSPARCE:
           break;
         case SpeciesId.ZACIAN:
+        // biome-ignore lint/suspicious/noFallthroughSwitchClause: intentional (?)
```
MokaStitcher: wow that is confusing, but I believe intentional, yes.
Best way to fix it would probably be renaming the sprite files tbh, don't understand why they are named behemoth in the first place. Probably not for this PR though, but a todo or issue might be in order

src/event-bus.ts
```diff
@@ -35,6 +35,7 @@ class EventBus extends Phaser.Events.EventEmitter {
     return super.addListener(event, fn, context);
   }
 
+  // biome-ignore lint/complexity/noBannedTypes: what should this even be replaced with?
   override removeListener(event: string | symbol, fn?: Function, context?: any, once?: boolean): this {
```
MokaStitcher: Do the same as for `off`? From my understanding both are interchangeable and `super.off` has the same signature as `super.removeListener`, so it shouldn't be an issue to do the same with our `off` and `removeListener`

src/modifier/modifier-type.ts
```diff
@@ -1609,13 +1610,13 @@ let modifierPoolThresholds = {};
 let ignoredPoolIndexes = {};
 
 let dailyStarterModifierPoolThresholds = {};
-let ignoredDailyStarterPoolIndexes = {}; // eslint-disable-line @typescript-eslint/no-unused-vars
+let _ignoredDailyStarterPoolIndexes = {}; // eslint-disable-line @typescript-eslint/no-unused-vars
```
MokaStitcher, line 1614: is eslint disable still needed? (times 3)

src/phases/abstract-pokemon-phase.ts
```diff
@@ -13,7 +13,7 @@ export abstract class PokemonPhase extends FieldPhase {
   public isPlayer: boolean;
   public fieldIndex: number;
 
-  constructor(battlerIndex?: BattlerIndex | number) {
+  constructor(battlerIndex: BattlerIndex | number) {
```
MokaStitcher, line 16: since it's no longer optional (is that correct?), is the `??` still needed below, as well as the the `battlerIndex === undefined`?

src/phases/command-phase.ts
```diff
@@ -307,18 +309,21 @@ export class CommandPhase extends FieldPhase {
           }
         }
         break;
+      }
+      // biome-ignore lint/suspicious/noFallthroughSwitchClause: intentional
```
MokaStitcher:
```suggestion
      // biome-ignore lint/suspicious/noFallthroughSwitchClause: intentional overlap between
      // `Pokemon` and `Run` commands, because they need the same the checks for being trapped
```
maybe?

src/phases/common-anim-phase.ts
```diff
@@ -17,6 +17,7 @@ export class CommonAnimPhase extends PokemonPhase {
   private readonly targetIndex?: BattlerIndex;
 
   constructor(anim: CommonAnim, battlerIndex?: BattlerIndex, targetIndex?: BattlerIndex) {
+    // @ts-expect-error - TODO: refactor `PokemonPhase`
```
MokaStitcher: humm.. So this would answer my question about the battlerIndex no longer being option in PokemonPhase.
Why was PokemonPhase being changed then?

src/phases/switch-summon-phase.ts
```diff
@@ -65,11 +65,11 @@ export class SwitchSummonPhase extends SummonPhase {
...
     if (!this.doReturn || (this.slotIndex !== -1 && !this.getAlliedParty()[this.slotIndex])) {
       if (this.isPlayer) {
-        return this.switchAndSummon();
-      } else {
-        time.delayedCall(750, () => this.switchAndSummon());
+        this.switchAndSummon();
         return;
       }
+      time.delayedCall(750, () => this.switchAndSummon());
+      return;
```
MokaStitcher:
```suggestion
      } else {
        time.delayedCall(750, () => this.switchAndSummon());
      }
      return;
```

now that is confusing. I think this might be clearer?

src/system/game-data.ts
```diff
@@ -36,6 +36,7 @@ import type { PlayerPokemon } from "#app/field/player-pokemon";
 import type { Pokemon } from "#app/field/pokemon";
 import { getGameMode } from "#app/game-mode";
 import { globalScene } from "#app/global-scene";
+// biome-ignore lint/style/noNamespaceImport: Something weird is going on here and I don't want to touch it
```
MokaStitcher: 😭

src/ui/handlers/starter-select-ui-handler.ts
```diff
@@ -1119,10 +1149,13 @@ export class StarterSelectUiHandler extends MessageUiHandler {
     const hasNonShiny = caughtAttr & DexAttr.NON_SHINY;
     if (starterAttributes.shiny && !hasShiny) {
       // shiny form wasn't unlocked, purging shiny and variant setting
+      // biome-ignore lint/performance/noDelete: TODO: is `delete` required?
```
MokaStitcher: hmm, these attributes are already optional apparently 
so I guess instead we could set them to `undefined`. Not 100% sure the code handles it properly everywhere though
<sub>although deleting also means less storage used (for now only local storage, but ideally in the future this would be synced with the server)

test/test-utils/gameWrapper.ts
```diff
@@ -43,14 +43,15 @@ export class GameWrapper {
     Pokemon.prototype.enableMask = () => null;
     Pokemon.prototype.cry = () => null as any;
     Pokemon.prototype.faintCry = (cb) => {
-      if (cb) cb();
+      if (cb) {
+        cb();
+      }
```
MokaStitcher: if we go with this notation, might as well
```suggestion
      cb?.();
```

test/ui/battle-info.test.ts
```diff
@@ -6,6 +6,7 @@ import { GameManager } from "#test/test-utils/gameManager";
 import Phaser from "phaser";
 import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
 
+// biome-ignore lint/correctness/noEmptyPattern: TODO: change this?
 vi.mock("../data/exp", ({}) => {
```
MokaStitcher, line 10: what even is this lol

MokaStitcher (commented):
Changes seem fine for the most part, unfortunately wasn't able to get biome to run on my machine so I couldn't test my questions/suggestions 💀

.github/workflows/biome.yml
DayKev: Maybe "Lint changed files with [X]" is better?

MokaStitcher: Oh so the GH workflow doesn't check all files?

biome.jsonc
DayKev: There's a few (I think mostly in MEs lol...). The reason I didn't enable this right away was because of uses where `default:` fell through intentionally into the first case, and I wasn't sure what to do about those. I didn't want to move them and copy the first case's code into them since that would be adding redundant code. I could just use a lint disable directive for those instances maybe?
<details><summary>Examples</summary>
<p>

```ts
/// encounter-test-utils.ts
  switch (optionNumber) {
    default:
    case 1:
      // no movement needed. Default cursor position
      break;
    case 2:
      uiHandler.processInput(Button.RIGHT);
      break;
    case 3:
      // ...

/// mystery-encounter-ui-handler.ts
      switch (this.optionsContainer.getAll()?.length) {
        default:
        case 3:
          success = this.handleTwoOptionMoveInput(button);
          break;
        case 4:
          // ...

      switch (this.encounterOptions.length) {
        default:
        case 2:
          optionText = addBBCodeTextObject(i % 2 === 0 ? 0 : 100, 8, "-", TextStyle.ME_OPTION_DEFAULT);
          break;
        case 3:
          // ...

/// encounter-phase-utils.ts
      switch (tier) {
        default:
        case BiomePoolTier.BOSS:
          ++bossEncountersByRarity[0];
          break;
        case BiomePoolTier.BOSS_RARE:
          // ...
```

</p>
</details>

src/data/abilities/ab-attrs/non-super-effective-immunity-ab-attr.ts
DayKev: Oh I forgot to add a comment. `TypeImmunityAbAttr`'s constructor had `immuneType: ElementalType | null` and then all the `AbAttr`s that extended it had `immuneType: ElementalType` in their constructors just to constrain the type to not allow `null`, but that's silly. Since this `AbAttr` is the unusual case I changed `TypeImmunityAbAttr` to not allow `null` and removed the now-unnecessary constructors from the other subclasses of `TypeImmunityAbAttr`. Will add the following comment:
```ts
    // This `AbAttr` ignores the superclass's `immuneType` field and overrides the methods that make use of it
    // TODO: refactor this?
    super(null!, condition);
```

src/data/mystery-encounters/mystery-encounter-option.ts
DayKev: Not sure, a quick initial look indicated it might require changing a decent bit of the code but I didn't look too closely into it (partially because MEs are in the "future" pile of "code to be thoroughly refactored").

src/data/mystery-encounters/mystery-encounters.ts
DayKev: Looks like it was used in some commented-out logging. I'll just comment out this too for now since we'll probably want to re-use it later when refactoring MEs.

src/modifier/modifier-type.ts
DayKev: Yeah, ESLint's "no-unused-vars" rule is still enabled because it's needed for imports (Biome doesn't have multi-line disable directives added until 2.0 which isn't out yet, and we use them for the TSDoc imports).
```jsonc
// biome.jsonc
"noUnusedImports": "off" // Handled by ESLint until multi-line disable is added (in Biome 2.0)
```

src/phases/switch-summon-phase.ts
DayKev: Yeah if I pull the `return` out of the inner `if` block entirely the functionality stays the same, will do.
```ts
      if (this.isPlayer) {
        this.switchAndSummon();
      } else {
        time.delayedCall(750, () => this.switchAndSummon());
      }
      return;
```

src/phases/common-anim-phase.ts
DayKev: Looking closer, this phase should probably just not extend `PokemonPhase`.

DayKev: Then again, `PokemonHealPhase` actually makes use of a couple of fields/methods from `PokemonPhase`...

DayKev: Really this is just an absolute mess of code being used improperly...

DayKev: Okay I found the least disruptive change: move the `battlerIndex = battlerIndex ?? ...` from `PokemonPhase` to this phase, since this is the only one that actually needed it. Then it can be removed from `PokemonPhase`.

src/phases/command-phase.ts
DayKev: Can't separate it into multiple lines since the `biome-ignore` line needs to be right above it. Maybe: `"Run" and "Pokemon" cases share checks for trapping`?

src/event-bus.ts
innerthunder: According to https://biomejs.dev/linter/rules/no-banned-types/, `Function` needs to be more strongly typed. This override seems useless in general, though.

test/test-utils/gameWrapper.ts
DayKev: Yeah that's better.

test/ui/battle-info.test.ts
DayKev: 🤷

src/ui/handlers/starter-select-ui-handler.ts
DayKev: If it reduces storage space then we should probably keep the `delete`s, though I assume that's because we currently just chuck raw `json` into the save data which is... not ideal. So maybe change the `TODO`s to "refactor save data to not be raw json objects"?

.github/workflows/biome.yml
DayKev: The Biome workflow only checks files changed by the PR.
Actually the ESLint workflow seems to check the whole repo, should I change the Biome workflow to match?

src/ui/handlers/starter-select-ui-handler.ts
MokaStitcher: It's not really save data it's the game memorizing the ability, nature, form, gender, etc of the various starters. It's completely separate of the actual save data, but yes in this case it's being stored as raw json (which is mostly fine since it's only local right now)

Definitely things will have to be changed if we want to store it on the server but that's a long term thing at the moment. Actual save data and server request optimization should come first, but that's a separate issue (and not as simple as raw json in this case). It will also heavily depends on what we do server side

All this to say, adding a comment is fine if you want to but maybe don't call it save data because starter preferences are their own independent thing right now and refactoring save data is a much wider and kind of different issue

.github/workflows/biome.yml
MokaStitcher: I think it would make sense yes.
When committing/running locally it's good to only check changes, but we might as well be thorough on GH

biome.jsonc
MokaStitcher: oh boy ok, let's leave it as that for now I guess :')

src/phases/common-anim-phase.ts
MokaStitcher: :') sounds good 👍

.github/workflows/biome.yml
MokaStitcher: image
would fixing that be possible too? not sure what's up with it

src/data/animations/anim-config-schema.ts
innerthunder: It should be ignored here. `then` is a JSON Schema keyword https://ajv.js.org/json-schema.html#if-then-else

.github/workflows/biome.yml
DayKev: It because I enabled the ruleset that requires GitHub workflows to succeed before a PR can be merged (we're still able to bypass ofc). I created an updated ruleset that I'll swap over to once this PR is merged.

src/data/abilities/ab-attrs/move-type-change-ab-attr.ts
innerthunder: I'm fine with the notation personally, though I could see it being confusing for future contributors.

src/event-bus.ts
DayKev: The overrides seem to be here to add additional type safety.

src/data/pokemon-species-form.ts
DayKev: Updated the comment to `intentional (TODO: rename sprite files?)`. Is that good?

index.css
DayKev: I'll update the comment to `TODO: add fallback font family` then?

src/data/mystery-encounters/mystery-encounters.ts
```diff
@@ -201,19 +200,18 @@ export function initMysteryEncounters() {
...
     });
 
-    encounterBiomeTableLog += `${getBiomeName(biome).toUpperCase()}: [${biomeEncounters
-      .map((type) => MysteryEncounterType[type].toString().toLowerCase())
-      .sort()
-      .join(", ")}]\n`;
+    // encounterBiomeTableLog += `${getBiomeName(biome).toUpperCase()}: [${biomeEncounters
+    //   .map((type) => MysteryEncounterType[type].toString().toLowerCase())
+    //   .sort()
+    //   .join(", ")}]\n`;
```
innerthunder, line 214: This debug code should probably be moved to another function (test util?)

src/data/trainer-config.ts
```diff
@@ -1224,7 +1225,7 @@ export class TrainerConfig {
    * @param battleBgm the string representation of the battle bgm
    * @returns The updated TrainerConfig instance.
    * **/
-  initForEvilTeamLeader(title: string, name: string, rematch: boolean = false, battleBgm: string): TrainerConfig {
+  initForEvilTeamLeader(title: string, name: string, rematch: boolean, battleBgm: string): TrainerConfig {
```
innerthunder, line 1223: It'd be nice if these args were rearranged to restore the defaults instead

src/data/trainer-config.ts
```diff
@@ -1463,7 +1464,7 @@ export class TrainerConfig {
    * @param variant - The variant of the trainer to determine the specific title.
    * @returns - The title of the trainer.
    **/
-  getTitle(trainerSlot: TrainerSlot = TrainerSlot.NONE, variant: TrainerVariant): string {
+  getTitle(trainerSlot: TrainerSlot, variant: TrainerVariant): string {
```
innerthunder, line 1462: Same here

src/modifier/modifier-type.ts
```diff
@@ -740,7 +741,8 @@ enum AttackTypeBoosterItem {
 
 export class AttackTypeBoosterModifierType
   extends PokemonHeldItemModifierType
-  implements GeneratedPersistentModifierType {
+  implements GeneratedPersistentModifierType
+{
```
innerthunder, line 746: Love that this doesn't fight with Prettier on the formatting anymore :joy:

innerthunder (dismissed):
Changes seem fine, at least in the scope of this PR 👍

src/data/trainer-config.ts
DayKev: Tbh this param shouldn't even have a default probably.

src/data/trainer-config.ts
DayKev: Same, shouldn't have had a default in the first place.

src/data/mystery-encounters/mystery-encounters.ts
DayKev: I think it was used when creating the ME framework / the MEs themselves, it's not really a test util.

src/ui/handlers/starter-select-ui-handler.ts
DayKev: Ah true starter prefs are local-only save data.

I could change the comment to "Optimizes local storage size".

MokaStitcher: yeah it's just to clarify things because "save data" can mean a lot of things :p

src/data/mystery-encounters/mystery-encounters.ts
MokaStitcher: yes, it was a debug thing that balance team wanted to get an overview of the biome -> ME table since the codebase only uses ME -> biome links (or the other way around, but same idea). It can still be useful to have, a good use case for another debug env variable maybe

innerthunder (dismissed):

lefthook.yml
```diff
@@ -4,23 +4,23 @@ pre-commit:
...
         - merge
         - rebase
-      priority: 2
 
 pre-push:
   commands:
     eslint:
       glob: "*.{js,ts,jsx,tsx}"
-      run: npx eslint {push_files}
+      run: npx biome ci --changed --no-errors-on-unmatched --diagnostic-level=error {push_files}
```
MokaStitcher:
```suggestion
      run: npx biome ci --changed --no-errors-on-unmatched --diagnostic-level=error {push_files}
```

MokaStitcher (dismissed):

src/data/abilities/ab-attrs/move-type-change-ab-attr.ts
flx-sta: I find this notation clearer than the `&&` one

flx-sta (dismissed):
The speed is crazy!

Is it expected that there is so many warnings ?

DayKev:
> The speed is crazy!
> 
> Is it expected that there is so many warnings ?

Yeah, I set a few rules to "warn" rather than "error" because it would require more extensive refactoring to fix them (and/or they're areas of the code I'm not confident in fixing for this PR, which I'm trying to keep to mostly the simpler fixes since it's already a large PR).

PR #1157, opened 2025-05-09
[Dev] Add and apply import aliases

DayKev:
N/A

No more relative imports.

- All imports converted to absolute imports.
- Import aliases created for (nearly) every directory.
- Organize imports applied to most files.
- ESLint plugin + rule added to enforce absolute imports (won't automatically apply import aliases unfortunately).

Relevant files: [eslint.config.js](https://github.com/Despair-Games/poketernity/blob/af2ba1072d14912ba0ad2e811bbde93abbfd5bee/eslint.config.js), [package.json](https://github.com/Despair-Games/poketernity/blob/0b9ff45bdf941efdd28d976c479c2d148dd2f88a/package.json), [tsconfig.json](https://github.com/Despair-Games/poketernity/blob/0b9ff45bdf941efdd28d976c479c2d148dd2f88a/tsconfig.json)
You might be able to browse the diffs on GitHub with [this filter](https://github.com/Despair-Games/poketernity/pull/1157/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&show-viewed-files=true#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0), otherwise you'll have to use VSCode's GitHub plugin (or the images provided below).
<details><summary>Diff images</summary>
<p>

image
image
image

</p>
</details> 

## Checklist


eslint.config.js
```diff
@@ -45,6 +46,11 @@ const eslintRules = {
   "@typescript-eslint/consistent-type-imports": "error", // Enforces type-only imports wherever possible
   "@typescript-eslint/no-import-type-side-effects": "error", // Typescript turns `import { type X } from Y` into `import {} from Y` but completely removes `import type { X } from Y`
   "@typescript-eslint/ban-ts-comment": ["error", { "ts-check": false, "ts-expect-error": "allow-with-description" }], // Disallow usage of `@ts-ignore`
+  "no-relative-import-paths/no-relative-import-paths": [
+    // Enforces absolute paths only (for example, converts "./data/moves/move-attrs/call-move-attr" to "#moves/call-move-attr")
```
DayKev:
```suggestion
    // Enforces absolute paths only (for example, converts "./data/moves/move-attrs/call-move-attr" to "#app/data/moves/move-attrs/call-move-attr")
```

Tempo-anon (dismissed):
1286 files changed lgtm trust

innerthunder (dismissed):

tsconfig.json
```diff
@@ -22,9 +22,50 @@
...
+      "#moves/*": ["./data/moves/move-attrs/*.ts", "./data/moves/move-conditions/*.ts", "./data/moves/*.ts"],
+      "#mystery-encounters/*": [
+        "./data/mystery-encounters/utils/*.ts",
+        "./data/mystery-encounters/requirements/*.ts",
+        "./data/mystery-encounters/encounters/*.ts",
+        "./data/mystery-encounters/*.ts"
+      ],
+      "#phases/*": ["./phases/*.ts"],
+      "#pipelines/*": ["./pipelines/*.ts"],
+      "#system/*": [
```
MokaStitcher, line 50: Is adding all the subfolders needed here or would something like `./system/**/*.ts` do the trick? (Same for UI and ME folders if yes, so that we don't need to edit this if we add any new subfolder)

DayKev: Unfortunately not:
image

PR #1160, opened 2025-05-10
[UI][P3] Fix option select cursor misalignement when scrolling down

MokaStitcher:

The cursor no longer overlaps with the text when a menu is resized dynamically upon scrolling

Currently the cursor placement in option select is updated in `setCursor` when the cursor index is changed.
When scrolling down/up only the scrolling cursor is updated, not the base cursor, so `setCursor` is not called, and thus the cursor is not moved. This is an issue when scrolling down results in the window being resized to fit wider options, resulting in the cursor overlapping the text.

- Move the code responsible to place the cursor in a dedicated function
- Call this function in `setCursor` like before, as well as in `updateSizeForOptions` so that the cursor is moved when the window is resized

## Screenshots/Videos

Before: scrolling down the nature list of a Pokemon (need many natures unlocked for this behavior to kick in, also depends on language probably)
image

After:
image

## How to test the changes?

- New game
- Find a Pokemon with a lot of natures unlocked
- Open its menu > Manage Natures
- Scroll down the list and make sure the cursor gets moved if the window becomes larger

## Checklist

### DIFF
--- a/src/ui/handlers/base-option-select-ui-handler.ts
+++ b/src/ui/handlers/base-option-select-ui-handler.ts
@@ -191,6 +191,10 @@ export abstract class BaseOptionSelectUiHandler<T extends OptionSelectItem> exte
     }
   }
 
+  /**
+   * Place the cursor in front of the currently selected option.
+   * Initializes the cursor sprite if it doesn't exist.
+   */
   private updateCursorPlacement() {
     if (!this.cursorObj) {
       this.cursorObj = globalScene.add.image(0, 0, "cursor");
--- a/src/ui/handlers/base-option-select-ui-handler.ts
+++ b/src/ui/handlers/base-option-select-ui-handler.ts
@@ -181,9 +181,28 @@ export abstract class BaseOptionSelectUiHandler<T extends OptionSelectItem> exte
     );
 
     this.optionSelectBg.setSize(bgWidth, bgHeight);
+
+    if (this.cursorObj) {
+      this.updateCursorPlacement();
+    }
+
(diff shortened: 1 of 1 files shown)


DayKev (dismissed):
Interestingly this didn't happen when wrapping around by scrolling up (at least on the Nature selection menu).

src/ui/handlers/base-option-select-ui-handler.ts
```diff
@@ -181,11 +181,30 @@ export abstract class BaseOptionSelectUiHandler<T extends OptionSelectItem> exte
...
+    if (this.cursorObj) {
+      this.updateCursorPlacement();
+    }
+
     if (this.config?.onResize) {
       this.config.onResize(bgWidth, bgHeight);
     }
   }
 
+  private updateCursorPlacement() {
```
Tempo-anon, line 198: nit: should have a comment

Tempo-anon (dismissed):

src/ui/handlers/base-option-select-ui-handler.ts
MokaStitcher: Thought it was fine for private elements not to have doc, but will do

MokaStitcher:
> Interestingly this didn't happen when wrapping around by scrolling up (at least on the Nature selection menu).

This is because when wrapping around the cursor is changed from 0 to the bottom of the list, so its position gets updated.
But from this point the bug would happen when scrolling up if there was a wider element somewhere in the middle of the list lol

PR #1161, opened 2025-05-10
[Save Data] Move unlocked natures and IVs from DexData to StarterData

MokaStitcher:
> [!WARNING]
> This is the first of 2 PRs making changes to the Dex / System Data, meaning partial loss of compatibility with existing saves.
Since the impacted elements (natures and ivs) are of little importance to current development this PR introduces a temporary "migration" of existing data which consists of unlocking all the elements instead of transferring the current unlocks. This means current saves work fine, and the current `everything` save file can still be used to unlock everything
The next PR will remove that migrator and fully force a reset of current system data, as well as provide a new, clean, `everything` save file.

- existing save files: already unlocked starters will have all their natures unlocked and IVs maxed
- fresh files: starters now get all neutral natures unlocked by default, instead of a single random one
- everything else should be the same except less space used for save file storage

Currently the nature and IVs data is stored in `systemData.dexData`, meaning there is a nature and IV entry for each of the 1100+ Pokémon species. However, unlocked natures and IVs are only relevant when it comes to starters
So, by moving this information from `systemData.dexData` to `systemData.starterData` we change those 1100+ entries to 587, one for each of the starters, resulting in roughly half the storage space used for these attributes

- Move nature and ivs from DexData to StarterData
- Move nature and ivs initialization in `game-data` accordingly
- Update IV scanner and container showing information of newly caught or hatched Pokémon accordingly
- Add temporary data migration unlocking all natures and maxxing out all starter IVs on existing save files only 
> [!NOTE]
> This "migration" does **not** remove the old data from `dexData`, it only populates `starterData`, meaning if you go from this branch back to another one, your old data is preserved for now (and also the save data is larger).
- **UPDATE** Removed the 1 offset in the nature attribute. It is not needed and meant that the attribute for all natures being unlocked was 26 bits long rather than 25

## Screenshots/Videos

<details><summary>Existing save file</summary>

Other branches:
image
This branch:
image

</details>

<details><summary>New save file</summary>

After: 15 IV, single neutral nature unlocked:
image
After: 15 IV, all neutral natures unlocked:
image

</details>

<details><summary>StarterData update when catching/hatching mons</summary>

Natures still get shown as new or not when hatching/catching. IVs get compared to existing IVs when using scanner:
image
image

Catching an Elekid and an Electabuzz:
image
image
Elekid's IVs are the max of IVs across both catches. Elekid can choose from 2 natures.
Electabuzz only has IVs from the Electabuzz catch, and a single nature
image
image
</details>

<details><summary>Storage</summary>

**Size of exported fresh file**
Before: 159KB - After: 142KB
**Size of exported advanced file**
Before: 187KB - After: 217KB (because of the temporary migration which doesn't delete the elements from dexData)

</details> 

## How to test the changes?
- Load the game with an existing save file, and go to starter select
  - All starters should have maxxed IVs and all natures unlocked

- Load the game in a private window to start a fresh file
  - All starters should have a single neutral nature unlocked, and IVs of 15
  - Start a run, catch Pokemon and check for IV and natures:
     - that they are shown properly as new/improved or not, depending
     - that they are unlocked properly back in starter selection
  - Refresh the page to force loading the data. The unlocks should be preserved

You might want to do extra testing around Pokémon with baby forms. Catching a Pikachu should unlock nature and update IVs for both Pikachu and Pichu, while catching a Pichu should only unlocks those things for Pichu itself. Another thing to potentially test for is eggs.

Here are some overrides that can be useful:
```
  STARTING_HELD_ITEMS_OVERRIDE: [{name: "IV_SCANNER", count: 2}],
  //ENEMY_SPECIES_OVERRIDE: SpeciesId.ELECTABUZZ,
  EGG_FREE_GACHA_PULLS_OVERRIDE: true,
  EGG_IMMEDIATE_HATCH_OVERRIDE: true,
  FREE_CANDY_UPGRADE_OVERRIDE: true,
  POKEBALL_OVERRIDE: {
    active: true,
    pokeballs: {
      [PokeballType.POKEBALL]: 5,
      [PokeballType.GREAT_BALL]: 0,
      [PokeballType.ULTRA_BALL]: 0,
      [PokeballType.MASTER_BALL]: 50,
    },
  }
```

## Checklist


src/phases/select-starter-phase.ts
```diff
@@ -74,7 +74,10 @@ export class SelectStarterPhase extends Phase {
         starterGender = Overrides.GENDER_OVERRIDE;
       }
 
-      const starterIvs = dexData[speciesId].ivs.slice(0);
+      // Get ivs from the root species in case of an override to a non starter species
```
Tempo-anon, line 77: Technically this is new functionality when overriding to a different species right?

src/system/game-data.ts
```diff
@@ -374,6 +375,17 @@ export class GameData {
           }
         }
 
+        // TODO: Temporary starter data migration, to be removed later
+        for (const starterData of Object.values(systemData.starterData)) {
+          if (!starterData.natureAttr || !starterData.ivs) {
+            const unlocked = starterData.abilityAttr !== 0;
+            // As a placeholder migration, unlock all natures
+            starterData.natureAttr = unlocked ? 67108862 : 0;
```
Tempo-anon: nit: I know this is temporary code but is that 67108862 value worth keeping around as a constant to represent the complete natureAttr? Also since there are 25 natures in game, shouldn't this be 2^25 (33554432)? 2^26 is 67108864

src/system/game-data.ts
```diff
@@ -1459,16 +1452,33 @@ export class GameData {
   private initStarterData(): void {
     const starterData: StarterData = {};
 
+    // Pick a random neutral nature for each of the default starters
```
Tempo-anon: While I like this in terms of flavor and uniqueness, I wonder if the extra complexity is needed rather than setting all the starters to a single nature. Or granting all the initial starter all neutral natures

src/system/game-data.ts
```diff
@@ -1459,16 +1452,33 @@ export class GameData {
...
+
       starterData[speciesId] = {
         moveset: null,
         eggMoves: 0,
         candyCount: 0,
         candyProgress: 0,
-        abilityAttr: defaultStarterSpecies.includes(speciesId) ? AbilityAttr.ABILITY_1 : 0,
+        abilityAttr: isDefaultStarter ? AbilityAttr.ABILITY_1 : 0,
         passiveAttr: 0,
+        natureAttr: isDefaultStarter ? 1 << (defaultStarterNatures.shift()! + 1) : 0,
```
Tempo-anon: nit: maybe a comment would be good because `1 << (defaultStarterNatures.shift()! + 1)` is not clear to me at a glance

Tempo-anon (dismissed):
Looks fine to me for the most part. Just some nits

src/system/game-data.ts
```diff
@@ -1459,16 +1452,33 @@ export class GameData {
...
+    const defaultStarterNatures: Nature[] = [];
+    globalScene.executeWithSeedOffset(
+      () => {
+        const neutralNatures = [Nature.HARDY, Nature.DOCILE, Nature.SERIOUS, Nature.BASHFUL, Nature.QUIRKY];
+        for (let s = 0; s < defaultStarterSpecies.length; s++) {
+          defaultStarterNatures.push(randSeedItem(neutralNatures));
+        }
+      },
+      0,
+      "default",
```
DayKev: Doesn't using `"default"` as a seed override mean every player will have the same "randomized" set of natures for their initial starters? If we're going to keep the "get a random neutral nature" for the starters I think it shouldn't use `executeWithSeedOffset`, or we should just unlock all neutral natures for the initial starters.

DayKev:
Updated the test (see commit message).

src/system/game-data.ts
MokaStitcher: For some reason the flags are offset by 1, so 3554432 doesn't work. I don't know if updating that should be part of this PR though, for now I left it as is because I'm scared to mess things up lol
But I don't think we want to keep that as a constant, it was just a very lazy way to do it in this case. I'll see for a cleaner way when creating the proper everything file

edit: nevermind, I removed the +1 everywhere when manipulating the natureattr, I don't see why it would be needed

DayKev:
Tested catching a Raichu with IVs set to `0, 5, 10, 15, 20, 25` and both Pikachu and Pichu had the correct IVs (and the same nature) in starter select. Reversing the IVs and catching another one also updated the IVs for both properly. 👍

DayKev (dismissed):

src/phases/select-starter-phase.ts
MokaStitcher: Not really? Not that the previous behavior was documented or predictable in any way. And this one is still not predictable, to me it's kind of the same except on average it will result in higher IVs
Before if you did an override to Charizard it would look for Charizard's ivs, and now Charizard doesn't have IVs so it uses Charmander's, which are always the same, or better than what Charizard's would have been

MokaStitcher: Feel free to elaborate on what you'd want done in this regard, update the override to make this behavior clear?

src/system/game-data.ts
MokaStitcher: No clue, that code wasn't changed, just moved

MokaStitcher: got removed and replaced with unlocking all neutral natures

src/phases/select-starter-phase.ts
Tempo-anon: Ah okay. What happens if I choose Charmander as a starter but I use a species override as Lapras?

MokaStitcher: It will use lapras' IVs since the species override is applied earlier in that function (the IVs will be 0 if Lapras hasn't been unlocked yet)

src/constants/game-constants.ts
```diff
@@ -24,6 +24,11 @@ export const IV_MIN = 0;
  */
 export const IV_MAX = 31;
 
+/**
+ * Value used for the IVs of starters unlocked by default on a fresh file.
+ */
+export const IV_DEFAULT = 15;
```
flx-sta: I think this constants name should be more specialized that it's only for `starters`.
Should also be the actual array I'd say. Maybe we want to alter specific IV aspects in the future and not the whole array.

```suggestion
export const IV_DEFAULT_UNLOCKED_STARTERS = [15, 15, 15, 15, 15, 15];
```

DayKev: I think I'd go with `DEFAULT_STARTER_IVS`.

MokaStitcher: Agree about the name, I was just trying to match the existing constants. Will update tomorrow
Unsure about storing the whole array though.
After having looked at the heap so much I think we need to be a bit more mindful about the amount of data we keep in memory at all times. Sure that single array won't make much of a difference but I don't know about keeping an array that is only ever relevant the very first time the game is ran for a potential future use that is very unlikely to happen

src/system/game-data.ts
```diff
@@ -371,6 +372,18 @@ export class GameData {
...
+        const allNaturesAttr = Math.pow(2, getEnumLength(Nature)) - 1;
+        for (const starterData of Object.values(systemData.starterData)) {
+          if (!starterData.natureAttr || !starterData.ivs) {
+            const unlocked = starterData.abilityAttr !== 0;
+            // As a placeholder migration, unlock all natures
+            starterData.natureAttr = unlocked ? allNaturesAttr : 0;
+            // As a placeholder migration, max out all ivs
+            starterData.ivs = Array(6).fill(unlocked ? IV_MAX : IV_MIN);
+          }
+        }
```
flx-sta, line 357: Shouldn't this be part of the migration tooling?

src/ui/handlers/starter-select-ui-handler.ts
```diff
@@ -3333,9 +3335,10 @@ export class StarterSelectUiHandler extends MessageUiHandler {
 
     if (species) {
       const dexEntry = globalScene.gameData.dexData[species.speciesId];
-      const abilityAttr = globalScene.gameData.starterData[species.speciesId].abilityAttr;
+      const starterEntry = globalScene.gameData.starterData[species.speciesId];
 
-      const caughtAttr = globalScene.gameData.dexData[species.speciesId]?.caughtAttr || BigInt(0);
+      const abilityAttr = starterEntry.abilityAttr;
+      const caughtAttr = dexEntry?.caughtAttr || BigInt(0);
```
flx-sta: Always prefer `??` over `||` unless there is a specific reason to use `||`

```suggestion
      const caughtAttr = dexEntry?.caughtAttr ?? BigInt(0);
```

test/system/dex-data.test.ts
```diff
@@ -44,13 +45,17 @@ describe("Dex Data", () => {
...
 
-      const unlockedNatures = gameData.getNaturesForAttr(dexData.natureAttr);
-      expect(unlockedNatures.length).toBe(1);
-      expect(neutralNatures.includes(unlockedNatures[0])).toBeTruthy();
+      // Starters get all neutral natures unlocked
+      const unlockedNatures = gameData.getNaturesForAttr(starterData.natureAttr);
+      expect(unlockedNatures.length).toBe(neutralNatures.length);
+      for (const starterNature of unlockedNatures) {
+        expect(neutralNatures).toContain(starterNature);
+      }
```
flx-sta: If you want to avoid the `for of` loop here use `.every()`

```suggestion
      expect(unlockedNatures.every(nature => neutralNatures.includes(nature))).toBe(true);
```
<sub> I personally find expects in loops odd </sub>

src/system/game-data.ts
MokaStitcher: Not worth the trouble, this is extremely temporary, will be removed in my next PR (more details in the PR description itself)

flx-sta (dismissed):

test/system/dex-data.test.ts
DayKev: We should be careful about how we format the `expect`s, they should be as informative as possible if they fail. If this fails it'll just show "expected `false` to be `true`".

DayKev: It's better to change this to
```ts
expect(unlockedNatures).toEqual(neutralNatures);
```

DayKev: image
vs
image
(Replaced "Quirky with Adamant" in the neutral natures array for demonstration purposes)

DayKev (dismissed):

test/system/dex-data.test.ts
flx-sta: Alternatively you can also define your own failure messages as the 2nd argument in `expect`

flx-sta (approved):

Tempo-anon (approved):

PR #1167, opened 2025-05-12
[P2] Protect's condition now considers full move history

### LINKED ISSUE #1165: [Bug] Protect and other moves using `ProtectAttr` don't calculate fail chance correctly
Protect should be able to examine the Pokémon's entire move history if needed, but the `getLastXMoves` call erroneously does not pass `-1` as the first param, meaning it only return the pokemon's last move. This means even if a Pokemon used Protect 3 times in a row and all of them succeeded, the success chance of using it again would be `1/3` instead of `1/27`.
```ts
export class ProtectAttr extends AddBattlerTagAttr {
  constructor(tagType: BattlerTagType = BattlerTagType.PROTECTED) {
    super(tagType, true);
  }

    return (user, _target, _move): boolean => {
      let timesUsed = 0;
      // this is incorrect, should be `user.getLastXMoves(-1)`
      const moveHistory = user.getLastXMoves();
      let turnMove: TurnMove | undefined;

      while (moveHistory.length) {
        turnMove = moveHistory.shift();
        if (!turnMove?.move?.hasAttr(ProtectAttr) || turnMove?.result !== MoveResult.SUCCESS) {
          break;
        }
        timesUsed++;
      }
      if (timesUsed) {
        return !user.randSeedInt(Math.pow(3, timesUsed));
      }
      return true;
    };
  }
}
```

innerthunder:

Protect et al. should now have the proper success rate after consecutive successful uses:

| Uses | Success Rate |
|--------|--------|
| 0 | 1 |
| 1 | 1/3 |
| 2 | 1/9 |
| 3 | 1/27 |
| 4 | ... |

Fixes #1165 

In `protect-attr.ts`:
- Fixed `ProtectAttr#getCondition` to use the user's full move history and filter out virtual moves
- General cleanup of condition function logic

## How to test the changes?

`npm run test protect` (3 new tests)

## Checklist

  - [?] Have I made sure that any UI change works for both UI themes (dark and light)?


DayKev (dismissed):

test/moves/protect.test.ts
```diff
@@ -130,4 +131,36 @@ describe("Moves - Protect", () => {
     expect(enemyPokemon).toHaveMoveResult(MoveResult.SUCCESS);
     expect(leadPokemon).toHaveMoveResult(MoveResult.FAIL);
   });
+
+  it.each([
+    [1 / 3, 1],
+    [1 / 9, 2],
+    [1 / 27, 3],
+  ])(
+    "should have a success rate of %d after being used %d consecutive time(s) successfully",
```
flx-sta, line 140: Here is a really useless suggestion.
If you wanna add the `s` for `times` conditionally :D

```suggestion
  it.each([
    [1 / 3, 1, ""],
    [1 / 9, 2, "s"],
    [1 / 27, 3, "s"],
  ])(
    "should have a success rate of %d after being used %d consecutive time%s successfully",
```

Feel free to do with that what you want

src/data/moves/move-attrs/protect-attr.ts
flx-sta, line 1: NIT: Can we add the example success-rate table to the class doc?

flx-sta (dismissed):

src/data/moves/move-attrs/protect-attr.ts
flx-sta:
```
| Uses | Success Rate |
|------|--------------|
|  0   |      1       |
|  1   |    1/3       |
|  2   |    1/9       |
|  3   |   1/27       |
|  4   |    ...       |
```

test/moves/protect.test.ts
innerthunder: I prefer the current version. Less clutter in the input arrays 😅

DayKev (approved):

PR #1174, opened 2025-05-15
[Bug] Prevent `min > max` in `clamp` function

DayKev:
Less bugs?

Phaser's implementation of `clamp` prevented the `max` value from being less than the `min` value but ours did not, causing a change in behavior.

- If `min > max`, set `max = min` in `clamp()`.
- Revert implementation of `clamp` to match Phaser's.
- Manually prevent the `max` value from being negative in `pokemon-data.ts`.

## Checklist


src/utils/common-utils.ts
```diff
@@ -181,6 +181,10 @@ export function coerceArray<T>(input: T | T[]): T[] {
  * @returns The clamped value, between `min` and `max`
  */
 export function clamp(value: number, min: number, max: number): number {
+  if (min > max) {
+    console.error(`Min (${min}) > max (${max}) in clamp function!`);
+    max = min;
+  }
```
Tempo-anon: Are there any cases where this ever happens in the codebase? I feel like this should never happen and a crash would at least reveal where it's happening while an error message in the console might be missed. That being said, `return min` as an early return is also better style than setting a parameter imo

flx-sta: I'm confused. 
Where does the phaser math clamp prevent that?
https://github.com/phaserjs/phaser/blob/c62baca677399e098ce21bb49757974c4742895b/src/math/Clamp.js#L19-L24
```ts
var Clamp = function (value, min, max)
{
    return Math.max(min, Math.min(max, value));
};
```

We can swap max/min to have it identical

----

If we go through with `min < max` I'd rather have it crash with an error, than just giving a warning and doing something I might not have intended it to do.

----

Edit: I now say the discussion on discord.
That's wild that the behavior is different...

src/utils/common-utils.ts
```diff
@@ -181,6 +181,10 @@ export function coerceArray<T>(input: T | T[]): T[] {
...
   return Math.min(Math.max(value, min), max);
```
flx-sta:
```suggestion
  return Math.max(min, Math.min(max, value));
```

test/utils/common-utils.test.ts
```diff
@@ -16,7 +18,7 @@ describe("Utils - Common Utils", () => {
       expect(actual).toBe(min);
     });
 
-    it.each([-3, 0, 3])("should not alter values between min and max (%i)", (value) => {
+    it.each(range)("should not alter values between min and max (%i)", (value) => {
```
flx-sta, line 21: I just chose 1 negative neutral, 0 and one positive neutral. But I guess this works too as these tests don't take much time and resources

test/utils/common-utils.test.ts
flx-sta, line 1: If we go through with the `min < max` approach, we should also have a test that covers that behavior

flx-sta (changes requested):

src/utils/common-utils.ts
DayKev: > Are there any cases where this ever happens in the codebase?

The code to set the `formIndex` in `PokemonData` was erroneously sending `-1` for the `max` value sometimes, which would be more than the `min` value of `0`. See [e39eba2](https://github.com/Despair-Games/poketernity/pull/1174/commits/e39eba285be608ee46dbbbad56578e7de1ef4804)

> I feel like this should never happen and a crash would at least reveal where it's happening while an error message in the console might be missed. That being said, `return min` as an early return is also better style than setting a parameter imo

Previous behavior was essentially returning `min` when `min > max`, just by order of the `Math.min/Math.max` calls rather than directly.

I changed it to not modify the param.

MokaStitcher: huh, which species don't have any form? That doesn't seem normal

flx-sta:
I’m still unsure about the whole min > max behavior. It seems like a user error and shouldn’t be handled by the method but rather throw an error as it’s simply an invalid use of the function.

src/utils/common-utils.ts
DayKev: It was any pokemon like Mewtwo or Ditto that didn't have any forms. Since it was doing `forms.length - 1` the max would be `-1` for these pokemon.
It's always been like this, but the issue was masked because until flx's change it was using `max(min(formIndex, forms.length - 1), 0)` before that.
The issue was in `pokemon-data.ts`.
```ts
// original
this.formIndex = Math.max(Math.min(source.formIndex, getPokemonSpecies(this.species).forms.length - 1), 0);

// change 1 (~3 weeks ago) (phaser's clamp being the same implementation of `max(min(...), ...)`)
this.formIndex = Phaser.Math.Clamp(source.formIndex, 0, getPokemonSpecies(this.speciesId).forms.length - 1);

// change 2 (flx adds `clamp()` with reversed `min(max(...), ...)` implementation)
this.formIndex = clamp(source.formIndex, 0, getPokemonSpecies(this.speciesId).forms.length - 1);

// change 3 (from this PR, also reverting `clamp()` to phaser's implementation)
this.formIndex = clamp(source.formIndex, 0, Math.max(getPokemonSpecies(this.speciesId).forms.length - 1, 0));

DayKev:
> I’m still unsure about the whole min > max behavior. It seems like a user error and shouldn’t be handled by the method but rather throw an error as it’s simply an invalid use of the function.

I can change it to `throw new Error("...");` instead of returning `min`.

flx-sta (dismissed):

flx-sta (approved):

Tempo-anon (approved):

PR #1151, opened 2025-05-07
[Refactor] Change `PokemonSummonData` to an interface

DayKev:
## What are the changes the user will see?
Transformed Pokémon will stay transformed when loading save data.

## Why am I making these changes?
Code cleanup.
Fixes a bug with Imposter/Transform.

## What are the changes from a developer perspective?
- `PokemonSummonData` changed from a class to an interface.
- `Pokemon#summonData` is now always initialized in the `Pokemon` class constructor via `resetSummonData()`.
- `PokemonData` constructor refactored/simplified.
- Removed `null` checks for `summonData`.
- Various tests updated.

## How to test the changes?
`npm run test:silent`

## Checklist
- [x] Otherwise: **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test:silent`)
  - [x] Have I created new automated tests (`npm run test:create`) or updated existing tests related to the PR's changes?

PR #1177, opened 2025-05-15
[Save Data] Clean up shiny variants and passive attributes in system data

### LINKED ISSUE #1162: [Save Data] Fix Shiny variants dex data
Currently the shiny and variant data is stored in 5 flags in dex attributes.
- non shiny unlocked
- shiny unlocked
- base variant unlocked
- rare variant unlocked
- epic variant unlocked

Currently the "base variant unlocked" flag is flipped when a non shiny Pokemon is caught, which results in the base shiny variant always getting considered unlocked as soon the shiny unlocked flag is true as well.
This is not the behavior we want, only properly unlocked variants should be considered unlocked.

Due to variants being added after the fact those 3 flags where added on top of the shiny/non shiny ones, resulting in the shiny flag being redundant. The following flags should be enough, make things simpler, and slightly reduced storage space:
- non shiny unlocked
- base variant unlocked
- rare variant unlocked
- epic variant unlocked

<sub>Updating this will mess up current system data, break backwards compatibility and will require creating a new `everything` save data, but no data migration. As such this should happen before any kind of public or private release.

MokaStitcher:
> [!WARNING]
> With this PR, existing system data is no longer compatible. The local storage key for system data, session data, starter preferences, run history and seen dialogue have been changed so when launching the game you will immediately be on a fresh file, but your old data will still be here in local storage should you need it and won't get overridden.
While the old system data is now incompatible and session data was made incompatible in #1151, the others can simply be renamed in local storage if you wish to keep using the information within them.

- When catching a rare or epic shiny variant, the base shiny variant is no longer unlocked
- The fact that passives are enabled or not is no longer saved on the server, but locally (see #1196 )
- ⚠️ Existing system data no longer compatible
- System data, session data, run history data, starter preferences and seen dialogues and tutorial data are now stored under a poketernity specific key in local storage, effectively resetting them (without deleting current data) and making them fully separate to potential saves from other forks
- Smaller save data

Other minor changes:
- Hatch summary now displays new non shiny catch the same way as a new shiny catch
- Starter Select shows variant icons from least rare to rarest left to right instead of right to left
- Session data naming convention changed from `SessionData, SessionData1, ..., SessionData4` to `SessionData1, SessionData2, ..., SessionData5`

Fixes #1162 
- The fact that the base variant gets unlocked by getting another was never intended, and doesn't actually work unless the non shiny form was also caught, leading to player confusion
- The current dex data stores unlocked (non-)shiny variant with 5 flags (non shiny, shiny, variant 1, variant 2, variant 3), while 4 are enough (non shiny, variant 1, variant 2, variant 3). The benefit from having an extra shiny flag is to make checking that any shiny form is unlocked easier, but this benefit is often nullified by the need to check the individual variants anyways. In the end it's complicating things more than helping them almost everywhere. And making the dex attr bigger
- Disabling passives is a very niche feature which does not justify the storage space used (more details in #1196)

#### Shiny dex attributes:
- Remove the DexAttr.SHINY flag and reorder the other flags
- Update all code accessing dex attributes to acomodate this change
   - Create helper function `gameData.getUnlockedVariantsAttr` which returns an array of the unlocked variants from a given dexAttr
   - Rewrite the cycle shiny code in starter select to make use of this

#### Unlocked Passive update (more details in #1196)
- Delete StarterData.passiveAttr and replace it with a new flag in AbilityAttr.
- Add a `passiveEnabled` field in Starter Preferences
- Update starter select to accommodate the changes, adding extra safeguard for non baby starters.

#### Save data storage update
- add new GameDataType for starter preferences
- `game-data.getDataTypeKey`: 
   - rename to `getLocalStorageKey`
   - move it to `account.ts` to prevent circular dependencies caused by the keys depending on the username
   - update the function and make every part of the code accessing local storage use this function to retrieve the key.
   - Add `pkty/` prefix to every data type that was missing it
   - Change the session data key slot number for less player confusion when exporting/importing data
- Remove the temporary starter data migration introduced in #1161 
- Remove the temporary session data migration introduced in #1013

#### Updated "everything" save file:
- every Starter gets all natures, egg moves, max ivs, abilities, passive, cost reduction, 100 meta candy and 1 win ribbon
- every Pokemon gets all its forms, genders and existing variants marked as caught and seen (only the variants that currently have sprite get unlocked, unlike before)
This required changing some starter select tests which were assuming that Bulbasaur had all 3 variants unlocked, which is no longer the case
- 500 of each type of voucher (except the 25 ones apparently, sorry 🤷 )
- all "unlockables" unlocked (endless, mini black hole, eviolite)
- all achievements unlocked
- all vouchers unlocked

## Screenshots/Videos

<details><summary>Catch panel and stuff</summary>

Catching a new shiny variant:
image

Catching a non shiny mon when only its shiny form was caught before:
image

Hatching a non shiny mon when only its shiny form was caught before:
image
image

</details> 

<details><summary>Starter Select</summary>

Mons with epic and rare variants don't get the base variant unlocked by default:
image

"Everything" save: only existing variants are unlocked. Non baby starters don't get the passive background. Variant icons ordered with base > rare > epic
image
</details> 

Login screen still properly retrieves usernames with data in local storage:
image

File names of exported session data of the first and last slot:
image
before would have been `sessionData0_Guest.pkty.txt` and `sessionData4_Guest.pkty.txt` 

</details> 

Size of exported save files (before #1161 and after this):
Fresh file: 159KB -> 136KB
Everything file: 195KB -> 160KB

## How to test the changes?

- New game. Only the default starters should be unlocked.
- Override to some Pokemon species, catch them repeatedly in different gender, form, shiny variant to make sure new unlocks are correctly shown as new, and others not
- You can also hatch eggs
- Check periodically that starter select has all the correct information from the mons you captured, and cycling through shiny forms works as intended no matter which variants are unlocked or not
- Play with unlocking passives and enabling/disabling them, making sure the UI updates properly (can use override `FREE_CANDY_UPGRADE_OVERRIDE: true`)

- Menu > Manage Data > Import Data. Import the everything save file and play around as well

## Checklist

  - [X] Have I created new automated tests (`npm run test:create`) or updated existing tests related to the PR's changes?


src/data/dex-attributes.ts
```diff
@@ -1,12 +1,11 @@
...
-  MALE: 4n,
-  FEMALE: 8n,
-  DEFAULT_VARIANT: 16n,
-  VARIANT_2: 32n,
-  VARIANT_3: 64n,
-  DEFAULT_FORM: 128n,
+  SHINY_BASE_VARIANT: 2n,
+  SHINY_RARE_VARIANT: 4n,
+  SHINY_EPIC_VARIANT: 8n,
+  MALE: 16n,
```
Tempo-anon, line 9: Having the shinies grouped together is nice but I wonder if it's better to have them at the end in case there might ever be more variants than just 3.

src/data/dex-attributes.ts
```diff
@@ -1,12 +1,11 @@
...
+  FEMALE: 32n,
+  DEFAULT_FORM: 64n,
```
Tempo-anon, line 15: Can you quickly remind me what `DEFAULT_FORM` is for?

src/field/pokemon.ts
```diff
@@ -616,8 +616,16 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-    ret |= this.variant >= 2 ? DexAttr.VARIANT_3 : this.variant === 1 ? DexAttr.VARIANT_2 : DexAttr.DEFAULT_VARIANT;
+    if (!this.shiny) {
+      ret |= DexAttr.NON_SHINY;
+    } else {
+      ret |=
+        this.variant >= 2
+          ? DexAttr.SHINY_EPIC_VARIANT
+          : this.variant === 1
+            ? DexAttr.SHINY_RARE_VARIANT
+            : DexAttr.SHINY_BASE_VARIANT;
```
Tempo-anon: chained ternaries :pikamad:

src/system/game-data.ts
```diff
@@ -1852,35 +1838,32 @@ export class GameData {
...
-        ret |= DexAttr.SHINY;
 
-        if (attr & DexAttr.VARIANT_3) {
-          ret |= DexAttr.VARIANT_3;
-        } else if (attr & DexAttr.VARIANT_2) {
-          ret |= DexAttr.VARIANT_2;
-        } else {
-          ret |= DexAttr.DEFAULT_VARIANT;
-        }
+    if (optimistic) {
```
Tempo-anon: the function `getSpeciesDefaultDexAttr` should have some documentation. I'm not sure what `optimistic` means off the bat

src/data/dex-attributes.ts
MokaStitcher: Not possible, currently the forms are at the end and they have varying lengths depending on species
The only way to allow that would be to move the form to their own attribute

src/data/dex-attributes.ts
MokaStitcher: See previous answer, it's the flag for the base form of a species. If a species has multiple forms they will appear after that flag

src/system/game-data.ts
MokaStitcher: Lots of game data should have doc yes
It means it'll return the highest tier variant, as opposed to returning the non shiny form, or lowest tier variant if non shiny is not unlocked

src/data/dex-attributes.ts
Tempo-anon: Hm, do you have some examples? Does this mean a Pokemon like Unown will have a different saved dex attr compared to something like a Bulbasaur or Charizard?

MokaStitcher: Well yes
Bulbasaur would have
`XXXX XX 1` to mark the default (base) form as unlocked 
An unown with say
`XXXX XX 010001` means that the default form (A) is not unlocked but B and F are

Tempo-anon: Oh hmmm. So the data structure only allows a single arbitrarily length field. My thought process was that an arbitrary length list would make sense for shinies since some pokemon only have 1 while some have 3 and in case in the future adding more shiny versions would result in having to change the data structure completely and break all existing save data

src/data/dex-attributes.ts
Tempo-anon: Gotcha. The example you gave above (with Unown) made it clear to me. What follows is a binary sequence where each bit corresponds to whether or not a specific form is owned

flx-sta:
@MokaStitcher  Conflicts to resolve

src/data/dex-attributes.ts
MokaStitcher: true, but I think 3 shinies is already more than enough both for artist workload and for players if someone wanted to unlock them all :p

src/system/game-data.ts
```diff
@@ -84,22 +85,23 @@ const saveKey = "x0i2O7WRiANTqPmZ"; // Temporary; secure encryption is not yet n
...
+      // TODO: add LS_PREFIX and remove the speciesId temporary migration in parseSessionData
+      return `${ret}_${loggedInUser?.username}`;
     }
     case GameDataType.SETTINGS:
       return SETTINGS_LS_KEY;
     case GameDataType.TUTORIALS:
       return TUTORIALS_LS_KEY;
     case GameDataType.SEEN_DIALOGUES:
-      return "seenDialogues";
+      return `${LS_PREFIX}/seenDialogues`;
```
flx-sta: This should probably be a constant like `SETTINGS_LS_KEY` or `TUTORIALS_LS_KEY`

src/ui/handlers/starter-select-ui-handler.ts
```diff
@@ -2266,95 +2265,77 @@ export class StarterSelectUiHandler extends MessageUiHandler {
...
-            return this.switchToVariant(starterPrefs, variant);
-          }
+
+    const variants = [
+      DexAttr.NON_SHINY,
+      DexAttr.SHINY_BASE_VARIANT,
+      DexAttr.SHINY_RARE_VARIANT,
+      DexAttr.SHINY_EPIC_VARIANT,
+    ];
+    const currentShinyVariant = isNil(starterPrefs.variant) ? props.variant : starterPrefs.variant;
```
DayKev: I think this could be simplified to this?
```suggestion
    const currentShinyVariant = starterPrefs.variant ?? props.variant;
```

src/system/game-data.ts
```diff
@@ -1825,7 +1847,12 @@ export class GameData {
     doUpdateIvs(speciesId);
   }
 
-  getSpeciesCount(dexEntryPredicate: (entry: DexEntry) => boolean): number {
+  /**
+   * Go through all species in dex data and counts those that fit the given criteria.
+   * @param dexEntryPredicate - Function that should return `true` if the given {@linkcode DexEntry} should be counted.
+   * @returns the number of Species (all Pokemon) that fit the given predicated in the dex data.
```
DayKev:
```suggestion
   * @returns the number of Species (all Pokemon) that fit the given predicate in the dex data.
```

src/system/game-data.ts
```diff
@@ -1836,7 +1863,12 @@ export class GameData {
     return speciesCount;
   }
 
-  getStarterCount(dexEntryPredicate: (entry: DexEntry) => boolean): number {
+  /**
+   * Go through all starters in dex data and counts those that fit the given criteria.
+   * @param dexEntryPredicate - Function that should return `true` if the given {@linkcode DexEntry} should be counted.
+   * @returns the number of Starters that fit the given predicated in the dex data.
```
DayKev:
```suggestion
   * @returns the number of Starters that fit the given predicate in the dex data.
```

src/system/game-data.ts
```diff
@@ -1848,54 +1880,75 @@ export class GameData {
...
   }
 
-  getSpeciesDexAttrProps(_species: PokemonSpecies, dexAttr: bigint): DexAttrProps {
+  /**
+   * Transforms a dex attribute (see {@linkcode DexAttr}) into a human readable object.
+   * Assumes that the input only has a single turned on flag per category (gender, form, variant),
+   * for example as returned by {@linkcode getSpeciesDefaultDexAttr} or obtained from a {@linkcode Pokemon}.
+   * @param _species - unused. TODO: remove
+   * @param dexAttr - The dex attribute to consider.
+   * @returns
```
DayKev: Missed the `@returns`.

src/system/game-data.ts
```diff
@@ -1940,7 +2002,71 @@ export class GameData {
...
+   */
+  public getFormIndex(attr: bigint): number {
+    if (!attr || attr < DexAttr.DEFAULT_FORM) {
+      return 0;
+    }
+    let f = 0;
+    while (!(attr & this.getFormAttr(f))) {
+      f++;
+    }
+    return f;
```
DayKev:
```suggestion
    let form = 0;
    while (!(attr & this.getFormAttr(form))) {
      form++;
    }
    return form;
```

src/system/game-data.ts
```diff
@@ -1940,7 +2002,71 @@ export class GameData {
...
+   */
+  getFormAttr(formIndex: number): bigint {
+    return BigInt(1 << formIndex) * DexAttr.DEFAULT_FORM;
+  }
+
+  /**
+   * Get the luck value associated to a given dex attribute.
+   * Currently 0 if no shiny unlocked, 1, 2 or 3 depending on the highest unlocked shiny variant.
+   * @param dexAttr - The dex attribute to consider.
+   * @returns The luck, between 0 and 3.
```
DayKev, line 2023:
```suggestion
   * @returns The luck, between 0 and 3.
   * @todo Decouple variants (and shinies entirely?) from luck
```

flx-sta (dismissed):

src/ui/handlers/starter-select-ui-handler.ts
```diff
@@ -3369,21 +3378,22 @@ export class StarterSelectUiHandler extends MessageUiHandler {
...
-        ? DexAttr.NON_SHINY
-        : DexAttr.SHINY;
+      if (shiny !== undefined ? !shiny : !(shiny = oldProps?.shiny)) {
+        this.dexAttrCursor |= DexAttr.NON_SHINY;
+      }
+      this.dexAttrCursor |= (variant !== undefined ? !variant : !(variant = oldProps?.variant))
+        ? DexAttr.SHINY_BASE_VARIANT
+        : variant === 1
+          ? DexAttr.SHINY_RARE_VARIANT
+          : DexAttr.SHINY_EPIC_VARIANT;
```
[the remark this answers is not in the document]
flx-sta: Is this actually a true statement from the AI?

test/system/set-pokemon-caught.test.ts
```diff
@@ -174,17 +175,15 @@ describe("Dex Data - Set Pokemon caught", () => {
...
+    expect(bulbaVariantsAfter).toHaveLength(1);
+    expect(bulbaVariantsAfter).toContain(DexAttr.SHINY_RARE_VARIANT);
 
     [ivyDexData, venuDexData].forEach((dexData) => {
       expect(dexData.caughtAttr & DexAttr.NON_SHINY).toBeFalsy();
-      expect(dexData.caughtAttr & DexAttr.SHINY).toBeTruthy();
-      expect(dexData.caughtAttr & DexAttr.DEFAULT_VARIANT).toBeFalsy();
-      expect(dexData.caughtAttr & DexAttr.VARIANT_2).toBeTruthy();
-      expect(dexData.caughtAttr & DexAttr.VARIANT_3).toBeFalsy();
+      const variantsAfter = gameData.getUnlockedVariantsAttr(bulbaDexData.caughtAttr);
```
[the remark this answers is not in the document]
flx-sta: ~~This seems false~~

src/ui/handlers/starter-select-ui-handler.ts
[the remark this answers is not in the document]
MokaStitcher: I hate setSpeciesDetails so much... But yes. Kind of. Before the default variant was set even if the mon was not shiny. Now it's not anymore so should ensure that
Trying to rewrite this without the nested ternary is hurting my brain though

test/system/set-pokemon-caught.test.ts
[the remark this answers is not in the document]
MokaStitcher: no it's right (the suggestion)

MokaStitcher:
~~found issues while playtesting. marking as draft for now~~
Fixed. Would appreciate any playtesting in case other parts of the code using magic numbers to check the dex attributes got missed

flx-sta:
image

Didn't notice anything out of the ordinary so far 
(Been catching back and forth. But **man!** has the game gotten hard with the new level scaling)

[pkty_data_Guest.pkty.txt](https://github.com/user-attachments/files/20377268/pkty_data_Guest.pkty.txt)

DayKev (dismissed):
The pokemon icons are animating weirdly in the egg hatch summary; they all start out bouncing (except for the shiny for some reason lol) unlike on beta, but otherwise 👍 
<p>

https://github.com/user-attachments/assets/398b7764-b0c9-4c7b-9d0c-49eb57112f6c

</p>
</details>

MokaStitcher:
> The pokemon icons are animating weirdly in the egg hatch summary; they all start out bouncing (except for the shiny for some reason lol) unlike on beta, but otherwise 👍
> 

Hmm, they don't seem to have new characteristics that should trigger the bouncing so that's not normal.
I'll look into it, thank you

src/account.ts
```diff
@@ -25,8 +25,9 @@ export function updateUserInfo(): Promise<[boolean, number]> {
         hasAdminRole: false,
       };
       let lastSessionSlot = -1;
-      for (let s = 0; s < 5; s++) {
-        if (localStorage.getItem(`sessionData${s ? s : ""}_${loggedInUser.username}`)) {
+      for (let s = 0; s < SAVE_SLOT_LIMIT; s++) {
+        // TODO: we should use getDataTypeKey(GameDataType.SESSION, s) but it creates a circular dependency
+        if (localStorage.getItem(`${SESSION_DATA_LS_KEY_PREFIX}${s ? s : ""}_${loggedInUser.username}`)) {
```
MokaStitcher: The easiest way to fix this circular dependency that I can see is to add the username as a parameter of getDataTypeKey
Which would kind of make sense but also be weird because everywhere it's called it would be given `loggedInUser.username` as a value

Another option would be to move getDataTypeKey to the account file lol
Maybe that's the way actually. Rename it `get(Account|User)DataStorageKey` or something 🤔

src/utils/save-data-utils.ts
```diff
@@ -0,0 +1,49 @@
...
+    case GameDataType.SEEN_DIALOGUES:
+      return SEEN_DIALOGUE_LS_KEY;
+    // The other data types have the username appended to the storage key
+    case GameDataType.SYSTEM:
+      prefix = SYSTEM_DATA_LS_KEY_PREFIX;
+      break;
+    case GameDataType.SESSION: {
+      prefix = SESSION_DATA_LS_KEY_PREFIX;
+      if (slotId) {
+        prefix += slotId;
```
DayKev: I wonder if it should just unconditionally append the slot id to the prefix value.

src/ui/handlers/starter-select-ui-handler.ts
```diff
@@ -1842,32 +1856,36 @@ export class StarterSelectUiHandler extends MessageUiHandler {
...
                 handler: () => {
                   if (Overrides.FREE_CANDY_UPGRADE_OVERRIDE || candyCount >= passiveCost) {
-                    starterData.passiveAttr |= PassiveAttr.UNLOCKED | PassiveAttr.ENABLED;
+                    starterData.abilityAttr |= AbilityAttr.PASSIVE;
                     if (!Overrides.FREE_CANDY_UPGRADE_OVERRIDE) {
                       starterData.candyCount -= passiveCost;
                     }
                     this.pokemonCandyCountText.setText(`x${starterData.candyCount}`);
+                    // TODO: should handle request failure better, and not allow to buy another update until it finishes
                     globalScene.gameData.saveSystem().then((success) => {
```
DayKev, line 1876: Can this not be `await`ed here?

src/system/game-data.ts
```diff
@@ -1848,54 +1843,75 @@ export class GameData {
...
+  /**
+   * Transforms a dex attribute (see {@linkcode DexAttr}) into a human readable object.
+   * Assumes that the input only has a single turned on flag per category (gender, form, variant),
+   * for example as returned by {@linkcode getSpeciesDefaultDexAttr} or obtained from a {@linkcode Pokemon}.
+   * @param _species - unused. TODO: remove
+   * @param dexAttr - The dex attribute to consider.
+   * @returns {@linkcode DexAttrProps} corresponding to the given dex attribute.
+   */
+  public getSpeciesDexAttrProps(_species: PokemonSpecies, dexAttr: bigint): DexAttrProps {
+    const female = !(dexAttr & DexAttr.MALE); // TODO is that correct for genderless species?
```
DayKev:
```suggestion
    const female = !(dexAttr & DexAttr.MALE); // TODO: is that correct for genderless species?
```

DayKev (dismissed):

src/account.ts
DayKev: `getLocalStorageKey` seems fine as a name for the function. Putting it here is also fine, it's a bit awkward to place anyway.

src/ui/handlers/starter-select-ui-handler.ts
MokaStitcher: we're not in an async function here and there so many issues with how these requests (and others) are handled, it'll be for another PR
<sub>Looking forward to yeeting all this and rewriting the handler from scratch :pikasmile:

src/utils/save-data-utils.ts
MokaStitcher: We definitely could, we could also make it not 0 indexed to make things less confusing for players when exporting data. Will look into that

src/account.ts
DayKev: Now that you moved the function you can use it here.

MokaStitcher: I- :plains:

MokaStitcher: done, sorry about that 😂

src/account.ts
```diff
@@ -46,3 +57,40 @@ export function updateUserInfo(): Promise<[boolean, number]> {
     });
   });
 }
+
+/**
+ * Retrieve the local storage key used to store the given data type.
+ * For System data, Session data, Run history and starter preferences the key depends on the username.
+ *
+ * Note: needs to be in this file to prevent circular dependencies due to `updateDateUserInfo` calling it.
```
DayKev:
```suggestion
 * Note: needs to be in this file to prevent circular dependencies due to `updateUserInfo` calling it.
```

DayKev (dismissed):

DayKev (approved):

flx-sta (approved):

PR #1180, opened 2025-05-16
 [Checkup][Ability] Guts

### LINKED ISSUE #648: [Ability][Checkup] Guts
## Ability Requirements
- [x] Documentation
- [x] Tests
- [x] While a Pokémon with this Ability has a status condition (non-volatile status), its Attack is increased by 50%. 
- [x] Additionally, a Pokémon with this Ability does not lose Attack due to burn. 
- [x] The Attack boost from Guts no longer activates if a Pokémon is frozen and uses a move that thaws it out, such as Flame Wheel or Flare Blitz. 

## Related Bugs
- Activates when a Pokemon is Frozen 

## References
- [Bulbapedia - Guts](https://bulbapedia.bulbagarden.net/wiki/Guts_(Ability))

flx-sta:

n/a

Closes #648 

- Guts test coverage

## Screenshots/Videos

n/a

## How to test the changes?

`npm run test:silent`

## Checklist


test/abilities/guts.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    override.statusEffect(StatusEffect.BURN);
+
+    await classicMode.startBattle([SpeciesId.FEEBAS]);
+    const player = field.getPlayerPokemon();
+    const enemy = field.getEnemyPokemon();
+    move.use(MoveId.TACKLE);
+    await game.toEndOfTurn();
+
+    expect(player.hasNonVolatileStatusEffect()).toBe(true);
+    expect(enemy).toHaveTakenDamage(5);
```
Tempo-anon, line 128: It's hard to tell that this test actually prevents damage reduction without seeing how much damage it would have done before being burned unless you link a relevant calculation like 
```ts
/**
 *  Lvl 9 0 Atk Guts Burned Magikarp Tackle vs. Lvl 10 0 HP / 0 Def Feebas: 5-6 (18.5 - 22.2%) -- possible 5HKO
 * Possible damage amounts: (5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 6)
  *  Lvl 9 0 Atk Guts Paralyzed Magikarp Tackle vs. Lvl 10 0 HP / 0 Def Feebas: 5-6 (18.5 - 22.2%) -- possible 5HKO
 * Possible damage amounts: (5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 6)
 */
```
or do a before/after comparison (and at lower level it's hard to see burn + guts being a 3x multiplier compared to just burned)

Tempo-anon (dismissed):
Looks mostly good to me. Just had a comment about a test

test/abilities/guts.test.ts
flx-sta: I've added coverage with simulating the burn damage reduction evaluation

test/abilities/guts.test.ts
```diff
@@ -0,0 +1,159 @@
...
+  it("should not apply an attack boost with no status effect", async () => {
+    const { field, classicMode, move } = game;
+
+    await classicMode.startBattle(SpeciesId.FEEBAS);
+    const player = field.getPlayerPokemon();
+    move.use(MoveId.SPLASH);
+    await game.toEndOfTurn();
+    const playerAtk = player.getStat(Stat.ATK);
+
+    expect(player.hasNonVolatileStatusEffect()).toBe(false);
```
DayKev: Don't we have a `hasStatusEffect` or w/e matcher?

DayKev (dismissed):

test/abilities/guts.test.ts
MokaStitcher: yes 
```suggestion
    expect(player).toHaveStatusEffect(StatusEffect.NONE);
```

or potentially 

```suggestion
    expect(player).toHaveStatusEffect(StatusEffect.NONE, { ignoreMockAbility: true });
```

MokaStitcher (commented):
How do we want guts to interact with Comatose?
Mainline doesn't have multiple abilities so we can do whatever we want in this case but it would be good to clarify if comatose should count as having a status effect for the purpose of guts, and add a test to ensure the implementation matches that

PigeonBar (dismissed):

DayKev:
> How do we want guts to interact with Comatose? Mainline doesn't have multiple abilities so we can do whatever we want in this case but it would be good to clarify if comatose should count as having a status effect for the purpose of guts, and add a test to ensure the implementation matches that

I think Comatose + Guts is fine, it gives a bit more use to a niche ability.

PR #1181, opened 2025-05-16
[Dev][GH] Merge lint workflows into one workflow file

flx-sta:

n/a

These files all have the same triggers and cover the same context.
To reduce the necessary resources for actions I've merged them into one file

- create lint.yml workflow
    - merge biome.yml
    - merge dependencies.yml
    - merge eslint.yml

## Screenshots/Videos

n/a

## How to test the changes?

See "Checks"

## Checklist


Tempo-anon (dismissed):

DayKev:
Does this actually change anything? Comparing the workflow output from this PR vs one of my PRs and it's the exact same.

DayKev:
> Does this actually change anything? Comparing the workflow output from this PR vs one of my PRs and it's the exact same.

[Biome workflow from kebab-case PR](https://github.com/Despair-Games/poketernity/actions/runs/15040250567/job/42270046963?pr=1171)
[Biome workflow from this PR](https://github.com/Despair-Games/poketernity/actions/runs/15076878298/job/42386221766?pr=1181)

[ESLint workflow from kebab-case PR](https://github.com/Despair-Games/poketernity/actions/runs/15040250588/job/42270046984?pr=1171)
[ESLint workflow from this PR](https://github.com/Despair-Games/poketernity/actions/runs/15076878298/job/42386221697?pr=1181)

DayKev:
There might be a way to create an "environment set up" job and then have the linting jobs require it to finish first (so they still run in parallel but don't all also run their own set-up).

flx-sta:
> There might be a way to create an "environment set up" job and then have the linting jobs require it to finish first (so they still run in parallel but don't all also run their own set-up).

~~Would you be fine to take this as a first step for this optimization?~~

EDIT: I can run them all in the same job. This way they can't run in parallel anymore but they should be much faster never the less as they don't all need their own environment booted

DayKev:
Actually, I wonder what will happen if one of the earlier linters catches an error; will the later linters not run then?

flx-sta:
> Actually, I wonder what will happen if one of the earlier linters catches an error; will the later linters not run then?

They probably wont

----

EDIT: Added a `continue-on-error: true` flag to the steps 

-> turns out that's not a solution as it's reporting a success at the end

flx-sta:
Adding another step that checks the results of each lint seems to solve that issue

Example failure:
https://github.com/Despair-Games/poketernity/actions/runs/15078252716/job/42390282022

DayKev (approved):

DayKev:
Updated the ruleset to account for the workflow changes.

PR #1193, opened 2025-05-17
[Dev] Enable `lint/suspicious/noPrototypeBuiltins` Biome rule

DayKev:
N/A

Enabling recommended rule: https://biomejs.dev/linter/rules/no-prototype-builtins/

`tsconfig.json` updated to add `"ES2022.Object"` to `"lib"`, granting access to [`Object.hasOwn`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn).

`Object.hasOwn(thing, prop)` should be used instead of `thing.hasOwnProperty(prop)`.

## Checklist

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -2656,7 +2656,7 @@ export default class BattleScene extends SceneBase {
 
   validateAchv(achv: Achievement, ...args: unknown[]): boolean {
     if (
-      (!Object.hasOwn(this.gameData.achvUnlocks.hasOwnProperty, achv.id) || Overrides.ACHIEVEMENTS_REUNLOCK_OVERRIDE)
+      (!Object.hasOwn(this.gameData.achvUnlocks, achv.id) || Overrides.ACHIEVEMENTS_REUNLOCK_OVERRIDE)
       && achv.validate(...args)
     ) {
       this.gameData.achvUnlocks[achv.id] = new Date().getTime();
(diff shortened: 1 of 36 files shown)


Tempo-anon (dismissed):

src/battle-scene.ts
```diff
@@ -2656,12 +2656,12 @@ export default class BattleScene extends SceneBase {
 
   validateAchv(achv: Achievement, ...args: unknown[]): boolean {
     if (
-      (!this.gameData.achvUnlocks.hasOwnProperty(achv.id) || Overrides.ACHIEVEMENTS_REUNLOCK_OVERRIDE)
+      (!Object.hasOwn(this.gameData.achvUnlocks.hasOwnProperty, achv.id) || Overrides.ACHIEVEMENTS_REUNLOCK_OVERRIDE)
```
PigeonBar:
```suggestion
      (!Object.hasOwn(this.gameData.achvUnlocks, achv.id) || Overrides.ACHIEVEMENTS_REUNLOCK_OVERRIDE)
```

PigeonBar (dismissed):

src/battle-scene.ts
DayKev: Wtf, how did that happen... that line doesn't look like something that should have messed with the regex.

PR #1194, opened 2025-05-18
[Refactor] Replace `enum`s with `const` objects (part 1)

DayKev:
Less memory usage hopefully.

> [!NOTE]
> current "Battle Style" setting will be messed up by the change, go to the settings (First tab) and move the cursor to another option and back to reset it to a correct value

TypeScript enums are very inefficient.
<details><summary>Input TypeScript</summary>
<p>

```ts
enum ExampleEnum {
  one = 1,
  two,
  three,
}

const ExampleConst = {
  one: 1,
  two: 2,
  three: 3,
} as const;
```

</p>
</details> 
<details><summary>Output Javascript</summary>
<p>

```js
"use strict";
var ExampleEnum;
(function (ExampleEnum) {
    ExampleEnum[ExampleEnum["one"] = 1] = "one";
    ExampleEnum[ExampleEnum["two"] = 2] = "two";
    ExampleEnum[ExampleEnum["three"] = 3] = "three";
})(ExampleEnum || (ExampleEnum = {}));
const ExampleConst = {
    one: 1,
    two: 2,
    three: 3,
};
```

</p>
</details> 

TypeScript enums are replaced with const objects.
<details><summary>Example</summary>
<p>

```ts
export const BattleType = {
  WILD: 1,
  TRAINER: 2,
  CLEAR: 3,
  MYSTERY_ENCOUNTER: 4,
} as const;

export type BattleType = (typeof BattleType)[keyof typeof BattleType];
```

</p>
</details>

They can be used mostly the same. A couple of differences however:
- Instead of `getTSEnumValues(EnumName)` use `Object.values(EnumName)` and instead of `getTSEnumKeys(EnumName)` use `Object.keys(EnumName)`
- Use `function fn(param: typeof EnumName.Value)` instead of `function fn(param: EnumName.Value)` if you want to use specific enum values as a type hint
- To emulate reverse mapping use the new `enumValueToKey(EnumName, EnumName.Value)` function

Changed enums so far: `AbAttrFlag` through to `ChallengeType`, excluding `AbilityId` and `BattlerIndex` (the former is too big and will likely be its own PR, and the latter needs special handling due to the way it's [ab]used in the codebase).

Most enums were changed to start at `1` instead of `0` unless the `0` value was used specifically as a "none" value or changing the enum values would break things (`TODO`s were left for those).

## Checklist
- [-] The PR is self-contained and cannot be split into smaller PRs?
  - I mean technically I could make 100 smaller PRs for each enum conversion but that's a bit silly. I'll probably make 3 or 4 PRs for this.


src/data/abilities/ab-attrs/post-item-lost-apply-battler-tag-ab-attr.ts
```diff
@@ -15,12 +15,12 @@ export class PostItemLostApplyBattlerTagAbAttr extends PostItemLostAbAttr {
   }
 
   override apply(pokemon: Pokemon, simulated: boolean): boolean {
-    if (!pokemon.getTag(this.tagType)) {
-      if (!simulated) {
-        pokemon.addTag(this.tagType);
-      }
-      return true;
+    if (pokemon.getTag(this.tagType)) {
```
flx-sta: shouldn't this be `hasTag()`?

src/utils/common-utils.ts
```diff
@@ -232,3 +232,27 @@ export function calcAccuracyMultiplier(userAccStage: number, targetEvaStage: num
...
+ * @param val - The value to get the key of
+ * @returns The name of the key with the specified value
+ * @example
+ * const thing = {
+ *   one: 1,
+ *   two: 2,
+ * } as const;
+ * console.log(enumValueToKey(thing, thing.two)); // output: "two"
+ */
+export function enumValueToKey(input: object, val: number): string {
```
flx-sta: Should this really be called `enumValueToKey`? Technically it's no longer an enum but a const

src/data/abilities/ab-attrs/post-item-lost-apply-battler-tag-ab-attr.ts
DayKev: Oh yeah, I didn't look closely at the condition, just inverted it. Will fix.

src/utils/common-utils.ts
DayKev: We're still using them as `enum`s. That is why I renamed the old functions to `getTSEnum[Key|Value|Length]` though (and specified in the TSDocs), to make it explicit what kind of `enum` they're working on.

src/data/moves/move-attrs/swap-arena-tags-attr.ts
```diff
@@ -7,7 +7,7 @@ import type { Move } from "#moves/move";
 import { MoveEffectAttr } from "#moves/move-effect-attr";
 import i18next from "i18next";
 
-export const courtChangeArenaTags = [
+export const courtChangeArenaTags = Object.freeze<ArenaTagType[]>([
```
MokaStitcher:
```suggestion
const courtChangeArenaTags = Object.freeze<ArenaTagType[]>([
```

src/data/mystery-encounters/utils/encounter-phase-utils.ts
```diff
@@ -1087,7 +1087,10 @@ export function calculateMEAggregateStats(baseSpawnWeight: number) {
       // Otherwise, roll encounter
 
       const roll = randSeedInt(256);
-      validMEfloorsByBiome.set(BiomeId[currentBiome], (validMEfloorsByBiome.get(BiomeId[currentBiome]) ?? 0) + 1);
+      validMEfloorsByBiome.set(
+        enumValueToKey(BiomeId, currentBiome),
+        (validMEfloorsByBiome.get(enumValueToKey(BiomeId, currentBiome)) ?? 0) + 1,
+      );
```
MokaStitcher:
```suggestion
      const biomeKey = enumValueToKey(BiomeId, currentBiome);
      validMEfloorsByBiome.set(biomeKey, (validMEfloorsByBiome.get(biomeKey) ?? 0) + 1);
```

src/data/mystery-encounters/utils/encounter-phase-utils.ts
```diff
@@ -1123,7 +1126,10 @@ export function calculateMEAggregateStats(baseSpawnWeight: number) {
             : tierValue > rareThreshold
               ? ++numEncounters[2]
               : ++numEncounters[3];
-        encountersByBiome.set(BiomeId[currentBiome], (encountersByBiome.get(BiomeId[currentBiome]) ?? 0) + 1);
+        encountersByBiome.set(
+          enumValueToKey(BiomeId, currentBiome),
+          (encountersByBiome.get(enumValueToKey(BiomeId, currentBiome)) ?? 0) + 1,
+        );
```
MokaStitcher:
```suggestion
        const biomeKey = enumValueToKey(BiomeId, currentBiome);
        encountersByBiome.set(biomeKey, (encountersByBiome.get(biomeKey) ?? 0) + 1);
```

src/enums/ab-attr-flag.ts
```diff
@@ -110,229 +110,231 @@ import type { WonderSkinAbAttr } from "#abilities/wonder-skin-ab-attr";
 /* eslint-enable @typescript-eslint/no-unused-vars */
 // -- end tsdoc imports --
 
-export enum AbAttrFlag {
+export const AbAttrFlag = {
   /** @see {@linkcode AbAttr} */
-  UNSPECIFIED,
+  UNSPECIFIED: -1,
```
MokaStitcher, line 115: is there a reason to use -1 rather than 0?

src/enums/ab-attr-flag.ts
```diff
@@ -110,229 +110,231 @@ import type { WonderSkinAbAttr } from "#abilities/wonder-skin-ab-attr";
 /* eslint-enable @typescript-eslint/no-unused-vars */
 // -- end tsdoc imports --
 
-export enum AbAttrFlag {
+export const AbAttrFlag = {
```
MokaStitcher, line 113: with "enums" as large as this, I wonder if it would make sense to add tests that check that there are no duplicate values, it seems easy to make a mistake that wouldn't get caught

src/enums/ability-apply-mode.ts
```diff
@@ -1,8 +1,10 @@
-export enum AbilityApplyMode {
+export const AbilityApplyMode = {
   /** Applies abilities as normal, without restrictions */
-  DEFAULT,
+  DEFAULT: 1,
   /** Only applies abilities that were previously applied in the current battle */
```
MokaStitcher, line 4:
```suggestion
  /** Only applies abilities that were previously revealed in the current battle */
```

MokaStitcher: or is "applied" correct because an ability is considered revealed after it's been applied once? 🤔

src/enums/anim-blend-type.ts
```diff
@@ -1,5 +1,8 @@
-export enum AnimBlendType {
-  NORMAL,
-  ADD,
-  SUBTRACT,
-}
+/** @todo make this start at 1 */
```
MokaStitcher, line 1: not for this PR?

src/enums/anim-frame-target.ts
```diff
@@ -6,6 +6,7 @@ export const AnimFrameTarget = {
   /**
    * Affects the animation's source or start point, e.g.
    * a Pokemon using a move.
+   * @todo make this start at 1
```
MokaStitcher, line 9: same question

src/enums/battle-style.ts
```diff
@@ -3,7 +3,9 @@
  * - 'Switch' - The option to switch the active pokemon at the start of a battle will be displayed.
  * - 'Set' - The option to switch the active pokemon at the start of a battle will not display.
  */
-export enum BattleStyle {
-  SWITCH,
-  SET,
-}
+export const BattleStyle = {
```
MokaStitcher, line 6: hmm, this is going to mess up the existing settings right? (hopefully won't cause issues with the UI itself)

src/enums/biome-id.ts
```diff
@@ -1,39 +1,41 @@
...
-  SLUM,
-  SNOWY_FOREST,
-  STEAM_VENT,
-  CHARGESTONE_CAVE,
-  ISLAND = 40,
-  LABORATORY,
-  END = 50,
-}
+export const BiomeId = {
+  TOWN: 1,
```
MokaStitcher: I don't know why but I feel like there must be somewhere in the code using the assumption that town is 0 :')

src/field/pokemon.ts
```diff
@@ -3471,7 +3471,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
       return false;
     }
 
-    const stubTag = new BattlerTag(tagType, 0, 0);
+    const stubTag = new BattlerTag(tagType, BattlerTagLapseType.CUSTOM, 0);
```
MokaStitcher, line 3484: so before it defaulted to faint, is it ok to default to custom now?

src/modifier/init-modifier-types.ts
```diff
@@ -211,22 +211,15 @@ export function initModifierTypes() {
...
-        randBerryType = BerryType.LEPPA;
-      } else {
-        randBerryType = berryTypes[randSeedInt(berryTypes.length - 3) + 2];
+      // Sitrus, Lum, and Leppa have higher weight to be chosen
+      const berryTypes: BerryType[] = [BerryType.SITRUS, BerryType.LUM, BerryType.LEPPA];
+      for (const value of Object.values(BerryType)) {
+        berryTypes.push(value);
       }
-      return new BerryModifierType(randBerryType);
+      return new BerryModifierType(randSeedItem(berryTypes));
```
MokaStitcher: I'm not sure this is fully equivalent. Before there was like 1/2 to have either sitrus, lum and leppa
I think this would be equivalent if there were only 3 other berry type, but that's not the case

MokaStitcher (commented):
not done but have to run

src/data/moves/move-attrs/swap-arena-tags-attr.ts
DayKev: This is imported in `init-moves.ts` for the move condition.

src/enums/ab-attr-flag.ts
DayKev: Possibly. I wonder if there's a way to just force there to not be duplicate values in the first place. :thinking:

src/enums/ab-attr-flag.ts
DayKev: I only want to use `0` for values that are explicitly some sort of "none" value. I'm not sure if this qualifies.

src/enums/ability-apply-mode.ts
DayKev: No idea, I think you'd have to ask inner.

src/enums/anim-blend-type.ts
DayKev: Changing `AnimFrameTarget` requires changing all of the animation json files, so that one is definitely not for this PR lol. I'm not sure about `AnimBlendType`, I can double check.

src/enums/battle-style.ts
DayKev: Oh, I'll test it to see if it breaks anything.

src/enums/biome-id.ts
DayKev: It didn't seem like it, and there were already Biomes with non-consecutive IDs before this PR so hopefully everything should be handled automatically.

src/field/pokemon.ts
DayKev: This is a "fake" tag that's only passed to the `AbAttr`s below to see if they trigger, and they don't check the lapse type.

src/modifier/init-modifier-types.ts
DayKev: It should be. Before: Sitrus, Lum and Leppa had 2 rolls to be chosen instead of 1, and all other berries had 1 roll to be chosen. I replicated that with a simpler setup.
Basically, each berry was mapped to a number from `0-11`, with Sitrus being mapped to `0` and `1`, Lum to `2` and `3`, Leppa to `4` and `5`, and the other six to `6-11`.
`0-5` -> Sitrus/Lum/Leppa, `6-11` -> roll again `0-5` (though I don't think rolling again matters since `0-5` and `6-11` are equivalently sized intervals) and pick that entry from the berry type array (with Sitrus, Lum, and Leppa removed from the array, that's what the `-3 + 2` is for).

src/enums/biome-id.ts
DayKev: Just kidding, apparently something was. 🙃

src/enums/battle-style.ts
DayKev: Seems to work fine if you had it as "set" style before the change. I'll test what happens if you had it as "switch" before the change.

DayKev: It causes a warning in the console but still works fine, and updates properly when you change it; the error handling catches it correctly.

src/enums/anim-blend-type.ts
DayKev: Yeah, the animation json files specify `0|1|2` for `blendType` so they'd all have to be updated for this one too.

src/modifier/init-modifier-types.ts
DayKev: Actually hm, I thought there were only 9 berries.

DayKev: Reverted it since I'm not trying to make balance changes here.

DayKev: ...Well, actually lemme check something.

DayKev: Okay, after double checking it should be reverted to the old behavior... what a ridiculous function.

DayKev: Before it sliced off the first 2 entries and the last entry in `BerryType` (which were Sitrus, Lum, and Leppa respectively) and then got a random number between those values to return. I simplified it/made it more understandable by just directly removing those berries from the possible values instead.

DayKev: Made an issue to propose the simplification #1220

src/modifier/init-modifier-types.ts
```diff
@@ -211,10 +211,13 @@ export function initModifierTypes() {
 
   modifierTypes.BERRY = () =>
     new ModifierTypeGenerator((_party: Pokemon[], pregenArgs?: any[]) => {
-      if (pregenArgs && pregenArgs.length === 1 && pregenArgs[0] in BerryType) {
+      if (pregenArgs && pregenArgs.length === 1 && Object.values(BerryType).includes(pregenArgs[0])) {
         return new BerryModifierType(pregenArgs[0] as BerryType);
       }
-      const berryTypes = getTSEnumValues(BerryType);
+      const berryTypes = Object.values(BerryType).filter((v) => {
+        const removeBerries: BerryType[] = [BerryType.SITRUS, BerryType.LUM, BerryType.LEPPA] as const;
```
MokaStitcher:
```suggestion
      const removeBerries: BerryType[] = [BerryType.SITRUS, BerryType.LUM, BerryType.LEPPA] as const;
      const berryTypes = Object.values(BerryType).filter((v) => {
```
just so that it doesn't recreate the array for every berry

MokaStitcher (dismissed):
lgtm

src/enums/biome-id.ts
MokaStitcher: not surprised since Town is unlike other biomes 😓

src/data/moves/move-attrs/swap-arena-tags-attr.ts
MokaStitcher: Oh I didn't notice it was used at another spot in the file, my bad 👍

src/enums/ab-attr-flag.ts
MokaStitcher: that would be ideal yes 🤔

src/enums/battle-style.ts
MokaStitcher: updated the PR description with that information, for reference
if it was to 'set' before I assume it is now 'switch' lol

DayKev: Yeah if you had it as "set" before this PR it'll swap to "switch", and if you had it as "switch" it stays as switch and prints a warning to the console when you open the settings menu.

src/modifier/init-modifier-types.ts
DayKev: Ah, good point.

src/enums/ab-attr-flag.ts
DayKev: There's a TS-ESLint rule to prevent duplicate TypeScript enum values... https://typescript-eslint.io/rules/no-duplicate-enum-values/
There doesn't appear to be anything for objects though, unfortunately.

MokaStitcher: I didn't know it was possible to have duplicate values in enums, that's wild :')

MokaStitcher (dismissed):

src/enums/ab-attr-flag.ts
DayKev: Yeah it just partially overwrites earlier entries.
```ts
enum ExampleEnum {
  one = 1,
  two = 1,
}

console.log(ExampleEnum[1]); // "two"
console.log(ExampleEnum["1"]); // "two"
console.log(ExampleEnum["one"]); // 1
console.log(ExampleEnum.one); // 1
console.log(ExampleEnum.two); // 1
console.log(ExampleEnum["two"]); // 1
```

src/utils/common-utils.ts
flx-sta: I think the function name is still misleading, as the input is never actually an enum. In fact you are stating that it shouldn't have enums as input.

I understand that they are still "considered" enums but idk. Just my 10ct.
Ultimately I can bare with it

src/utils/common-utils.ts
```diff
@@ -232,3 +232,27 @@ export function calcAccuracyMultiplier(userAccStage: number, targetEvaStage: num
...
+  for (const [key, value] of Object.entries(input)) {
+    if (val === value) {
+      return key;
+    }
+  }
+  console.error("Invalid value passed to `enumValueToKey`!");
+  return "Invalid enum value!";
```
flx-sta: This should report an Error I'd say. Not some odd string return with a console-error only.
If you throw an error make sure to add `never` as return type

flx-sta (dismissed):
LGTM.
2 minor questions/concerns

src/utils/common-utils.ts
DayKev: I wasn't sure if I should throw an error or not for this. I can switch it over.

src/utils/common-utils.ts
DayKev: I'm not sure what a better name would be. It's not designed to work with anything that isn't being used as an enum.

src/utils/common-utils.ts
DayKev: Turns out that revealed a bug lmao, guess it's the right approach.

src/data/init/init-anims.ts
```diff
@@ -92,7 +93,7 @@ export async function populateAnims() {
                 Number.parseFloat(values[3]),
                 Number.parseInt(values[4]) === 1,
                 Number.parseInt(values[6]) === 1,
-                Number.parseInt(values[5]),
+                Number.parseInt(values[5]) as AnimBlendType,
```
flx-sta, line 96: May I ask why only this one gets a type?

src/data/moves/move-attrs/swap-arena-tags-attr.ts
flx-sta: Since those are constants that are used somewhere else they should go into a `constants/arena-tag-constants.ts` file and be put into `UPPSER_SNAKE_CASE`

src/data/moves/move-attrs/swap-arena-tags-attr.ts
```diff
@@ -26,49 +26,46 @@ export const courtChangeArenaTags = [
...
+  constructor() {
     super(true);
-    this.swappableTags = SwapTags;
   }
 
   override applyEffect(user: Pokemon, _target: Pokemon, _move: Move): boolean {
-    const tagPlayerTemp = globalScene.arena.getTags((t) => this.swappableTags.includes(t.tagType), ArenaTagSide.PLAYER);
-    const tagEnemyTemp = globalScene.arena.getTags((t) => this.swappableTags.includes(t.tagType), ArenaTagSide.ENEMY);
+    const playerTags = globalScene.arena.getTags((t) => courtChangeArenaTags.includes(t.tagType), ArenaTagSide.PLAYER);
+    const enemyTags = globalScene.arena.getTags((t) => courtChangeArenaTags.includes(t.tagType), ArenaTagSide.ENEMY);
```
flx-sta: NIT: `globalScene.arena` is used a lot here. Consider using the globalScene destructuring
```ts
const { arena } = globalScene;
```

src/data/mystery-encounters/utils/encounter-phase-utils.ts
```diff
@@ -1020,7 +1020,7 @@ export function calculateMEAggregateStats(baseSpawnWeight: number) {
   const numRuns = 1000;
   let run = 0;
   const biomes = Object.keys(BiomeId).filter((key) => Number.isNaN(Number(key)));
-  const alwaysPickTheseBiomes = [
+  const alwaysPickTheseBiomes: readonly BiomeId[] = [
```
flx-sta, line 1023: Q: When to use `Object.freeze()` and when to use `readonly T[]`?

<sub> We should try to keep it consistent </sub>

src/enums/ab-attr-flag.ts
flx-sta: Maybe we should create a _factory_ utils function for these _enums_ ? (I hate to call them that, as they aren't enums anymore.)

Which would basically do a quick wrap in a set, check if length is identical and return back the original `as const`?

```ts
export function createEnum<T>(array: T[]) {
  const set = new Set(array);
  
  if (set.size !== array.length) {
    throw new Error("Duplicate bla bla yada yada");
  }

  return array as const;
}
```

src/enums/ai-type.ts
```diff
@@ -1,5 +1,7 @@
-export enum AiType {
-  RANDOM,
-  SMART_RANDOM,
-  SMART,
-}
+export const AiType = {
+  RANDOM: 1,
+  SMART_RANDOM: 2,
+  SMART: 3,
+} as const;
```
flx-sta, line 5: Are there any comments we could add to explain those AI types better?
(I know, out of scope. Gathering information here)

src/enums/battle-command.ts
```diff
@@ -6,13 +6,17 @@ import type { CommandPhase } from "#phases/command-phase";
 
 /**
  * Commands that can be executed from a {@linkcode CommandPhase}.
+ *
+ * @todo Decouple values from cursor positioning
```
flx-sta, line 10: What does that mean?

src/enums/battle-command.ts
```diff
@@ -6,13 +6,17 @@ import type { CommandPhase } from "#phases/command-phase";
...
 // Note: Do not re-order this without a good reason! The cursor positioning
 // relies on their values being as they are now.
-export enum BattleCommand {
-  FIGHT,
-  BALL,
-  POKEMON,
-  RUN,
-  TERA,
-}
+export const BattleCommand = {
```
flx-sta, line 14: Should it also start at `1` or was that just connected to anim files?

src/utils/common-utils.ts
```diff
@@ -232,3 +232,27 @@ export function calcAccuracyMultiplier(userAccStage: number, targetEvaStage: num
...
+ * @throws An `Error` if an invalid enum value is passed to the function
+ */
+export function enumValueToKey(input: object, val: number | string): string | never {
+  for (const [key, value] of Object.entries(input)) {
+    if (val === value) {
+      return key;
+    }
+  }
+  throw new Error(`Invalid value passed to \`enumValueToKey\`! Value: ${val}`);
+}
```
flx-sta: You should auto-inject the type through generic tpying to persist the actual type of return

```suggestion
export function enumValueToKey<T extends Record<string, string | number>>(input: T, val: T[keyof T]): keyof T {
  for (const [key, value] of Object.entries(input)) {
    if (val === value) {
      return key as keyof T;
    }
  }
  throw new Error(`Invalid value passed to \`enumValueToKey\`! Value: ${val}`);
}
```

| Before | After |
|--------|--------|
| image | image |

src/phases/summon-phase.ts
```diff
@@ -283,8 +284,9 @@ export class SummonPhase extends PartyMemberPokemonPhase {
 
     if (
       !this.loaded
-      || [BattleType.TRAINER, BattleType.MYSTERY_ENCOUNTER].includes(globalScene.currentBattle.battleType)
-      || globalScene.currentBattle.waveIndex % 10 === 1
+      || battleType === BattleType.TRAINER
+      || battleType === BattleType.MYSTERY_ENCOUNTER
```
flx-sta, line 288: Q: Why not stick with the `includes` pattern here?

src/data/init/init-anims.ts
DayKev: It's the only one that needed to be cast to prevent errors.

src/data/mystery-encounters/utils/encounter-phase-utils.ts
DayKev: `Object.freeze` for global constants, temporary variables are just `as const` (or in this case `readonly T[]` to prevent errors).

src/enums/ai-type.ts
DayKev: The idea is that "random" should be more or less true random, "smart" should try to make the correct play, and "smart random" should be halfway between the other two... I guess.

src/enums/battle-command.ts
DayKev: The UI uses the enum values to position the cursor, which means the enum values can't be changed or the cursor breaks.

src/utils/common-utils.ts
DayKev: Ah, good idea.

src/phases/summon-phase.ts
DayKev: Causes a type error.

DayKev: If it was more than 2 values I'd use a buffer variable (like `const battleTypes: readonly BattleType[] = [...];`) to keep the `.includes`.

src/data/mystery-encounters/utils/encounter-phase-utils.ts
flx-sta: I think we should keep it consistent with always Object.freeze but we can vote on that. It's not urgent and shouldn't hold up this PR

src/enums/battle-command.ts
flx-sta: lol. Yeah that's kinda wild

src/utils/common-utils.ts
flx-sta: Feel free to experiment with _better_ or different approaches but you get the goal of it

src/enums/ai-type.ts
innerthunder: I believe there's more detail on the different types in the comments of `EnemyPokemon.getNextMove`. Knowing the difference between `SMART_RANDOM` and `SMART` probably requires more than a one-line explanation, so I think it'd be best to just add a link to `getNextMove` and refer to that.

The AI Rework doesn't make use of this enum, though, so it might not be worth the trouble anyway.

src/enums/ab-attr-flag.ts
DayKev: I don't think we should be adding runtime cost for this.

MokaStitcher (approved):

flx-sta (approved):

PR #1195, opened 2025-05-18
[UI][Refactor][P3] Add mode attribute to Starter Select

MokaStitcher:

Going up from the sort filter to the start button no longer shows the last hovered Pokemon's information

- Found this bug
- Right now starter select ui handler does not keep track of its current state cleanly. It has a boolean for filter mode, but for everything else it relies on checking the visibility of the various cursors sprites to figure out its state. Since the bug was caused by moving from the filtering state to the "cursor on start button" state, this is a good opportunity to clean this up

- create a const `StarterSelectMode` to keep track of the following states (not an enum because PR 1194):
   - handler not active
   - moving the cursor around in the starter grid
   - in filtering mode
   - moving the cursor in the selected starters (party)
   - cursor on the start button
- Create `currentMode` attribute to keep track of the current mode. Remove redundant `filterMode` attribute
- Create a method `setMode` to switch from a mode to another
- Call `setMode` (hopefully) everywhere it was needed
- Replace checks of the current state using the visibility of some UI objects with a check of the current mode

## Screenshots/Videos

Before, cursor on "start" button but showing a mon's information (the pokemon's icon is also jumping up and down):
image

After:
image

## How to test the changes?

- New Game
- For the bug:
  - Add Pokemon(s) to the team
  - Move your cursor to the team on the right
  - Go up to the "Sort" Filter
  - Go up to wrap around to the "Start" button.
- Everything else: play around in starter select, moving from an element to the another with or without Pokemon in the party

## Checklist


src/ui/handlers/starter-select-ui-handler.ts
```diff
@@ -90,6 +90,16 @@ interface LanguageSetting {
   starterInfoYOffset?: number;
 }
 
+const StarterSelectMode = {
+  NONE: 0,
+  SELECTION: 1,
+  FILTER: 2,
+  PARTY: 3,
+  START: 4,
+};
```
innerthunder:
```suggestion
const StarterSelectMode = {
  NONE: 0,
  SELECTION: 1,
  FILTER: 2,
  PARTY: 3,
  START: 4,
} as const;
```
Using `AnimFrameTarget` as an example...

Before:

image

After:

image

MokaStitcher: ah, thought it was only needed if it was exported, thank you!

innerthunder (commented):
Not sure if this is introduced by this PR, but the filters for Gen, Type, and Caught currently apply even when not selected:

https://github.com/user-attachments/assets/ca3ce35b-557f-4b55-8800-ed725cfafe23

This only happens if no filters are selected for the category. I didn't notice any other potential issues while playtesting.

MokaStitcher:
> Not sure if this is introduced by this PR, but the filters for Gen, Type, and Caught currently apply even when not selected:

Unless I misundertand this is the intended behavior, to mimic the pre-filters version of starter select where it was possible to go from a gen to the next by simply pressing up/down

However if you press action when hovering any of the options you can turn them on and off which allows you to filter multiple gens or types at once
image

innerthunder:
> > Not sure if this is introduced by this PR, but the filters for Gen, Type, and Caught currently apply even when not selected:
> 
> Unless I misundertand this is the intended behavior, to mimic the pre-filter version of starter select where it was possible to from a gen to the next by pressing simply pressing up/down

I think it's good to have the filters previewed this way, but I personally expected them to reset if they're not selected after escaping from the filter. It's especially unintuitive with the other filters not working the same way.

This is all out of scope for this PR, though, so I'll approve :+1:

innerthunder (dismissed):

src/ui/handlers/starter-select-ui-handler.ts
```diff
@@ -90,6 +90,16 @@ interface LanguageSetting {
   starterInfoYOffset?: number;
 }
 
+const StarterSelectMode = {
```
Tempo-anon, line 94: Is this just a const to keep track of which part of the starter select the cursor is at? Would be good to have some doc explaining what each one is

src/ui/handlers/starter-select-ui-handler.ts
```diff
@@ -90,6 +90,16 @@ interface LanguageSetting {
...
+  NONE: 0,
+  SELECTION: 1,
+  FILTER: 2,
+  PARTY: 3,
+  START: 4,
```
DayKev: Based on @Tempo-anon's comment. I don't feel strongly either way for this (the property names seem good enough, except maybe "selection").
```suggestion
/** Possible cursor states */
const StarterSelectMode = {
  /** Handler is not active */
  NONE: 0,
  /** A Pokemon in the starter grid is selected */
  SELECTION: 1,
  /** A filter option is selected */
  FILTER: 2,
  /** A Pokemon in the player Party section is selected */
  PARTY: 3,
  /** The "Start" button is selected */
  START: 4,
```

DayKev (dismissed):
Moved the cursor around "randomly" and everything seemed to work. 👍

src/ui/handlers/starter-select-ui-handler.ts
MokaStitcher: Thank you. Remembered that when i started rewriting Starter Select way back when I went with `STARTER_GRID` with is clearer I think, so switched to that

DayKev (approved):

PR #1196, opened 2025-05-18
[Save Data] Merge Starter Data's passive and ability attributes

MokaStitcher:
> [!WARNING]
> this targets #1177 because both affect system data and I'd rather not make a 3rd PR that breaks system data 
Original Warning from that PR:
> With this PR, existing system data is no longer compatible. The local storage key for system data, run history and seen dialogue have been changed so when launching the game you will immediately be on a fresh file, but your old data will still be here in local storage should you need it and won't get overridden.
While the old system data is now incompatible, the other 2 can simply be renamed in local storage if you wish to keep using the information within them.

- The fact that passives are enabled or not is no longer saved on the server, but locally
- Nothing else except slightly smaller save data

Currently StarterData has an attribute for unlocked abilities, with bit flags for ability 1, ability 2 and hidden ability.
It has a separate attribute for passive, with a bit flag for it being unlocked, and one for it being enabled.
This raises multiple concerns:
- Unlocked passives are enabled by default. Disabling a passive would only get used in very niche situations. Most players will never disable any passive, making all these "passive enabled" flags take server storage space for very little to no benefit
- This is currently the only player preference (aside from movesets) stored in the starter data, on the server. It would make more sense for it to behave like the other preferences for ability, shinyness, form, nature and gender and be kept locally.
- While in the future it would be good to store those starter preferences in the server, they should IMO be regrouped together, not scattered across the attributes that denote whether they are unlocked or not.

- Delete `StarterData.passiveAttr` and replace it with a new flag in AbilityAttr.
I chose this approach over simply changing the passiveAttr to a boolean to give us I hope more flexibility in potential future passive-related updates. However I'm happy to hear arguments for using a boolean instead
- Add `passiveEnabled` field to starter preferences in starter select.
- Add extra safeguards in starter select to ensure that non baby starters like Pikachu do not get access to their baby's passive.
- Update the `everything` and `empty save` save files in test utils. With extra elements in it compared to #1177:
   - 500 of each type of voucher (except the 25 ones apparently, sorry 🤷 )
   - All achievements unlocked
   - All vouchers unlocked
   - All starters get a ribbon

## Screenshots/Videos
<details><summary>everything save file</summary>

image
image
image
image

</details> 

Size of exported save files (before https://github.com/Despair-Games/poketernity/pull/1161 and after this):
```txt
               before 1161  | after 1177 |  after this
Fresh file:         159KB   |    142KB   |    136KB
Everything file:    195KB   |    159KB   |    160KB (despite adding the elements mentioned above compared to 1177)
```

## How to test the changes?

- New game. Only the default starters should be unlocked.
- Buy passives (can use override `FREE_CANDY_UPGRADE_OVERRIDE: true`), disable some of them
- Start a run and make sure the Pokemon that should have their passive do
- Go back to starter select and make sure that the passives you disabled are still disabled
- Override to some Pokemon species, catch them repeatedly and make sure the abilities show as new or not as they should
- You can also hatch eggs
- Check periodically that starter select has all the correct information from the mons you captured
- Menu > Manage Data > Import Data. Import the everything save file and play around as well

## Checklist


flx-sta (approved):

Tempo-anon (approved):
Maybe add a todo in the game data saving part for moving passive status from local storage? Though this seems like an edge case since unlocking then disabling is very rarely done

DayKev:
If you disable a starter's passive and then use one of the attr cycling button (tera, shiny, etc) it'll incorrectly make the passive text no longer grayed out. The passive is still disabled, and moving the cursor off the starter and then back will show the correct text state again.

DayKev:
> If you disable a starter's passive and then use one of the attr cycling button (tera, shiny, etc) it'll incorrectly make the passive text no longer grayed out. The passive is still disabled, and moving the cursor off the starter and then back will show the correct text state again.

Just checked and this doesn't happen in 1177.

MokaStitcher:
> If you disable a starter's passive and then use one of the attr cycling button (tera, shiny, etc) it'll incorrectly make the passive text no longer grayed out. The passive is still disabled, and moving the cursor off the starter and then back will show the correct text state again.

Thank you for the catch, will look into it. SetSpeciesDetails is so confusing and messy, I probably missed something up in there (this is why I want to rewrite starter select from scratch lol)

DayKev:
> Thank you for the catch, will look into it. SetSpeciesDetails is so confusing and messy, I probably missed something up in there (this is why I want to rewrite starter select from scratch lol)

Yeah, I've noticed that there is an extreme amount of nesting and other nonsense going on in UI files that can make things hard to follow.

MokaStitcher:
> Maybe add a todo in the game data saving part for moving passive status from local storage? Though this seems like an edge case since unlocking then disabling is very rarely done

This todo is not specific to passive info. Other starter preferences are much more important to sync than passives being disabled.
Made an issue instead, but this should only be done if we have a solid plan for optimizing storage imo #1205

flx-sta:
~~This sounds almost like what Copilot didn't comment, but "noted" as "low confidence"~~

image

MokaStitcher:
> This sounds almost like what Copilot didn't comment, but "noted" as "low confidence"
> 

Didn't even see you had run it
But no what it said wasn't the issue

DayKev:
Huh, interesting. It was right that there was an issue (and managed to identify the fix, use `this` instead of `options`), but it was wrong about what the issue was and what the outcome of the bug would be (maybe? I'm not sure what it means by "runtime error", does that mean "crash" or just "incorrect behavior" or...?).

DayKev (approved):

PR #1200, opened 2025-05-19
[Refactor] `AbAttr.getTriggerMessage` now uses `unknown[]` for additional args

innerthunder:

N/A

For parity with `AbAttr.apply`

- Changed `AbAttr.getTriggerMessage`'s signature to now use `...args: unknown[]` instead of `...args: any[]`. Subclasses' signatures have also been changed to follow suit.
- Added docs to `AbAttr.getTriggerMessage`

## Checklist

  - [n/a] Have I created new automated tests (`npm run test:create`) or updated existing tests related to the PR's changes?
  - [?] Have I made sure that any UI change works for both UI themes (dark and light)?


DayKev (dismissed):

src/data/abilities/ab-attrs/ab-attr.ts
```diff
@@ -39,7 +39,14 @@ export abstract class AbAttr {
     return false;
   }
 
-  getTriggerMessage(_pokemon: Pokemon, _abilityName: string, ..._args: any[]): string | null {
+  /**
+   * @param pokemon - The {@linkcode Pokemon} with this ability
+   * @param abilityName - The name of the {@linkcode Ability} with this attribute
```
MokaStitcher:
```suggestion
   * @param abilityName - The localized name of the {@linkcode Ability} with this attribute
```
(just so it's clear that there's no need to worry about localization at this point)

src/data/abilities/ab-attrs/ab-attr.ts
```diff
@@ -39,7 +39,14 @@ export abstract class AbAttr {
...
+   * @param args - Additional parameters for the trigger message. **NOTE**: This should be a subset
+   * of the additional parameters (i.e. `args`) given in {@linkcode apply}.
+   * @returns A message to play when the ability applies successfully, or `null` if no message should play.
+   */
+  getTriggerMessage(_pokemon: Pokemon, _abilityName: string, ..._args: unknown[]): string | null {
```
MokaStitcher:
```suggestion
  public getTriggerMessage(_pokemon: Pokemon, _abilityName: string, ..._args: unknown[]): string | null {
```
I think we decided somewhere (?) to always write out access modifiers, meaning all subclasses that override it should also add the `public`.
Don't feel like making all those suggestions so will trust you to use search/replace to do that :p

<sub>but I can't find that discussion anymore so I'm doubting my memory

MokaStitcher (dismissed):
just a nit and doubt about general code style

src/data/abilities/ab-attrs/ab-attr.ts
DayKev: Yeah. It wasn't reasonable to add a lint rule for it though because that would be like a billion errors in every file, so it's just been "add them as you add/edit methods".

MokaStitcher: Since this will affect manh files anyway it would be a good opportunity to do the same with the other methods of this class then

src/data/abilities/ab-attrs/ab-attr.ts
```diff
@@ -35,19 +55,37 @@ export abstract class AbAttr {
...
 
-  addCondition(condition: AbAttrCondition): AbAttr {
+  /**
+   * Adds a condition to this attribute. If the condition isn't met,
+   * the attribute will not apply its effects.
+   * @param condition - The {@linkcode AbAttrCondition | condition} to add
+   * @returns `this`
+   * @see {@linkcode Ability.conditionalAttr}
+   */
+  public addCondition(condition: AbAttrCondition): this {
```
DayKev: Hm, seems like this is more `setCondition` rather than `addCondition`.
Also `extraCondition` seems like it should just be named `condition`?

DayKev (dismissed):

MokaStitcher:
Added all the missing `public` in ab attr overrides. Feel free to rollback the commit if you think it's polluting that PR too much

Also renamed `addCondition` to `setCondition`.
Left `extraCondition` as is because I wasn't sure about it. I'm guessing the extra is meant to express that it's not necessarily the only condition for the attribute, since children classes can define their own separate conditions or something. not sure, don't feel strongly either way

<sub> edit: works better if i actually push the rename commit

DayKev (approved):

PR #1202, opened 2025-05-19
[Refactor] Clean up `RunHistoryUiHandler#processInput`

DayKev:
Success sound plays properly when pushing `Button.ACTION`.

Code cleanup.

Reorganized the code to be more readable.
Early return when handling `Button.ACTION` is replaced with setting the `success` variable.
Unused `error` variable removed.

## How to test the changes?
Press buttons in the Run History menu.

## Checklist


joao-krauchenco-mottu (approved):

src/ui/handlers/ui-handler.ts
```diff
@@ -154,6 +154,7 @@ export abstract class UiHandler {
 
   public abstract processInput(button: Button): boolean;
 
+  /** @deprecated */
   public getUi() {
```
MokaStitcher, line 158: Technically if we ever want to split the global scene we should keep using getUi instead of accessing the globalScene directly
But it's a long way away so :shrug:

DayKev: We would be able to use `const { ui } = this.scene;` in that case to achieve the same thing, which is why I added the `@deprecated` tag.

MokaStitcher: which `this?` ui handlers have no scene attribute
Or do you mean we would add that at this point

DayKev: Yeah we'd probably re-add it when adding more scenes. Unless we decide to keep all the handlers as singletons but that seems like it'd be harder to work with.

DayKev: I suppose we might also have it so `globalScene` always points to the active scene, and only ever have one active scene at a time. Though that still obsoletes the `getUi` function.

src/ui/handlers/run-history-ui-handler.ts
```diff
@@ -153,12 +154,9 @@ export class RunHistoryUiHandler extends MessageUiHandler {
       }
     }
 
-    if (success) {
-      ui.playSelect();
-    } else if (error) {
-      ui.playError();
-    }
-    return success || error;
+    success ? ui.playSelect() : ui.playError();
```
MokaStitcher: This is not correct, with this the game will play the error sound for any button press that is not expected by the UI, like left or right.
So even if it is a bit counterintuitive (and should probably be refactored to be made clearer) the way it was done before should stay in some capacity. We don't want to play a sound effect for every input, only for inputs that do something and succeed, and those that do something but fail for a reason or another (for example line 112).
So either we should bring back the error variable and set it to true line 112, or it's also fine to just set success to `false` in this case and never play the error sound effect, because arguably that error should in theory never happen

MokaStitcher (dismissed):
the sound effect behavior is incorrect, see comment

src/ui/handlers/ui-handler.ts
MokaStitcher: I suppose we'll cross that bridge when we get there :p 
The UI refactor is still not finished and that will likely impact that decision as well

src/ui/handlers/run-history-ui-handler.ts
```diff
@@ -96,36 +96,37 @@ export class RunHistoryUiHandler extends MessageUiHandler {
...
-            runEntry.isVictory,
-          );
-        } else {
-          return false;
-        }
-        success = true;
-        return success;
+
+    const handleAction = () => {
+      // TODO: should be `.getCursor()`?
```
MokaStitcher: I don't believe so, unless we override `getCursor()` to return `scrollCursor + cursor`

MokaStitcher (dismissed):
still not fully sure about deprecating `getUi` but otherwise lgtm

src/ui/handlers/run-history-ui-handler.ts
DayKev: Oh this comment should have been "... in place of .cursor". Though idk if that would actually change anything after taking a look now since it actually just directly returns `.cursor` anyway, so I'll just remove the comment.

MokaStitcher (approved):

f-fsantos (approved):

PR #775, opened 2025-03-02
[Refactor] Removed a vast majority of achievements + unused 'score' property

frutescens:
## What are the changes the user will see?

Money, item, and hatch-related achievements will no longer be accessible since they are now deleted. It also removes the 'score' property from achievements since that isn't used at the moment. Because vouchers no longer have a score property, they are now all common vouchers. 

## Why am I making these changes?

I want to move the unlocking process of achievements to their own scene as an early, first step into breaking the giant scene object of the game and adding event handling to the process. This just prunes the number of achievements that have to be refactored and updated in order to do so into a reasonable amount of achievements. 

## What are the changes from a developer perspective?

- Many achievements deleted.
- achv-category has been pruned. <-- note: will be changed into timing based categories later on
- 'score' property removed

## Screenshots/Videos

image

image

## How to test the changes?

View Menu > Achievements/Vouchers

## Checklist

- [x] Otherwise: **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?

- [x] Are all unit tests still passing? (`npm run test:silent`)
  - [x] Have I created new automated tests (`npm run test:create`) or updated existing tests related to the PR's changes?
- [x] Have I provided screenshots/videos of the changes (if applicable)?
  - [x] Have I made sure that any UI change works for both UI themes (dark and light)?

PR #1203, opened 2025-05-19
[UI][P2] Fix achievements screen scrolling behavior

### LINKED ISSUE #785: [Bug] Achievements UI breaks when scrolling up past the first row
### Describe the bug

On the Achievements UI, if the cursor is on the first row and the user presses Up, it makes the UI enter a glitchy state where it displays an incomplete list of achievements. This also causes a console error without crashing the game.

No bugs occur when scrolling Left or Right (the cursor simply wraps around as normal) or when scrolling Down (the cursor simply jumps to the final achievement). No bugs occur either when using the related Voucher Rewards UI.

### Reproduction

Title.

### Expected behavior

If the user presses Up while on the first row, then either nothing should happen, or the cursor should wrap around to the last row.

### Screenshots / Videos

https://github.com/user-attachments/assets/aa0b7197-006f-49a1-9281-aed0511ad9a0

### Session export file

_No response_

### User data export file

_No response_

### Additional context

Most likely caused by #775 causing the number of achievements to no longer fill up the 4 rows of achievements that the UI is designed to handle.

MokaStitcher:

Achievements UI no longer breaks when:
- Pressing up from the top of the achievement page
- Exiting after scrolling down in voucher page then coming back to the screen
Achievements UI no longer has unused window

Fixes #785 and more
- The scrolling bugs were always here, but unseen until #775, when the number of achievements was reduced
- In some other achievement related PR the unused score attribute was removed from the achievements class, but the window that was showing the score didn't get removed.

- Remove the scoreContainer that is no longer needed, and the code that resized the title window to make room for it in the achievement screen
- The handler now uses the existing ScrollableGridHandler to take care of input processing and navigating the grid

## Screenshots/Videos

**Before:**
scrolling up:
cheevos_before
exiting from bottom of voucher screen and coming back blocks you until the cancel button is pressed again:
image

**After:** no scrolling issue and no more empty window
image

## How to test the changes?

- Menu > Achievements
- Move the cursor around the Achievements page, wrapping around, etc.
- Switch to the Vouchers page, move around, scroll down
- Switch back and forth between the pages after scrolling down / moving the cursor to various values and make sure the cursors and cheevo/voucher details update properly when doing so
- Same thing but exit the screen and re open it. It should always go back to the achievements menu when opening

## Checklist


src/ui/handlers/achievements-ui-handler.ts
```diff
@@ -11,14 +11,17 @@ import type { Voucher } from "#system/voucher";
...
+import { ScrollableGridController } from "#ui/scrollable-grid-controller";
 import { addTextObject } from "#ui/text-utils";
 import { addWindow } from "#ui/ui-theme";
 import i18next from "i18next";
 
-enum Page {
-  ACHIEVEMENTS,
-  VOUCHERS,
-}
+const Page = {
```
Tempo-anon, line 19: Totally out of scope for this PR, but should an achievements-ui-handler also be handling things like vouchers? Should something like that deserve its own class/handler? If not and achievements-ui-handler will handle more things in the future maybe it can be renamed to be something more general?

src/ui/handlers/achievements-ui-handler.ts
```diff
@@ -35,9 +38,7 @@ export class AchievementsUiHandler extends MessageUiHandler {
   private iconsBg: Phaser.GameObjects.NineSlice;
   private icons: Phaser.GameObjects.Sprite[];
 
-  private titleBg: Phaser.GameObjects.NineSlice;
   private titleText: Phaser.GameObjects.Text;
-  private scoreContainer: Phaser.GameObjects.Container;
```
Tempo-anon, line 40: Are `titleBg` and `scoreContainer` no longer needed?

Tempo-anon (dismissed):

src/ui/handlers/achievements-ui-handler.ts
MokaStitcher: See pr description
If we ever add a score back we can adapt the UI again

src/ui/handlers/achievements-ui-handler.ts
MokaStitcher: Feel free to suggest a better name
These were separate before but got merged to free up room in the main menu

flx-sta (dismissed):

src/ui/handlers/achievements-ui-handler.ts
```diff
@@ -11,14 +11,17 @@ import type { Voucher } from "#system/voucher";
...
 import { addWindow } from "#ui/ui-theme";
 import i18next from "i18next";
 
-enum Page {
-  ACHIEVEMENTS,
-  VOUCHERS,
-}
+const Page = {
+  ACHIEVEMENTS: 0,
+  VOUCHERS: 1,
```
DayKev: I'd like all the enums to start at `1` (unless there is an explicit "none" value where `enum.none` is equivalent to it being `nil`).
```suggestion
  ACHIEVEMENTS: 1,
  VOUCHERS: 2,
```

DayKev (dismissed):

src/ui/handlers/achievements-ui-handler.ts
DayKev: Having them together like this makes sense, they're pretty similar, and it seems like there's a lot of reusable code.

DayKev: Not sure what a good name for the class/file is though.

MokaStitcher: > Having them together like this makes sense, they're pretty similar, and it seems like there's a lot of reusable code.

agreed. It could probably be refactored even further for even less duplicate code too

I think achievements make sense as a global term, the vouchers are just rewards earned from some of those achievements like clearing classic, (+ boss fights that would represent too many achievements to have in the achievements list). Maybe the "vouchers" tab could be renamed though, but not sure to what either

PR #1206, opened 2025-05-21
[Test] Make `species` param of `startBattle` required

DayKev:
N/A

More reliable tests (no unexpected random species spawns breaking a test).

- The `species` param of `ClassicModeHelper#runToSummon`, `#startBattle`, `ChallengeModeHelper#runToSummon` and `#startBattle` is now required.
- The deprecated `GameManager#startBattle` function is removed.

## How to test the changes?
`npm run test:silent`

## Checklist


test/abilities/arena-trap.test.ts
```diff
@@ -29,25 +29,10 @@ describe("Abilities - Arena Trap", () => {
       .enemyMoveset(MoveId.TELEPORT);
   });
 
-  // TODO: Enable test when Issue #935 is addressed
```
Tempo-anon, line 32: The TODO is outdated but is the test still warranted?

test/abilities/disguise.test.ts
```diff
@@ -30,12 +30,11 @@ describe("Abilities - Disguise", () => {
       .battleType("single")
       .enemySpecies(SpeciesId.MIMIKYU)
       .enemyMoveset(MoveId.SPLASH)
-      .starterSpecies(SpeciesId.REGIELEKI)
       .moveset([MoveId.SHADOW_SNEAK, MoveId.VACUUM_WAVE, MoveId.TOXIC_THREAD, MoveId.SPLASH]);
   });
 
   it("takes no damage from attacking move and transforms to Busted form, takes 1/8 max HP damage from the disguise breaking", async () => {
-    await game.classicMode.startBattle();
+    await game.classicMode.startBattle([SpeciesId.REGIELEKI]);
```
Tempo-anon: I'm noticing a few Pokemon showing up commonly for tests, primarily REGIELEKI (for fast Pokemon), SHUCKLE (for slow Pokemon), and FEEBAS (as a generic pokemon). Wonder if we should have consts for these in the tests so someone not as familiar with specific Pokemon will find it easier to use tests since otherwise a lot of tests use Pokemon that are just on the test writer's mind at the time

Tempo-anon (dismissed):

test/abilities/arena-trap.test.ts
DayKev: No, the test was checking for incorrect behavior (the original issue didn't realize Gamefreak changed the move/ability interaction in LGPE+).

test/abilities/disguise.test.ts
DayKev: Possibly.
```ts
export const FAST_POKEMON = SpeciesId.NINJASK;
export const SLOW_POKEMON = SpeciesId.SHUCKLE;
export const GENERIC_POKEMON = SpeciesId.MAGIKARP;
```
I wonder if there's other good/relevant "archetypes" that should be added.

innerthunder: Maybe some Pokemon with high physical/special defense to reduce faint risk in some cases? I tend to use Aggron for high Defense / Blissey for high Sp. Def fairly often.

Tempo-anon: I like those ideas though there are some cases where comments should be noted if there are any special requirements i.e. using Ninjask as a fast Pokemon that's not electric type. I used to use Regirock as a high defense mon for testing but sand stream passive has messed me up before :plains:

test/test-utils/game-manager.ts
```diff
@@ -280,13 +281,12 @@ export class GameManager {
...
   /**
-   * @deprecated Use `game.classicMode.startBattle()` or `game.dailyMode.startBattle()` instead
-   *
    * Transitions to the start of a battle.
    * @param species - Optional array of species to start the battle with.
    * @returns A promise that resolves when the battle is started.
+   * @deprecated Use `game.[mode].startBattle()` instead
    */
-  async startBattle(species?: SpeciesId[]) {
+  async startBattle(species: SpeciesId[]) {
```
innerthunder: Thoughts on making this a spread argument (i.e. using `...species` instead)?

innerthunder (dismissed):

test/abilities/disguise.test.ts
DayKev: Yeah maybe we should avoid using Legendary/etc Pokémon (that is, any Pokémon that would automatically be considered a boss Pokémon and have their passive enabled when used as an enemy). I guess Ninjask is the fastest non-boss Pokémon then.

test/test-utils/game-manager.ts
DayKev: Ah, good idea.

DayKev: Should I do it in this PR or make a follow-up?

innerthunder: > Should I do it in this PR or make a follow-up?

I think it's fine to add to this PR.

I guess this exposes another concern: what should happen if an empty array is passed to this function? Should it be allowed at all?

DayKev: Well the whole point of the PR is to prevent that. I could do this:
```ts
startBattle(species: SpeciesId, ...extraSpecies: SpeciesId[]) {
  const startingSpecies = [species, ...extraSpecies];
```

innerthunder: That could work. It's probably better than the current version, even, because as far as I can tell this is still valid atm:
```ts
await game.classicMode.startBattle([]);
```

innerthunder (dismissed):

test/test-utils/game-manager.ts
```diff
@@ -240,12 +239,13 @@ export class GameManager {
    * @param encounterType if specified, will expect encounter to have been spawned
    * @param species Optional array of species for party.
```
MokaStitcher:
```suggestion
   * @param species Optional array of species for party. If not provided, a party of Feebas, Magikarp and Feebas is used.
```

test/abilities/disguise.test.ts
MokaStitcher: not sure where we would put these consts. An alternative could be to edit the documentation for `startBattle` to add those suggested starters, as well as a warning about legendaries and passives. Although that's an issue for enemy Pokemon, not player Pokemon, so not relevant to startBattle
It could also be put as a comment in the test boilerplate

test/abilities/magic-guard.test.ts
```diff
@@ -296,7 +296,7 @@ describe("Abilities - Magic Guard", () => {
 
   /*
   it("Magic Guard does not prevent self-damage from confusion", async () => {
```
MokaStitcher: why is this commented out? Should it be marked as `it.todo` instead?

test/abilities/magic-bounce.test.ts
```diff
@@ -197,7 +197,7 @@ describe("Abilities - Magic Bounce", () => {
 
   it("should not bounce back curse", async () => {
     game.override.starterSpecies(SpeciesId.GASTLY);
```
MokaStitcher: that's not needed, right?
```suggestion
```

test/battle/battle.test.ts
```diff
@@ -182,7 +182,7 @@ describe("Test Battle Phase", () => {
     game.override.moveset([MoveId.TACKLE]);
     game.override.enemyAbility(AbilityId.HYDRATION);
     game.override.enemyMoveset([MoveId.TACKLE, MoveId.TACKLE, MoveId.TACKLE, MoveId.TACKLE]);
-    await game.classicMode.startBattle();
+    await game.classicMode.startBattle(SpeciesId.FEEBAS);
```
MokaStitcher:
```suggestion
    await game.classicMode.startBattle(SpeciesId.MEWTWO);
```
starterSpecies is overriden to Mewtwo line 178, so that line should be removed

We may want to take a closer look at the starter species override VS this `generateStarter` helper. If there are meaningful differences we should note them and make them clear in the documentation. If we always want to use startBattle rather than the override, maybe `startBattle` should fail or at least print a warning if there is a species override?

test/battle/battle.test.ts
```diff
@@ -196,7 +196,7 @@ describe("Test Battle Phase", () => {
     game.override.enemyAbility(AbilityId.HYDRATION);
     game.override.enemyMoveset([MoveId.TAIL_WHIP, MoveId.TAIL_WHIP, MoveId.TAIL_WHIP, MoveId.TAIL_WHIP]);
     game.override.battleType("single");
-    await game.classicMode.startBattle();
+    await game.classicMode.startBattle(SpeciesId.FEEBAS);
```
MokaStitcher:
```suggestion
    await game.classicMode.startBattle(SpeciesId.MEWTWO);
```
same

test/battle/battle.test.ts
```diff
@@ -297,7 +297,7 @@ describe("Test Battle Phase", () => {
     game.override.startingWave(3);
     game.override.moveset([moveToUse]);
     game.override.enemyMoveset([MoveId.TACKLE, MoveId.TACKLE, MoveId.TACKLE, MoveId.TACKLE]);
-    await game.classicMode.startBattle();
+    await game.classicMode.startBattle(SpeciesId.FEEBAS);
```
MokaStitcher:
```suggestion
    await game.classicMode.startBattle(SpeciesId.MEWTWO);
```
another

test/battle/battle.test.ts
```diff
@@ -277,7 +277,7 @@ describe("Test Battle Phase", () => {
     game.override.startingWave(3);
     game.override.moveset([moveToUse]);
     game.override.enemyMoveset([MoveId.TACKLE, MoveId.TACKLE, MoveId.TACKLE, MoveId.TACKLE]);
-    await game.classicMode.startBattle([SpeciesId.DARMANITAN, SpeciesId.CHARIZARD]);
+    await game.classicMode.startBattle(SpeciesId.DARMANITAN, SpeciesId.CHARIZARD);
```
MokaStitcher:
```suggestion
    await game.classicMode.startBattle(SpeciesId.MEWTWO, SpeciesId.CHARIZARD);
```
again (since the starter species override only applies to the first mon, this ensure the same behavior as before)
TBH the species here probably don't matter but I'm worried this mix and match of ways to override the starter species could cause issues

test/moves/belly-drum.test.ts
```diff
@@ -41,7 +41,7 @@ describe("Moves - BELLY DRUM", () => {
   // Bulbapedia Reference: https://bulbapedia.bulbagarden.net/wiki/Belly_Drum_(move)
 
   test("raises the user's ATK stat stage to its max, at the cost of 1/2 of its maximum HP", async () => {
-    await game.startBattle([SpeciesId.MAGIKARP]);
+    await game.classicMode.startBattle(SpeciesId.MAGIKARP);
```
MokaStitcher, line 43: the starter species override in `beforeEach` can be removed

test/moves/clangorous-soul.test.ts
```diff
@@ -37,7 +37,7 @@ describe("Moves - Clangorous Soul", () => {
   //Bulbapedia Reference: https://bulbapedia.bulbagarden.net/wiki/Clangorous_Soul_(move)
 
   it("raises the user's ATK, DEF, SPATK, SPDEF, and SPD stat stages by 1 each at the cost of 1/3 of its maximum HP", async () => {
-    await game.startBattle([SpeciesId.MAGIKARP]);
+    await game.classicMode.startBattle(SpeciesId.MAGIKARP);
```
MokaStitcher, line 39: the starter species override in `beforeEach` can be removed

test/moves/fillet-away.test.ts
```diff
@@ -38,7 +38,7 @@ describe("Moves - FILLET AWAY", () => {
   //Bulbapedia Reference: https://bulbapedia.bulbagarden.net/wiki/fillet_away_(move)
 
   test("raises the user's ATK, SPATK, and SPD stat stages by 2 each, at the cost of 1/2 of its maximum HP", async () => {
-    await game.startBattle([SpeciesId.MAGIKARP]);
+    await game.classicMode.startBattle(SpeciesId.MAGIKARP);
```
MokaStitcher, line 40: the starter species override in `beforeEach` can be removed

test/moves/follow-me.test.ts
```diff
@@ -34,7 +34,7 @@ describe("Moves - Follow Me", () => {
   });
 
   test("move should redirect enemy attacks to the user", async () => {
-    await game.classicMode.startBattle([SpeciesId.AMOONGUSS, SpeciesId.CHARIZARD]);
+    await game.classicMode.startBattle(SpeciesId.AMOONGUSS, SpeciesId.CHARIZARD);
```
MokaStitcher: the starter species override in `beforeEach` can be removed

test/moves/freeze-dry.test.ts
```diff
@@ -83,7 +82,7 @@ describe("Moves - Freeze-Dry", () => {
       .enemyMoveset(MoveId.SPLASH)
       .starterSpecies(SpeciesId.MAGIKARP)
```
MokaStitcher:
```suggestion
```

test/moves/freeze-dry.test.ts
```diff
@@ -83,7 +82,7 @@ describe("Moves - Freeze-Dry", () => {
...
       .moveset([MoveId.SOAK, MoveId.FREEZE_DRY]);
-    await game.classicMode.startBattle();
+    await game.classicMode.startBattle(SpeciesId.FEEBAS);
```
MokaStitcher:
```suggestion
    await game.classicMode.startBattle(SpeciesId.MAGIKARP);
```

test/moves/gigaton-hammer.test.ts
```diff
@@ -33,7 +33,7 @@ describe("Moves - Gigaton Hammer", () => {
   });
 
   it("can't be used two turns in a row", async () => {
-    await game.classicMode.startBattle();
+    await game.classicMode.startBattle(SpeciesId.FEEBAS);
```
MokaStitcher, line 35: the starter species override in beforeEach can be removed

test/moves/magic-coat.test.ts
```diff
@@ -176,7 +176,7 @@ describe("Moves - Magic Coat", () => {
 
   it("should not bounce back curse", async () => {
     game.override.starterSpecies(SpeciesId.GASTLY);
```
MokaStitcher:
```suggestion
```

test/moves/spotlight.test.ts
```diff
@@ -31,7 +31,7 @@ describe("Moves - Spotlight", () => {
   });
 
   test("move should redirect attacks to the target", async () => {
-    await game.classicMode.startBattle([SpeciesId.AMOONGUSS, SpeciesId.CHARIZARD]);
+    await game.classicMode.startBattle(SpeciesId.AMOONGUSS, SpeciesId.CHARIZARD);
```
MokaStitcher: the starter species override in beforeEach can be removed

MokaStitcher (commented):
I'm wondering about the relevance of keeping the starter species override in this context.
Two issues:
- We can set a species in startBattle and miss that there is a starter override that will force the use of another species
- the starter species override only applies to the first Pokemon, adding to the confusion if both are used together

I understand the need of startBattle since the override doesn't allow for multiple species, but I'm not fond of how this separates the species information from the other overrides, making it harder to see at a glance all parameters of a test. (this is not new, but now that starterSpecies override is almost fully deprecated, it will be the case for every test). I think ideally the starter species override would be updated to allow multiple species, rather than relying on a separate helper function
(I'm also not 100% sure the species override and startBattle have the exact same result)

I see a few options, not sure which one is best or if there's another better one
- make the starterSpecies override helper for tests deprecated, or remove it
- update the starter species override to allow multiple species
- re-allow to give no species in startBattle, making it use the override if it exists, and fail otherwise
- print a warning or fail the test if startBattle is used in conjunction with the starterSpecies override

test/abilities/disguise.test.ts
Tempo-anon: I'm okay with moving defining test constants into a separate PR but here's what I was imagining a constants file like

```ts
/**
 * Ninjask is chosen because it is the fastest non legendary
 */
export const FAST_POKEMON = SpeciesId.NINJASK;
/**
 * Shuckle is chosen as it has the lowest speed and also high defenses
 */
export const SLOW_POKEMON = SpeciesId.SHUCKLE;
/**
 * Explain reason why we use magikarp as generic pokemon
 */
export const GENERIC_POKEMON = SpeciesId.MAGIKARP;
/**
 * Aggron is chosen because it has the highest defense for a non mega/legendary
 */
export const HIGH_DEF_POKEMON = SpeciesId.AGGRON;
```

flx-sta (dismissed):

test/abilities/magic-guard.test.ts
flx-sta: https://github.com/pagefaultgames/pokerogue/pull/3253

It's been like this forever apparently

DayKev:
> I'm wondering about the relevance of keeping the starter species override in this context. Two issues:
> 
> * We can set a species in startBattle and miss that there is a starter override that will force the use of another species
> * the starter species override only applies to the first Pokemon, adding to the confusion if both are used together
> 
> I understand the need of startBattle since the override doesn't allow for multiple species, but I'm not fond of how this separates the species information from the other overrides, making it harder to see at a glance all parameters of a test. (this is not new, but now that starterSpecies override is almost fully deprecated, it will be the case for every test). I think ideally the starter species override would be updated to allow multiple species, rather than relying on a separate helper function (I'm also not 100% sure the species override and startBattle have the exact same result)
> 
> I see a few options, not sure which one is best or if there's another better one
> 
> * make the starterSpecies override helper for tests deprecated, or remove it
> * update the starter species override to allow multiple species
> * re-allow to give no species in startBattle, making it use the override if it exists, and fail otherwise
> * print a warning or fail the test if startBattle is used in conjunction with the starterSpecies override

I'll make it prevent tests from running if the species override is used for non-daily mode tests.
The species override could do with being updated, but that's for a different PR.

test/abilities/disguise.test.ts
flx-sta: If you are going to introduce those make sure to update the create test template file

test/test-utils/helpers/challenge-mode-helper.ts
```diff
@@ -32,10 +32,16 @@ export class ChallengeModeHelper extends GameManagerHelper {
...
   /**
    * Runs the Challenge game to the summon phase.
-   * @param gameMode - Optional game mode to set.
+   * @param species - The species to start the battle with. At least one must be specified, up to 6 max.
    * @returns A promise that resolves when the summon phase is reached.
    */
-  async runToSummon(species?: SpeciesId[]) {
+  async runToSummon(species: SpeciesId, ...extraSpecies: SpeciesId[]): Promise<void> {
+    if (overrides.STARTER_SPECIES_OVERRIDE) {
+      throw new Error(
```
MokaStitcher, line 40: wondering if throwing an error or using `expect(overrides.STARTER_SPECIES_OVERRIDE)` to not be 0 would provide easier to parse error 🤔

MokaStitcher (dismissed):

test/test-utils/helpers/challenge-mode-helper.ts
DayKev: Ah, using an `expect` might be better. I'll take a look.

flx-sta (dismissed):

test/test-utils/helpers/challenge-mode-helper.ts
DayKev: There's no real difference since I'd have to add a custom error message to the `expect` anyway.
image
image

PR #1209, opened 2025-05-21
[P2][Checkup][Ability] Bad Dreams 💤

### LINKED ISSUE #437: [Ability][Checkup] Bad Dreams
## Ability Requirements
- [x] Documentation
- [x] Tests
- [x] At the end of each turn, each adjacent sleeping opponent takes damage equal to 1/8 of its maximum HP.
- [x] Bad Dreams inflicts damage after a Pokémon that is drowsy due to Yawn falls asleep (and thus inflicts damage to a Pokémon the same turn it falls asleep due to drowsiness). 
- [x] Bad Dreams applies after a Pokémon has the opportunity to be woken up by Hydration, Shed Skin, or Healer. 

## References
- [Bulbapedia - Bad Dreams](https://bulbapedia.bulbagarden.net/wiki/Bad_Dreams_(Ability))

flx-sta:

- **[P2]** The `Bad Dreams` ability will no longer damage Pokemon that will wake up in the same turn e.g. by `Hydration`, `Shed Skin` or `Healer` 
- **[P2]** Bad Dreams will now hurt a Pokemon on the first sleep turn after "Drowsy" (from Yawn)

Closes #437 

- Bad Dreams test coverage
- Change `applyAbAttrs` return value
- Use new return value from `applyAbAttrs` to evaluate if a Pokemon will wake up in the same turn

## Screenshots/Videos

n/a

## How to test the changes?

`pnpm test:silent`

## Checklist


src/data/abilities/ab-attrs/post-turn-hurt-if-sleeping-ab-attr.ts
```diff
@@ -1,4 +1,7 @@
+import { applyAbAttrs } from "#abilities/apply-ab-attrs";
 import { PostTurnAbAttr } from "#abilities/post-turn-ab-attr";
+import { PostTurnResetStatusAbAttr } from "#abilities/post-turn-reset-status-ab-attr";
+import { PostTurnStatusHealAbAttr } from "#abilities/post-turn-status-heal-ab-attr";
```
Tempo-anon: PostTurnStatusHealAbAttr is only on Poison Heal to heal the user for 1/8 of their HP if they are poisoned/toxic'd. I don't think this is what you want

src/data/abilities/ab-attrs/post-turn-hurt-if-sleeping-ab-attr.ts
```diff
@@ -1,4 +1,7 @@
+import { applyAbAttrs } from "#abilities/apply-ab-attrs";
 import { PostTurnAbAttr } from "#abilities/post-turn-ab-attr";
+import { PostTurnResetStatusAbAttr } from "#abilities/post-turn-reset-status-ab-attr";
```
Tempo-anon: PostTurnResetStatusAbAttr is for healer, shed skin, and hydration

src/data/abilities/ab-attrs/post-turn-hurt-if-sleeping-ab-attr.ts
```diff
@@ -36,3 +40,30 @@ export class PostTurnHurtIfSleepingAbAttr extends PostTurnAbAttr {
...
+//#region Helpers
+
+/**
+ * Check if a Pokemon will wake up this turn by simulating the PostTurnAbAttrs applications.
+ * @param pokemon - The Pokemon to check
+ * @returns `true` if the Pokemon will wake up this turn, `false` otherwise
+ */
+function willWakeUpThisTurn(pokemon: Pokemon) {
+  // Will wake up from Hydration ability + Rain
+  const results = applyAbAttrs<PostTurnStatusHealAbAttr>(AbAttrFlag.POST_TURN, pokemon, true);
```
Tempo-anon: You probably want `PostTurnResetStatusAbAttr` which clears a status rather than `PostTurnStatusHealAbAttr` which heals HP if status'd

src/data/abilities/ab-attrs/post-turn-hurt-if-sleeping-ab-attr.ts
```diff
@@ -21,6 +24,7 @@ export class PostTurnHurtIfSleepingAbAttr extends PostTurnAbAttr {
         opp.hasStatusEffect(StatusEffect.SLEEP)
         && !opp.hasAbilityWithAttr(AbAttrFlag.BLOCK_NON_DIRECT_DAMAGE)
         && !opp.switchOutStatus
+        && !willWakeUpThisTurn(opp)
```
Tempo-anon: Does this retain the rng value for things like healer/shed skin? What about if rain is removed after the check but before the application?

Tempo-anon (dismissed):

src/data/abilities/ab-attrs/post-turn-hurt-if-sleeping-ab-attr.ts
flx-sta: For Healer & Shed Skin I can guarantee that the rng is is retained, as they didn't trigger properly before I added the spy for `enemy.randSeedInt`

flx-sta: Added coverage for all 3 abilities NOT triggering

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -5,24 +5,48 @@ import { AbilityApplyMode } from "#enums/ability-apply-mode";
...
+
+//#endregion
 //#region Exports
 
-export function applyAbAttrs<TAttr extends AbAttr = never>(
+/**
+ * The function for using the {@linkcode AbilityApplyMode.DEFAULT | DEFAULT} ability mode
+ * @param abAttrFlag - The {@linkcode AbAttrFlag} to apply
+ * @param params - The parameters for the given attribute's `apply` function
+ * @returns An array of {@linkcode AppliedAbAttr | applied ability attributes}
```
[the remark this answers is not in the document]
flx-sta: no

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -5,24 +5,48 @@ import { AbilityApplyMode } from "#enums/ability-apply-mode";
...
+): AppliedAbAttr<TAttr>[] {
   return applyAbAttrsInternal<TAttr>({ canApplyOnly: true }, abAttrFlag, ...params);
 }
 
 /**
  * Obtains the function to apply abilities corresponding to the given mode
  * @param mode - The {@linkcode AbilityApplyMode} determining how abilities are applied
  * @returns The function to apply abilities based on the mode:
- * - {@linkcode AbilityApplyMode.DEFAULT} - Applies abilities without restriction
+ * - {@linkcode AbilityApplyMode.DEFAULT DEFAULT} - Applies abilities without restriction
```
DayKev:
```suggestion
 * - {@linkcode AbilityApplyMode.DEFAULT | DEFAULT} - Applies abilities without restriction
```

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -5,24 +5,48 @@ import { AbilityApplyMode } from "#enums/ability-apply-mode";
...
  *     (as long as they meet conditions to apply).
- * - {@linkcode AbilityApplyMode.REVEALED} - Only applies abilities that have
+ * - {@linkcode AbilityApplyMode.REVEALED REVEALED} - Only applies abilities that have
```
DayKev:
```suggestion
 * - {@linkcode AbilityApplyMode.REVEALED | REVEALED} - Only applies abilities that have
```

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -5,24 +5,48 @@ import { AbilityApplyMode } from "#enums/ability-apply-mode";
...
  *     previously applied in the current battle.
- * - {@linkcode AbilityApplyMode.IGNORE} - Does nothing and returns an empty array.
+ * - {@linkcode AbilityApplyMode.IGNORE IGNORE} - Does nothing and returns an empty array.
```
DayKev:
```suggestion
 * - {@linkcode AbilityApplyMode.IGNORE | IGNORE} - Does nothing and returns an empty array.
```

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -5,24 +5,48 @@ import { AbilityApplyMode } from "#enums/ability-apply-mode";
...
+  /** The (simulated) result of applying the attribute */
+  result: boolean;
+  /** The (optional) message to show if the attribute was applied */
+  message: string | null;
+}
+
+//#endregion
 //#region Exports
 
-export function applyAbAttrs<TAttr extends AbAttr = never>(
```
DayKev, line 10: Why did you remove the `= never`? That was what enforced usage of the typing.

test/@types/vitest.d.ts
```diff
@@ -40,14 +41,17 @@ declare module "vitest" {
...
-     * @param options - The {@linkcode ToHaveUsedMoveMatcherOptions} (optional)
+     * @param options - (Optional) The {@linkcode ToHaveUsedMoveMatcherOptions}
      * @see {@linkcode Pokemon.getLastXMoves}
      */
     toHaveUsedMove(expected: MoveId, options?: ToHaveUsedMoveMatcherOptions): void;
 
     /**
      * Matcher to check if a {@linkcode Pokemon} stat is as expected
      *
+     * _Stat means the stat before modifiers are applied._
```
DayKev, line 52:
```suggestion
     * _This checks the stat before modifiers are applied._
```

test/@types/vitest.d.ts
```diff
@@ -57,24 +61,26 @@ declare module "vitest" {
     /**
      * Matcher to check if a {@linkcode Pokemon Pokemon's} effective stat is as expected
      *
-     * *Effective stat means the stat after all the stat modifiers are applied.*
+     * _Effective stat means the stat after all the stat modifiers are applied._
```
DayKev, line 64:
```suggestion
     * _This checks the stat after all the stat modifiers are applied._
```

test/test-utils/matchers/to-have-taken-damage-matcher.ts
```diff
@@ -1,7 +1,17 @@
 import { getPokemonNameWithAffix } from "#app/messages";
 import { isPokemonInstance, receivedStr } from "#test/test-utils/test-utils";
+import { toDmgValue } from "#utils/common-utils";
 import type { MatcherState, SyncExpectationResult } from "@vitest/expect";
 
+//#region Types
+
+export interface ToHaveTakenDamageMatcherOptions {
+  /** Whether to skip the internal {@linkcode toDmgValue} call. @defaultValue false */
+  skipToDmgValue?: boolean;
```
DayKev, line 10: Is there a scenario where we want this? All damage values are supposed to have `toDmgValue` applied to them, it's a bug if they don't.

flx-sta: I don’t think so, no. I just thought to give ultimate flexibility

DayKev: I guess it doesn't really matter.

test/@types/vitest.d.ts
```diff
@@ -57,24 +61,26 @@ declare module "vitest" {
...
+     * _If you want to check the stat **before** modifiers are applied, use {@linkcode toHaveStat}._
      *
      * @param stat - The {@linkcode EffectiveStat} to check
      * @param expectedValue - The expected value of the {@linkcode stat}
-     * @param options - The {@linkcode ToHaveStatMatcherOptions} (optional)
+     * @param options - (Optional) The {@linkcode ToHaveStatMatcherOptions}
```
PigeonBar:
```suggestion
     * @param options - (Optional) The {@linkcode ToHaveEffectiveStatMatcherOptions}
```

test/abilities/bad-dreams.test.ts
```diff
@@ -0,0 +1,146 @@
...
+    move.use(MoveId.YAWN);
+    await game.toEndOfTurn();
+
+    expect(enemy).toHaveBattlerTagType(BattlerTagType.DROWSY);
+    expect(enemy).toHaveFullHp();
+
+    for (let i = 0; i < 2; i++) {
+      move.use(MoveId.SPLASH);
+      await game.toEndOfTurn();
+    }
```
PigeonBar: Since the player already spent 1 turn using Yawn, it should only take 1 more turn for the opponent to fall asleep.

In fact, the behaviour being tested is currently bugged, and a drowsy Pokemon does not take damage on the turn that it falls asleep. I think the main reason is because DrowsyTag sets the Sleep status using `asPhase = true`, which causes the Pokemon to fall asleep during a separate phase, after Bad Dreams has attempted to activate. (The `asPhase = true` is required for the game to display "enemy Pokemon fell asleep!".) So, this may need to be a `.edgeCase()` for now.

https://github.com/Despair-Games/poketernity/blob/e9ed9dc41b75f3d4dc8db8fed24d8d88460054b2/src/data/battler-tags/drowsy-tag.ts#L36

src/data/abilities/ab-attrs/post-turn-hurt-if-sleeping-ab-attr.ts
```diff
@@ -36,3 +39,20 @@ export class PostTurnHurtIfSleepingAbAttr extends PostTurnAbAttr {
...
+//#region Helpers
+
+/**
+ * Check if a Pokemon will wake up this turn by simulating the {@linkcode AbAttrFlag.POST_TURN} applications
+ * and checking for a {@linkcode PostTurnResetStatusAbAttr} with a result of `true`.
+ *
+ * @param pokemon - The Pokemon to check
+ * @returns `true` if the Pokemon will wake up this turn, `false` otherwise
+ */
+function willWakeUpAtEndOfTurn(pokemon: Pokemon): boolean {
```
PigeonBar: This function sometimes does not work because Shed Skin and Healer activate randomly, so this function can receive a different RNG roll than the RNG roll that actually determines whether the Pokemon wakes up. 

I think a potential alternate solution would be to make this attr no longer extend `PostTurnAbAttr`, and instead make it apply separately after all `PostTurnAbAttr`s have been applied, so that Bad Dreams applies after Hydration, Healer, and Shed Skin have all had a chance to apply.

test/abilities/bad-dreams.test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  it.each([
+    ["Hydration", AbilityId.HYDRATION, WeatherType.RAIN],
+    ["Shed Skin", AbilityId.SHED_SKIN, WeatherType.SANDSTORM],
+    ["Healer", AbilityId.HEALER, WeatherType.NONE],
+  ])(
+    "should not damage if enemy is woken up by '%s' ability in the same turn",
+    async (_abilityName, abilityId, weatherType) => {
+      const { override, classicMode, field, move } = game;
+      override.weather(weatherType).enemyAbility(abilityId).enemyPassiveAbility(AbilityId.OVERCOAT); // overcoat to prevent sandstorm damage
```
PigeonBar:
```suggestion
      override.weather(weatherType)
        .battleType("double")
        .enemyAbility(abilityId)
        .enemyPassiveAbility(AbilityId.OVERCOAT); // overcoat to prevent sandstorm damage
```
Healer activates on the Pokemon's ally, not on the Pokemon itself, so Healer requires a double battle to activate.

test/abilities/bad-dreams.test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+      await classicMode.runToSummon(SpeciesId.DARKRAI);
+      const enemy = field.getEnemyPokemon();
+      vi.spyOn(enemy, "randSeedInt").mockReturnValueOnce(0); // Make sure that Shed Skin/Healer always triggers.
+      enemy.trySetStatus(StatusEffect.SLEEP);
```
PigeonBar:
```suggestion
      enemy.trySetStatus(StatusEffect.SLEEP);
      for (const pokemon of game.scene.getEnemyParty()) {
        vi.spyOn(pokemon, "randSeedInt").mockReturnValueOnce(0); // Make sure that Shed Skin/Healer always triggers.
      }
```
The RNG function needs to be mocked after setting the status; otherwise, it sets the status to have a turns remaining count of 0.

test/abilities/bad-dreams.test.ts
```diff
@@ -0,0 +1,146 @@
...
+  ])(
+    "should damage if enemy is not woken up by '%s' ability in the same turn",
+    async (_abilityName, abilityId, weatherType) => {
+      const { override, classicMode, field, move } = game;
+      override.weather(weatherType).enemyAbility(abilityId).enemyPassiveAbility(AbilityId.OVERCOAT); // overcoat to prevent sandstorm damage
+
+      await classicMode.runToSummon(SpeciesId.DARKRAI);
+      const enemy = field.getEnemyPokemon();
+      vi.spyOn(enemy, "randSeedInt").mockReturnValueOnce(3); // Make sure that Shed Skin/Healer never triggers.
+      enemy.trySetStatus(StatusEffect.SLEEP, false, null, Number.MAX_SAFE_INTEGER);
```
PigeonBar:
```suggestion
      enemy.trySetStatus(StatusEffect.SLEEP, false, null, Number.MAX_SAFE_INTEGER);
      for (const pokemon of game.scene.getEnemyParty()) {
        vi.spyOn(pokemon, "randSeedInt").mockReturnValueOnce(3); // Make sure that Shed Skin/Healer never triggers.
      }
```

test/abilities/bad-dreams.test.ts
```diff
@@ -0,0 +1,146 @@
...
+  });
+
+  it.each([
+    ["Shed Skin", AbilityId.SHED_SKIN, WeatherType.SANDSTORM],
+    ["Healer", AbilityId.HEALER, WeatherType.NONE],
+  ])(
+    "should damage if enemy is not woken up by '%s' ability in the same turn",
+    async (_abilityName, abilityId, weatherType) => {
+      const { override, classicMode, field, move } = game;
+      override.weather(weatherType).enemyAbility(abilityId).enemyPassiveAbility(AbilityId.OVERCOAT); // overcoat to prevent sandstorm damage
```
PigeonBar:
```suggestion
      override.weather(weatherType)
        .battleType("double")
        .enemyAbility(abilityId)
        .enemyPassiveAbility(AbilityId.OVERCOAT); // overcoat to prevent sandstorm damage
```

test/abilities/bad-dreams.test.ts
```diff
@@ -0,0 +1,146 @@
...
+      await game.toEndOfTurn();
+    }
+
+    expect(enemy).toHaveStatusEffect(StatusEffect.SLEEP);
+    expect(enemy).toHaveTakenDamage(enemy.getMaxHp() / 8);
+  });
+
+  it.each([
+    ["Hydration", AbilityId.HYDRATION, WeatherType.RAIN],
+    ["Shed Skin", AbilityId.SHED_SKIN, WeatherType.SANDSTORM],
```
PigeonBar: Is there a reason why Sandstorm is being used?

src/data/abilities/ab-attrs/post-turn-hurt-if-sleeping-ab-attr.ts
innerthunder: I honestly think it'd be better to leave an `edgeCase` and move on. This isn't the sort of solution we want to continue using for event order (#1211)

test/abilities/bad-dreams.test.ts
flx-sta: Probably copy-paste error

test/abilities/bad-dreams.test.ts
flx-sta: Added a workaround

src/data/abilities/apply-ab-attrs.ts
flx-sta: What do you mean that this enforces it? Providing a default does the opposite of enforcing it

src/data/abilities/ab-attrs/bad-dreams-ab-attr.ts
```diff
@@ -0,0 +1,52 @@
...
+    this._flags.add(AbAttrFlag.BAD_DREAMS);
+  }
+
+  public override apply(pokemon: Pokemon, simulated: boolean): boolean {
+    let isApplied = false;
+
+    for (const opp of pokemon.getOpponents()) {
+      const isAsleep = opp.hasStatusEffect(StatusEffect.SLEEP);
+      const blocksNonDirectDamage = opp.hasAbilityWithAttr(AbAttrFlag.BLOCK_NON_DIRECT_DAMAGE);
+      const willFallAsleep = opp.getTag<DrowsyTag>(BattlerTagType.DROWSY)?.turnCount === 1; // TODO: Workaround because Drowsy sets the sleep status AFTER applying bad dreams due to "asPhase = true"
```
MokaStitcher:
```suggestion
      // TODO: Workaround because Drowsy sets the sleep status AFTER applying bad dreams due to "asPhase = true"
      const willFallAsleep = opp.getTag<DrowsyTag>(BattlerTagType.DROWSY)?.turnCount === 1;
``` 

Not sure I understand what actually is to do. Do this in another way by moving the time bad dream gets applied? (Or the time drowsy get applied?)

src/data/abilities/ab-attrs/bad-dreams-ab-attr.ts
```diff
@@ -0,0 +1,52 @@
...
+
+      console.log(
+        `isAsleep: ${isAsleep}, willFallAsleep: ${willFallAsleep}, blocksNonDirectDamage: ${blocksNonDirectDamage}`,
+      );
+
```
MokaStitcher:
```suggestion
```

src/data/abilities/apply-ab-attrs.ts
MokaStitcher: Except if that default is `never`, the compiler will complain if it's not changed. So using never as default enforces usage of the typing
image

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -5,24 +5,48 @@ import { AbilityApplyMode } from "#enums/ability-apply-mode";
...
+/**
+ * The function for using the {@linkcode AbilityApplyMode.DEFAULT | DEFAULT} ability mode
+ * @param abAttrFlag - The {@linkcode AbAttrFlag} to apply
+ * @param params - The parameters for the given attribute's `apply` function
+ * @returns An array of {@linkcode AppliedAbAttr | applied ability attributes}
+ *
+ * @see {@linkcode applyAbAttrsInternal}
+ */
+export function applyAbAttrs<TAttr extends AbAttr>(
```
MokaStitcher:
```suggestion
export function applyAbAttrs<TAttr extends AbAttr = never>(
```

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -31,53 +55,58 @@ export function getAbApplyFunc(mode: AbilityApplyMode) {
...
  * @param params The parameters for the given attribute's `apply` function. This should include:
  * - `pokemon`: The {@linkcode Pokemon} with the ability
  * - `simulated`: If `true`, suppresses changes to game state when applying.
  * - Any additional necessary arguments for the specific attribute type
- * @returns The message(s) displayed when the ability applies
+ * @returns An array of {@linkcode AppliedAbAttr | applied ability attributes}
  * @see {@linkcode AbAttr}
  */
-function applyAbAttrsInternal<TAttr extends AbAttr = never>(
+function applyAbAttrsInternal<TAttr extends AbAttr>(
```
MokaStitcher:
```suggestion
function applyAbAttrsInternal<TAttr extends AbAttr = never>(
```

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -86,28 +115,47 @@ function applyAbAttrsInternal<TAttr extends AbAttr = never>(
...
       }
+
       if (result) {
-        const message = attr.getTriggerMessage(pokemon, ability.name, ...args);
+        message = attr.getTriggerMessage(pokemon, ability.name, ...args);
+
         if (message) {
           if (!simulated) {
             globalScene.phaseManager.queueMessagePhase(message);
           }
```
MokaStitcher:
```suggestion
        if (message && !simulated) {
          globalScene.phaseManager.queueMessagePhase(message);
```

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -31,53 +55,58 @@ export function getAbApplyFunc(mode: AbilityApplyMode) {
...
  */
-function applyAbAttrsInternal<TAttr extends AbAttr = never>(
+function applyAbAttrsInternal<TAttr extends AbAttr>(
   abFilterOptions: AbilityFilterOptions,
   abAttrFlag: AbAttrFlag,
   ...params: Parameters<TAttr["apply"]>
-): string[] {
-  const messages: string[] = [];
+): AppliedAbAttr<TAttr>[] {
+  const applied: AppliedAbAttr<TAttr>[] = [];
```
MokaStitcher: this should probably be called `result(s)`, `applicationResults`, or similar, `applied` sounds like a boolean

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -31,53 +55,58 @@ export function getAbApplyFunc(mode: AbilityApplyMode) {
...
+    const matchingAttrs = ability.getAttrs<TAttr>(abAttrFlag).filter((attr) => {
       const condition = attr.getCondition();
       return !condition || condition(pokemon);
     });
 
     matchingAttrs.forEach((attr) => {
       globalScene.phaseManager.setPhaseQueueSplice();
-
+      let message: AppliedAbAttr<TAttr>["message"] = null;
       const result = attr.apply(pokemon, simulated, ...args);
```
MokaStitcher: And this, which is a boolean, could be named `applied`, or `applies` instead of `result`

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -5,24 +5,48 @@ import { AbilityApplyMode } from "#enums/ability-apply-mode";
 import type { AbilityFilterOptions } from "#types/ability-filter-options";
 import { queueShowAbility } from "#utils/ability-utils";
 
+//#region ApplyAbAttrsResult
+
+interface AppliedAbAttr<TAttr extends AbAttr> {
```
MokaStitcher: Not sure about this name since if I understand correctly even ability that do not get applied will have and `AppliedAbAttr` for them, but with `AppliedAbAttr.result` being `false`

Maybe something like `AbilityResult`, `AbilityApplyResult`, `AbilityApplicationResult` or something like that? We a have similar thing for moves, with `MoveResult`, `AttackMoveResult`, etc.

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -5,24 +5,48 @@ import { AbilityApplyMode } from "#enums/ability-apply-mode";
...
+  /** The ability attribute */
+  attr: TAttr;
+  /** The (simulated) result of applying the attribute */
+  result: boolean;
```
MokaStitcher: Again not fully sure about this name, something like `applies` or `applied` seems better to me.

Not entirely sure the comment is super clear either, if i understand correctly `result` here is the return value from the ability's `apply` function, which says this: 
```
   * @returns `true` if this attribute applies successfully. If {@linkcode showAbility} is enabled,
   * and this apply call is not simulated, returning `true` activates the ability's flyout
   * and {@linkcode getTriggerMessage | trigger message} (if applicable)
```
So all in all I'd say something like that
```suggestion
  /** Whether the ability was applied (or would have be applied, if the check wasn't simulated) */
  applied: boolean;
```

People more familiar with the ability code probably would have better insight to provide

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -86,28 +115,47 @@ function applyAbAttrsInternal<TAttr extends AbAttr = never>(
...
     });
   });
 
-  return messages;
+  return applied;
 }
 
-function applyRevealedAbAttrs<TAttr extends AbAttr = never>(
+/**
+ * The function for using the {@linkcode AbilityApplyMode.REVEALED | REVEALED} ability mode
```
MokaStitcher:
```suggestion
 * Apply abilities with the {@linkcode AbilityApplyMode.REVEALED | REVEALED} ability mode.
```

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -86,28 +115,47 @@ function applyAbAttrsInternal<TAttr extends AbAttr = never>(
...
+function applyRevealedAbAttrs<TAttr extends AbAttr>(
   abAttrFlag: AbAttrFlag,
   ...params: Parameters<TAttr["apply"]>
-): string[] {
+): AppliedAbAttr<TAttr>[] {
   return applyAbAttrsInternal<TAttr>({ canApplyOnly: true, revealedOnly: true }, abAttrFlag, ...params);
 }
 
+/**
+ * The function for using the {@linkcode AbilityApplyMode.IGNORE | IGNORE} ability mode
```
MokaStitcher:
```suggestion
 * Apply abilities with the {@linkcode AbilityApplyMode.IGNORE | IGNORE} ability mode.
```

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -5,24 +5,48 @@ import { AbilityApplyMode } from "#enums/ability-apply-mode";
...
+  /** The (optional) message to show if the attribute was applied */
+  message: string | null;
+}
+
+//#endregion
 //#region Exports
 
-export function applyAbAttrs<TAttr extends AbAttr = never>(
+/**
+ * The function for using the {@linkcode AbilityApplyMode.DEFAULT | DEFAULT} ability mode
```
MokaStitcher:
```suggestion
 * Apply abilities with the {@linkcode AbilityApplyMode.DEFAULT | DEFAULT} ability mode.
```

<sub>should this be moved down in the file next to applyRevealedAbAttrs and ignoreAbAttrs? (or them moved up)

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -31,53 +55,58 @@ export function getAbApplyFunc(mode: AbilityApplyMode) {
...
   }
 }
 
 //#endregion
-//#region Helpers
+//#region Internal Functions
 
 /**
  * Applies a Pokemon's ability attributes of matching type
+ * @template TAttr The specific ability attribute type.
```
MokaStitcher:
```suggestion
 * @typeParam TAttr - The specific ability attribute type.
```
should also add the `-` in the `@param`s

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -86,28 +115,47 @@ function applyAbAttrsInternal<TAttr extends AbAttr = never>(
...
-): string[] {
+): AppliedAbAttr<TAttr>[] {
   return applyAbAttrsInternal<TAttr>({ canApplyOnly: true, revealedOnly: true }, abAttrFlag, ...params);
 }
 
+/**
+ * The function for using the {@linkcode AbilityApplyMode.IGNORE | IGNORE} ability mode
+ * @returns an empty array
+ */
+function ignoreAbAttrs<TAttr extends AbAttr>(
```
MokaStitcher: should maybe be renamed applyIgnoredAbAttrs?
Sure it's a weird but if we're making a separate function just for this we might as well make it consistent with the others

src/phases/turn-end-phase.ts
```diff
@@ -40,6 +40,7 @@ export class TurnEndPhase extends FieldPhase {
           );
         }
         applyAbAttrs<PostTurnAbAttr>(AbAttrFlag.POST_TURN, pokemon, false);
+        applyAbAttrs<PostTurnAbAttr>(AbAttrFlag.BAD_DREAMS, pokemon, false); // TODO: THIS IS A WORKAROUND SO BAD DREAMS DOESN'T HURT POKEMON WAKING UP IN THE SAME TURN. Has to be fixed with #1211
```
MokaStitcher:
```suggestion
       // TODO: Temporary workaround so that bad dreams doesn't hurt Pokemon waking up in the same turn. Has to be fixed with #1211
        applyAbAttrs<PostTurnAbAttr>(AbAttrFlag.BAD_DREAMS, pokemon, false);
```
Guess that answers my question from earlier

test/test-utils/matchers/to-have-taken-damage-matcher.ts
```diff
@@ -1,7 +1,17 @@
...
+}
+
+//#endregion
+
 /**
  * Matcher to check if a Pokemon has taken a specific amount of damage
```
MokaStitcher:
```suggestion
 * Matcher to check if a Pokemon has taken a specific amount of damage.
 * Unless specified, will run the expected damage value through {@linkcode toDmgValue}
 * to round it down and make it a minimum of 1.
```
for clarity's sake

test/abilities/solar-power.test.ts
```diff
@@ -116,11 +114,11 @@ describe("Abilities - Solar Power", () => {
     await classicMode.startBattle(SpeciesId.CHARMANDER);
 
     const playerPkm = field.getPlayerPokemon();
-    const enemeyPokemon = field.getEnemyPokemon();
+    const enemyPkm = field.getEnemyPokemon();
```
MokaStitcher, line 117: I would've renamed playerPkm rather than this one if the issue is consistency, but doesn't matter

test/abilities/bad-dreams.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    const { classicMode, field, move } = game;
+    await classicMode.startBattle(SpeciesId.DARKRAI);
+    const enemy = field.getEnemyPokemon();
+    move.use(MoveId.SPLASH);
+    await game.toEndOfTurn();
+
+    expect(enemy).toHaveFullHp();
+  });
+
+  it("should do 1/8 max-hp damage to drowsy enemies falling asleep in the same turn", async () => {
```
MokaStitcher, line 61: would be nice to have a test for comatose as well.
In this case the behavior is defined and Bad Dreams should apply to Pokemon with Comatose

MokaStitcher (commented):
Code review only

src/data/abilities/ab-attrs/bad-dreams-ab-attr.ts
flx-sta: The time drowsy transitions over into sleep happens AFTER bad-dreams was applied

MokaStitcher: I understand that, but not what exactly is "to do" (since it doesn't say)

flx-sta: Drowsy should be setting the Sleep status BEFORE bad dreams is applied. It's supposed to hurt the pokemon as soon as it falls asleep

src/data/abilities/apply-ab-attrs.ts
flx-sta: I was torn the same way with the naming. I can rename it to `applyIgnoredAbAttrs` but it sounds misleading.

Maybe `applyNoAbAttrs`?

src/data/abilities/ab-attrs/bad-dreams-ab-attr.ts
flx-sta: FYI, the additional `applyAbAttr()` call in turn-end phase with "Pokemon waking up in the same turn" has nothing to do with this one

src/data/abilities/apply-ab-attrs.ts
flx-sta: That's actually a good idea. This would also make it easier in my mind to name the interface something with `Result`. I simply adopted the name from the original function body

src/data/abilities/apply-ab-attrs.ts
flx-sta: It shouldn't be moved down since this is part of the `Exports` region

src/data/abilities/apply-ab-attrs.ts
flx-sta: Id' prefer to stick to `@template`. My internal tsdoc highlighter can connect `TAttr` then. If I use `@typeParam` (which is from the official tsdoc though), it doesn't work.

test/abilities/solar-power.test.ts
flx-sta: I'm not even sure why I renamed this tbh

src/data/abilities/apply-ab-attrs.ts
MokaStitcher: the `@template`s were replaced by `@typeParam`s in #1216 because they raise warnings due to invalid syntax for tsdoc as you said.
If we want to use `template` we should decide this as a group, add it as a custom tag in the tsdoc config and replace them all back. We don't want to have a mix
I'd rather we stick to typeParam for now and make a different PR if we decide to use `@template` instead

test/abilities/solar-power.test.ts
MokaStitcher: i was confused too 😂

MokaStitcher (dismissed):
Found an issue with the drowsy thing, sorry

Steps to reproduce
- Turn 1: Yawn
- Turn 2: Uproar (without killing, or getting killed)
=> At the end of turn 2 the opponent will not fall asleep because of uproar, but will receive the damage from bad dreams

I imagine the current check for drowsy turn count can also fail in other cases like terrain or ability/move from a double's partner preventing the drowsiness from taking effect

Maybe on top of the drowsy turn count we should double check with `pokemon.canSetStatus` if it will be able to take effect or not. It maybe won't cover every single edge but hopefully most of them would be caught by this


<sub> also unresolved [this conversation](https://github.com/Despair-Games/poketernity/pull/1209#discussion_r2116481844) in case you missed it

flx-sta:
> Found an issue with the drowsy thing, sorry

<details><summary>Original Comment </summary>
<p>

> Steps to reproduce
> 
>     * Turn 1: Yawn
> 
>     * Turn 2: Uproar (without killing, or getting killed)
>       => At the end of turn 2 the opponent will not fall asleep because of uproar, but will receive the damage from bad dreams
> 
> 
> I imagine the current check for drowsy turn count can also fail in other cases like terrain or ability/move from a double's partner preventing the drowsiness from taking effect
> 
> Maybe on top of the drowsy turn count we should double check with `pokemon.canSetStatus` if it will be able to take effect or not. It maybe won't cover every single edge but hopefully most of them would be caught by this

</p>
</details> 

@MokaStitcher Thanks for the detailed testing

----

### Yawn -> Uproar | No damage

<details><summary>[🎥 Show Video]</summary>
<p>

https://github.com/user-attachments/assets/40f753c4-ece8-4b02-a34c-a3eb337352e5

</p>
</details> 

### Yawn -> Splash | Damage 

<details><summary>[🎥 Show Video]</summary>
<p>

https://github.com/user-attachments/assets/54bc1d99-356c-4e07-81be-6525f3f34234

</p>
</details>

src/data/abilities/ab-attrs/bad-dreams-ab-attr.ts
```diff
@@ -11,17 +13,25 @@ import i18next from "i18next";
...
-      if (
-        opp.hasStatusEffect(StatusEffect.SLEEP)
-        && !opp.hasAbilityWithAttr(AbAttrFlag.BLOCK_NON_DIRECT_DAMAGE)
-        && !opp.switchOutStatus
-      ) {
+      const isAsleep = opp.hasStatusEffect(StatusEffect.SLEEP);
+      const blocksNonDirectDamage = opp.hasAbilityWithAttr(AbAttrFlag.BLOCK_NON_DIRECT_DAMAGE);
+      // TODO: Workaround because Drowsy sets the sleep status AFTER applying bad dreams due to "asPhase = true"
+      const willFallAsleep =
+        opp.getTag<DrowsyTag>(BattlerTagType.DROWSY)?.turnCount === 1 && opp.canSetStatus(StatusEffect.SLEEP);
```
MokaStitcher:
```suggestion
        opp.getTag<DrowsyTag>(BattlerTagType.DROWSY)?.turnCount === 1 && opp.canSetStatus(StatusEffect.SLEEP, true);
```

`canSetStatus` needs to be called quietly otherwise it will trigger a "not falling asleep" message, causing it to appear twice (the video you took stopped a turn early so we didn't get to see it lol)
<sub> the message of uproar causing sleep to fail is not correct for the drowsy/uproar interaction but that's an issue with uproar or drowsy, so unrelated lol

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -5,24 +5,50 @@ import { AbilityApplyMode } from "#enums/ability-apply-mode";
 import type { AbilityFilterOptions } from "#types/ability-filter-options";
 import { queueShowAbility } from "#utils/ability-utils";
 
+//#region ApplyAbAttrsResult
+
+/** The result of applying an ability attribute, whether simulated or actual. */
+interface ApplyAbAttrResult<TAttr extends AbAttr> {
+  /** The ability attribute that was processed */
+  attr: TAttr;
+  /** Whether the ability was applied (or would have be applied, if the check wasn't simulated) */
```
DayKev:
```suggestion
  /** Whether the ability was applied (or would have been applied, if the check wasn't simulated) */
```

src/data/abilities/apply-ab-attrs.ts
```diff
@@ -86,28 +117,47 @@ function applyAbAttrsInternal<TAttr extends AbAttr = never>(
...
 
-function applyRevealedAbAttrs<TAttr extends AbAttr = never>(
+/**
+ * Apply abilities with the {@linkcode AbilityApplyMode.REVEALED | REVEALED} ability mode.
+ * @typeParam TAttr - The specific ability attribute type.
+ * @param abAttrFlag - The {@linkcode AbAttrFlag} to apply
+ * @param params - The parameters for the given attribute's `apply` function
+ * @returns An array of {@linkcode ApplyAbAttrResult | applied ability attributes}
+ */
+function applyRevealedAbAttrs<TAttr extends AbAttr>(
```
DayKev:
```suggestion
function applyRevealedAbAttrs<TAttr extends AbAttr = never>(
```

src/phases/turn-end-phase.ts
```diff
@@ -40,6 +40,8 @@ export class TurnEndPhase extends FieldPhase {
           );
         }
         applyAbAttrs<PostTurnAbAttr>(AbAttrFlag.POST_TURN, pokemon, false);
+       // TODO: Temporary workaround so that bad dreams doesn't hurt Pokemon waking up in the same turn. Has to be fixed with #1211
+        applyAbAttrs<PostTurnAbAttr>(AbAttrFlag.BAD_DREAMS, pokemon, false);
```
DayKev, line 44: I don't think we should be adding workarounds that are just going to need to be removed later. We don't have a playerbase we need to worry about, so it's fine if something is broken until we get around to that particular refactor.

flx-sta: hm...
So do you want to remove it and add an `.edgeCase()` marker?

The problem with those fixes is that I don't know how long it would take, since it's related to the exec-order

MokaStitcher (dismissed):

MokaStitcher (dismissed):

src/data/abilities/ab-attrs/bad-dreams-ab-attr.ts
```diff
@@ -11,17 +13,25 @@ import i18next from "i18next";
 /**
  * Attribute to damage all sleeping opponents by 1/8 of their max hp at the end of turn.
  * Used for {@link https://bulbapedia.bulbagarden.net/wiki/Bad_Dreams_(Ability) | Bad Dreams}.
- * @extends PostTurnAbAttr
+ * @extends AbAttr
```
DayKev, line 16:
```suggestion
 * @extends AbAttr
 * @todo This should extend `PostTurnAbAttr` but currently does not as a workaround until proper ability timing is implemented.
```

DayKev (dismissed):

src/phases/turn-end-phase.ts
DayKev: Might as well just leave it and add a comment, since it's already done.
https://github.com/Despair-Games/poketernity/pull/1209#discussion_r2130578808

PR #1215, opened 2025-05-22
[P2][Checkup][Ability] Sturdy

### LINKED ISSUE #193: [Ability][Checkup] Sturdy
## Ability Requirements
- [x] Documentation
- [x] Tests
- [x] Sturdy makes the Pokémon unaffected by one-hit knockout moves. 
- [x] In addition to Sturdy's previous effect, Sturdy now works like a Focus Sash, causing the Pokémon to have 1 HP remaining if another Pokémon's attack or confusion damage would have brought it from full health to 0 HP. 
- [x] Sturdy will activate as long as a Pokémon has full HP before the attack, and therefore it can activate multiple times if the Pokémon is brought back to full health, or if the Pokémon has a maximum of 1 HP.
- [x] If a Pokémon with Sturdy is holding a Focus Sash, Sturdy will activate first. 
- [x] A Pokémon with Mold Breaker, Teravolt or Turboblaze will ignore Sturdy. 
- [x] When a multistrike move hits a Pokémon with Sturdy, Sturdy's effect can only activate for hits while the user has full HP.
- [x] Destiny Bond and Perish Song will not cause Sturdy to activate. 
- [x] Should not proc on sacrificial moves

## References
- [Bulbapedia - Sturdy](https://bulbapedia.bulbagarden.net/wiki/Sturdy_(Ability))

### LINKED ISSUE #523: [P2] Sturdy Activates when the Actual Damage done to a Boss Pokemon on the Initial Hit is greater or equal to their Actual Health
Copied over from PKR 4199

### Describe the bug

When breaking Boss segments, Actual Damage required to break multiple boss segments is calculated by
x1 for the 1st Segment + x2 for the 2nd Segment + x4 for the 3rd Segment.

**FYI:** After it's calculated, if the actual damage does not pass the threshold of x1 + x2  or + x4, etc** (where it did enough damage to break at least one segment, but doesn't actually do enough damage to break another), **the damage number displayed by [Damage Numbers: Fancy] will only display the resulting damage the Boss Pokémon took for those segments to have broken, ignoring the Actual Damage required to break that/those segments.

-Sturdy seems to look at the Actual Damage done on the first hit when checking if Sturdy should be activated (where Sturdy pops up and it survives on 1 HP)
-This can result in a Boss that has had Sturdy "activated", and on that turn, survives a second hit from an ally with 1 HP left.
 
**In practice, this means that for a boss with Sturdy, two segments, and 26 Health;**
( x1 + x2 segment damage needs to be done to break both segments.  which translates to {(HP/Segment number) + 2(HP/Segment Number) = Actual Damage required}  -->  13 + 26 = 39 )
(When **between 13 and 38** Actual Damage is done, then the first segment will break, and the Damage Numbers will show "13". )

**a) If Actual Damage done is greater or equal to 26, then Sturdy activates, and because 39 isn't exceeded, the next segment is left intact.**
b) If Actual Damage done is greater than 13 and does not exceed 26, then Sturdy does not activate, and the next segment is left intact.

In a), Sturdy has been activated, and if the next hit would KO it, the Actual damage of that hit is reduced to let it survive on 1 HP.

Link to Discord Post: (Incorrect Sturdy proc on boss with multihit move)
https://discord.com/channels/1125469663833370665/1283829540388470807

### Session export file

_No response_

### User data export file

_No response_

### Expected behavior

Sturdy shouldn't "activate" if the Pokémon does not have a full health bar, and it should only activate for the hit that would otherwise KO it in doubles.

### Screenshots / Videos

Night Shade deals 32 damage to a Boss Geodude with 26 HP or less, Sturdy "Activates"

https://github.com/user-attachments/assets/f432118c-366a-46dc-8de6-9f89eddb7bd4

Sturdy Activates for the turn in a double battle.

https://github.com/user-attachments/assets/f021ef2d-99ef-4711-96fb-1d5a9044add4

Tusk is level 25 and Boss Geodude(Sturdy) has 26 health. Because Night Shade did 25 damage to a 26 Health Geodude(Sturdy), Sturdy does not "Activate"

https://github.com/user-attachments/assets/8634f904-d70a-449f-84aa-605b2f95c47f

LEVEL 25, 26 HP GEO (NO STURDY)

### What OS did you observe the bug on?

Other

### If other please specify

gameplay

### Which browser do you use?

Other

### If other please specify

gameplay

### Additional context

> The boss segment logic is weirdly placed in between different parts of the logic for Sturdy, so the overall sequence for damage calcs looks something like
> 1. Raw damage (for the first hit) is calculated disregarding Sturdy, endure tokens, etc.
> 2. If the raw damage is greater than the target's HP, the target (if it has Sturdy) adds a "sturdy" tag to itself that lasts until the end of the turn
> 3. Boss segment logic is applied, which can reduce the incoming damage to hit an HP threshold
> 4. If the target has a "sturdy" tag and more than 1 HP, reduce the incoming damage to `target.hp - 1`
> 
> The ability popup shows when (2) applies, but that doesn't account for (3) reducing the damage, and if (3) reduces the damage, then (4) can still apply to the next hit on that turn
- Innerthunder

False Swipe Damage Suppression Occurs after Actual Damage is calculated, resulting in massive damage suppression whenever it would otherwise reduce a Boss to 1 HP
https://discord.com/channels/1125469663833370665/1232405429083050065

flx-sta:

- The `Sturdy` ability will now trigger on self-inflicted confusion damage (given that it would go from 100% HP to 0% HP)

- Closes #193 
- **[P2]** Closes #523 (Ability flyout issue will still be present!)
- **[P2]** Fixes: Sturdy now procs on self-inflicted dmg (e.g. confusion)

- Sturdy ability test coverage enhancement.
- Add `SACRIFICIAL_MOVES`move constants.
- Add `toHaveHp` matcher
- Add `toHaveFainted` matcher
- Add `applyAbAttrs<PreDefendFullHpEndureAbAttr>(AbAttrFlag.PRE_DEFEND_FULL_HP_ENDURE, /*...*/);` call to `ConfusedTag.lapse` call

## Screenshots/Videos

<details><summary>[🎥 Boss enduring]</summary>
<p>

https://github.com/user-attachments/assets/1e651fe6-a9b4-45a8-80cf-c1c3230297bc

</p>
</details> 

<details><summary>[🎥 Enduring confusion self-damage]</summary>
<p>

https://github.com/user-attachments/assets/090b6231-c3ae-430f-a9d8-8631d37e1bea

</p>
</details> 

## How to test the changes?

`pnpm test:silent`

## Checklist


flx-sta:
~~Taking a look at #523 too~~

Exceeds the range of a Checkup PR

test/abilities/sturdy.test.ts
```diff
@@ -22,55 +44,224 @@ describe("Abilities - Sturdy", () => {
...
+
+    move.use(MoveId.DESTINY_BOND);
+    await move.forceEnemyMove(MoveId.TACKLE);
+    await game.toEndOfTurn();
+
+    expect(player).toHaveFainted();
+    expect(enemy).toHaveFainted();
+  });
+
+  it.each(sacrificialMoves)("should not proc on sacrificial '%s' move", async (_enemyMoveName, enemyMoveId) => {
```
Tempo-anon, line 251: nit: should clarify this is user using a sacrificial move (or is the term self KO?)

test/abilities/sturdy.test.ts
```diff
@@ -22,55 +44,224 @@ describe("Abilities - Sturdy", () => {
...
+    move.use(MoveId.SPLASH);
+    await move.forceEnemyMove(enemyMoveId);
+    await game.toEndOfTurn();
 
-    const enemyPokemon: EnemyPokemon = game.scene.getEnemyParty()[0];
-    expect(enemyPokemon.hp).toBe(0);
-    expect(enemyPokemon.isFainted()).toBe(true);
+    expect(enemy).toHaveFainted();
   });
 });
```
Tempo-anon, line 304: wonder if there should be tests
* When it comes to bosses
* When it comes to status damage (though with the way the game works that only applies to Shedinja since that's the only time % damage = entire hp)
* For not interacting with wonder guard

src/data/abilities/ab-attrs/pre-defend-full-hp-endure-ab-attr.ts
```diff
@@ -5,6 +5,10 @@ import type { Pokemon } from "#field/pokemon";
 import type { Move } from "#moves/move";
 import type { NumberHolder } from "#utils/common-utils";
 
+/**
+ * If the pokemon with this ab-attr is full hp and hit with a move that would kill it, it will survive with 1 hp left.
+ * @see {@link https://bulbapedia.bulbagarden.net/wiki/Sturdy_(Ability) | Sturdy Ability - Bulbapedia}
+ */
 export class PreDefendFullHpEndureAbAttr extends PreDefendAbAttr {
```
Tempo-anon: wonder if it makes sense to just name this SturdyAbAttr or something 🤔

src/data/abilities/ab-attrs/pre-defend-full-hp-endure-ab-attr.ts
```diff
@@ -5,6 +5,10 @@ import type { Pokemon } from "#field/pokemon";
 import type { Move } from "#moves/move";
 import type { NumberHolder } from "#utils/common-utils";
 
+/**
+ * If the pokemon with this ab-attr is full hp and hit with a move that would kill it, it will survive with 1 hp left.
```
Tempo-anon: I think it will be beneficial to add additional information here like how it interacts with boss bars (no mainline correlation) and sturdy (does not work unlike mainline)

Tempo-anon (dismissed):
While the PR does a checkup, I think it's worth adding in the actual change in the PR title as well (sturdy now procs on self confusion hit)

test/abilities/sturdy.test.ts
flx-sta: I _stole_ the term `sacrificial` from the Attr name. I thought it pretty clear

test/abilities/sturdy.test.ts
flx-sta: There should probably be tests for that, Good point.
It's possible that my boss testing will be lackluster though, as there seems to be quite some bugs involved in that

src/data/abilities/ab-attrs/pre-defend-full-hp-endure-ab-attr.ts
flx-sta: Sure, why not. I'll just double check it's usages

src/data/abilities/ab-attrs/sturdy-ab-attr.ts
```diff
@@ -5,7 +5,20 @@ import type { Pokemon } from "#field/pokemon";
 import type { Move } from "#moves/move";
 import type { NumberHolder } from "#utils/common-utils";
 
-export class PreDefendFullHpEndureAbAttr extends PreDefendAbAttr {
+/**
+ * If the pokemon with this ab-attr is full hp and hit with a move that would 1-hit faint it, it will survive with 1 hp left.
+ * Wonder Guard _overrules_ this ab-attr.
+ *
+ * #### Boss Pokemon
```
DayKev: Is this just indirect documentation of the bug?
The ability prevents fainting if the user would go from full HP to 0 HP in one hit, causing the user to survive with 1 HP left. Boss bars make that more difficult (basically increasing the pokemon's effective HP, but not actual HP), but shouldn't change how the ability activates.

src/data/battler-tags/confused-tag.ts
```diff
@@ -60,39 +66,56 @@ export class ConfusedTag extends BattlerTag {
...
-      const damage = this.getDamage(pokemon);
-      if (damage > 0) {
+      const damageHolder = new NumberHolder(this.getDamage(pokemon));
+
+      if (damageHolder.value > 0) {
         globalScene.phaseManager.queueMessagePhase(i18next.t("battlerTags:confusedLapseHurtItself"));
-        pokemon.damageAndUpdate(damage);
+
+        if (pokemon.isFullHp()) {
+          // `applyAbAttrs` may modify `damageHolder` by reference to adjust the damage value.
```
DayKev:
```suggestion
```
These would just be nuisance comments if we had them before every single `applyAbAttrs` call.

Edit: Also this is just inaccurate anyway, the `AbAttr` doesn't modify the damage dealt, it just applies a `BattlerTag` that handles the actual damage modification.

src/data/battler-tags/confused-tag.ts
```diff
@@ -60,39 +66,56 @@ export class ConfusedTag extends BattlerTag {
...
+          applyAbAttrs<SturdyAbAttr>(
+            AbAttrFlag.PRE_DEFEND_FULL_HP_ENDURE,
```
DayKev: Since you renamed the `AbAttr` you should rename the matching `AbAttrFlag` too.

src/data/battler-tags/sturdy-tag.ts
```diff
@@ -10,8 +10,7 @@ import i18next from "i18next";
 /**
  * Tag to manage messages related to the enduring effect of
  * {@link https://bulbapedia.bulbagarden.net/wiki/Sturdy_(Ability) | Sturdy}.
- * This is added to Pokemon with the ability whenever they receive
- * lethal attack damage from full HP.
+ * This is added to Pokemon with the ability whenever they receive lethal attack damage from full HP.
  * @extends BattlerTag
  */
 export class SturdyTag extends BattlerTag {
```
DayKev, line 16: ~Hm, is this `BattlerTag` really needed for the message to appear at the right time? It seems to me we should just be able to use `appendToPhase(messagePhase, PhaseId.MOVE_EFFECT)`.~
~https://www.youtube.com/watch?v=dEUO7E_ozQA~

Edit: This tag seems to be what's applying the actual effects of Sturdy, meaning this TSDoc is inaccurate (it doesn't just handle the messages).

test/abilities/sturdy.test.ts
```diff
@@ -25,50 +63,312 @@ describe("Abilities - Sturdy", () => {
...
+
+    move.use(MoveId.SPLASH);
+    await phaseInterceptor.to("PostKnockoutPhase", true);
+
+    expect(enemy).toHaveFainted();
+  });
+
+  it("should not proc on 'Destiny Bond' move", async () => {
+    const { classicMode, field, move } = game;
+    vi.spyOn(DestinyBondTag.prototype, "lapse");
```
DayKev, line 230: You never checked that these functions weren't called (I assume that's the intention of spying on them?).

test/abilities/sturdy.test.ts
```diff
@@ -25,50 +63,312 @@ describe("Abilities - Sturdy", () => {
...
-    expect(enemyPokemon.isFainted()).toBe(true);
+    expect(enemy.bossSegmentIndex).toBe(1);
+    expect(enemy).toHaveHp(1);
   });
 
-  test("Sturdy pokemon should be immune to OHKO moves", async () => {
-    await game.classicMode.startBattle(SpeciesId.LUCARIO);
-    game.move.select(MoveId.FISSURE);
-    await game.phaseInterceptor.to("PostActionPhase");
+  it("should not proc when 'Wonder Guard' ability is present too", async () => {
```
DayKev, line 287: Is this just testing that the flyout doesn't show incorrectly?

There should be a test showing that Sturdy doesn't activate if a Wonder Guard pokemon takes damage from an attack (it should faint).

test/test-utils/matchers/to-have-fainted-matcher.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (!isPokemonInstance(received)) {
+    return {
+      pass: this.isNot,
+      message: () => `Expected Pokemon, but got ${receivedStr(received)}!`,
+    };
+  }
+
+  const { hp } = received;
+  const maxHp = received.getMaxHp();
+  const pass = received.isFainted() === true && hp === 0;
```
DayKev:
```suggestion
  const pass = received.isFainted() === true && hp === 0;
```
`Pokemon#isFainted` is already checking for the pokemon having 0 HP.

DayKev:
Commented on #523 with info on fixing the bugs from it. The functionality bug should be a simple fix, the ability flyout is more complicated but also just a P3 so 🤷.

src/data/abilities/ab-attrs/sturdy-ab-attr.ts
flx-sta: This isn't documenting the bug, it's documenting the actual expected behavior with an example of how it would be expected to 1hit it.
If it's not clear enough let me know. It's rather complicated thus hard to summarize in few words without going too much into the weeds

DayKev: I don't think it makes sense to document how boss bars work here, they should be documented at the fields/methods/etc related to them. You don't need to know about boss bars to understand Sturdy because there's no interaction between them.

flx-sta:
> Commented on #523 with info on fixing the bugs from it. The functionality bug should be a simple fix, the ability flyout is more complicated but also just a P3 so 🤷.

Added the fix for that to the PR

src/data/battler-tags/sturdy-tag.ts
```diff
@@ -8,10 +8,9 @@ import type { Pokemon } from "#field/pokemon";
 import i18next from "i18next";
 
 /**
- * Tag to manage messages related to the enduring effect of
+ * Tag to behavior related to the enduring effect of
```
DayKev:
```suggestion
 * Tag to implement behavior related to the enduring effect of
```

test/abilities/sturdy.test.ts
```diff
@@ -25,50 +62,311 @@ describe("Abilities - Sturdy", () => {
...
+    await move.forceEnemyMove(enemyMoveId);
+    await game.toEndOfTurn();
+
+    expect(enemy).toHaveFainted();
   });
 
-  test("Sturdy doesn't activate when user is not at full HP", async () => {
-    await game.classicMode.startBattle(SpeciesId.LUCARIO);
+  it("should proc properly on Boss Pokemon and deplete all hp-segments but the last", async () => {
+    // Known issue: Boss Pokemon sturdy triggers at the end of the first hp-segment, instead of the last (See #523)
```
DayKev, line 284:
```suggestion
```

DayKev (dismissed):

test/abilities/sturdy.test.ts
DayKev:
```suggestion
    // Known issue: Boss Pokemon sturdy flyout triggers at the end of the first hp-segment, instead of the last (See #523)
```

src/data/abilities/ab-attrs/sturdy-ab-attr.ts
flx-sta: But this is specific to the Sturdy behavior. It's not universal boss explanation but how sturdy interacts with boss segments. It's unique to PKTY

DayKev: Not really? You're just explaining how much damage is needed to KO a Pokémon with boss bars. Sturdy cares if a Pokémon goes from full HP to 0 HP; yes it takes more damage to do it when the Pokémon has boss bars because they "absorb" some of the damage but there's no actual interaction between Sturdy and the boss bars.
Sturdy sees a Pokémon going from full HP to 0 HP and sets it to 1 HP, it doesn't see the boss bars.
Boss bars are just one of many methods to effectively reduce the damage a Pokémon is taking.

flx-sta: @DayKev Where should i put this explanation/documentation instead then?

DayKev: Somewhere boss bars are defined/handled (maybe in `EnemyPokemon#damage`, at L544?).
https://github.com/Despair-Games/poketernity/blob/ee5269b594e7deafc4cb68a0e3f08a169eba4a61/src/field/enemy-pokemon.ts#L544-L546

src/constants/move-constants.ts
```diff
@@ -5,3 +5,12 @@ export const TERA_MOVES = Object.freeze([MoveId.TERA_BLAST, MoveId.TERA_STARSTOR
 
 /** An array of all pledge {@linkcode MoveId | MoveIds}. */
 export const PLEDGE_MOVES = Object.freeze([MoveId.GRASS_PLEDGE, MoveId.FIRE_PLEDGE, MoveId.WATER_PLEDGE]);
+
+/** An array of all sacrificial {@linkcode MoveId | MoveIds}. */
+export const SACRIFICIAL_MOVES = Object.freeze([
```
innerthunder, line 10: Should this also include Healing Wish and Lunar Dance?
<sub>Given they require a non-fainted Pokemon in the user's party, you'd probably have to modify tests a bit if they're added</sub>

src/data/abilities/ab-attrs/sturdy-ab-attr.ts
```diff
@@ -5,10 +5,16 @@ import type { Pokemon } from "#field/pokemon";
 import type { Move } from "#moves/move";
 import type { NumberHolder } from "#utils/common-utils";
 
-export class PreDefendFullHpEndureAbAttr extends PreDefendAbAttr {
+/**
+ * If the pokemon with this ab-attr is full hp and hit with a move that would 1-hit faint it, it will survive with 1 hp left.
+ * Wonder Guard _overrules_ this ab-attr.
```
innerthunder:
```suggestion
 * If the pokemon with this ab-attr is full hp and hit with a move that would 1-hit faint it, it will survive with 1 hp left (_unless it also has the ability Wonder Guard_).
```
It's not immediately clear what "overruling" the attribute entails. You may want to break this up to reduce line length as well.

innerthunder (dismissed):

test/abilities/sturdy.test.ts
```diff
@@ -25,50 +62,311 @@ describe("Abilities - Sturdy", () => {
...
-    game.override.ability(AbilityId.MOLD_BREAKER);
+  it.each([
+    ["Poison", StatusEffect.POISON],
+    ["Toxic", StatusEffect.TOXIC],
+    ["Burn", StatusEffect.BURN],
+  ])("should not proc on '%s' Status Effect damage", async () => {
+    const { override, classicMode, field, move } = game;
+    override
+      .enemySpecies(SpeciesId.SHEDINJA)
+      .enemyPassiveAbility(AbilityId.WONDER_GUARD)
```
DayKev: Wonder Guard already prevents Sturdy from activating, so this test isn't actually checking that status damage doesn't proc Sturdy.
I think it should be impossible for both Burn and Poison to OHKO a Pokémon anyway since they deal %-based damage, and after checking Bulbapedia Toxic apparently can only do a max of `15/16` of a Pokémon's max HP (which I don't think we do correctly actually lol).
So in the end, these status tests are unnecessary.

test/abilities/sturdy.test.ts
```diff
@@ -25,50 +62,311 @@ describe("Abilities - Sturdy", () => {
...
-    expect(enemyPokemon.hp).toBe(0);
-    expect(enemyPokemon.isFainted()).toBe(true);
+    move.use(MoveId.SPLASH);
+    await game.toEndOfTurn();
+
+    expect(SturdyAbAttr.prototype.apply).not.toHaveBeenCalled();
+    expect(enemy).toHaveFainted();
   });
+
+  it.each(damagingBattlerTagTypes)(
```
DayKev: These moves/effects do %-based damage and so should never be able to OHKO a Pokémon. If they did, that would be a bug.

src/data/abilities/ab-attrs/sturdy-ab-attr.ts
DayKev:
```suggestion
 * If the pokemon with this `AbAttr` is full HP and is hit with a move that would 1-hit faint it,
 * it will survive with 1 HP left (_unless it also has the ability Wonder Guard_).
```
Modified inner's suggestion to break it into 2 lines.

src/constants/move-constants.ts
flx-sta: Probably, yeah.
https://pokemon.fandom.com/wiki/Category:Sacrificing_Moves

test/abilities/sturdy.test.ts
```diff
@@ -25,50 +62,311 @@ describe("Abilities - Sturdy", () => {
...
+
+      expect(enemy).toHaveHp(1);
+      expect(enemy).toHaveFullHp();
+      expect(enemy).toHaveBattlerTag(battlerTagType);
+
+      move.use(MoveId.SPLASH);
+      await game.toEndOfTurn();
+
+      expect(SturdyAbAttr.prototype.apply).not.toHaveBeenCalled();
+      expect(TrappedTag.prototype.lapse).toHaveBeenCalled;
```
flx-sta:
```suggestion
      expect(TrappedTag.prototype.lapse).toHaveBeenCalled();
```

test/abilities/sturdy.test.ts
flx-sta: But is it wrong to cover it in a test? I mean, look at MEs, who knows what other ideas might be added in the future

test/abilities/sturdy.test.ts
flx-sta: Not sure why I used Wonder-Guard here in the first place tbh.
Also not sure if I should keep this test at all if I were to remove it as it would just be redundant to  `"should not proc on '%s' damaging trap battler-tag ",`

test/abilities/sturdy.test.ts
DayKev: If we introduce a new effect like that then a test for that effect should be made, in which case these tests would still be redundant.

DayKev: These tests do have a time cost (~450ms per test for me). If we're testing something relevant that's fine, we shouldn't worry about it. These tests aren't testing for anything that's possible in the game though.

Also you duplicated most of the `BattlerTag` tests as well.
image

DayKev (approved):

PR #1216, opened 2025-05-23
[Dev] Move `typedoc` settings from `tsconfig.json` to `typedoc.json`

DayKev:
N/A

Improved `typedoc` behavior and TSDocs.

- Moved `typedoc` settings from `tsconfig.json` to `typedoc.json`.
- Added `tsdoc.json` to define custom tags.
  - Added block tags: `@todo`, `@extends`
  - Added inline tags: `@linkcode`
- Cleaned up a bunch of TSDocs that were using invalid syntax.

## How to test the changes?
`npm run docs`

## Checklist

### DIFF
--- a/src/data/animations/anim-config-schema.ts
+++ b/src/data/animations/anim-config-schema.ts
@@ -123,33 +123,20 @@ function getNumberArrayKeyFrameSetSchema(
 const animPropSchema: JSONSchemaType<AnimProp> = {
   type: "object",
   properties: {
-    /**
-     * For battle animations, the origin point for keyframes is defined
-     * along the line connecting the start point ("source") and end point ("target").
-     * The `u`-value is the fraction of the distance between the start and end point
-     * the origin point is away from the source. If `u = 0`, then the origin point is
-     * the source; if `u = 1`, then the origin point is the target.
-     */
     u: getNumberKeyFrameSetSchema({
       type: "number",
       minimum: 0,
       maximum: 1,
     }),
 
-    /** The horizontal coordinate relative to the keyframe's origin point. */
     x: getNumberKeyFrameSetSchema({
       type: "number",
     }),
 
-    /**
-     * The vertical coordinate relative to the keyframe's origin point.
-     * An increase in `y` will move the sprite downward.
-     */
     y: getNumberKeyFrameSetSchema({
       type: "number",
     }),
 
-    /** Horizontal scale factor (%) */
     scaleX: {
       ...getNumberKeyFrameSetSchema({
         type: "number",
@@ -158,7 +145,6 @@ const animPropSchema: JSONSchemaType<AnimProp> = {
       nullable: true,
     },
 
-    /** Vertical scale factor (%) */
     scaleY: {
       ...getNumberKeyFrameSetSchema({
         type: "number",
@@ -167,7 +153,6 @@ const animPropSchema: JSONSchemaType<AnimProp> = {
       nullable: true,
     },
 
-    /** The alpha value for the animated sprite, in the range [0, 255] */
     alpha: {
       ...getNumberKeyFrameSetSchema({
         type: "number",
@@ -177,12 +162,6 @@ const animPropSchema: JSONSchemaType<AnimProp> = {
       nullable: true,
     },
 
-    /**
-     * The rotation angle of the sprite in degrees.
-     * Phaser uses a right-hand clockwise rotation system, where 0 is right,
-     * 90 is down, and -90 is up. The value of this should be in the interval
-     * [-180, 180].
-     */
     angle: {
       ...getNumberKeyFrameSetSchema({
         type: "number",
@@ -192,22 +171,16 @@ const animPropSchema: JSONSchemaType<AnimProp> = {
       nullable: true,
     },
 
-    /** If `true`, flips the sprite horizontally */
     mirror: {
       ...getBooleanKeyFrameSetSchema({ type: "boolean" }),
       nullable: true,
     },
 
-    /** If `false`, hides the sprite */
     visible: {
       ...getBooleanKeyFrameSetSchema({ type: "boolean" }),
       nullable: true,
     },
 
-    /**
-     * The blend mode to specify how the sprite is rendered on the canvas
-     * @see {@link https://docs.phaser.io/api-documentation/constant/blendmodes}
-     */
     blendType: {
       ...getNumberKeyFrameSetSchema({
         type: "integer",
@@ -216,17 +189,11 @@ const animPropSchema: JSONSchemaType<AnimProp> = {
       nullable: true,
     },
 
-    /**
-     * If this keyframe is for a graphic, specifies the tile index used
-     * for the graphic during the tween. This is only relevant for VFX
-     * properties.
-     */
     graphicFrame: {
       ...getNumberKeyFrameSetSchema({ type: "integer" }),
       nullable: true,
     },
 
-    /** A tone to pipeline over the animated sprite (RGBA, normalized) */
     tone: {
       ...getNumberArrayKeyFrameSetSchema({
         type: "array",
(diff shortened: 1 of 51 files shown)


src/data/animations/anim-config-schema.ts
Tempo-anon, line 1: I feel like this removes a lot of good info

Tempo-anon (dismissed):

flx-sta (commented):
What's up with all the removals of the comments in post-summon-stat-stage-change?

src/data/animations/anim-config-schema.ts
MokaStitcher: These seem to be duplicate comments from those in the anim config file. Those that weren't were put there
Duplicate comments means it's even more likely that they get out of sync, so I'm fine with removing them from one place or the other. Don't know if Innerthunder has an opinion

src/data/animations/anim-config-schema.ts
```diff
@@ -357,44 +303,21 @@ const animTimedEventSchema: JSONSchemaType<AnimTimedEvent> = {
...
-     * This is required for all {@linkcode MoveAnim | MoveAnims}.
-     */
     moveId: {
       type: "integer",
       enum: getTSEnumValues(MoveId),
       nullable: true,
     },
 
-    /** The name of the tileset used for the animation. */
+    // The name of the tileset used for the animation.
```
MokaStitcher, line 312: is there a reason this comment was kept?
is it because it's different than the one in anim config? (`/** The file name for VFX assets used in this animation */`)

src/data/battler-tags/grounded-tag.ts
```diff
@@ -12,7 +12,8 @@ import i18next from "i18next";
 
 /**
  * Tag for effects that ground the source, allowing Ground-type moves to hit them.
- * @description `IGNORE_FLYING`: Persistent grounding effects (i.e. from Smack Down and Thousand Waves)
+ *
+ * `IGNORE_FLYING`: Persistent grounding effects (i.e. from Smack Down and Thousand Waves)
```
MokaStitcher: why is this called grounded but the tag type is `IGNORE_FLYING`. That comment was extremely confusing
```suggestion
 * Corresponds to `BattlerTagType.IGNORE_FLYING`
```
or a `@linkcode`, or `@see`?

src/data/mystery-encounters/mystery-encounter.ts
```diff
@@ -578,7 +579,8 @@ export class MysteryEncounterBuilder implements Partial<IMysteryEncounter> {
    */
 
   /**
-   * @statif Defines the type of encounter which is used as an identifier, should be tied to a unique MysteryEncounterType
+   * Defines the type of encounter which is used as an identifier, should be tied to a unique {@linkcode MysteryEncounterType}.
+   *
    * NOTE: if new functions are added to {@linkcode MysteryEncounter} class
```
MokaStitcher, line 584: <sub>what does this mean, this seems to be cut off

src/data/terrain.ts
```diff
@@ -9,8 +9,8 @@ import i18next from "i18next";
 
 /**
  * Class representing Terrain effects
- * @var terrainType - The {@linkcode TerrainType} that is being represented
- * @var turnsLeft - How many turns the terrain still has left
+ * @param terrainType - The {@linkcode TerrainType} that is being represented
```
MokaStitcher: huh, can `@param` be used for class attributes?
Or should these be moved to the attributes themselves, and the constructor?

src/data/weather.ts
```diff
@@ -11,8 +11,8 @@ import i18next from "i18next";
 
 /**
  * Class representing Weather effects
- * @var weatherType - The {@linkcode WeatherType} that is being represented
- * @var turnsLeft - How many turns the weather still has left (0 if immutable)
+ * @param weatherType - The {@linkcode WeatherType} that is being represented
```
MokaStitcher: same question

src/phase-manager.ts
```diff
@@ -56,7 +56,7 @@ interface ToTitleScreenInit {
 }
 
 interface ToLoginScreenInit {
-  /** Whether to show text. @default true */
+  /** Whether to show text. @defaultValue `true` */
```
MokaStitcher, line 59: <sub> that comment left me as confused as before as to what that text is supposed to be lol

MokaStitcher:
> What's up with all the removals of the comments in post-summon-stat-stage-change?

you must be looking at the wrong file. Assuming you're talking about https://github.com/Despair-Games/poketernity/pull/1216#discussion_r2104970867

src/data/animations/anim-config-schema.ts
flx-sta: @innerthunder

innerthunder: A link to the respective interface field would be fine enough while avoiding duplicate comments, e.g.

```ts
const keyframeOptions = {
  /** @see {@linkcode AnimKeyFrame.duration} */
  duration: {
    type: "integer",
    minimum: 0,
    nullable: true,
  },
  // ...
}
```

src/data/animations/anim-config-schema.ts
DayKev: Yeah I wasn't sure which comment was best (or both?).

src/data/animations/anim-config-schema.ts
DayKev: The comments are still shown via intellisense or w/e.
<details><summary>Image</summary>
<p>

image

</p>
</details>

src/data/terrain.ts
DayKev: Sort of. Actually they can be combined between the class definition and the constructor. With the way we use move and ability attr classes it's necessary to put all the TSDocs on the class definition itself (since there's no `new MoveAttr()` the constructor TSDocs don't show up), but with normal classes we can use the intended method, so I'll change this and the `Weather` class docs.

DayKev: Example:
image
image
image

src/phase-manager.ts
DayKev: 🤷 
I was just fixing focused on fixing syntax, I didn't look too closely at the content of most of the docs.

src/data/mystery-encounters/mystery-encounter.ts
DayKev: The comment was just never finished in the first place, like most of the ME docs...

src/data/animations/anim-config-schema.ts
DayKev: > A link to the respective interface field would be fine enough while avoiding duplicate comments, e.g.
> 
> ```ts
> const keyframeOptions = {
>   /** @see {@linkcode AnimKeyFrame.duration} */
>   duration: {
>     type: "integer",
>     minimum: 0,
>     nullable: true,
>   },
>   // ...
> }
> ```

Also that wouldn't work, the TSDocs on the interface (`AnimTimedEvent` etc) will override any TSDocs you attempt to put here. That `@see` would never show up on hover, meaning the `@linkcode` is pointless as it won't be clickable.

src/data/battler-tags/grounded-tag.ts
DayKev: How about this?
```ts
/**
 * Tag for effects that ground the source (i.e. from Smack Down and Thousand Waves),
 * allowing Ground-type moves to hit them.
 *
 * @see {@linkcode BattlerTagType.IGNORE_FLYING}
 * @extends BattlerTag
 */
```

src/phase-manager.ts
flx-sta: https://github.com/Despair-Games/poketernity/blob/26698b3f5252295950947d88b48e06c04bbe33fa/src/phases/login-phase.ts#L44-L46

[locales/en/menu.json](https://github.com/Despair-Games/poketernity-locales/blob/926dcd5ca1a07a979e5e4b383ef6b0a89858a860/en/menu.json#L9)
```ts
  "logInOrCreateAccount": "Log in or create an account to start. No email required!",
```

So apparently if it's `true` it shows that text, and if `false` not

src/data/battler-tags/grounded-tag.ts
MokaStitcher: looks good

src/phase-manager.ts
MokaStitcher: :plains:

src/data/animations/anim-config-schema.ts
MokaStitcher: not sure either

MokaStitcher (approved):

PR #1218, opened 2025-05-23
[Dev] Biome Lint: remove `.` from `biome-ci` script

flx-sta:

n/a

Without that `.` in the command, the annotations will also properly display in the `Files` tab in any PR

- remove `.` from `biome-ci` script

## Screenshots/Videos

The screenshot is form this check:
https://github.com/Despair-Games/poketernity/actions/runs/15220473644/job/42815037181

image

## How to test the changes?

You can't really. You'll just have to trust Moka and me that we visually confirmed it working

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "eslint-ci": "eslint .",
     "biome": "biome check --write --changed --no-errors-on-unmatched",
     "biome:unsafe": "biome check --write --unsafe --changed --no-errors-on-unmatched",
-    "biome-ci": "biome ci --diagnostic-level=error --reporter=github --no-errors-on-unmatched .",
+    "biome-ci": "biome ci --diagnostic-level=error --reporter=github --no-errors-on-unmatched",
     "biome-all": "biome check --write --no-errors-on-unmatched",
     "biome-all:unsafe": "biome check --write --unsafe --no-errors-on-unmatched",
     "prettier": "prettier . --write",


src/data/battler-tags/type-boost-tag.ts
```diff
@@ -4,6 +4,7 @@ import type { BattlerTagType } from "#enums/battler-tag-type";
 import type { ElementalType } from "#enums/elemental-type";
 import type { MoveId } from "#enums/move-id";
 import type { Pokemon } from "#field/pokemon";
+import { AnimFrameTarget } from "#enums/anim-frame-target";
```
MokaStitcher: Lol, this doesn't cause biome to fail, I think only eslint reacts on unused imports
What you can do instead is remove the `type` from one of the other imports

.github/workflows/lint.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       - name: Run Biome-Lint
         run: npm run biome-ci
         id: biome_lint
-        continue-on-error: true
+        # continue-on-error: true
```
MokaStitcher:
```suggestion
        continue-on-error: true
```
With the annotations working I think it's fine to continue on error

flx-sta: yeah, but I wanted to make sure that we aren't seeing any eslint annotations by accident

DayKev (approved):

MokaStitcher (approved):
:rocket:

PR #1219, opened 2025-05-24
[AI-Rework] Add effect scores for moves with `BattlerTag` effects

innerthunder:

The AI should more accurately evaluate moves that apply volatile status conditions and other temporary effects.

<details><summary>All affected moves</summary>
<p>

- Aqua Ring
- Bind et al.
- Follow Me, Rage Powder, and Spotlight
- Charge
- G-Max Chi Strike
- Disable
- Dragon Cheer
- Yawn and G-Max Snooze
- Electrify
- Encore
- Magnet Rise
- Focus Energy
- Foresight et al.
- Outrage et al.
- Glaive Rush
- Smack Down and Thousand Arrows
- Grudge
- Heal Block and Psychic Noise
- Helping Hand
- Imprison
- Attract and G-Max Cuddle
- Ingrain
- Jaw Lock
- Laser Focus
- Leech Seed and Sappy Seed
- Magic Coat
- Nightmare
- No Retreat
- Octolock
- Perish Song
- Powder
- Protect et al.
- Rollout and Ice Ball
- Roost
- Salt Cure
- Snatch
- Stockpile
- Syrup Bomb
- Tar Shot
- Taunt
- Telekinesis
- Throat Chop
- Torment
- Trapping moves (e.g. Spider Web, Spirit Shackle)
- Uproar
- Moves that may make the target flinch
- Moves that may confuse the target
- Moves that require a recharging turn, e.g. Hyper Beam

</p>
</details> 

Part of the AI Rework

- `AddBattlerTagAttr` is now an abstract class. Subclasses are required to provide an override for `getEffectScore`
- Most instances of `AddBattlerTagAttr` in `init-moves.ts` have been replaced with bespoke subclasses with various scoring functions. Pre-existing subclasses also have new effect scores (WIP)
- Glaive Rush's tags -- `RECEIVE_DOUBLE_DAMAGE` and `ALWAYS_GET_HIT` -- have been consolidated into a single `GLAIVE_RUSH` tag
- Added some missing instances of `failOnOverlap` and other `AddBattlerTagOptions` fields
- Renamed `TrapAttr` to `BindingAttr` -- `TrapAttr` is now used for non-damaging trapping effects, e.g. from Spider Web and Spirit Shackle

## How to test the changes?

- [x] `npm run typecheck`
- [x] `npm run depcruise`
- [x] `npm run biome` -- should show no errors
- [x] `npm run test[:silent] move-effect-scores/`
- [x] `npm run test[:silent] move-condition-scores/`

## Checklist


src/@types/status.ts
```diff
@@ -10,6 +10,11 @@ export interface Status {
    * @defaultValue 0
    */
   toxicTurnCount: number;
+  /**
+   * The number of turns the affected Pokemon has slept since
+   * @defaultValue 0
+   */
+  turnsAsleep: number;
```
flx-sta: I think this should instead be `totalSleepTurns` so one can calc the asleep turns from `total - remaining`

innerthunder: I discussed this with @DayKev. Calculating asleep turns this way is inaccurate when the asleep Pokemon has [Early Bird](https://bulbapedia.bulbagarden.net/wiki/Early_Bird_(Ability))

flx-sta: fun fun fun fun funfun fun fun fun funfun fun fun fun funfun fun fun fun funfun fun fun fun funfun fun fun fun funfun fun fun fun funfun fun fun fun funfun fun fun fun funfun fun fun fun fun

src/@types/status.ts
```diff
@@ -10,6 +10,11 @@ export interface Status {
    * @defaultValue 0
    */
   toxicTurnCount: number;
+  /**
+   * The number of turns the affected Pokemon has slept since
```
MokaStitcher:
```suggestion
   * The number of turns the affected Pokemon has been sleeping
```

Should we specify at which point this number is incremented? I assume at the end of each sleeping turn?

src/data/abilities/ab-attrs/pre-apply-battler-tag-immunity-ab-attr.ts
```diff
@@ -20,12 +20,10 @@ export class PreApplyBattlerTagImmunityAbAttr extends PreApplyBattlerTagAbAttr {
...
   }
 
-  public override apply(_pokemon: Pokemon, simulated: boolean, tag: BattlerTag, cancelled: BooleanHolder): boolean {
+  public override apply(_pokemon: Pokemon, _simulated: boolean, tag: BattlerTag, cancelled: BooleanHolder): boolean {
     if (this.immuneTagTypes.includes(tag.tagType)) {
       cancelled.value = true;
-      if (!simulated) {
-        this.battlerTag = tag;
-      }
+      this.battlerTag = tag;
```
MokaStitcher, line 25: this change scares me lol, can you share the reasoning behind it?

src/data/moves/move-attrs/center-of-attention-attr.ts
```diff
@@ -0,0 +1,42 @@
...
+import type { Move } from "#moves/move";
+
+/**
+ * Attribute to apply the {@link https://bulbapedia.bulbagarden.net/wiki/Center_of_attention | Center of Attention}
+ * volatile status condition. Can apply to the user or selected target.
+ * @extends AddBattlerTagAttr
+ */
+export class CenterOfAttentionAttr extends AddBattlerTagAttr {
+  constructor(selfTarget: boolean = true) {
+    super(BattlerTagType.CENTER_OF_ATTENTION, selfTarget, { failOnOverlap: true });
```
MokaStitcher: the `failOnOverlap` wasn't present in init-moves before, but I think it's more correct like this?
I guess it doesn't really matter because a very specific set of moves would have to be used to end up in an overlap situation lol

src/data/moves/move-attrs/charge-attr.ts
```diff
@@ -0,0 +1,29 @@
...
+import type { Pokemon } from "#field/pokemon";
+import { AddBattlerTagAttr } from "#moves/add-battler-tag-attr";
+import type { Move } from "#moves/move";
+
+/**
+ * Attribute to apply the power-boosting effect of {@link https://bulbapedia.bulbagarden.net/wiki/Charge_(move) | Charge}.
+ * Doubles the power of the user's next Electric-type attack.
+ * @extends AddBattlerTagAttr
+ */
+export class ChargeAttr extends AddBattlerTagAttr {
```
MokaStitcher: hmm, that name could create confusion with charging moves attributes like `InstantChargeAttr`.
Charging moves are initialized with something also called `chargeAttr` in init-moves:
`new ChargingAttackMove(MoveId.DIG).chargeAttr(SemiInvulnerableAttr, BattlerTagType.UNDERGROUND),`

Although I think in this case ChargeAttr makes sense, it could also be renamed `ChargedAttr` to match the Battler Tag Type, and hopefully avoid that confusion.
Otherwise the aforementionned attributes and function should maybe be renamed to `xxxChargingAttr` and `.chargingAttr`?

src/data/moves/move-attrs/confuse-attr.ts
```diff
@@ -13,8 +15,12 @@ import i18next from "i18next";
  * @extends AddBattlerTagAttr
  */
 export class ConfuseAttr extends AddBattlerTagAttr {
```
MokaStitcher, line 16: <sub>If ChargeAttr gets renamed ChargedAttr, this also could be renamed ConfusedAttr

src/data/moves/move-attrs/disable-attr.ts
```diff
@@ -0,0 +1,35 @@
...
+import type { Pokemon } from "#field/pokemon";
+import { AddBattlerTagAttr } from "#moves/add-battler-tag-attr";
+import type { Move } from "#moves/move";
+
+/**
+ * Attribute for the effects of {@link https://bulbapedia.bulbagarden.net/wiki/Disable_(move) | Disable}.
+ * Disables the target's last-used move for 4 turns.
+ * @extends AddBattlerTagAttr
+ */
+export class DisableAttr extends AddBattlerTagAttr {
```
MokaStitcher, line 14: `Disabled`?

src/data/moves/move-attrs/drowsy-attr.ts
```diff
@@ -0,0 +1,28 @@
...
+import { AddBattlerTagAttr } from "#moves/add-battler-tag-attr";
+import type { Move } from "#moves/move";
+
+/**
+ * Attribute to apply the {@link https://bulbapedia.bulbagarden.net/wiki/Status_condition#Drowsy | Drowsy}
+ * status condition to the target.
+ * @extends AddBattlerTagAttr
+ */
+export class DrowsyAttr extends AddBattlerTagAttr {
+  constructor(isAttack: boolean = false, effectChanceOverride?: number) {
```
MokaStitcher, line 14: since there are only two uses of DrowsyAttr, is there a reason to go through a `isAttack` parameter rather than simply have a `failOnOverlap` one?

src/data/moves/move-attrs/electrify-attr.ts
```diff
@@ -0,0 +1,37 @@
...
+import type { Pokemon } from "#field/pokemon";
+import { AddBattlerTagAttr } from "#moves/add-battler-tag-attr";
+import type { Move } from "#moves/move";
+
+/**
+ * Attribute to apply {@link https://bulbapedia.bulbagarden.net/wiki/Electrify_(move) | Electrify's}
+ * effect. Converts the target's moves to Electric-type for the rest of the turn.
+ * @extends AddBattlerTagAttr
+ */
+export class ElectrifyAttr extends AddBattlerTagAttr {
```
MokaStitcher, line 19: Still to match the battler tag type, `Electrified`?

src/data/moves/move-attrs/glaive-rush-attr.ts
```diff
@@ -0,0 +1,28 @@
+import { MINOR_EFFECT_SCORE_PENALTY } from "#constants/ai-constants";
+import { BattlerTagType } from "#enums/battler-tag-type";
+import type { EnemyPokemon } from "#field/enemy-pokemon";
+import type { Pokemon } from "#field/pokemon";
+import { AddBattlerTagAttr } from "#moves/add-battler-tag-attr";
+import type { Move } from "#moves/move";
+
+/**
+ * Attribute to apply {@link https://bulbapedia.bulbagarden.net/wiki/Glaive_Rush_(move) | Glaive Rush's}
+ * secondary effect. Causes attacks against the user to deal double damage
```
MokaStitcher:
```suggestion
 * secondary effects. Causes attacks against the user to deal double damage
```

src/data/moves/move-attrs/grudge-attr.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Attribute to apply {@link https://bulbapedia.bulbagarden.net/wiki/Grudge_(move) | Grudge's} effect.
+ * If the user is knocked out by another Pokemon this turn, this
+ * depletes the PP of the attacker's last used move.
+ * @extends AddBattlerTagAttr
+ */
+export class GrudgeAttr extends AddBattlerTagAttr {
+  constructor() {
+    super(BattlerTagType.GRUDGE, true, { turnCountMin: 1, failOnOverlap: true });
```
MokaStitcher, line 16: the failOnOverlap is new, is it correct?

src/data/moves/move-attrs/snatch-attr.ts
```diff
@@ -0,0 +1,35 @@
...
+import { AddBattlerTagAttr } from "#moves/add-battler-tag-attr";
+import type { Move } from "#moves/move";
+
+/**
+ * Attribute to apply {@link https://bulbapedia.bulbagarden.net/wiki/Snatch_(move) | Snatch's} effect.
+ * For the rest of the turn, the user steals the effects of
+ * any {@linkcode MoveFlags.SNATCHABLE | snatchable} moves used by other Pokemon.
+ * @extends AddBattlerTagAttr
+ */
+export class SnatchAttr extends AddBattlerTagAttr {
```
MokaStitcher, line 16: `SnatchingAttr`? (i've probably missed a few others with this issue at this point, will be the last one I point out ^^")

src/data/moves/move-attrs/stockpile-attr.ts
```diff
@@ -0,0 +1,28 @@
...
+import { AddBattlerTagAttr } from "#moves/add-battler-tag-attr";
+import type { Move } from "#moves/move";
+
+/**
+ * Attribute to apply the effects of {@link https://bulbapedia.bulbagarden.net/wiki/Stockpile_(move) | Stockpile}.
+ * Adds a {@linkcode BattlerTagType.STOCKPILING | STOCKPILING} stack to the user,
+ * which increases the user's defenses and powers up their next use of Spit Up or Swallow.
+ * @extends AddBattlerTagAttr
+ */
+export class StockpileAttr extends AddBattlerTagAttr {
```
MokaStitcher, line 15: or not -> `StockpilingAttr`?

src/data/moves/move-attrs/stockpile-attr.ts
```diff
@@ -0,0 +1,28 @@
...
+    super(BattlerTagType.STOCKPILING, true);
+  }
+
+  /**
+   * Grants a {@link MINOR_EFFECT_SCORE_BONUS | minor bonus} for each of
+   * Spit Up and Swallow that the user knows.
+   */
+  public override getRawEffectScore(user: EnemyPokemon, _target: Pokemon, _move: Move): number {
+    const synergyMoves = [MoveId.SPIT_UP, MoveId.SWALLOW];
+    return synergyMoves.reduce((total, mv) => total + (user.hasMove(mv) ? MINOR_EFFECT_SCORE_BONUS : 0), 0);
```
MokaStitcher:
```suggestion
    return synergyMoves.filter((mv) => user.hasMove(mv)).length * MINOR_EFFECT_SCORE_BONUS;
```
seems easier to understand and I think it's the same result? Don't know about efficiency though

src/data/moves/move-attrs/telekinesis-attr.ts
```diff
@@ -0,0 +1,31 @@
...
+import type { Move } from "#moves/move";
+
+/**
+ * Attribute to apply {@link https://bulbapedia.bulbagarden.net/wiki/Telekinesis_(move) | Telikinesis'}
+ * effect. Attacks against the target bypass accuracy checks for 3 turns.
+ * @extends AddBattlerTagAttr
+ */
+export class TelekinesisAttr extends AddBattlerTagAttr {
+  constructor() {
+    super(BattlerTagType.TELEKINESIS, false, { failOnOverlap: true });
```
MokaStitcher, line 14:
```suggestion
    super(BattlerTagType.TELEKINESIS, false, { failOnOverlap: true, turnCountMin: 3 });
```
there was this before in init moves, is it not needed?

src/data/moves/move-attrs/torment-attr.ts
```diff
@@ -0,0 +1,21 @@
...
+import type { Move } from "#moves/move";
+
+/**
+ * Attribute to apply the effects of {@link https://bulbapedia.bulbagarden.net/wiki/Torment_(move) | Torment}.
+ * Prevents the target from using the same move twice in a row.
+ * @extends AddBattlerTagAttr
+ */
+export class TormentAttr extends AddBattlerTagAttr {
+  constructor() {
+    super(BattlerTagType.TORMENT, false, { failOnOverlap: true });
```
MokaStitcher:
```suggestion
  constructor(isAttack = false) {
    super(BattlerTagType.TORMENT, false, { failOnOverlap: !isAttack });
```
For G_MAX_MELTDOWN, assuming this is the same as other similar moves where it shouldn't fail if it's an attacking move
Would it also need the `lastHitOnly` option because of that? In which case
```suggestion
  constructor(isAttack = false) {
    super(BattlerTagType.TORMENT, false, {
      failOnOverlap: !isAttack,
      lastHitOnly: isAttack,
    });
```

src/data/init/init-moves.ts
```diff
@@ -3222,13 +3262,13 @@ export function initMoves() {
       .attr(AddEntryHazardTagAttr, ArenaTagType.SHARP_STEEL),
     new AttackMove(MoveId.G_MAX_MELTDOWN, ElementalType.STEEL, MoveCategory.PHYSICAL, 80, -1, 3, -1, 0, 8)
       .gMaxMove(SpeciesId.MELMETAL)
-      .attr(AddBattlerTagAttr, BattlerTagType.TORMENT),
+      .attr(TormentAttr),
```
MokaStitcher:
```suggestion
      .attr(TormentAttr, true),
```

src/field/pokemon.ts
```diff
@@ -1217,6 +1226,16 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     return Math.floor(ret);
   }
 
+  /**
+   * @param target - The {@linkcode} to compare Speed against
+   * @param estimate - If `true`, estimates the target's Speed only based on revealed information
```
MokaStitcher: Technically effective stats are never "revealed information" but i'm not sure how to phrase it better so
```suggestion
   * @param estimate - If `true`, estimates the target's Speed only based on revealed ability information
```

MokaStitcher (commented):
Still need to look through the tests
Not sure about naming conventions we should be using for the attributes and battler tag types, but it's partly outside the scope of this PR

src/data/abilities/ab-attrs/pre-apply-battler-tag-immunity-ab-attr.ts
innerthunder: Before this change, I ran into a runtime error while running the Yawn scoring test. I guess it not being defined for `getTriggerMessage` now causes issues while evaluating the move?

In any case, this should be harmless. `battlerTag` is only used for `getTriggerMessage`, which simulated calls suppress.

src/data/moves/move-attrs/center-of-attention-attr.ts
innerthunder: Hmm... I assumed as much, but overlapping "center of attention" effects gets weird, especially with Spotlight in the mix. This was added mainly so that Follow Me / Rage Powder will fail instead of give no feedback when the user is already Spotlighted.

src/data/moves/move-attrs/confuse-attr.ts
innerthunder: I prefer `ConfuseAttr` since, unlike `ChargedAttr`, it always confuses a Pokemon other than the user.

<sub>and `ChargeAttr` is only being renamed because of conflicts with other parts of the move code</sub>

src/data/moves/move-attrs/drowsy-attr.ts
innerthunder: A couple reasons I can think of:
- `isAttack` makes the criteria for when `failOnOverlap` is added more apparent
- I'm not sure how we want G-max Snooze to interact with multi-hit enhancers. If it should only apply its effect on the last hit, `isAttack` makes that easier to enforce.

src/data/moves/move-attrs/electrify-attr.ts
innerthunder: My intent was to name these bespoke attributes after their respective moves, not necessarily the tags they apply.

src/data/moves/move-attrs/grudge-attr.ts
innerthunder: We'd basically have to check its interaction with Instruct to be sure, but logically it makes sense to me. Otherwise, the second use of Grudge just plays the move's animation and does nothing after.

src/data/moves/move-attrs/telekinesis-attr.ts
innerthunder: `TelekinesisTag` always sets its turn count to 3, so it's unnecessary to define the turn count here.

src/@types/status.ts
innerthunder: Probably not necessary since it's updated at the same time as the other fields here

test/ai/move-effect-scores/focus-energy.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  it("should be preferred over moves with low-impact", async () => {
+    await game.classicMode.startBattle([SpeciesId.MAGIKARP]);
+
+    const enemy = game.field.getEnemyPokemon();
+    expect(enemy).toPreferSelectingMove(MoveId.FOCUS_ENERGY);
+  });
+
+  it("should have greater incentive when the user has a move with a high critical hit ratio", async () => {
```
MokaStitcher, line 40: I think this test would need a comparison of a moveset without a move with high crit ratio to be able to show that there is greater incentive

test/ai/move-effect-scores/heal-block.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  });
+
+  it("should be preferred over low-impact moves", async () => {
+    await game.classicMode.startBattle([SpeciesId.MAGIKARP]);
+
+    const enemy = game.field.getEnemyPokemon();
+    expect(enemy).toPreferSelectingMove(MoveId.HEAL_BLOCK);
+  });
+
+  it("should be strongly preferred if the opponent has Triage", async () => {
```
MokaStitcher: here we have the same issue, there's no test that shows it not being favored except the "already under heal block's effect" one, so this is not proving that Triage is what made it favored

test/ai/move-effect-scores/powder.test.ts
```diff
@@ -0,0 +1,60 @@
...
+      .battleType("single")
+      .enemySpecies(SpeciesId.MAGIKARP)
+      .enemyAbility(AbilityId.BALL_FETCH)
+      .enemyMoveset([MoveId.POWDER, MoveId.TACKLE, MoveId.SPLASH])
+      .ability(AbilityId.BALL_FETCH)
+      .startingLevel(100)
+      .enemyLevel(100);
+  });
+
+  it("should be preferred when the target is known to have a Fire-type attack", async () => {
```
MokaStitcher, line 34: similar here.
I think playing a first turn without the move revealed, and showing it's not favored, then revealing and showing it's favored would do the trick

test/ai/move-effect-scores/powder.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  it("should be preferred when the target is a Fire-type Pokemon", async () => {
+    await game.classicMode.startBattle([SpeciesId.CHARMANDER]);
+
+    const enemy = game.field.getEnemyPokemon();
+    expect(enemy).toPreferSelectingMove(MoveId.POWDER);
+  });
+
+  it("should not be preferred when the target is known to not have any Fire-type attack", async () => {
```
MokaStitcher, line 53: can a similar thing be done here, or would it already not be preferred on the turn before revealing the moves?

test/ai/move-effect-scores/protect.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      .battleType("single")
+      .enemySpecies(SpeciesId.MAGIKARP)
+      .enemyAbility(AbilityId.BALL_FETCH)
+      .enemyMoveset(baseMoveset)
+      .ability(AbilityId.BALL_FETCH)
+      .startingLevel(100)
+      .enemyLevel(100);
+  });
+
+  it("should be preferred over moves with low impact", async () => {
```
MokaStitcher, line 35: similar thing, no comparison point to a scenario with "high impact" moves

test/ai/move-effect-scores/smack-down.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      .battleType("single")
+      .enemySpecies(SpeciesId.DIGLETT)
+      .enemyAbility(AbilityId.BALL_FETCH)
+      .enemyMoveset([MoveId.SMACK_DOWN, MoveId.BULLDOZE, MoveId.AERIAL_ACE, MoveId.SPLASH])
+      .ability(AbilityId.BALL_FETCH)
+      .startingLevel(100)
+      .enemyLevel(100);
+  });
+
+  it("should be preferred if the user is Ground-type against a Flying-type target", async () => {
```
MokaStitcher: similar thing

test/ai/move-effect-scores/yawn.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      .battleType("single")
+      .enemySpecies(SpeciesId.MAGIKARP)
+      .enemyAbility(AbilityId.BALL_FETCH)
+      .enemyMoveset([MoveId.YAWN, MoveId.TACKLE, MoveId.SPLASH])
+      .ability(AbilityId.BALL_FETCH)
+      .startingLevel(100)
+      .enemyLevel(100);
+  });
+
+  it("should be preferred when the target can be made drowsy", async () => {
```
MokaStitcher, line 35: again

MokaStitcher (commented):
Some nits about tests, otherwise lgtm

test/ai/move-effect-scores/focus-energy.test.ts
innerthunder: A lot of the "greater incentive" tests compare against Super Fang because it's guaranteed 50% damage (assuming the target is at full HP), which translates to an expected Attack Score of 1.25. To be preferred over Super Fang, a move's expected total score needs to be above that threshold.

In this case, Focus Energy's expected ES is only 1.5 when the user has a high crit ratio move, and 0.5 otherwise. It can only surpass Super Fang with a high crit ratio move present.

test/ai/move-effect-scores/powder.test.ts
innerthunder: I changed the player Pokemon to Charmander in this test. That way, we already know that Powder would be preferred based on the results of the "Fire-type target" test

src/data/moves/move-attrs/add-battler-tag-attr.ts
```diff
@@ -63,138 +65,9 @@ export class AddBattlerTagAttr extends ChanceBasedMoveEffectAttr {
...
-  }
-
-  override getTargetBenefitScore(user: Pokemon, target: Pokemon, move: Move): number {
-    let moveChance = this.getMoveChance(user, target, move);
-    if (moveChance < 0) {
-      moveChance = 100;
-    }
-    return Math.floor(this.getTagTargetBenefitScore() * (moveChance / 100));
-  }
+  public abstract override getRawEffectScore(user: EnemyPokemon, target: Pokemon, move: Move): number;
```
Tempo-anon, line 71: Should the target benefit scores (-16, -5, -3, -2, -1, 0, 3, 5) be moved to a constants file with names for each?

src/data/moves/move-attrs/binding-attr.ts
```diff
@@ -0,0 +1,29 @@
...
+    super(tagType, false, { turnCountMin: 4, turnCountMax: 5 });
+  }
+
+  /**
+   * Grants a 65%(+1) bonus if the user has a {@linkcode FAVORABLE_MATCHUP_SCORE_THRESHOLD | favorable matchup score}
+   * against all opponents.
+   */
+  public override getRawEffectScore(user: EnemyPokemon, _target: Pokemon, _move: Move): number {
+    if (user.getOpponents().every((opp) => user.getMatchupScore(opp) > FAVORABLE_MATCHUP_SCORE_THRESHOLD)) {
+      return this.getRandomScore(user, 65);
```
Tempo-anon, line 25: 65 is a magic number here

src/data/moves/move-attrs/charged-attr.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  /** Grants 30%(+2) if the user knows an Electric-type attack and isn't already "Charged" */
+  public override getRawEffectScore(user: EnemyPokemon, _target: Pokemon, _move: Move): number {
+    const userKnowsElectricAttack = user
+      .getAttackMoves(true)
+      .some((mv) => user.getMoveType(mv) === ElementalType.ELECTRIC);
+
+    return userKnowsElectricAttack && !user.hasTag(BattlerTagType.CHARGED)
+      ? this.getRandomScore(user, 30, MAJOR_EFFECT_SCORE_BONUS)
```
Tempo-anon, line 26: 30 is another magic number here and something I expect might be changed in the future since charge is a move that is unique and interesting in concept but in practice seems pretty useless.

src/data/moves/move-attrs/confuse-attr.ts
```diff
@@ -13,8 +15,12 @@ import i18next from "i18next";
...
-  constructor(selfTarget?: boolean) {
-    super(BattlerTagType.CONFUSED, selfTarget, { turnCountMin: 2, turnCountMax: 5 });
+  constructor(failOnOverlap: boolean = false) {
```
Tempo-anon: This is missing the current implementation in beta of accounting for axe kick

src/data/moves/move-attrs/crit-boost-stackable-attr.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Unlike Focus Energy's effect, this effect may stack with itself.
+ * @extends AddBattlerTagAttr
+ */
+export class CritBoostStackableAttr extends AddBattlerTagAttr {
+  constructor() {
+    super(BattlerTagType.CRIT_BOOST_STACKABLE, true);
+  }
+
+  /**
+   * Has a 30% chance to grant (+1).
```
Tempo-anon, line 20: Should this be a % to increase score? The boost is guaranteed so this should always be a positive unless the user is already at +3 crit or a high enough crit stage that it won't affect their moves.

src/data/moves/move-attrs/dragon-cheer-attr.ts
```diff
@@ -0,0 +1,36 @@
...
+   * Grants a 55%(+1) bonus.
+   * Also grants a {@link MINOR_EFFECT_SCORE_BONUS | minor bonus} if the target is Dragon-type.
+   * Also grants a minor bonus if the target has a move with an increased critical hit ratio.
+   */
+  public override getRawEffectScore(user: EnemyPokemon, target: Pokemon, _move: Move): number {
+    const targetIsDragon = target.isOfType(ElementalType.DRAGON, true, true);
+    const targetHasHighCrit = target.getAttackMoves(true).some((mv) => mv.hasAttr(HighCritAttr));
+
+    return (
+      this.getRandomScore(user, 55)
```
Tempo-anon: 55% feels like a weird number since it's not a nice round multiple of 10

src/data/moves/move-attrs/electrify-attr.ts
```diff
@@ -0,0 +1,37 @@
...
+  constructor() {
+    super(BattlerTagType.ELECTRIFIED, false, { failOnOverlap: true });
+  }
+
+  /**
+   * Grants 60%(+2) if the user or its ally has an immunity to Electric-type moves,
+   * either from its typing or ability. Otherwise, grants a {@link MINOR_EFFECT_SCORE_PENALTY | minor penalty}
+   */
```
Tempo-anon: I feel like there should be a major penalty if the target is faster than the user. Maybe also some benefit for the user/partner being resistant rather than completely immune to electric?

src/data/moves/move-attrs/focus-energy-attr.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Increases the user's critical hit ratio by 2 stages.
+ * @extends AddBattlerTagAttr
+ */
+export class FocusEnergyAttr extends AddBattlerTagAttr {
+  constructor() {
+    super(BattlerTagType.CRIT_BOOST, true, { failOnOverlap: true });
+  }
+
+  /**
+   * Has a 50% chance to grant (+1).
```
Tempo-anon: Since this is not stackable, the move should not be used at all if the user already has the effect

src/data/moves/move-attrs/heal-block-attr.ts
```diff
@@ -0,0 +1,32 @@
...
+  constructor(isAttack: boolean = false, turnCount: number = 5) {
+    super(BattlerTagType.HEAL_BLOCK, false, {
+      failOnOverlap: !isAttack,
+      turnCountMin: turnCount,
+    });
+  }
+
+  /**
+   * Grants 40%(+1), with an additional {@link MINOR_EFFECT_SCORE_BONUS | minor bonus}
+   * if the target has the ability {@link AbilityId.TRIAGE | Triage}.
```
Tempo-anon: Does this account for if the target is already heal blocked?

src/data/moves/move-attrs/add-battler-tag-attr.ts
innerthunder: The common scores are already set in `constants/ai-constants.ts`

src/data/moves/move-attrs/binding-attr.ts
innerthunder: This is a pretty common pattern... Should these chances be set as a `const` within this method?

src/data/moves/move-attrs/focus-energy-attr.ts
innerthunder: `failOnOverlap` being `true` already covers this via the move's condition score; there's also a test for it in `move-effect-scores/focus-energy.test.ts`

Tempo-anon: Oh cool, I had this question for a few other instances of failing on overlap so good to know it's covered

src/data/moves/move-attrs/binding-attr.ts
Tempo-anon: Oh I think that might be a good idea. I was thinking of either having it as a constant like 
```ts
public override getRawEffectScore(user: EnemyPokemon, _target: Pokemon, _move: Move): number {
  const chanceToAddScore = 65;
```

or maybe just adding it to the method params (with the default being 100 otherwise)
```ts
public override getRawEffectScore(user: EnemyPokemon, _target: Pokemon, _move: Move, chance: number = 65): number {
```

innerthunder: I personally don't think it adds much to the code's readability. `getRandomScore`'s docs already explain the meaning of the chance value well enough. If others feel strongly about including them, I'll probably go with your first option

src/data/moves/move-attrs/confuse-attr.ts
innerthunder: Will be fixed once beta is merged into `ai-rework` (after this and #1258)

src/data/moves/move-attrs/dragon-cheer-attr.ts
innerthunder: I can change it to 50 for now. The specific score chances are fairly rough; they can always be adjusted later.

src/data/moves/move-attrs/heal-block-attr.ts
innerthunder: It does now (this wasn't covered for Psychic Noise specifically)

src/data/moves/move-attrs/binding-attr.ts
DayKev: Given that everything is already documented I don't think we need a `const` or anything like that.

src/data/moves/move-attrs/crit-boost-stackable-attr.ts
DayKev: It should probably not be incentivized to use the move if the pokemon is already at max crit stages, though I'm not sure how easy it is to check that.

src/field/pokemon.ts
```diff
@@ -448,6 +448,15 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     return this.status?.toxicTurnCount ?? 0;
   }
 
+  /**
+   * The number of turns this Pokemon has slept since afflicted with the
+   * {@linkcode StatusEffect.SLEEP | SLEEP} condition.
+   * @defaultValue 0
+   */
+  public get turnsAsleep(): number {
+    return this.status?.turnsAsleep ?? -1;
```
DayKev: Is the default supposed to be `0` or `-1`?

src/field/pokemon.ts
```diff
@@ -1217,6 +1226,16 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     return Math.floor(ret);
   }
 
+  /**
+   * @param target - The {@linkcode} to compare Speed against
```
```suggestion
   * @param target - The {@linkcode Pokemon} to compare Speed against
```

src/data/moves/move-attrs/crit-boost-stackable-attr.ts
```diff
@@ -21,9 +20,17 @@ export class CritBoostStackableAttr extends AddBattlerTagAttr {
    * Has a 30% chance to grant (+1).
    * Grants an additional (+1) if the user has a move with an increased critical hit ratio
    */
-  public override getRawEffectScore(user: EnemyPokemon, _target: Pokemon, _move: Move): number {
-    const userHasHighCritMove = user.getMoveset().some((mv) => mv.getMove().hasAttr(HighCritAttr));
+  public override getRawEffectScore(user: EnemyPokemon, target: Pokemon, _move: Move): number {
+    const maxCritStage = user.getMoveset().reduce((maxStage, mv) => {
+      /** @todo The active ally of `target` isn't accounted for here */
+      const critStage = target.getCritStage(user, mv.getMove(), true);
+      return critStage > maxStage ? critStage : maxStage;
```
DayKev:
```suggestion
      return Math.max(critStage, maxStage);
```

src/data/moves/move-attrs/focus-energy-attr.ts
```diff
@@ -21,8 +20,17 @@ export class FocusEnergyAttr extends AddBattlerTagAttr {
    * Grants an additional (+1) if the user has a move with an increased critical hit ratio.
    */
   public override getRawEffectScore(user: EnemyPokemon, _target: Pokemon, _move: Move): number {
-    const userHasHighCritMove = user.getMoveset().some((mv) => mv.getMove().hasAttr(HighCritAttr));
+    const opponent = user.getOpponents()[0];
+    /** @todo This is only based on the user's first opponent */
+    const maxCritStage = user.getMoveset().reduce((maxStage, mv) => {
+      const critStage = opponent?.getCritStage(user, mv.getMove(), true) ?? 0;
+      return critStage > maxStage ? critStage : maxStage;
```
DayKev:
```suggestion
      return Math.max(critStage, maxStage);
```

src/data/moves/move-attrs/faint-countdown-attr.ts
```diff
@@ -32,4 +35,14 @@ export class FaintCountdownAttr extends AddBattlerTagAttr {
...
   }
+
+  /**
+   * Grants a {@link MINOR_EFFECT_SCORE_BONUS | minor bonus} if the target is trapped by any effect.
+   * If the target is allied to the user, this bonus is multiplied by -1.
+   * @todo `allyTargetMultiplier` should be used more widely for multi-target moves
+   */
+  public override getRawEffectScore(user: EnemyPokemon, target: Pokemon, _move: Move): number {
+    const allyTargetMultiplier = target.isOpponent(user) ? 1 : -1;
+    return allyTargetMultiplier * (target.isTrapped() ? MINOR_EFFECT_SCORE_BONUS : 0);
```
DayKev, line 46: Hm, does this pretend that the move is single-target or is this run on every other pokemon on the field?

src/data/moves/move-attrs/floating-attr.ts
```diff
@@ -0,0 +1,41 @@
...
+   */
+  public override getRawEffectScore(user: EnemyPokemon, target: Pokemon, _move: Move): number {
+    const pokemon = this.selfTarget ? user : target;
+    if (!pokemon.isGrounded()) {
+      return 0;
+    }
+
+    const numGroundTypeOpponents = pokemon
+      .getOpponents()
+      .filter((opp) => opp.isOfType(ElementalType.GROUND, true)).length;
```
DayKev: Shouldn't it check for the opponent having ground type moves (which will also effectively include a check for the opponent being a ground type due to the virtual simulated moves being the same type as the pokemon's type)?
Same question for similar attrs (like `GroundingAttr`).

src/data/moves/move-attrs/heal-block-attr.ts
```diff
@@ -0,0 +1,38 @@
...
+   */
```
DayKev: Should it consider if the opponent has any healing moves (Recover, Bitter Blade, etc)?

src/data/moves/move-attrs/trap-attr.ts
```diff
@@ -1,14 +1,37 @@
...
+    });
+  }
+
+  /**
+   * Grants 60%(+1) if the user has a {@link STRONG_MATCHUP_SCORE_THRESHOLD | strong matchup}
+   * against all opponents.
+   * @todo Should this only be granted in single battles?
+   */
+  public override getRawEffectScore(user: EnemyPokemon, _target: Pokemon, _move: Move): number {
+    const userMatchupScores = user.getOpponents().map((opp) => user.getMatchupScore(opp));
```
DayKev, line 28: Can't you use `Pokemon#getAverageMatchupScore`?

src/data/moves/move-attrs/laser-focus-attr.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  /**
+   * Grants 40%(+1) if the user has a move of at least moderate
+   * {@linkcode Pokemon.getExpectedAttackScore | EAS} against an opponent
+   */
+  public override getRawEffectScore(user: EnemyPokemon, _target: Pokemon, _move: Move): number {
+    const opponents = user.getOpponents();
+    const userHasThreatMove = user
+      .getAttackMoves(true)
+      .some((attack) => opponents.some((opp) => user.getExpectedAttackScore(opp, attack) >= 1));
```
DayKev, line 27: If the user has a move that will KO the opponent, there's probably no reason to use this move to force a crit, so maybe it should be `eas >= 1 && eas < ko_score`?

src/data/moves/move-attrs/leech-seed-attr.ts
```diff
@@ -10,4 +14,17 @@ export class LeechSeedAttr extends AddBattlerTagAttr {
   constructor() {
     super(BattlerTagType.SEEDED);
   }
+
+  /**
+   * Grants a 70%(+1) bonus.
```
DayKev, line 18: Maybe it should be 100% to give a +1 if the opponent is trapped?
Or just a flat +1 if the opponent is trapped, on top of the base 70% for a +1.

src/data/moves/move-attrs/no-retreat-attr.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Attribute to apply {@link https://bulbapedia.bulbagarden.net/wiki/No_Retreat_(move) | No Retreat's}
+ * self-trapping effect.
+ * @extends AddBattlerTagAttr
+ */
+export class NoRetreatAttr extends AddBattlerTagAttr {
+  constructor() {
+    super(BattlerTagType.NO_RETREAT, true);
+  }
+
+  /** Grants a {@link MINOR_EFFECT_SCORE_PENALTY | minor penalty} if the user isn't already trapped. */
```
DayKev: I wonder if the penalty should also be negated if the user has a good MUS vs the opponent?

src/data/moves/move-attrs/protect-attr.ts
```diff
@@ -24,17 +31,62 @@ export class ProtectAttr extends AddBattlerTagAttr {
...
-      return !user.randSeedInt(Math.pow(3, lastNonUse));
-    };
+  public override getCondition(): MoveCondition {
+    return new ProtectCondition();
+  }
+
+  /**
+   * The Effect Score bonus from this effect consists of 3 components, and only applies
+   * if the user did not use a "protection" move successfully last turn:
+   * - A flat {@link MINOR_EFFECT_SCORE_BONUS | minor bonus}
```
DayKev, line 40: Should the AI be so eager to use basic Protect?

src/data/moves/move-attrs/quash-attr.ts
```diff
@@ -1,18 +1,33 @@
...
+ * Attribute to apply {@link https://bulbapedia.bulbagarden.net/wiki/Quash_(move) | Quash's}
+ * effect. Forces the target to move last in turn order.
+ * @extends AddBattlerTagAttr
+ */
 export class QuashAttr extends AddBattlerTagAttr {
   constructor() {
     super(BattlerTagType.QUASHED, false, { failOnOverlap: true });
   }
 
+  /** @todo Should this be converted to a scored {@linkcode MoveCondition}? */
```
DayKev, line 21: If the condition fails then the move automatically gets a score penalty doesn't it?

src/data/moves/move-attrs/taunt-attr.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Prevents the target from selecting or using Status moves for 4 turns.
+ * @extends AddBattlerTagAttr
+ */
+export class TauntAttr extends AddBattlerTagAttr {
+  constructor() {
+    super(BattlerTagType.TAUNT, false, { failOnOverlap: true, turnCountMin: 4 });
+  }
+
+  /**
+   * Grants (+1) + 60%(+1) if the target isn't expected to have any threatening attacks
```
DayKev, line 19: Should the AI be so eager to use Taunt?

src/data/moves/move-attrs/exposed-move-attr.ts
```diff
@@ -32,4 +36,26 @@ export class ExposedMoveAttr extends AddBattlerTagAttr {
 
     return true;
   }
+
+  /**
+   * Grants an Effect Score bonus based on the following:
+   * - If the user or its ally has a move that would bypass a type immunity from the target
+   * because of this effect, this grants a {@link MINOR_EFFECT_SCORE_BONUS | minor bonus}.
+   * - Otherwise, if the user or its ally has a move with less than 80 base accuracy, this grants 50%(+1).
```
DayKev: Foresight et al only cause moves used against the target to ignore the target's positive evasion stat stages, so instead of checking for low accuracy moves for this boost it should check if the target has any evasion stage boosts.

src/data/moves/move-attrs/uproar-attr.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  /**
+   * Grants (-1) + (# of asleep allies) - (# of asleep enemies).
+   * The total score from this attribute cannot be lower than (-2).
+   */
+  public override getRawEffectScore(user: EnemyPokemon, _target: Pokemon, _move: Move): number {
+    const otherAsleepPokemon = globalScene
+      .getField(true)
+      .filter((p) => p !== user && p.hasStatusEffect(StatusEffect.SLEEP));
```
DayKev:
```suggestion
      .filter((p) => p !== user && p.hasStatusEffect(StatusEffect.SLEEP, false, true));
```

src/data/moves/move-attrs/add-battler-tag-if-boosted-attr.ts
```diff
@@ -1,18 +1,29 @@
...
  * Attribute to apply a battler tag to the target if they have had their stats boosted this turn.
  * @extends AddBattlerTagAttr
  */
-export class AddBattlerTagIfBoostedAttr extends AddBattlerTagAttr {
-  constructor(tag: BattlerTagType) {
-    super(tag, false, { turnCountMin: 2, turnCountMax: 5 });
+export class ConfuseIfBoostedAttr extends ConfuseAttr {
+  constructor() {
+    super(false);
   }
```
DayKev: The default is already `false`.
```suggestion
```

src/data/moves/move-attrs/confuse-attr.ts
```diff
@@ -29,4 +35,25 @@ export class ConfuseAttr extends AddBattlerTagAttr {
...
+   * @see {@linkcode getEffectScore}
+   */
+  public override getRawEffectScore(user: EnemyPokemon, target: Pokemon, move: Move): number {
+    if (target.isSafeguarded(user)) {
+      /**
+       * Status moves "fail" if this is the case, but since Safeguard is checked
+       * in {@linkcode applyEffect}, it needs to be accounted for here instead
+       * of within the move's Condition Score calculation.
+       * @todo This can be removed if the Safeguard check is reorganized into a condition
+       */
```
DayKev: This is not a TSDoc.
```suggestion
      /*
       * Status moves "fail" if this is the case, but since Safeguard is checked
       * in `applyEffect`, it needs to be accounted for here instead
       * of within the move's Condition Score calculation.
       * TODO: This can be removed if the Safeguard check is reorganized into a condition
       */
```

src/data/moves/move-attrs/telekinesis-attr.ts
```diff
@@ -0,0 +1,31 @@
+import { BattlerTagType } from "#enums/battler-tag-type";
+import type { EnemyPokemon } from "#field/enemy-pokemon";
+import type { Pokemon } from "#field/pokemon";
+import { AddBattlerTagAttr } from "#moves/add-battler-tag-attr";
+import type { Move } from "#moves/move";
+
+/**
+ * Attribute to apply {@link https://bulbapedia.bulbagarden.net/wiki/Telekinesis_(move) | Telikinesis'}
```
DayKev:
```suggestion
 * Attribute to apply {@link https://bulbapedia.bulbagarden.net/wiki/Telekinesis_(move) | Telekinesis'}
```

DayKev (commented):
image

src/data/moves/move-attrs/faint-countdown-attr.ts
innerthunder: It's run for all Pokemon on the field (including the user itself)

src/data/moves/move-attrs/quash-attr.ts
innerthunder: Yes, but as written the condition is always met during AI move selection. The `QUASHED` tag doesn't persist between turns, and the turn manager is populated with the player's commands before `EnemyCommandPhase`

src/data/moves/move-attrs/protect-attr.ts
innerthunder: Well, the expected effect score for Protect under this formula is 1.3 (or 2 if the user has Moody/Speed Boost/etc.). For reference, this is slightly better than Super Fang against a full-HP target (50% HP damage -> 1.25 EAS).  This seems reasonable to me in early turns when the AI may want to collect information on the players' moves, but it may also become easy to exploit once the enemy reveals it has Protect.

src/data/moves/move-attrs/taunt-attr.ts
innerthunder: In practice the target "not having threatening attacks" is a steep condition. I think it's reasonable for the AI to assume the  target intends to use status moves either to set up or provide utility if that's the case.

On the other hand, I feel like this could be improved by also checking if the user outspeeds the target. Maybe the flat bonus should only apply if that's the case?

src/data/moves/move-attrs/faint-countdown-attr.ts
DayKev: So the results get added together?

innerthunder: Yes, e.g. if one opponent and one ally are trapped, the total bonus will be 0.

src/data/moves/move-attrs/laser-focus-attr.ts
innerthunder: I suppose, but this is already implied considering the KO move will always be selected over this move.

src/data/moves/move-attrs/leech-seed-attr.ts
innerthunder: I'll just add it as an alternate condition for the "defensively favored" bonus

src/data/moves/move-attrs/trap-attr.ts
innerthunder: This checks that all opponents are above the threshold, so the behavior would change

innerthunder:
Moving to draft until #1283 is merged

innerthunder:
image

Tests are now passing again 😄

MokaStitcher:
> image
> Tests are now passing again 😄

Nice!
Would this also solve the issues mentioned in the tera PR or is it unrelated?

innerthunder:
> Nice! Would this also solve the issues mentioned in the tera PR or is it unrelated?

Yes, there was an issue with `Pokemon.getMoveset` returning empty arrays that was introduced in #1258 and fixed here (see https://github.com/Despair-Games/poketernity/pull/1219/commits/af93ea3f2a225d00d68a557fb6ba24bcb49481cd).

src/data/moves/move-attrs/crit-boost-stackable-attr.ts
```diff
@@ -21,9 +20,17 @@ export class CritBoostStackableAttr extends AddBattlerTagAttr {
    * Has a 30% chance to grant (+1).
```
MokaStitcher, line 20: Seems like this doc is outdated

src/data/moves/move-attrs/dragon-cheer-attr.ts
```diff
@@ -25,12 +24,21 @@ export class DragonCheerAttr extends AddBattlerTagAttr {
    */
   public override getRawEffectScore(user: EnemyPokemon, target: Pokemon, _move: Move): number {
```
MokaStitcher, line 25: Can't comment on it but the doc is outdated as well

src/data/moves/move-attrs/floating-attr.ts
```diff
@@ -34,7 +34,7 @@ export class FloatingAttr extends AddBattlerTagAttr {
 
     const numGroundTypeOpponents = pokemon
```
MokaStitcher: Should it be `numOpponentsWithGroundMoves`?
 the doc should probably be updated as well

src/data/moves/move-attrs/focus-energy-attr.ts
```diff
@@ -21,8 +20,17 @@ export class FocusEnergyAttr extends AddBattlerTagAttr {
    * Grants an additional (+1) if the user has a move with an increased critical hit ratio.
    */
   public override getRawEffectScore(user: EnemyPokemon, _target: Pokemon, _move: Move): number {
```
MokaStitcher, line 22: Outdated doc

src/data/moves/move-attrs/no-retreat-attr.ts
```diff
@@ -17,6 +17,7 @@ export class NoRetreatAttr extends AddBattlerTagAttr {
 
   /** Grants a {@link MINOR_EFFECT_SCORE_PENALTY | minor penalty} if the user isn't already trapped. */
```
MokaStitcher: Outdated doc

src/data/moves/move-conditions/first-move-condition.ts
```diff
@@ -8,4 +8,4 @@ import type { MoveConditionFunc } from "#types/move-condition-func";
  * @param _move - (Unused) The {@linkcode Move} being used
  * @returns `true` if the condition is met
  */
-export const firstMoveCondition: MoveConditionFunc = (user, _target, _move) => user.summonData.waveTurnCount <= 1;
+export const firstMoveCondition: MoveConditionFunc = (user, _target, _move) => user.summonData.waveTurnCount === 1;
```
MokaStitcher: Could this interact weirdly with ME mons who get a free turn before the battle starts?

src/data/moves/move-conditions/protect-condition.ts
```diff
@@ -32,7 +32,7 @@ const protectCondition: MoveConditionFunc = (user, _target, _move) => {
   const moveHistory = user.getLastXMoves(-1).filter((mv) => !mv.virtual);
   // Note: This can't check for `ProtectAttr` directly as it would create a circular dependency
   const lastNonUse = moveHistory.findIndex(
```
MokaStitcher, line 44: Can this be renamed, I'm confused what this checks

MokaStitcher (commented):
Some raw effect score functions have outdated doc after being changed

src/data/moves/move-conditions/protect-condition.ts
DayKev: It's supposed to be the index of the last used move that failed or isn't a Protect move. Not sure how to rename it though without it being ridiculously named. Maybe just adding a TSDoc would be enough, since it's only used nearby?

src/data/moves/move-conditions/first-move-condition.ts
innerthunder: As far as I know, those are just free attacks – they don't update the wave turn counter

MokaStitcher (approved):

PR #1222, opened 2025-05-26
[AI-Rework] Remove `appliesScoreOnKO` option

innerthunder:

N/A

While implementing scores, I found that it was generally safe to assume `appliesScoreOnKO` has a 1:1 correspondence with self-targeted effects (that is, an effect applies its scoring on KO if and only if the effect is self-targeted). With the option removed, effect-scoring code should be less error-prone.

- Removed `MoveAttrOptions.appliesScoreOnKO` and the getter `MoveAttr.appliesScoreOnKO`
- Attributes are now filtered via their `selfTarget` property when the move is expected to KO the target
- Updated some docs for `Move`'s new scoring functions.

## How to test the changes?

- [x] `npm run typecheck`
- [x] `npm run depcruise`
- [x] `npm run biome` -- no new errors or warnings

## Checklist

  - [n/a] Have I created new automated tests (`npm run test:create`) or updated existing tests related to the PR's changes?

### DIFF
--- a/src/data/moves/move.ts
+++ b/src/data/moves/move.ts
@@ -768,10 +768,10 @@ export abstract class Move implements Localizable {
    * This means that the ES for an attack should rarely exceed (+1) and almost
    * never exceed (+2). Status moves may have a higher ES in comparison but
    * should typically be limited to (+3) or lower.
-   * @param user the {@linkcode Pokemon} using the move
-   * @param target the {@linkcode Pokemon} targeted by the move
-   * @param isKnockOut `true` if the move is already known to KO the target (default `false`)
-   * @param isFail `true` if the move is already known to fail or have no effect (default `false`)
+   * @param user - The {@linkcode Pokemon} using the move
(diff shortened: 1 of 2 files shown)


src/data/moves/move.ts
```diff
@@ -834,7 +838,7 @@ export abstract class Move implements Localizable {
 
     let attrs: MoveAttr[] = this.attrs;
     if (isKnockOut) {
-      attrs = attrs.filter((attr) => attr.appliesScoreOnKO);
+      attrs = attrs.filter((attr) => attr.selfTarget);
```
DayKev, line 841: So if a move would knock out the target, only self targeting attrs are scored?

innerthunder: Yes; this reflects the logic of `MoveEffectAttr.canApply`. Non-`MoveEffect` attributes are usually taken into account during damage calcs.

DayKev (approved):

PR #1223, opened 2025-05-26
[Dev] Add tag-based deployment

f-fsantos:
Enables tag-based deploys.

### DIFF
--- a/.github/workflows/deploy-tag.yml
+++ b/.github/workflows/deploy-tag.yml
@@ -33,4 +33,5 @@ jobs:
           ssh-keyscan -H ${{ secrets.BETA_SSH_HOST }} >> ~/.ssh/known_hosts
       - name: Deploy build on server
         run: |
+          ssh ${{ secrets.BETA_SSH_USER }}@${{ secrets.BETA_SSH_HOST }} "mkdir -p ${{ secrets.BETA_DESTINATION_DIR }}/poketernity/${{ env.TAG_NAME }}/"
           rsync --del --no-times --checksum -vrm dist/* ${{ secrets.BETA_SSH_USER }}@${{ secrets.BETA_SSH_HOST }}:${{ secrets.BETA_DESTINATION_DIR }}/poketernity/${{ env.TAG_NAME }}/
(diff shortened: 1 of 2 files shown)


innerthunder (approved):
https://poketernity.com/test/

Deployment seems stable 👍

flx-sta (approved):
Can’t wait for this having to be updated for pnpm 😅

MokaStitcher:
That's awesome 🚀 
Couple questions:
- will there be a way to automatically delete those deployments as well?
- is anyone able to push tags right now, or only members of the org? (the latter would probably be idéal just in case)

Wondering if we should force a simple tag pattern like `dev-whatever` or `beta-whatever` to be able to exclude tags like the release tags

DayKev (approved):
Isn't `build` supposed to create a `manifest.json`? Or is it not working because the app expects to be at the root directory and you're using subdirectories instead of subdomains? Is it not possible to automatically create subdomains?
image

f-fsantos:
> That's awesome 🚀 Couple questions:
> 
> * will there be a way to automatically delete those deployments as well?
> * is anyone able to push tags right now, or only members of the org? (the latter would probably be idéal just in case)
> 
> Wondering if we should force a simple tag pattern like `dev-whatever` or `beta-whatever` to be able to exclude tags like the release tags

Deployment cleanup is an issue, but only limited by the server storage. I need to figure out what the limit is before imposing something too lax or too strict. Probably has to be solved server-side.. Discussion to follow on a separate thread.

We should only allow org members to push tags, regardless of the feature. Releases are a good point, we might want to either exclude those, or filter the tag as per your suggestion.

@MokaStitcher

f-fsantos:
> Isn't `build` supposed to create a `manifest.json`? Or is it not working because the app expects to be at the root directory and you're using subdirectories instead of subdomains? Is it not possible to automatically create subdomains? image

@DayKev 

I thought subdomains were going to be harder to handle from a certificate standpoint, [but that seems to not be entirely the case.](https://serverfault.com/a/1012254)

I will take some extra time to see what would change to work with subdomains instead. Independently of that, we might still face issues due to the server-side file structure organization.

MokaStitcher:
> Releases are a good point, we might want to either exclude those, or filter the tag as per your suggestion.

Do you want to look at that for this PR? If not we can merge it probably

PR #1224, opened 2025-05-26
[P2][Refactor] Update+docs `BattlerTag` methods, fix Intimidate

DayKev:
Intimidate will activate properly if one of the opposing Pokémon has a substitute out but not the other.

General code cleanup.

- Added TSDocs and `public` to all the `BattlerTag`-related methods in the `Pokemon` class.
- `Pokemon#findTags` can now take a type argument.
- Replaced many uses of `findTag` and `getTag` with `hasTag`.
- Added some groupings of tags as constants and replaced various `instanceof BattlerTagClass` with `isType<BattlerTagClass>(...BATTLER_TAG_GROUP)`.
- Replaced a mistaken early return in Intimidate's `AbAttr` with `continue`.
- Dragon Cheer's `BattlerTag` class now properly uses `BattlerTagType.DRAGON_CHEER` instead of `BattlerTagType.CRIT_BOOST`.
- Removed some unnecessary `if` checks when lapsing tags.
- Removed Leech Seed tag lapsing from `resetSummonData` (why was that ever there in the first place...?).

## Checklist


test/moves/whirlwind.test.ts
```diff
@@ -42,7 +42,7 @@ describe("Moves - Whirlwind", () => {
 
     await game.toEndOfTurn();
 
-    expect(staraptor.findTag((t) => t.tagType === BattlerTagType.FLYING)).toBeDefined();
+    expect(staraptor.hasTag(BattlerTagType.FLYING)).toBeTruthy();
```
Tempo-anon: totally nit: the name of the battler tag is a bit confusing. Maybe something like IN_AIR or MIDAIR

Tempo-anon (dismissed):
Is there a difference between `return super.end();` and `super.end(); return;`?

Change (fix) for intimidate should be in the PR title

DayKev:
> Is there a difference between `return super.end();` and `super.end(); return;`?

Not really. One of the Biome recommended rules prefers the latter https://biomejs.dev/linter/rules/no-void-type-return/ so I figured might as well.

test/moves/whirlwind.test.ts
DayKev: "Midair" seems good.

DayKev: I'll rename it in my other PR so I don't have to fix conflicts lol.

Tempo-anon: I also made #1229 :>

src/field/pokemon.ts
```diff
@@ -3466,8 +3471,14 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
 
-  canAddTag(tagType: BattlerTagType): boolean {
-    if (this.getTag(tagType)) {
+  // #region Battler Tag methods
+
+  /**
+   * @param tagType - The {@linkcode BattlerTagType | type of tag} to check
+   * @returns Whether the specified tag can be added to the pokemon
+   */
+  public canAddTag(tagType: BattlerTagType): boolean {
```
innerthunder, line 3488: There are a few issues/questions with this worth looking into for this PR or a follow-up:
- This doesn't check the respective tag's `canAdd` function, which is unintuitive given the method's name. It seems like extracting the stub's specific type for this would be difficult, though.
- Abilities are always simulated. Is there any case where they shouldn't be simulated?
- Do the battler tag immunity abilities actually need a full tag to determine immunity? Would these work if they were only given the tag type?

src/field/pokemon.ts
```diff
@@ -3523,28 +3542,52 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
 
-  findTag<T extends BattlerTag = BattlerTag>(tagFilter: (tag: BattlerTag) => boolean): T | undefined {
-    return this.summonData.tags.find((t) => tagFilter(t)) as T | undefined;
+  /**
+   * Filters the {@linkcode BattlerTag}s attached to the pokemon and returns the first result
+   * @param tagFilter - The filter to apply to the tag list
+   * @returns The first {@linkcode BattlerTag} found, or `undefined` if none is found
+   */
+  public findTag<T extends BattlerTag = BattlerTag>(tagFilter: (tag: T) => boolean): T | undefined {
+    return this.summonData.tags.find((t: T) => tagFilter(t)) as T | undefined;
```
```suggestion
    return this.summonData.tags.find(tagFilter) as T | undefined;
```
I also wonder if `tagFilter` could be made into a type guard to avoid the casting, but I feel like it wouldn't be so simple...

src/field/pokemon.ts
```diff
@@ -3523,28 +3542,52 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
 
-  findTags(tagFilter: (tag: BattlerTag) => boolean): BattlerTag[] {
-    return this.summonData.tags.filter((t) => tagFilter(t));
+  /**
+   * Filters the {@linkcode BattlerTag}s attached to the pokemon and returns the result
+   * @param tagFilter - The filter to apply to the tag list
+   * @returns An array of {@linkcode BattlerTag}s matching the input filter
+   */
+  public findTags<T extends BattlerTag = BattlerTag>(tagFilter: (tag: T) => boolean): T[] {
+    return this.summonData.tags.filter((t: T) => tagFilter(t)) as T[];
```
```suggestion
    return this.summonData.tags.filter(tagFilter) as T[];
```

src/field/pokemon.ts
```diff
@@ -3554,7 +3597,12 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     return !!tag;
   }
 
-  lapseTags(lapseType: BattlerTagLapseType): void {
+  /**
+   * Lapses all {@linkcode BattlerTag}s attached to the pokemon with the specified lapse type
+   * and then removes the ones that should be removed
+   * @param lapseType - The {@linkcode BattlerTagLapseType} to trigger
```
innerthunder, line 3610:
```suggestion
   * @param lapseType - The {@linkcode BattlerTagLapseType} to trigger
   * @see {@linkcode BattlerTag.lapse}
```

src/field/pokemon.ts
```diff
@@ -3568,7 +3616,12 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
       });
   }
 
-  removeTag(tagType: BattlerTagType): boolean {
+  /**
+   * Removes a {@linkcode BattlerTag} from the pokemon without calling its `lapse()` method
+   * @param tagType - The {@linkcode BattlerTagType} to remove
+   * @returns Whether the specified tag was attached to the pokemon before being removed
```
innerthunder, line 3630:
```suggestion
   * @returns Whether the specified tag was attached to the pokemon before being removed
   * @see {@linkcode BattlerTag.onRemove}
```

src/field/pokemon.ts
```diff
@@ -3578,26 +3631,42 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+  /**
+   * Removes all {@linkcode BattlerTag}s matching the input filter without calling their `lapse()` methods
+   * @param tagFilter - The filter to apply
+   */
+  public findAndRemoveTags(tagFilter: (tag: BattlerTag) => boolean): void {
     const tags = this.summonData.tags;
     const tagsToRemove = this.findTags(tagFilter);
     for (const tag of tagsToRemove) {
-      tag.turnCount = 0;
+      tag.turnCount = 0; // TODO: is this necessary?
```
innerthunder: I highly doubt this is necessary

src/field/pokemon.ts
```diff
@@ -3648,9 +3717,12 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-        return (tag as MoveRestrictionBattlerTag) !== null;
+    const restrictingTags = this.findTags<MoveRestrictionBattlerTag>((t) =>
+      t.isType<MoveRestrictionBattlerTag>(...RESTRICTING_TAG_TYPES),
+    );
+    for (const tag of restrictingTags) {
+      if (tag.isMoveTargetRestricted(moveId, user, target)) {
+        return tag !== null; // TODO: this doesn't seem correct, `Pokemon#findTags` can't return `null`
       }
     }
     return false;
```
```suggestion
    return restrictingTags.some((tag) => tag.isMoveTargetRestricted(moveId, user, target));
```
`tag !== null` should always be `true`, so this is equivalent.

src/data/abilities/ab-attrs/post-defend-move-disable-ab-attr.ts
```diff
@@ -5,6 +5,7 @@ import type { Pokemon } from "#field/pokemon";
 import type { Move } from "#moves/move";
 
 export class PostDefendMoveDisableAbAttr extends PostDefendAbAttr {
+  // TODO: create a generic way to do ability chances, like with moves
```
flx-sta, line 8: NIT: add a doc for `chance` so people know it can be set to `-1` for `always`

src/data/abilities/ab-attrs/post-summon-stat-stage-change-ab-attr.ts
```diff
@@ -40,8 +40,8 @@ export class PostSummonStatStageChangeAbAttr extends PostSummonAbAttr {
     for (const opponent of pokemon.getOpponents()) {
       const cancelled = new BooleanHolder(false);
       if (this.intimidate) {
-        if (opponent.getTag(BattlerTagType.SUBSTITUTE)) {
-          return false;
+        if (opponent.hasTag(BattlerTagType.SUBSTITUTE)) {
+          continue;
```
flx-sta, line 44: I assume this is related to:

> Intimidate will activate properly if one of the opposing Pokémon has a substitute out but not the other.

src/field/pokemon.ts
```diff
@@ -3523,28 +3542,52 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
   /**
    * Helper function to check if a Pokemon has any of the input tag types.
    * @param tagTypes - The battler tag types to search for
-   * @returns `true` if the Pokemon has at least one of the battler tags, `false` otherwise
+   * @returns Whether the Pokemon has at least one of the input battler tags
    */
   public hasTag(...tagTypes: BattlerTagType[]): boolean {
+    if (tagTypes.length === 0) {
+      console.warn("`Pokemon#hasTag` called with no parameters!");
+      console.trace();
```
flx-sta: Should those stay? or were they just debugging?

src/field/pokemon.ts
flx-sta: Typeguard like `is T`?

src/field/pokemon.ts
DayKev: Not sure what to do about this. `.hasTag` should never be called with no params passed in because that's pointless, but removing the rest param would be disruptive. Is there a way to make TypeScript complain if nothing is passed to the function without changing the params? `Pokemon#hasTag(tagType, ...tagTypes)` doesn't work either because then you can't do `.hasTag(...SOME_TAG_TYPES)`.

src/data/abilities/ab-attrs/post-summon-stat-stage-change-ab-attr.ts
DayKev: Yeah.

src/field/pokemon.ts
DayKev: Yeah it more checks if any tags can be applied to the pokemon (and if the specified tag is already on the pokemon).

Passing an actual tag instead of the tag types is completely unnecessary lol, not sure why it was done like that in the first place. I can change that rq.

DayKev: Actually now I see why it's like this, they're using the `BattlerTag#getDescriptor` method for the display message...

DayKev: I don't think that works though? Not sure, this is weird.

src/field/pokemon.ts
flx-sta: I would just throw an Error, or return `false` with a warning (no trace?). Anything else would require much more unnecessary refactoring

src/field/pokemon.ts
DayKev: Apparently it is. The Sky Drop test starts failing without this line.

src/field/pokemon.ts
DayKev: Well I want the trace so we can find out where it happened and fix the mistake. Although I guess it'll be easy to ctrl+f `hasTag()`, so maybe it's unnecessary.

flx-sta: Having `console.trace` in production seems odd to me. 
I'd Rather have an error thrown.

flx-sta (dismissed):

src/field/pokemon.ts
innerthunder: I don't think it works *here* since the stub is just a `BattlerTag`, but it doesn't have to with everything simulated.

The thing is, while working on the AI rework I ran into runtime errors when calling this in an effect score function. I fixed it by removing the `simulated` check in `PreApplyBattlerTagImmunityAbAttr.apply`:

```ts
public override apply(_pokemon: Pokemon, _simulated: boolean, tag: BattlerTag, cancelled: BooleanHolder): boolean {
  if (this.immuneTagTypes.includes(tag.tagType)) {
    cancelled.value = true;
    this.battlerTag = tag; // this is in an if (!simulated) block in beta
    return true;
  }

  return false;
}
```

You might want to make the same change here.

innerthunder: > * Do the battler tag immunity abilities actually need a full tag to determine immunity? Would these work if they were only given the tag type?

One thing we could do is have the attribute(s) store the tag type the way the tag is stored atm, then have `getTriggerMessage` generate the tag stub with `getBattlerTag(tagType, <some filler args>)`. It's not perfect, but it's at least better for garbage collection...

test/moves/whirlwind.test.ts
```diff
@@ -42,7 +42,7 @@ describe("Moves - Whirlwind", () => {
 
     await game.toEndOfTurn();
 
-    expect(staraptor.findTag((t) => t.tagType === BattlerTagType.MID_AIR)).toBeDefined();
+    expect(staraptor.hasTag(BattlerTagType.MID_AIR)).toBeTruthy();
```
flx-sta: NIT:
```suggestion
    expect(staraptor).toHaveBattlerTag(BattlerTagType.MID_AIR);
```

I must have missed that on Temps PR

flx-sta (dismissed):

src/data/battler-tags/damaging-trap-tag.ts
```diff
@@ -44,7 +44,7 @@ export abstract class DamagingTrapTag extends TrappedTag {
   }
 
   override canAdd(pokemon: Pokemon): boolean {
-    return !pokemon.getTag(...TRAPPED_BATTLER_TAG_TYPES) && !pokemon.getTag(BattlerTagType.SUBSTITUTE);
+    return !pokemon.hasTag(...TRAPPED_BATTLER_TAG_TYPES) && !pokemon.hasTag(BattlerTagType.SUBSTITUTE);
```
MokaStitcher:
```suggestion
    return !pokemon.hasTag(...TRAPPED_BATTLER_TAG_TYPES, BattlerTagType.SUBSTITUTE);
```
I think these can be merge?

src/data/battler-tags/grounded-tag.ts
```diff
@@ -31,8 +31,8 @@ export class GroundedTag extends BattlerTag {
     const wasNotGrounded =
       pokemon.isOfType(ElementalType.FLYING, true, true)
       || pokemon.hasAbility(AbilityId.LEVITATE)
-      || pokemon.getTag(BattlerTagType.FLOATING)
-      || pokemon.getTag(...SEMI_INVULNERABLE_BATTLER_TAG_TYPES);
+      || pokemon.hasTag(BattlerTagType.FLOATING)
+      || pokemon.hasTag(...SEMI_INVULNERABLE_BATTLER_TAG_TYPES);
```
MokaStitcher:
```suggestion
      || pokemon.hasTag(BattlerTagType.FLOATING, ...SEMI_INVULNERABLE_BATTLER_TAG_TYPES);
```
same here? Personally I don't mind keeping them separate for readability, but it was suggested to do this on another PR and I assume it saves on performance a little bit?

<sub>is it normal that SKY DROP is not here? who knows 😔

src/data/init/init-moves.ts
```diff
@@ -1858,8 +1858,8 @@ export function initMoves() {
       )
       .condition(
         (_user, target, _move) =>
-          isNil(target.getTag(BattlerTagType.INGRAIN))
-          && isNil(target.getTag(BattlerTagType.IGNORE_FLYING)),
+          !target.hasTag(BattlerTagType.INGRAIN)
+          && !target.hasTag(BattlerTagType.IGNORE_FLYING),
```
MokaStitcher:
```suggestion
          !target.hasTag(BattlerTagType.INGRAIN, BattlerTagType.IGNORE_FLYING),
```

src/data/moves/move-attrs/sky-drop-attr.ts
```diff
@@ -40,9 +40,9 @@ export class SkyDropAttr extends MoveEffectAttr {
       failOnGravityCondition(user, target, move)
       && target.isPlayer() !== user.isPlayer()
       && target.species.weight < 200
-      && !target.getTag(BattlerTagType.SUBSTITUTE)
-      && !target.getTag(...SEMI_INVULNERABLE_BATTLER_TAG_TYPES)
+      && !target.hasTag(BattlerTagType.SUBSTITUTE)
+      && !target.hasTag(...SEMI_INVULNERABLE_BATTLER_TAG_TYPES)
```
MokaStitcher:
```suggestion
      && !target.hasTag(BattlerTagType.SUBSTITUTE, ...SEMI_INVULNERABLE_BATTLER_TAG_TYPES)
```

src/field/pokemon.ts
```diff
@@ -1770,9 +1773,9 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
       this.hasTag(BattlerTagType.IGNORE_FLYING)
       || (!this.isOfType(ElementalType.FLYING, true, true)
         && !this.hasAbility(AbilityId.LEVITATE)
-        && !this.getTag(BattlerTagType.FLOATING)
-        && !this.getTag(...SEMI_INVULNERABLE_BATTLER_TAG_TYPES)
-        && !this.getTag(BattlerTagType.SKY_DROP))
+        && !this.hasTag(BattlerTagType.FLOATING)
+        && !this.hasTag(...SEMI_INVULNERABLE_BATTLER_TAG_TYPES)
+        && !this.hasTag(BattlerTagType.SKY_DROP))
```
```suggestion
        && !this.hasTag(BattlerTagType.FLOATING, BattlerTagType.SKY_DROP, ...SEMI_INVULNERABLE_BATTLER_TAG_TYPES))
```

src/field/pokemon.ts
```diff
@@ -4027,7 +4098,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
 
       this.setFrameRate(4);
 
-      const tag = SEMI_INVULNERABLE_BATTLER_TAG_TYPES.find((t) => this.getTag(t));
+      const tag = SEMI_INVULNERABLE_BATTLER_TAG_TYPES.find((t) => this.hasTag(t));
```
MokaStitcher: Would it be the same as doing `this.getTag(...SEMI_INVULNERABLE_BATTLER_TAG_TYPES)`?
then 
`this.removeTag(tag.tagType);`

src/field/pokemon.ts
```diff
@@ -4160,10 +4227,6 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
       moveHistory: [],
     };
     this.setSwitchOutStatus(false);
-    // TODO: why is this here? is it safe to remove this?
-    if (this.getTag(BattlerTagType.SEEDED)) {
-      this.lapseTag(BattlerTagType.SEEDED);
```
MokaStitcher, line 4165: did you confirm that removing this doesn't cause issues?

src/phases/move-phase.ts
```diff
@@ -803,7 +805,7 @@ export class MovePhase extends BattlePhase {
       return;
     }
 
-    if (this.pokemon.getTag(BattlerTagType.RECHARGING) || this.pokemon.getTag(BattlerTagType.INTERRUPTED)) {
+    if (this.pokemon.hasTag(BattlerTagType.RECHARGING) || this.pokemon.hasTag(BattlerTagType.INTERRUPTED)) {
```
MokaStitcher:
```suggestion
    if (this.pokemon.hasTag(BattlerTagType.RECHARGING, BattlerTagType.INTERRUPTED)) {
```

src/phases/post-summon-phase.ts
```diff
@@ -31,8 +30,7 @@ export class PostSummonPhase extends PokemonPhase {
     // If this is mystery encounter and has post summon phase tag, apply post summon effects
     if (
       globalScene.currentBattle.isBattleMysteryEncounter()
-      && pokemon.findTags((t) => t.isType<MysteryEncounterPostSummonTag>(BattlerTagType.MYSTERY_ENCOUNTER_POST_SUMMON))
-        .length > 0
+      && pokemon.hasTag(BattlerTagType.MYSTERY_ENCOUNTER_POST_SUMMON)
```
MokaStitcher:
```suggestion
```
we can probably even remove the check like it was done in other places?

src/constants/battler-tag-constants.ts
```diff
@@ -100,3 +100,23 @@ export const EXPOSED_TAG_TYPES = Object.freeze<BattlerTagType[]>([
   BattlerTagType.IGNORE_GHOST,
   BattlerTagType.IGNORE_DARK,
 ]);
+
+/** All {@linkcode BattlerTagType}s that boost the pokemon's highest stat */
+export const HIGHEST_STAT_BOOST_TAG_TYPES = Object.freeze<BattlerTagType[]>([
```
MokaStitcher, line 148: If I understand correctly this was added to be able to remove some `instanceof` in other parts of the code
Which means that all the battler tag types in this array have a BattlerTag implementation that extends the same base BattlerTag class
I think I'd like that parent class to be linked in the doc here, to make that information clear. And maybe the class' doc should also link to that list and state that subclasses should add their tag type in the list

Does that make sense?
If yes, it'd be the same for the others below (and maybe those above too if they were created on the same premise)

flx-sta (dismissed):

src/data/battler-tags/grounded-tag.ts
DayKev: Added a `TODO` just in case.

src/phases/post-summon-phase.ts
DayKev: Ah true, it didn't look the same so I didn't notice.

src/field/pokemon.ts
DayKev: I think so.

src/field/pokemon.ts
DayKev: Yeah, it's still removed when the pokemon switches out, and still triggers correctly on pokemon on the field.

src/data/battler-tags/grounded-tag.ts
MokaStitcher: I made an issue to look at that the other day so the Todo is not mandatory, but thanks 🙏

src/data/battler-tags/floating-tag.ts
```diff
@@ -11,7 +17,8 @@ import i18next from "i18next";
  * Tag that lifts the affected Pokemon into the air and provides immunity to Ground type moves.
  * Used for {@link https://bulbapedia.bulbagarden.net/wiki/Magnet_Rise_(move) | Magnet Rise}
  * and {@link https://bulbapedia.bulbagarden.net/wiki/Telekinesis_(move) | Telekinesis}
- * @extends TypeImmuneTag
+ *
+ * Tags that use or subclass this should be added to {@linkcode TYPE_IMMUNE_TAG_TYPES}
```
innerthunder, line 22:
```suggestion
  * @privateRemarks
  * Tags that use or subclass this should be added to {@linkcode TYPE_IMMUNE_TAG_TYPES}
```

src/data/battler-tags/grounded-tag.ts
```diff
@@ -28,11 +28,11 @@ export class GroundedTag extends BattlerTag {
    */
   override onAdd(pokemon: Pokemon) {
     const isSmackDownOrThousandArrows = [MoveId.SMACK_DOWN, MoveId.THOUSAND_ARROWS].includes(this.sourceMoveId);
+    // TODO: Should Sky Drop be here?
```
innerthunder: No, Pokemon in Sky Drop aren't grounded by the effect. From [Bulba](https://bulbapedia.bulbagarden.net/wiki/Smack_Down_(move)):

> Smack Down can hit Pokémon during the semi-invulnerable turns of [Fly](https://bulbapedia.bulbagarden.net/wiki/Fly_(move)), [Bounce](https://bulbapedia.bulbagarden.net/wiki/Bounce_(move)), and [Sky Drop](https://bulbapedia.bulbagarden.net/wiki/Sky_Drop_(move)). If it hits a Pokémon that is using Fly or Bounce, it will knock it down, cancelling the move; if it hits a Pokémon using or affected by Sky Drop, it will not be knocked down.

test/abilities/aroma-veil.test.ts
```diff
@@ -45,7 +45,7 @@ describe("Moves - Aroma Veil", () => {
     await game.toNextTurn();
 
     playerPokemon.forEach((p) => {
-      expect(p.getTag(BattlerTagType.HEAL_BLOCK)).toBeUndefined();
+      expect(p.hasTag(BattlerTagType.HEAL_BLOCK)).toBeFalsy();
```
innerthunder:
```suggestion
      expect(p).not.toHaveBattlerTag(BattlerTagType.HEAL_BLOCK);
```

test/moves/baton-pass.test.ts
```diff
@@ -98,12 +98,12 @@ describe("Moves - Baton Pass", () => {
     game.move.select(MoveId.BATON_PASS);
     game.setTurnOrder([BattlerIndex.ENEMY, BattlerIndex.PLAYER]);
     await game.phaseInterceptor.to("PostActionPhase");
-    expect(player1.findTag((t) => t.tagType === BattlerTagType.SALT_CURED)).toBeTruthy();
+    expect(player1.hasTag(BattlerTagType.SALT_CURED)).toBeTruthy();
```
innerthunder:
```suggestion
    expect(player1).toHaveBattlerTag(BattlerTagType.SALT_CURED);
```

test/moves/baton-pass.test.ts
```diff
@@ -98,12 +98,12 @@ describe("Moves - Baton Pass", () => {
...
     game.selectPartyPokemon(1);
     await game.toNextTurn();
 
-    expect(player2.findTag((t) => t.tagType === BattlerTagType.SALT_CURED)).toBeUndefined();
-  }, 20000);
+    expect(player2.hasTag(BattlerTagType.SALT_CURED)).toBeFalsy();
```
innerthunder:
```suggestion
    expect(player2).not.toHaveBattlerTag(BattlerTagType.SALT_CURED);
```

test/moves/octolock.test.ts
```diff
@@ -102,13 +102,13 @@ describe("Moves - Octolock", () => {
     const enemyPokemon = game.scene.getEnemyPokemon()!;
 
     // before Octolock - enemy should not be trapped
-    expect(enemyPokemon.findTag((t) => t instanceof TrappedTag)).toBeUndefined();
+    expect(enemyPokemon.hasTag(...TRAPPED_BATTLER_TAG_TYPES)).toBeFalsy();
```
innerthunder: We may want to update the `toHaveBattlerTag` matcher to allow multiple type inputs...

innerthunder (dismissed):
Some nits, otherwise lgtm

test/moves/octolock.test.ts
DayKev: Yeah. Though in this case it should probably just be checking for the tag set by Octolock.

innerthunder (approved):

flx-sta (approved):

PR #1225, opened 2025-05-27
[P3][Refactor] Remove `switch (true)`, update Heavy Slam

DayKev:
- **[P3]** The power calculation for Heavy Slam and Heat Crash is slightly more accurate.
- **[P3]** Pichu's weight and height are fixed.

`switch (true)` is dumb.

- Replaced `switch (true)` with formula in `compare-weight-power-attr.ts`.
- Replaced `switch (true)` with `switch (this.species.speciesId)` in `EnemyPokemon#generateAndPopulateMoveset`. (...)
- Fixed weight and height valuse for Pichu's forms.
  - Weight: was `61.5`, now `2`.
  - Height: was `1.4`, now `0.3`.

## Checklist


src/data/moves/move-attrs/compare-weight-power-attr.ts
```diff
@@ -18,24 +18,16 @@ export class CompareWeightPowerAttr extends VariablePowerAttr {
...
+    if (userWeight >= targetWeight * 5) {
+      power.value = 120;
+    } else if (userWeight >= targetWeight * 4) {
+      power.value = 100;
+    } else if (userWeight >= targetWeight * 3) {
+      power.value = 80;
+    } else if (userWeight >= targetWeight * 2) {
+      power.value = 60;
+    } else {
+      power.value = 40;
```
innerthunder:
```suggestion
    power.value = (clamp(Math.floor(userWeight / targetWeight), 1, 5) + 1) * 20;
```
May need a double-check, but I think this is equivalent.

DayKev: I'll write a test for it 👍 
Just gotta find some pokemon with the right weights...

innerthunder: Oh, this also should have a way to handle when `targetWeight` is 0 (the `if` block before this doesn't cover divide by 0 issues anymore)

src/data/init/init-species.ts
```diff
@@ -5607,7 +5607,7 @@ export function initSpecies() {
         ElementalType.ELECTRIC,
         null,
         1.4,
```
Tempo-anon: The height value is also incorrect. Looks like the 1.4 height and 61.5 weight values come from copy/pasting Ampharos's data

src/data/moves/move-attrs/compare-weight-power-attr.ts
```diff
@@ -14,29 +14,11 @@ export class CompareWeightPowerAttr extends VariablePowerAttr {
...
-        power.value = 80;
-        break;
-      case relativeWeight < 50.01:
-        power.value = 60;
-        break;
-      default:
-        power.value = 40;
-        break;
-    }
+    power.value = (clamp(Math.floor(userWeight / targetWeight), 1, 5) + 1) * 20;
```
Tempo-anon, line 31: While as a one liner it's less code, it's harder to see what it's doing at a glance. I'm not sure if my preferred solution is to replace this with all if/else or to have a comment/table here explaining the breakpoints

src/field/enemy-pokemon.ts
```diff
@@ -156,16 +156,16 @@ export class EnemyPokemon extends Pokemon {
   }
 
   override generateAndPopulateMoveset(formIndex?: number): void {
-    switch (true) {
-      case this.species.speciesId === SpeciesId.SMEARGLE:
+    switch (this.species.speciesId) {
```
Tempo-anon, line 159: More of a styling thing, but is a switch needed here for just 2 usecases?

Tempo-anon (commented):
* Pichu's incorrect data should also be mentioned in the PR title and desc. 
* Is Heavy Slam's effect functionally changed?

src/data/moves/move-attrs/compare-weight-power-attr.ts
DayKev: I could put the if blocks back.
@innerthunder thoughts?

src/field/enemy-pokemon.ts
DayKev: I think a `switch` is fine here, though it doesn't really matter.

DayKev:
> * Is Heavy Slam's effect functionally changed?

Yes, previously the breakpoints were incorrect (the comparisons were using `<` when they should have been using `<=`).

src/data/moves/move-attrs/compare-weight-power-attr.ts
innerthunder: A comment/table with breakpoints would be better

DayKev: Table added to TSDdoc of the class.

src/field/enemy-pokemon.ts
flx-sta: It's designed to extendable so I think it's a good use case

src/data/moves/move-attrs/compare-weight-power-attr.ts
```diff
@@ -1,42 +1,33 @@
...
+ *
  * @extends VariablePowerAttr
  */
 export class CompareWeightPowerAttr extends VariablePowerAttr {
   override apply(user: Pokemon, target: Pokemon, _move: Move, power: NumberHolder): boolean {
     const userWeight = user.getWeight();
     const targetWeight = target.getWeight();
 
-    if (!userWeight || userWeight === 0) {
+    if (!userWeight || !targetWeight) {
```
MokaStitcher: not sure those checks are needed since getWeight should always returns at least 0.1

MokaStitcher (dismissed):
lgtm

src/data/moves/move-attrs/compare-weight-power-attr.ts
flx-sta: agree

test/moves/heavy-slam.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    expect(moveSpy).toHaveLastReturnedWith(power);
+
+    game.switchPokemon(2);
+    await game.toNextTurn();
+
+    game.move.use(MoveId.HEAVY_SLAM);
+    await game.toNextTurn();
+    expect(game.field.getPlayerPokemon().getWeight()).toBe(weights[2]);
+    expect(moveSpy).toHaveLastReturnedWith(power);
+  });
```
flx-sta, line 67: FYI you don't **have** to buffer the `moveSpy`. You transform the original into a spy too.

> [!TIP]
> This is not a necessary change, just informational. You decide what you want to do with it. 

SO you could also do

```suggestion
  ])("should scale with weight differences (power $power)", async ({ power, species, weights }) => {
    await game.classicMode.startBattle(species[0], species[1], species[2]);
    vi.spyOn(Move.prototype, "calculateBattlePower");

    game.move.use(MoveId.HEAVY_SLAM);
    await game.toNextTurn();
    expect(game.field.getPlayerPokemon().getWeight()).toBe(weights[0]);
    expect(game.field.getEnemyPokemon().getWeight()).toBe(1);
    expect(Move.prototype.calculateBattlePower).toHaveLastReturnedWith(power);

    game.switchPokemon(1);
    await game.toNextTurn();

    game.move.use(MoveId.HEAVY_SLAM);
    await game.toNextTurn();
    expect(game.field.getPlayerPokemon().getWeight()).toBe(weights[1]);
    expect(Move.prototype.calculateBattlePower).toHaveLastReturnedWith(power);

    game.switchPokemon(2);
    await game.toNextTurn();

    game.move.use(MoveId.HEAVY_SLAM);
    await game.toNextTurn();
    expect(game.field.getPlayerPokemon().getWeight()).toBe(weights[2]);
    expect(Move.prototype.calculateBattlePower).toHaveLastReturnedWith(power);
  });
```

flx-sta (dismissed):

src/data/moves/move-attrs/compare-weight-power-attr.ts
```diff
@@ -1,42 +1,33 @@
...
       return false;
     }
```
innerthunder:
```suggestion
    if (targetWeight === 0) {
      power.value = 120;
      return true;
    }
```
`userWeight` doesn't need to be checked here at all; the only failsafe needed is for when `targetWeight` zeroes out the denominator. Also, I think it makes more sense for Heavy Slam to be set to max power when `targetWeight` is 0 than fail to adjust power entirely.

<sub>Also, I know others are saying this check isn't *really* needed, but we have similar failsafes in other places and it doesn't hurt to keep this one around imo</sub>

MokaStitcher: I don't mind keeping the / 0 failsafe. It should be on userWeight though, not targetWeight since the calculation is `(targetWeight / userWeight)`

I got lazy with my original comment but I'll write up my full reasoning then:
- The `targetWeight` check is not needed because we should always get at least 0.1 from `getWeight`, and even if we somehow got 0 it would end up at the max power value without having to change anything to the current implementation. I agree with Inner that this should be the result in this case.
- The `userWeight` check is technically not needed because `getWeight` will always return at least 0.1, but I don't mind keeping it as a failsafe

innerthunder: > I don't mind keeping the / 0 failsafe. It should be on userWeight though, not targetWeight since the calculation is `(targetWeight / userWeight)`

I think you're looking at the old code? The new formula uses `userWeight / targetWeight`.

MokaStitcher: Oh... you're right, sorry 💀

test/moves/heavy-slam.test.ts
DayKev: True, though I think it's simpler this way.

PR #1229, opened 2025-05-27
[Refactor] Rename FLYING battler tag to MID_AIR

Tempo-anon:

None

FLYING as a battler tag name is a bit confusing since we already have an ElementalType.FLYING and this tag is also used for Bounce

Renamed a enum

## Screenshots/Videos

n/a

## How to test the changes?

`npm run test`

## Checklist


src/constants/battler-tag-constants.ts
```diff
@@ -8,7 +8,7 @@ import { BattlerTagType } from "#enums/battler-tag-type";
 
 /** All {@linkcode BattlerTagType | BattlerTagTypes} that grant semi-invulnerability */
 export const SEMI_INVULNERABLE_BATTLER_TAG_TYPES = Object.freeze([
```
innerthunder: should `SKY_DROP` be added to this list?

innerthunder (dismissed):

flx-sta (dismissed):

src/constants/battler-tag-constants.ts
MokaStitcher: From a quick look through the code base the Sky drop tag is checked in some spots where the SEMI_INVULNERABLE_BATTLER_TAG_TYPES is checked as well, but not all, so doing that would change some interactions, and I don't know if it would be correct for all of them.
It's probably best to leave that for a different PR where each case can be looked at properly, I made an issue for it #1235

PR #1230, opened 2025-05-29
[UI] Initialize and destroy UI handlers dynamically

### LINKED ISSUE #1117: [Refactor] Initialize UI handlers dynamically

MokaStitcher:

- Less average memory usage
- Faster initial loading time
- Longer loading time for starter select UI (temporarily until the UI is reworked)
- Minor updates to the Statistics screen, see #1242 for details

The UI currently works by creating an instance of UiHandler for each UiMode during the BattleScene initialization and keeping them in memory at all times, simply hiding and showing their content when switching UiMode.
This causes a significant memory overhead because those handlers are never cleared from memory, even those like the login/registration form, or the starter select handler that is not needed at all while during a run.

This PR changes things so that most handlers are only initialized when they are needed the first time, and get destroyed based on various in game events.

There is more work to do for loading/memory optimization on specific handlers, but I have decided to leave them to separate PRs to avoid bloating this one more.

</details> 

### Core Changes
**[ui.ts](https://github.com/Despair-Games/poketernity/pull/1230/files#diff-80caad6552583ef3d93fb59112800cae742ea01316f854a0c2f5b251b4171f16) is the core of it:**
- Replace the `handlers` array with a `Map<UiMode, UiHandler>
- Create functions to add, delete and retrieve handlers based on the given UiMode
- In setup, only* the handler for the default mode (MESSAGE) is initialized to always be kept in memory. Buuuut...
     - ... the `POKEMON_RENAME` handler is also initialized to prevent a graphical bug caused by the `InputText`
     - ... And until these handlers are optimized in another PRs, all settings handlers are also initialized there and always kept in memory
- In theory because of the above, `getMessageHandler()` should always be defined. But this is not the case when the UI is destroyed then rebuilt from scratch (for example after a Save & Quit). In this case `getMessageHandler()` can be undefined during destruction, causing issues for some UIs that access it when clearing themselves. Ideally no handler should be doing that, but for now as a safeguard `getMessageHandler()` potentially returns `undefined`, and everything was updated to account for that.

### Additional Changes

<details><summary>[Click to show details]</summary>

#### Added logic to free some Ui Handlers from memory

- When UiMode is switched to the title screen, Login and Registration handlers get destroyed
- When an `ENCOUNTER_PHASE` event is fired (the event already exists):
   - destroy the handlers for starter select, challenge select, title screen, etc.
   - destroy the ME handler, Egg hatch and egg summary handler on the rationale that these will likely not be needed for ~10 waves, which seemed like a large enough number of waves to make the destruction and later reconstruction reasonable
   - destroy the FormChange scene handler (used for evolutions mostly) because it is also unlikely to be needed a lot in the course of a run
- When a `POST_GAME_OVER` even is fired (this is a new event): destroy all battle-related handlers.
- When the `MENU` mode is ended, destroy all handlers that can only be accessed through the menu (achievements, egg gacha, etc.). The settings handlers are not included in this.

Other handlers like option select, confirm, message and similar are never destroyed once initialized, as they are used extremely frequently

<sub>A lot of this logic could be replaced by splitting the game into different scenes with their own UI in the future, but for now, this seemed like an ok approach to things

</details> 

#### Other
- Added a `UI_DEBUG` env variable to log handlers getting initialized and destroyed
- Updated `GAME_STATS` ui handler to use less memory. Full details in #1242 but short version:
   - created a new `TextListContainer` that creates a single multiline Text object from an array of string.
   - Replaced the 36 Text objects in handlers by only 4 of these containers
   - Edited font styles to accept line spacing configuration and added new font styles for the spacing used in the stats handler

## Screenshots/Videos

Test run with UI debug messages:
image

Heap snapshots size. 1) on the title screen - 2) in starter select - 3) in battle - 4) after losing the run, back on title screen
Beta:
| `beta` (old)  | **this branch** (new) |
|--------|--------|
| image | image | 

<sub>would need to do more long term testing to see how things are after 30 or 50 of waves in both

## How to test the changes?

- optionally, set `UI_DEBUG` to 1 in  `.env.development`
- run the game, open and close various UIs, play some waves, lose or win, then open more UIs, play some more waves, etc
~~if things went well~~ when npm issues get solved, the latest version should be available [here](https://poketernity.com/dev-ui/)

## Checklist


src/enums/battle-scene-event-type.ts
```diff
@@ -11,6 +27,12 @@ export enum BattleSceneEventType {
    */
   BERRY_USED = "onBerryUsed",
 
+  /**
+   * Triggers when a run end either through winning or losing (but not save and quitting)
+   * @see {@linkcode RunEndEvent}
+   */
+  RUN_END = "onRunEnd",
+
```
flx-sta: Not sure if this event name is optimal. It could also be associated with "Running from a battle" by accident.

src/phases/level-up-phase.ts
```diff
@@ -50,7 +50,7 @@ export class LevelUpPhase extends PlayerPartyMemberPokemonPhase {
 
     const promptLevelUpStats = (): Promise<void> =>
       ui
-        .getMessageHandler()
+        .getMessageHandler()!
```
flx-sta: Why `!` here? I'd prefer not to introduce any new `!`

src/phases/scan-ivs-phase.ts
```diff
@@ -43,7 +43,7 @@ export class ScanIvsPhase extends PokemonPhase {
       const enemyIvs = enemyPokemon.ivs;
       // we are using getRootSpeciesId() here because we want to check against the baby form, not the mid form if it exists
       const currentIvs = gameData.starterData[enemyPokemon.species.getRootSpeciesId()].ivs;
-      const ivsToShow = ui.getMessageHandler().getTopIvs(enemyIvs, this.shownIvs);
+      const ivsToShow = ui.getMessageHandler()!.getTopIvs(enemyIvs, this.shownIvs);
```
flx-sta: Any way to avoid adding another `!` here?

src/phases/scan-ivs-phase.ts
```diff
@@ -69,7 +69,7 @@ export class ScanIvsPhase extends PokemonPhase {
               ui.setMessageMode();
               ui.clearText();
               new CommonBattleAnim(CommonAnim.LOCK_ON, pokemon, pokemon).play(false, () => {
-                ui.getMessageHandler()
+                ui.getMessageHandler()!
```
flx-sta: Any way to avoid adding another `!` here?

src/ui/handlers/command-ui-handler.ts
```diff
@@ -261,9 +265,9 @@ export class CommandUiHandler extends UiHandler {
   protected override clear(): void {
     this.commandsContainer.setVisible(false);
     const messageHandler = this.getUi().getMessageHandler();
-    if (messageHandler.ready) {
-      messageHandler.commandWindow.setVisible(false);
-      messageHandler.clearText();
+    if (messageHandler?.ready) {
+      messageHandler?.commandWindow.setVisible(false);
+      messageHandler?.clearText();
```
flx-sta: are those `?` necessary? I think the `if` before should cover telling the compiler it's set

```suggestion
      messageHandler.commandWindow.setVisible(false);
      messageHandler.clearText();
```

flx-sta:
Sorry saw too late that it's WIP/Draft

src/phases/level-up-phase.ts
MokaStitcher: Updated.
As for why: because I was being annoyed by having to check for it being defined when it should almost always be defined and got lazy 😭 
And doubly annoyed because why in the world is the messageUiHandler used to display level up stats and IVs...

<sub>The only time it could be undefined is when the full UI is in the process of being destroyed. And you'd think that's not an issue, but it is because some handlers do things on the message handler when they get cleared.
But IMO those handlers shouldn't even access the message handler like this, to make some ui elements (in)visible, it's very messy. At the very minimum it should use helper functions. But not the scope of this PR so we'll roll with it

flx-sta:
FYI: removed from project since the bug is already on there 👍 
Bug > PR in boards

.env.test
```diff
@@ -1,3 +1,4 @@
 VITE_I18N_DEBUG=0
 VITE_API_DEBUG=0
-VITE_MODIFIERS_DEBUG=0
\ No newline at end of file
+VITE_MODIFIERS_DEBUG=0
+VITE_UI_DEBUG=1
```
flx-sta, line 4: Debug in test? Is that intended?

src/enums/battle-scene-event-type.ts
```diff
@@ -11,6 +27,12 @@ export const BattleSceneEventType = {
    */
   BERRY_USED: "onBerryUsed",
 
+  /**
+   * Triggers when after run ends either through winning or losing (but not save and quitting)
```
flx-sta: NIT: 
```suggestion
   * Triggers after run ends. Either through winning or losing (but not through "Save & Quit")
```

src/loggers.ts
```diff
@@ -74,3 +74,9 @@ export function logModifiers(...args: any[]): void {
     console.log(...args);
   }
 }
+
+export function logUiEvent(...args: any[]): void {
```
flx-sta: NIT: docs

src/phases/post-game-over-phase.ts
```diff
@@ -29,14 +30,15 @@ export class PostGameOverPhase extends Phase {
           }
           globalScene.reset();
           globalScene.phaseManager.toTitleScreen({ eager: true });
+          globalScene.eventTarget.dispatchEvent(new GameOverEvent());
```
flx-sta, line 33: We should probably decide if we want to stick to using `globalScene.eventTarget` or migrate over to the global `eventBus` (that I introduced with PR #50)

https://github.com/Despair-Games/poketernity/blob/75fe4f1db77f80bc0dcb8a4ee5bff2e19ab9521d/src/event-bus.ts#L44

src/phases/scan-ivs-phase.ts
```diff
@@ -69,9 +75,7 @@ export class ScanIvsPhase extends PokemonPhase {
               ui.setMessageMode();
               ui.clearText();
               new CommonBattleAnim(CommonAnim.LOCK_ON, pokemon, pokemon).play(false, () => {
-                ui.getMessageHandler()
-                  .promptIvs(pokemon.id, pokemon.ivs, this.shownIvs)
-                  .then(() => this.end());
+                messageUiHandler.promptIvs(pokemon.id, pokemon.ivs, this.shownIvs).then(() => this.end());
```
flx-sta, line 78: Missing `catch()` here
 <sub> Probably out of scope for this PR </sub>

src/ui/components/text-list-container.ts
```diff
@@ -0,0 +1,125 @@
...
+  private maxLines: number;
+
+  private textObject: Phaser.GameObjects.Text | BBCodeText;
+
+  /**
+   * @param x - The x position of the TextObject in its parent container.
+   * @param y - The y position of the TextObject in its parent container.
+   * @param maxLines - The maximum number of lines to show at one.
+   * @param options - The text style {@linkcode TextListContainerOptions}.
+   */
```
flx-sta: Q: Not sure if the constructor has/needs docs

src/ui/handlers/game-stats-ui-handler.ts
```diff
@@ -214,18 +215,31 @@ const displayStats: DisplayStats = {
...
+ * Note: currently the stats are computed when the handler is shown and not updated while it's in use,
+ * meaning the playtime stat won't change until the handler is exited then re-opened.
+ */
 export class GameStatsUiHandler extends UiHandler {
+  private readonly ROWS_ON_SCREEN = 9;
+  private readonly NUM_COLUMNS = 2;
+  private readonly MAX_STATS_ON_SCREEN = this.ROWS_ON_SCREEN * this.NUM_COLUMNS;
+  private readonly MAX_CURSOR = Math.ceil(
+    (Object.keys(displayStats).length - this.MAX_STATS_ON_SCREEN) / this.NUM_COLUMNS,
+  );
```
flx-sta: NIT: 
1. non static properties are initialized in every instance thus might occupy more memory if multiple instances are created
2. non static properties shouldn't be `UPPER_SNAKE_CASE` but follow the standard `camelCase` naming

src/ui/ui.ts
```diff
@@ -128,79 +173,206 @@ export class UI extends Phaser.GameObjects.Container {
...
+    this.handlers.set(mode, handler);
+    handler.initialize();
+    return handler;
+  }
+
+  /**
+   * Destroy the handlers for the given modes if they exist, and remove them from the handlers list.
+   * @param modes - one or more {@linkcode UiMode}s
+   */
+  private deleteUiHandlers(modes: UiMode | UiMode[]): void {
```
flx-sta: We should consider making this a `vararg` instead.

```suggestion
  private deleteUiHandlers(...modes: UiMode[]): void {
```

src/ui/ui.ts
```diff
@@ -602,7 +800,7 @@ export class UI extends Phaser.GameObjects.Container {
     });
   }
 
-  revertMode(): Promise<boolean> {
+  public revertMode(): Promise<boolean> {
```
flx-sta, line 895: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -330,11 +515,11 @@ export class UI extends Phaser.GameObjects.Container {
       }
       showMessageAndCallback();
     } else {
-      this.getCurrentMessageHandler().showText(text, delay, callback, callbackDelay, prompt, promptDelay);
+      this.getCurrentMessageHandler()!.showText(text, delay, callback, callbackDelay, prompt, promptDelay);
     }
   }
 
-  showDialogue(
+  public showDialogue(
```
flx-sta, line 530: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -386,7 +571,7 @@ export class UI extends Phaser.GameObjects.Container {
     }
   }
 
-  shouldSkipDialogue(i18nKey: string): boolean {
+  public shouldSkipDialogue(i18nKey: string): boolean {
```
flx-sta, line 588: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -395,11 +580,11 @@ export class UI extends Phaser.GameObjects.Container {
     return false;
   }
 
-  getTooltip(): { visible: boolean; title: string; content: string } {
+  public getTooltip(): { visible: boolean; title: string; content: string } {
```
flx-sta, line 600: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -395,11 +580,11 @@ export class UI extends Phaser.GameObjects.Container {
...
     return { visible: this.tooltipContainer.visible, title: this.tooltipTitle.text, content: this.tooltipContent.text };
   }
 
-  showTooltip(title: string, content: string, overlap?: boolean): void {
+  public showTooltip(title: string, content: string, overlap?: boolean): void {
```
flx-sta: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -409,7 +594,7 @@ export class UI extends Phaser.GameObjects.Container {
     }
   }
 
-  editTooltip(title: string, content: string): void {
+  public editTooltip(title: string, content: string): void {
```
flx-sta, line 625: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -422,7 +607,7 @@ export class UI extends Phaser.GameObjects.Container {
     this.tooltipTitle.x = this.tooltipBg.width / 2;
   }
 
-  hideTooltip(): void {
+  public hideTooltip(): void {
```
flx-sta, line 639: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -460,11 +645,12 @@ export class UI extends Phaser.GameObjects.Container {
     }
   }
 
-  clearText(): void {
+  public clearText(): void {
```
flx-sta, line 680: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -460,11 +645,12 @@ export class UI extends Phaser.GameObjects.Container {
...
     this.getCurrentMessageHandler().clearText();
   }
 
-  setCursor(cursor: number): boolean {
+  // UNUSED?
+  public setCursor(cursor: number): boolean {
```
flx-sta: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -473,15 +659,15 @@ export class UI extends Phaser.GameObjects.Container {
     return changed;
   }
 
-  playSelect(): void {
+  public playSelect(): void {
```
flx-sta, line 687: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -473,15 +659,15 @@ export class UI extends Phaser.GameObjects.Container {
...
     globalScene.audioManager.playSound("ui/select");
   }
 
-  playError(): void {
+  public playError(): void {
```
flx-sta, line 694: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -473,15 +659,15 @@ export class UI extends Phaser.GameObjects.Container {
...
     globalScene.audioManager.playSound("ui/error");
   }
 
-  fadeOut(duration: number): Promise<void> {
+  public fadeOut(duration: number): Promise<void> {
```
flx-sta, line 703: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -499,7 +685,7 @@ export class UI extends Phaser.GameObjects.Container {
     });
   }
 
-  fadeIn(duration: number): Promise<void> {
+  public fadeIn(duration: number): Promise<void> {
```
flx-sta, line 726: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -518,33 +704,36 @@ export class UI extends Phaser.GameObjects.Container {
     });
   }
 
-  getMode(): UiMode {
+  public getMode(): UiMode {
```
flx-sta, line 748: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -518,33 +704,36 @@ export class UI extends Phaser.GameObjects.Container {
...
     return this.mode;
   }
 
-  setMessageMode(): Promise<void> {
+  public setMessageMode(): Promise<void> {
     return this.setMode<MessageUiHandler>(UiMode.MESSAGE);
   }
 
-  setMode<THandler extends UiHandler = never>(mode: UiMode, ...args: Parameters<THandler["show"]>): Promise<void> {
+  public setMode<THandler extends UiHandler = never>(
```
flx-sta, line 767: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -518,33 +704,36 @@ export class UI extends Phaser.GameObjects.Container {
     });
   }
 
-  getMode(): UiMode {
+  public getMode(): UiMode {
     return this.mode;
   }
 
-  setMessageMode(): Promise<void> {
+  public setMessageMode(): Promise<void> {
```
flx-sta, line 756: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -518,33 +704,36 @@ export class UI extends Phaser.GameObjects.Container {
...
-  setMode<THandler extends UiHandler = never>(mode: UiMode, ...args: Parameters<THandler["show"]>): Promise<void> {
+  public setMode<THandler extends UiHandler = never>(
+    mode: UiMode,
+    ...args: Parameters<THandler["show"]>
+  ): Promise<void> {
     return this.setModeInternal<THandler>(mode, true, false, false, ...args);
   }
 
-  setModeForceTransition<THandler extends UiHandler = never>(
+  public setModeForceTransition<THandler extends UiHandler = never>(
```
flx-sta, line 781: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -518,33 +704,36 @@ export class UI extends Phaser.GameObjects.Container {
...
     mode: UiMode,
     ...args: Parameters<THandler["show"]>
   ): Promise<void> {
     return this.setModeInternal<THandler>(mode, true, true, false, ...args);
   }
 
-  setModeWithoutClear<THandler extends UiHandler = never>(
+  public setModeWithoutClear<THandler extends UiHandler = never>(
```
flx-sta, line 795: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -518,33 +704,36 @@ export class UI extends Phaser.GameObjects.Container {
...
     mode: UiMode,
     ...args: Parameters<THandler["show"]>
   ): Promise<void> {
     return this.setModeInternal<THandler>(mode, false, false, false, ...args);
   }
 
-  setOverlayMode<THandler extends UiHandler = never>(
+  public setOverlayMode<THandler extends UiHandler = never>(
```
flx-sta, line 809: NIT: tsdoc

src/ui/ui.ts
```diff
@@ -634,7 +833,18 @@ export class UI extends Phaser.GameObjects.Container {
...
+    if (alwaysDestroyModes.includes(this.mode)) {
+      this.deleteUiHandlers(this.mode);
+    } else if (this.mode === UiMode.MENU) {
+      // When stopping the menu destroy all handlers that depend on it
+      this.deleteUiHandlers(mainMenuAccessedModes);
+    }
+  }
+
+  // UNUSED?
+  public revertModes(): Promise<void> {
```
flx-sta, line 946: NIT: tsdoc

test/test-utils/mocks/mock-console.ts
flx-sta, line 1: Should probably consider using `chalk` instead of all those ansii codes
<sub> Probably out of scope </sub>

src/enums/battle-scene-event-type.ts
DayKev:
```suggestion
   * Triggers after a run completes via winning or losing (but not through "Save & Quit").
```
I'd go with this.

src/phases/post-game-over-phase.ts
DayKev: I think it probably makes sense to transition to `eventBus`. Updated the TSDoc for `BattleScene#eventTarget` to specify as such:
[0001-Update-TSDocs-for-BattleScene-eventTarget.patch](https://github.com/user-attachments/files/20720982/0001-Update-TSDocs-for-BattleScene-eventTarget.patch)
<details><summary>Contents of the above patch file</summary>
<p>

```patch
From 8e0003d6729d42749d4c759a1f7c397e034bcbba Mon Sep 17 00:00:00 2001
From: NightKev <34855794+DayKev@users.noreply.github.com>
Date: Thu, 12 Jun 2025 23:42:13 -0700
Subject: [PATCH] Update TSDocs for `BattleScene#eventTarget`

---
 src/battle-scene.ts | 8 ++------
 1 file changed, 2 insertions(+), 6 deletions(-)

diff --git a/src/battle-scene.ts b/src/battle-scene.ts
index 78f4e79ba..48212ce6e 100644
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -270,12 +270,8 @@ export default class BattleScene extends SceneBase {
 
   /**
    * Allows subscribers to listen for events
-   *
-   * Current Events:
-   * - {@linkcode BattleSceneEventType.MOVE_USED} {@linkcode MoveUsedEvent}
-   * - {@linkcode BattleSceneEventType.TURN_INIT} {@linkcode TurnInitEvent}
-   * - {@linkcode BattleSceneEventType.TURN_END} {@linkcode TurnEndEvent}
-   * - {@linkcode BattleSceneEventType.NEW_ARENA} {@linkcode NewArenaEvent}
+   * @see `src/events/`
+   * @todo Migrate to using {@linkcode eventBus} instead
    */
   public readonly eventTarget: EventTarget = new EventTarget();
 
-- 
2.49.0.windows.1


```

</p>
</details>

src/ui/handlers/game-stats-ui-handler.ts
DayKev: Good point, just checked and these values are only used in this file so I moved them out of the class.
[0001-Move-constants-outside-of-GameStatsUiHandler.patch](https://github.com/user-attachments/files/20721210/0001-Move-constants-outside-of-GameStatsUiHandler.patch)
<details><summary>Contents of patch file</summary>
<p>

```patch
From 26441cdf009e2d5b14b791e5d6507836275386e7 Mon Sep 17 00:00:00 2001
From: NightKev <34855794+DayKev@users.noreply.github.com>
Date: Thu, 12 Jun 2025 23:59:48 -0700
Subject: [PATCH] Move constants outside of `GameStatsUiHandler`

---
 src/ui/handlers/game-stats-ui-handler.ts | 38 +++++++++++-------------
 1 file changed, 18 insertions(+), 20 deletions(-)

diff --git a/src/ui/handlers/game-stats-ui-handler.ts b/src/ui/handlers/game-stats-ui-handler.ts
index f9642440a..4b863973a 100644
--- a/src/ui/handlers/game-stats-ui-handler.ts
+++ b/src/ui/handlers/game-stats-ui-handler.ts
@@ -215,6 +215,11 @@ const displayStats: DisplayStats = {
   },
 };
 
+const ROWS_ON_SCREEN = 9;
+const NUM_COLUMNS = 2;
+const MAX_STATS_ON_SCREEN = ROWS_ON_SCREEN - NUM_COLUMNS;
+const MAX_CURSOR = Math.ceil((Object.keys(displayStats).length - MAX_STATS_ON_SCREEN) / NUM_COLUMNS);
+
 /**
  * Ui Handler for {@linkcode UiMode.GAME_STATS}.
  * Shows a list of account statistics over two columns.
@@ -223,13 +228,6 @@ const displayStats: DisplayStats = {
  * meaning the playtime stat won't change until the handler is exited then re-opened.
  */
 export class GameStatsUiHandler extends UiHandler {
-  private readonly ROWS_ON_SCREEN = 9;
-  private readonly NUM_COLUMNS = 2;
-  private readonly MAX_STATS_ON_SCREEN = this.ROWS_ON_SCREEN * this.NUM_COLUMNS;
-  private readonly MAX_CURSOR = Math.ceil(
-    (Object.keys(displayStats).length - this.MAX_STATS_ON_SCREEN) / this.NUM_COLUMNS,
-  );
-
   private gameStatsContainer: Phaser.GameObjects.Container;
 
   private statLabels: TextListContainer[];
@@ -259,25 +257,25 @@ export class GameStatsUiHandler extends UiHandler {
       .setPositionRelative(headerBg, 8, 4);
     this.gameStatsContainer.add(headerText);
 
-    const statsBgWidth = Math.floor((GAME_WIDTH - 2) / this.NUM_COLUMNS);
+    const statsBgWidth = Math.floor((GAME_WIDTH - 2) / NUM_COLUMNS);
     const statsBgHeight = Math.floor(GAME_HEIGHT - headerBg.height - 2);
 
-    for (let i = 0; i < this.NUM_COLUMNS; i++) {
+    for (let i = 0; i < NUM_COLUMNS; i++) {
       const xPosition = (statsBgWidth - 2) * i;
       const yPosition = headerBg.height;
-      const width = statsBgWidth + (i > 0 ? 2 : 0) + (i < this.NUM_COLUMNS - 1 ? 2 : 0);
+      const width = statsBgWidth + (i > 0 ? 2 : 0) + (i < NUM_COLUMNS - 1 ? 2 : 0);
       // Create the background window for each panel
       const statsBg = addWindow(xPosition, yPosition, width, statsBgHeight, false, false, i > 0 ? -3 : 0, 1);
       statsBg.setOrigin(0, 0);
 
       const statY = statsBg.y + 5;
       // Create a single text object for all labels to save on resources
-      const statsLabels = new TextListContainer(statsBg.x + 8, statY, TextStyle.STATS_LABEL, this.ROWS_ON_SCREEN);
+      const statsLabels = new TextListContainer(statsBg.x + 8, statY, TextStyle.STATS_LABEL, ROWS_ON_SCREEN);
       this.statLabels.push(statsLabels);
 
       // Create a single text object for all values to save on resources
       const statX = statsBg.x + statsBgWidth - 5;
-      const statsValues = new TextListContainer(statX, statY, TextStyle.STATS_VALUE, this.ROWS_ON_SCREEN, {
+      const statsValues = new TextListContainer(statX, statY, TextStyle.STATS_VALUE, ROWS_ON_SCREEN, {
         textAlign: "right",
       });
       this.statValues.push(statsValues);
@@ -327,9 +325,9 @@ export class GameStatsUiHandler extends UiHandler {
    */
   private initStatsDisplay() {
     // Init arrays containing labels and values for each column of statistics
-    const labels = new Array(this.NUM_COLUMNS);
-    const values = new Array(this.NUM_COLUMNS);
-    for (let col = 0; col < this.NUM_COLUMNS; col++) {
+    const labels = new Array(NUM_COLUMNS);
+    const values = new Array(NUM_COLUMNS);
+    for (let col = 0; col < NUM_COLUMNS; col++) {
       values[col] = [];
       labels[col] = [];
       this.statValues[col].setList(values[col]);
@@ -339,7 +337,7 @@ export class GameStatsUiHandler extends UiHandler {
     // Fill the arrays with computed stats and labels for each statistic
     Object.keys(displayStats).forEach((key, i) => {
       const stat = displayStats[key] as DisplayStat;
-      const column = i % this.NUM_COLUMNS;
+      const column = i % NUM_COLUMNS;
       const value = stat.sourceFunc(globalScene.gameData);
       const showStat = !stat.hidden || Number.isNaN(Number.parseInt(value)) || Number.parseInt(value);
       labels[column].push(showStat ? i18next.t(`gameStatsUiHandler:${stat.label_key}`) : "???");
@@ -351,7 +349,7 @@ export class GameStatsUiHandler extends UiHandler {
    * Update the stats currently on screen.
    */
   private updateStats(): void {
-    for (let col = 0; col < this.NUM_COLUMNS; col++) {
+    for (let col = 0; col < NUM_COLUMNS; col++) {
       this.statLabels[col].setCursor(this.cursor);
       this.statValues[col].setCursor(this.cursor);
     }
@@ -364,7 +362,7 @@ export class GameStatsUiHandler extends UiHandler {
     const showUpArrow = this.cursor > 0;
     this.arrowUp.setVisible(showUpArrow);
 
-    const showDownArrow = this.cursor < this.MAX_CURSOR;
+    const showDownArrow = this.cursor < MAX_CURSOR;
     this.arrowDown.setVisible(showDownArrow);
   }
 
@@ -382,11 +380,11 @@ export class GameStatsUiHandler extends UiHandler {
           if (this.cursor > 0) {
             success = this.setCursor(this.cursor - 1);
           } else if (this.cursor === 0) {
-            success = this.setCursor(this.MAX_CURSOR);
+            success = this.setCursor(MAX_CURSOR);
           }
           break;
         case Button.DOWN:
-          if (this.cursor < this.MAX_CURSOR) {
+          if (this.cursor < MAX_CURSOR) {
             success = this.setCursor(this.cursor + 1);
           } else {
             success = this.setCursor(0);
-- 
2.49.0.windows.1


```

</p>
</details>

src/ui/handlers/menu-ui-handler.ts
```diff
@@ -517,6 +511,7 @@ export class MenuUiHandler extends OptionSelectUiHandler {
       options: communityOptions,
       xOffset: this.optionSelectBg.displayWidth,
       yOffset: this.menuMessageBox.displayHeight + 1,
+      onResize: (w) => (this.communityWindowWidth = w),
```
DayKev:
```suggestion
      onResize: (w) => {
        this.communityWindowWidth = w;
        },
```
Resolves the [`lint/suspicious/noAssignInExpressions`](https://biomejs.dev/linter/rules/no-assign-in-expressions/) warning.

src/ui/handlers/message-ui-handler.ts
```diff
@@ -297,6 +297,14 @@ export abstract class MessageUiHandler extends AwaitableUiHandler {
...
       this.message.setText("");
     }
+    if (this.textTimer) {
+      this.textTimer.remove();
+      this.textTimer = null;
+    }
+    if (this.textCallbackTimer) {
+      this.textCallbackTimer.remove();
+      this.textCallbackTimer = null;
+    }
```
DayKev: I think the `if`s can be removed by doing this:
```suggestion
    this.textTimer?.remove();
    this.textTimer = null;
    this.textCallbackTimer?.remove();
    this.textCallbackTimer = null;
```

src/ui/ui.ts
```diff
@@ -259,23 +441,26 @@ export class UI extends Phaser.GameObjects.Container {
...
 
-  getCurrentHandler<H extends UiHandler = UiHandler>(): H {
-    return this.handlers[this.mode] as H;
+  public getCurrentHandler<H extends UiHandler = UiHandler>(): H {
+    return this.handlers.get(this.mode) as H;
   }
 
-  getMessageHandler(): BattleMessageUiHandler {
-    return this.handlers[UiMode.MESSAGE] as BattleMessageUiHandler;
+  public getMessageHandler(): BattleMessageUiHandler | undefined {
```
DayKev, line 456: What's the plan for this method in the future?

src/ui/ui.ts
```diff
@@ -128,79 +173,206 @@ export class UI extends Phaser.GameObjects.Container {
...
-      handler.initialize();
+
+    // Initialize the default handler
+    this.addUiHandler(this.mode);
+    // The settings Ui handlers are inefficient and take a while to initialize
+    // so for now we initialize them all during loading and never stop them.
+    for (const mode of settingsUiModes) {
+      this.addUiHandler(mode);
     }
+    // If the handler gets initialized just before being shown the input text flashes
```
DayKev:
```suggestion
    // If this handler gets initialized just before being shown the input text flashes
```

src/ui/ui.ts
```diff
@@ -128,79 +173,206 @@ export class UI extends Phaser.GameObjects.Container {
...
-
     globalScene.uiContainer.add(this.savingIcon);
+
+    // Register listener to new encounter events and run end events
+    globalScene.eventTarget.addEventListener(BattleSceneEventType.ENCOUNTER_PHASE, this.onNextEncounterEvent);
+    globalScene.eventTarget.addEventListener(BattleSceneEventType.POST_GAME_OVER, this.onGameOverEvent);
+  }
+
+  /**
+   * Instanciate a new UiHandler corresponding the the given mode.
```
DayKev:
```suggestion
   * Instantiate a new UiHandler corresponding the the given mode.
```

src/ui/ui.ts
```diff
@@ -211,10 +383,15 @@ export class UI extends Phaser.GameObjects.Container {
   public resetHandlers(): void {
     this.mode = DEFAULT_MODE;
     const currentHandler = this.getCurrentHandler();
-    for (const handler of this.handlers.filter((h) => h.active && h !== currentHandler)) {
-      handler.stop();
+    this.handlers.forEach((handler) => {
+      if (handler.active && handler !== currentHandler) {
+        handler.stop();
+      }
+    });
```
DayKev: I'd prefer to keep it this way (so that `handler` is defined as `const`) like it was originally.
```suggestion
    for (const handler of this.handlers) {
      if (handler.active && handler !== currentHandler) {
        handler.stop();
      }
    }
```

src/ui/ui.ts
```diff
@@ -259,23 +441,26 @@ export class UI extends Phaser.GameObjects.Container {
     globalScene.uiContainer.add(this.tooltipContainer);
   }
 
-  getCurrentHandler<H extends UiHandler = UiHandler>(): H {
-    return this.handlers[this.mode] as H;
+  public getCurrentHandler<H extends UiHandler = UiHandler>(): H {
```
DayKev, line 448: Should this return `| undefined` now?

.env.test
MokaStitcher: Yes.
I believe knowing what UI modes get set is almost as important as knowing Phase changes, which always get logged.
In tests in particular it seems like when Ui Handlers fail it doesn't usually result in the game crashing and the test failing immediately. Sometimes the test will fail at a later point that will make it hard to pinpoint that it was originally caused by a UI bug (this happened recently in the enum to const PR part 2). I hope this logging can help solve these issues better.
The other reason to have it enabled in tests is because UiMode changes are already being logged through the PhaseInterceptor
https://github.com/Despair-Games/poketernity/blob/8599cd1d2a0926e6bf13e985dff45c61ca934aa2/test/test-utils/phase-interceptor.ts#L358
... except it's only partial because it logs `ui.setMode` but not `ui.setOverlayMode`, `ui.setModeWithoutClear` or `ui.revertMode`. I removed that line in the PR and now the logging is done by the UI itself, and covers all mode changes.

The main reason I put this through an env variable rather than always enabled is because setMode(MESSAGE) is called a lot, so it could get overwhelming.
The other reason is because I also wanted to log handlers creation and destruction.
This gave me an idea though, and I've added two levels to the logging now, by default only the mode changes will be logged in test with the value of 1

test/test-utils/mocks/mock-console.ts
DayKev: Yeah, something to update in a separate PR.

src/phases/post-game-over-phase.ts
MokaStitcher: I agree about migrating but it's not for this PR to do, good point about adding the todo.
And since I was using an already existing BattleSceneEvent for the encounter phase it made sense to me to put the game over event as one as well, for consistency's sake

src/ui/ui.ts
DayKev: Agreed, in fact it's probably a good idea to go through the codebase and find more places this can be done easily (this pattern is most of the cause of the need for the `coerceArray` utility function iirc). Will add an issue for it.

Edit: Issue added #1256

src/phases/scan-ivs-phase.ts
MokaStitcher: oh yeah if we want to add missing catches everywhere that's definitely a gargantuan task, and out of scope for this :')
<sub>This reminded me I needed to create an issue to move those things out of the message handler though

src/ui/components/text-list-container.ts
MokaStitcher: why wouldn't it need doc? 
image

DayKev: There is a doc missing for the `textStyle` param. Not sure if that's what flx was referring to?

src/ui/handlers/game-stats-ui-handler.ts
MokaStitcher: 1) There is and always will be at most a single instance of GameStatsUiHandler in memory at once. This is why I put the values in it, rather than say in ui constants which would mean they would always be in memory for no reason.

2) Ah I see, knowing what I said in the point above though I'd appreciate both of your opinion on what's best to do, keeping them as is with updated case, changing them to static, or DayKev's suggestion. This will be useful to know for future handlers rework too

test/test-utils/mocks/mock-console.ts
MokaStitcher: definitely out of scope

src/ui/ui.ts
MokaStitcher: I did it this way cause handlers is a Map now so it can't be iterated like this, but iterating over `handlers.values()` should work and I prefer it too

src/ui/ui.ts
MokaStitcher: hmm, the handler for the current mode should always exist, except maybe during destruction. Even if it's the case it would only cause issues I think if another handler tried to access the current handler when being cleared. Which shouldn't happen because handlers get cleared when they are the current handler and about to get replaced, so it wouldn't make sense for them to call `getCurrentHandler` at that point.
I'm pretty sure we shouldn't have the same issues as with the MessageHandler so for now I'd rather work under the assumption it is always defined. If issues arise we'll have to figure something out because making it potentially undefined would make everything way more cumbersome, all to handle a single edge case, like with the MessageHandler

src/ui/handlers/game-stats-ui-handler.ts
DayKev: Hm, true.

I'd rather move them out than make them static fields. Keeping them as-is and just renaming to change the casing is fine too.
There's no need to export them from the UI constants file (they have no use outside this class right?).

src/ui/ui.ts
MokaStitcher: <sub>like everything related to text and dialogue, the plan is to cry

More seriously, not entirely sure what you mean
I think ideally we would get rid of it and completely rework the message handler, because the way it's being used is not great. But that impacts pretty much everything so it's a long way off.

DayKev: Was wondering if adding something like `/** @deprecated Planned to be removed in the future */`  made sense.

src/ui/ui.ts
innerthunder: Why was this resolved? "Instantiate" is correct.

DayKev: Moka just hasn't pushed the changes yet I think.

src/ui/ui.ts
MokaStitcher: The whole text and dialogue thing if thoroughly undocumented and has a ton of parameters which are extremely confusing. This function is basically a buffer for calling `showDialogue` on the currently active MessageUiHandler, but that's as much as I can tell considering MessageUiHandler is not documented either.
I think this will have to be for a text/dialogue refactor, because it is a whole mess, sorry

src/ui/ui.ts
```diff
@@ -128,79 +173,206 @@ export class UI extends Phaser.GameObjects.Container {
...
+    globalScene.eventTarget.addEventListener(BattleSceneEventType.ENCOUNTER_PHASE, this.onNextEncounterEvent);
+    globalScene.eventTarget.addEventListener(BattleSceneEventType.POST_GAME_OVER, this.onGameOverEvent);
+  }
+
+  /**
+   * Instanciate a new UiHandler corresponding the the given mode.
+   * @param mode - The {@linkcode UiMode} to consider
+   * @returns the {@linkcode UiHandler} for that mode
+   */
+  private createUiHandler(mode: UiMode): UiHandler {
```
innerthunder, line 271: Hmm... This may benefit from the `PhaseKey` pattern in #1250:

```ts
// At the top of this file
const UI_HANDLERS = {
  BattleMessageUiHandler,
  TitleUiHandler,
  CommandUiHandler,
  // ...
} as const;

export type UiHandlerMap = typeof UI_HANDLERS;

// can be in this file or in a separate @types file, depending on how widely this is used

export type UiHandlerKey = keyof UiHandlerMap;
```

Then this method can be changed to

```ts
private createUiHandler<K extends UiHandlerKey>(mode: K): UiHandler {
  const UiHandlerClass = UI_HANDLERS[mode];
  return new UiHandlerClass();
}
```

With this you'd have a more strongly typed `mode` input, and you could probably get rid of the `UiMode` enum as a whole.

MokaStitcher: this is interesting, thank you. Unfortunately one of the constructors requires a parameter, but that can probably be solved.
This could probably maybe the `setMode` functions a bit cleaner/more robust too

I think I'd like to look in to this, but probably in a different PR if that's fine (would be the next one after this one most likely)

innerthunder: The only problem with this from what I can tell is `OptionsSelectUiHandler` being used for multiple modes. This could be fixed by turning the "mode" of `OptionSelectUiHandler` into a constructor parameter, then adding:

```ts
// in the same file as 'UiHandlerKey'
export type UiHandlerConstructorMap = {
  [K in UiHandlerKey]: ConstructorParameters<UiHandlerMap[K]>;
}

// in this file

private createUiHandler<K extends UiHandlerKey>(
  mode: K,
  ...params: UiHandlerConstructorMap[K]
): UiHandlerMap[K] {
  const UiHandlerClass = UI_HANDLERS[mode];
  return new UiHandlerClass(...params);
}
```

<sub>I'd rather not add this for a single UI handler, but it works as a quick fix</sub>

MokaStitcher: I think we can actually completely remove the MenuOptionSelect mode and use the standard option select one, which would solve this issue
This was brought up in another PR but since I wasn't 100% sure of it it was left in for now. This is a good opportunity to look more closely into it I suppose :p

src/ui/ui.ts
```diff
@@ -386,7 +578,13 @@ export class UI extends Phaser.GameObjects.Container {
     }
   }
 
-  shouldSkipDialogue(i18nKey: string): boolean {
+  /**
+   * Check whether a dialogueshould be shown or not, based on the "skip seen dialogues" setting.
```
DayKev:
```suggestion
   * Check whether a dialogue should be shown or not, based on the "skip seen dialogues" setting.
```

src/ui/components/text-list-container.ts
flx-sta: I didn't notice that, no.

MokaStitcher: > There is a doc missing for the `textStyle` param. Not sure if that's what flx was referring to?

oh right, I changed the signature but not the doc :plains: -- fixed

src/ui/ui.ts
MokaStitcher: I'm not sure about deprecating it since there's currently no alternative, and we're not fully sure how/when/if it will changed. I'll add a todo for now I think

src/ui/ui.ts
MokaStitcher: Indeed, I usually do resolve things as I go and still had a bunch of tsdoc to write after this so it took a while longer to push everything :plains:

src/ui/ui.ts
flx-sta: How about just leveraging AI to write some stuff based on what's going on in there?

```suggestion
  /**
   * Displays a dialogue box with the given text and character name, handling localization, message paging, 
   * and dialogue-skipping logic. If the text contains `$`, it will be split into multiple dialogue pages.
   * If the keyOrText matches a translation key, the localized text will be used and optionally skipped if already seen.
   *
   * @param keyOrText - The dialogue text to display, or a localization key.
   * @param name - The name of the speaking character. If undefined, no name is shown.
   * @param delay - Optional. The delay before showing the dialogue, in milliseconds.
   * @param callback - Function to call after the dialogue (and all pages, if any) has been shown.
   * @param callbackDelay - Optional. Delay before calling the callback, in milliseconds.
   * @param promptDelay - Optional. Delay before showing the prompt to continue, in milliseconds.
   */
  public showDialogue(
```

src/ui/components/text-list-container.ts
```diff
@@ -0,0 +1,126 @@
...
+  private maxLines: number;
+
+  private textObject: Phaser.GameObjects.Text | BBCodeText;
+
+  /**
+   * @param x - The x position of the TextObject in its parent container.
+   * @param y - The y position of the TextObject in its parent container.
+   * @param textStyle - The {@linkcode TextStyle} to use.
+   * @param maxLines - The maximum number of lines to show at one.
+   * @param options - (Optionnal) Extra text styling options {@linkcode TextListContainerOptions}.
```
DayKev:
```suggestion
   * @param options - (Optional) Extra text styling options {@linkcode TextListContainerOptions}.
```

src/ui/components/text-list-container.ts
```diff
@@ -0,0 +1,126 @@
...
+import type BBCodeText from "phaser3-rex-plugins/plugins/bbcodetext";
+
+interface TextListContainerOptions {
+  /**
+   * Whether to use a {@linkcode BBCodeText} object instead of a basic {@linkcode TextObject}.
+   * You should only set this to `true` if BBCode is needed. The BBCode should be part of the provided strings.
+   */
+  useBBCode?: boolean;
+  /**
+   * Optional alignement for the text.
```
DayKev:
```suggestion
   * Optional alignment for the text.
```

DayKev (approved):

src/ui/ui.ts
DayKev: Bad/misleading docs are worse than no docs. Those docs would have to be investigated/verified, and this isn't even a function being modified by the PR.

src/ui/ui.ts
```diff
@@ -460,28 +674,33 @@ export class UI extends Phaser.GameObjects.Container {
...
     globalScene.audioManager.playSound("ui/error");
   }
 
-  fadeOut(duration: number): Promise<void> {
+  /**
+   * Fade out the Ui container to make it appear black.
+   * @param duration - duration of the fading animation, in ms.
+   * @returns Promise that resolves after the animation is done.
+   */
+  public fadeOut(duration: number): Promise<void> {
```
innerthunder, line 703: Should this and other methods returning `Promise<void>` be `async`?

innerthunder (approved):

src/ui/ui.ts
flx-sta: _They don't have to be, no.
They just have to be if they are using `await` too._

**It would probably not hurt though since it immediately adds awareness of the async behavior of the function.**

PR #1232, opened 2025-05-31
[Refactor] Pokemon IVs are now 6 randomly generated numbers

DayKev:
N/A

This weird method of generating IVs is not what mainline does, so there's no reason to keep it.

`generateIvsFromId` function removed and `Pokemon#generateIvs` added to replace it, which generates an array of 6 random numbers (each between `0-31` inclusive) using the Pokémon's `randSeedIntRange` method.

## Checklist


src/field/pokemon.ts
```diff
@@ -473,9 +473,15 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
    */
-  generateId(): void {
-    this.id = randSeedInt(4294967296);
+  public generateId(): void {
+    this.id = randSeedInt(4294967295);
+  }
+
+  /** @returns An array of 6 random numbers, each between `0-31` inclusive */
+  public generateIvs(): number[] {
+    return new Array(6).fill(null).map(() => this.randSeedIntRange(0, 31));
```
flx-sta: Don’t we have `IV_MIN` and `IV_MAX` constants ?

flx-sta (commented):
Question: the usage of rand**Seed**Int would guarantee the IV to be the same after every reload, right ?

src/field/pokemon.ts
DayKev: Oh yeah, I'll put those in.

DayKev:
> Question: the usage of rand**Seed**Int would guarantee the IV to be the same after every reload, right ?

It should, afaik.

flx-sta (approved):

src/data/egg.ts
```diff
@@ -265,7 +265,7 @@ export class Egg {
       ret.shiny = this._isShiny;
       ret.variant = this._variantTier;
 
-      const secondaryIvs = getIvsFromId();
+      const secondaryIvs = ret.generateIvs();
```
MokaStitcher, line 268: hmm, would this means inconsistent egg IVs if say you refresh after a hatch and make it hatch again on a different run?
Maybe it was already the case before, depending when the Id gets generated 🤔

DayKev: It should be the same [in]consistent behavior as before I think.

MokaStitcher (approved):

PR #1234, opened 2025-06-01
[Test] `MoveHelper#changeMoveset` disables moveset overrides

DayKev:
N/A

To prevent mistakes when writing tests.

The helper function used to manually modify a Pokémon's moveset in tests will now disable the matching override.

## Checklist


flx-sta:
What's the use case for this over simply forcing a move?

DayKev:
If you need to track the properties of the Pokémon's moveset over multiple turns (Sketch for example).

flx-sta (approved):

innerthunder (approved):

Issue #489, opened 2025-02-11
Bug: Summoning abilities don't activate when reloading a session.


Opaque02:
Copied over from PKR 1005

When reloading a session (battle), the game only activates summoning abilities like Drought, Intimidate, etc if they meet one of these 3 criteria:

- The battle is a trainer battle.
- It's the first wave of a biome for classic mode standards (1, 11, 21, etc)
- The battle wasn't loaded from the menu 

This logic doesn't include all scenarios where summoning abilities should trigger, like going from a single to a double battle or changing biomes in endless where it can happen on nearly any floor.

It's also built on a flaw: When reloading a session, all Pokemon on the field are sent out, including the ones who were supposed to already be there.

To fix this issue, I propose storing the current allied Pokemons on the field when saving a session. Then, when loading it, only activate summoning abilities of the Pokemon who weren't on the field when the session was saved.

This, together with #694 should fix most if not all bugs pertaining to reloading sessions.

Drought and Psychic surge are both activated when switching biomes in endless:
![image](https://github.com/pagefaultgames/pokerogue/assets/63069680/a0e25021-6c0b-4387-befc-e01455598538)

After reloading the fight neither activate:
![image](https://github.com/pagefaultgames/pokerogue/assets/63069680/fcda4478-9f02-4c8f-a71a-cdc44a616c54)

PR #1236, opened 2025-06-02
refactor: Rewrite logic to recall, switch, and summon Pokemon

### LINKED ISSUE #494: [BUG] Revival Blessing Breaks Double Battles
Copied over from PKR 2769

**Describe the bug**

When battling against two trainers in a double battle, if a Pokemon is revived with revival blessing while a slot is open, it instantly returns to the field. At the end of the turn, another pokemon is sent out, making the revived Pokemon untargetable, eventually leading to only attacking ones own Pokemon.

**Pokerogue PR #5141**

**To Reproduce**

1. Go to a trainer double battle
2. Defeat One Pokemon before the Revival Blessing Mon acts and leave it up, if it clicks Revival Blessing, the bug occurs
3. Make sure Trainer has an extra Pokemon to throw out.
4. See extra mon Error
5. Defeat all Pokemon
6. See Unwinnable Error

**Expected behavior**

When a Pokemon is revived, either it doesn't instantly return to the field, or doesn't allow for a second Pokemon to be added on top of it at end of turn.

Screenshots / Videos**

https://github.com/pagefaultgames/pokerogue/assets/174589289/e0c6c1a5-62f7-482e-add2-b034c35b9e19

Screenshot_20240703_164806_Samsung Internet

https://github.com/pagefaultgames/pokerogue/assets/174589289/709ab987-ee63-43cc-8f2d-6f7158646fdb

**Device**

Playing on Samsung Internet on a Galaxy [S24]

**Additional context**

### LINKED ISSUE #1207: [Refactor] Switching behavior
Requirements: https://github.com/Despair-Games/poketernity/issues/1207#issuecomment-2897039900
Code organization: https://github.com/Despair-Games/poketernity/issues/1207#issuecomment-2897134578

innerthunder:

- On the prompt to switch Pokemon at the start of a battle, if the Player selects "Yes", then cancels out of the Party menu without selecting a Pokemon, the game will now return to the switch prompt.
- Some minor improvements to animation timing

The current phase hierarchy for recalling, switching, and summoning Pokemon (namely `ReturnPhase`, `SwitchPhase`, `SwitchSummonPhase`, and `SummonPhase`) is extremely convoluted. Logic and data manipulation is woven in between animations composed of hanging Promises and other unmarked async calls. This PR aims to improve the readability and stability of the switch sequence by
- Separating logical and visual elements and reducing duplicate code with helper methods
- Centralizing the logical elements of the switch sequence within a single phase
- Promisifying all animations for the switch sequence so that they may more explicitly be run in sequence or in parallel.

- Removed `SwitchSummonPhase` and `ReturnPhase`
- Added `RecallPhase`, which plays animations for recalling a Pokemon and removes the Pokemon from the field and is essentially a replacement for `ReturnPhase`
- Rewrote `SwitchPhase` to handle all logical aspects of switching, including:
  - Resolving the index of the Pokemon to switch in
  - Applying all changes to game data as a result of the switch, including pre-switch Ability triggers and data transfers from Baton Pass and Shed Tail.
- Rewrote `SummonPhase` to only contain visual effects related to summoning a Pokemon (WIP)
- In `BattleScene`: Added a helper method to prompt the Player to select a Pokemon from their party. This is currently only used in `SwitchPhase`, but may be applied to other instances where the Party UI is required.

The end result: Instead of the `SummonPhase` > `SwitchSummonPhase` > `ReturnPhase` hierarchy, each phase in the `RecallPhase` > `SwitchPhase` > `SummonPhase` sequence is independent and can be scheduled independently. In theory, every  sequence where Pokemon are recalled, switched, and/or summoned can be implemented by a combination of one or more of the new phases.

## How to test the changes?

The changes in this PR are mainly visual and therefore require manual playtesting. Changes can be tested at https://poketernity.com/recall-switch-summon-v1/

## Checklist


innerthunder:
A few follow-up tasks I'm considering:
- Consolidate and simplify switch-out logic within Wimp Out / Emergency Exit and switching/phasing moves (we could look into porting https://github.com/pagefaultgames/pokerogue/pull/5825 for this)
- Fix logic for applying post-summon effects on reload (See #489, the current logic is oversimplified and may require some changes to save data to be fully accurate)
- Simplify Speed Boost implementation (`switchedInThisTurn` is redundant with `turnCount` / `waveTurnCount`)
- Investigate `switchOutStatus` Pokemon field (current implementation's pretty jank)

src/phase-manager.ts
```diff
@@ -628,6 +632,46 @@ export class PhaseManager {
...
+      case "before":
+        validatePhaseId();
+        this.prependToPhase(phaseKey!, ...phases);
+        break;
+      case "after":
+        validatePhaseId();
+        this.prependToPhase(phaseKey!, ...phases);
+        break;
+      default:
+        throw new Error(`Unknown queueBattlerSwitchOut.when: ${when}`);
```
DayKev: I don't think this is necessary, `when` already has the type union of `"eager" | "before" | "after"` so trying to pass a different value will result in TS emitting an error.
```suggestion
```

src/phases/base/pokemon-phase.ts
```diff
@@ -32,6 +33,15 @@ export abstract class PokemonPhase extends FieldPhase {
     return pokemon!;
   }
 
+  public getTrainerSlot(): TrainerSlot | undefined {
```
DayKev: Needs TSDoc.

src/phases/summon-phase.ts
```diff
@@ -1,120 +1,256 @@
...
 import { settings } from "#system/settings-manager";
 import type { PhaseKey } from "#types/phase-types";
+import { playTween } from "#utils/anim-utils";
 import i18next from "i18next";
 
-export class SummonPhase extends PartyMemberPokemonPhase {
+interface SummonPhaseOptions {
+  loaded?: boolean;
+  playTrainerAnim?: boolean;
+  delayPostSummon?: boolean;
```
DayKev: These should have TSDocs copied from the class fields.

src/phases/mystery-encounter-phases/battle-phase.ts
```diff
@@ -164,20 +167,20 @@ export class MysteryEncounterBattlePhase extends Phase {
...
         if (!availablePartyMembers[1].isOnField()) {
-          globalScene.phaseManager.createAndPushPhase("SummonPhase", 1);
+          globalScene.phaseManager.createAndPushPhase("SummonPhase", BattlerIndex.PLAYER_2, { delayPostSummon: true });
         }
       }
     } else {
       if (availablePartyMembers.length > 1 && availablePartyMembers[1].isOnField()) {
         globalScene.getPlayerField().forEach((pokemon) => pokemon.lapseTag(BattlerTagType.COMMANDED));
-        globalScene.phaseManager.createAndPushPhase("ReturnPhase", 1);
+        globalScene.phaseManager.createAndPushPhase("RecallPhase", 1);
```
DayKev: This should also be changed to `BattlerIndex.PLAYER_2` as well probably, to match the other changes.

src/phases/post-turn-status-effect-phase.ts
```diff
@@ -61,7 +60,6 @@ export class PostTurnStatusEffectPhase extends PokemonPhase {
     if (damage.value) {
       // Set preventEndure flag to avoid pokemon surviving thanks to focus band, sturdy, endure ...
       pokemon.damageAndUpdate(toDmgValue(damage.value), { preventEndure: true });
-      applyAbAttrs<PostDamageAbAttr>(AbAttrFlag.POST_DAMAGE, pokemon, false, damage.value);
```
DayKev, line 64: Is this handled somewhere else now or..?

DayKev (commented):
#1280

src/phases/post-turn-status-effect-phase.ts
innerthunder: It's already handled within `damageAndUpdate`. For some reason, it was applied twice here... I'm not sure what side effects came from that 😅

src/data/abilities/ab-attrs/post-damage-force-switch-ab-attr.ts
```diff
@@ -144,37 +149,28 @@ class ForceSwitchOutHelper {
...
        * For non-wild battles, it checks if the opposing party has any available Pokémon to switch in.
-       * If yes, the Pokémon leaves the field and a new SwitchSummonPhase is initiated.
+       * If yes, the Pokémon leaves the field and a new switch sequence is initiated.
        */
     } else if (battleType !== BattleType.WILD) {
       if (globalScene.getEnemyParty().filter((p) => p.isAllowedInBattle() && !p.isOnField()).length < 1) {
         return false;
       }
       if (switchOutTarget.hp > 0) {
         switchOutTarget.leaveField(this.switchType === SwitchType.SWITCH);
```
MokaStitcher: this `switchOutTarget.leaveField` is untouched while the one above (line 147) was removed, is it intended?

src/data/battler-tags/battler-tag.ts
```diff
@@ -158,8 +158,10 @@ export class BattlerTag {
    * Helper function that retrieves the source Pokemon object
    * @returns The source {@linkcode Pokemon} or `null` if none is found
```
MokaStitcher:
```suggestion
   * @returns The source {@linkcode Pokemon} or `undefined` if none is found
```

src/phase-manager.ts
```diff
@@ -628,6 +632,44 @@ export class PhaseManager {
...
+      case "eager":
+        this.unshiftPhase(...phases);
+        break;
+      case "before":
+        validatePhaseId();
+        this.prependToPhase(phaseKey!, ...phases);
+        break;
+      case "after":
+        validatePhaseId();
+        this.prependToPhase(phaseKey!, ...phases);
```
MokaStitcher: both "before" and "after" use `prependToPhase`, should it be this?
```suggestion
        this.appendToPhase(phaseKey!, ...phases);
```

src/phases/return-phase.ts
```diff
@@ -1,30 +0,0 @@
...
-  protected override onEnd(): void {
-    const pokemon = this.getPokemon();
-
-    pokemon.resetSprite();
-    pokemon.resetTurnData();
-    pokemon.resetSummonData();
-
-    globalScene.updateFieldScale();
-
-    globalScene.triggerPokemonFormChange(pokemon, SpeciesFormChangeActiveTrigger);
```
MokaStitcher, line 28: Have these been moved somewhere else?

From what I know of it, updateFieldScale is needed if the Pokemon that is being recalled/fainted was GMax, to put back the non-zoomed out field scale
The form change is I assume to reset battle specific forms so that they are not kept for the next time the pokemon is summoned (for mons like Cramorant or Aegislash)

src/phases/recall-phase.ts
```diff
@@ -0,0 +1,113 @@
...
+import { BattlerTagType } from "#enums/battler-tag-type";
+import { SwitchType } from "#enums/switch-type";
+import type { Pokemon } from "#field/pokemon";
+import { PokemonPhase } from "#phases/base/pokemon-phase";
+import { playTween } from "#utils/anim-utils";
+import i18next from "i18next";
+
+/**
+ * - Handles all VFX and SFX related to recalling a {@linkcode Pokemon}.
+ * - {@link Pokemon.leaveField | Removes the Pokemon from the field}.
```
MokaStitcher:
```suggestion
 * - Handles all VFX and SFX related to recalling a {@linkcode Pokemon} for player Pokemon at all times,
 * and for enemy Pokemon in trainer battles.
 * - {@link Pokemon.leaveField | Removes the Pokemon from the field}.
```

src/phases/summon-missing-phase.ts
```diff
@@ -1,15 +0,0 @@
...
-import { getPokemonNameWithAffix } from "#app/messages";
-import { SummonPhase } from "#phases/summon-phase";
-import i18next from "i18next";
-
-export class SummonMissingPhase extends SummonPhase {
-  public override readonly phaseName = "SummonMissingPhase";
-
-  protected override preSummon(): void {
-    globalScene.ui.showText(
-      i18next.t("battle:sendOutPokemon", { pokemonName: getPokemonNameWithAffix(this.getPokemon()) }),
```
MokaStitcher, line 11: this line is no longer in use?

src/phases/summon-phase.ts
```diff
@@ -1,120 +1,271 @@
...
 import { settings } from "#system/settings-manager";
-import type { PhaseKey } from "#types/phase-types";
+import { playTween } from "#utils/anim-utils";
 import i18next from "i18next";
 
-export class SummonPhase extends PartyMemberPokemonPhase {
-  public override readonly phaseName: PhaseKey = "SummonPhase";
+interface SummonPhaseOptions {
+  /**
+   * If `true`, summons the Pokemon as if loading into a wave
```
MokaStitcher, line 23: A bit confused about what this means. Is it for when it is the first wave right after loading an existing save file?
<sub>From the name I had thought it was related to the mon's sprite or something already being loaded but it seems like no

src/phases/summon-phase.ts
```diff
@@ -1,120 +1,271 @@
...
-        getPokemonNameWithAffix(partyMember),
-        partyMember,
-        getPokemonNameWithAffix(party[0]),
-        party[0],
+        `Swapped ${illegalPokemon.name} (index ${this.fieldIndex}) with ${this.getPokemon().name} (index ${legalIndex})`,
       );
+      return true;
+    }
+
+    if (this.getAlliedField().every((p) => !p.isAllowedInBattle())) {
```
MokaStitcher, line 161: this should always be true if we end up here, right? In which case the test is not needed

src/phases/summon-phase.ts
```diff
@@ -1,120 +1,271 @@
...
+
+    tweens.add({
+      targets: trainer,
+      x: -36,
+      duration: 1000,
+      onComplete: () => trainer.setVisible(false),
+    });
+
+    // Resolve 750 ms into the above Tween animation
+    await new Promise<void>((resolve) => time.delayedCall(750, resolve));
```
MokaStitcher, line 225: it seems a bit weird to mix Promises and async like this

src/phases/summon-phase.ts
```diff
@@ -252,56 +406,31 @@ export class SummonPhase extends PartyMemberPokemonPhase {
...
       x: "-=16",
       y: "+=16",
       alpha: 1,
       duration: 1000,
       ease: "Sine.easeIn",
       scale: pokemon.getSpriteScale(),
-      onComplete: () => {
-        pokemon.cry(pokemon.getHpRatio() > 0.25 ? undefined : { rate: 0.85 });
-        pokemon.getSprite().clearTint();
-        pokemon.resetSummonData();
```
MokaStitcher, line 266: this was not added below, is it intended?

src/phases/summon-phase.ts
```diff
@@ -140,87 +291,90 @@ export class SummonPhase extends PartyMemberPokemonPhase {
...
+      targets: pokemon,
+      duration: 250,
+      ease: "Sine.easeIn",
+      scale: pokemon.getSpriteScale(),
     });
+
+    pokemon.cry(pokemon.getHpRatio() > 0.25 ? undefined : { rate: 0.85 });
+    pokemon.getSprite().clearTint();
+    // required to load the proper assets when loading from save data
+    if (pokemon.summonData.speciesForm) {
```
MokaStitcher:
```suggestion
    // required to load the proper assets when loading from save data
    if (this.loaded && pokemon.summonData.speciesForm) {
```
this comment seems to imply that it's only needed when loading a save?

src/phases/switch-phase.ts
```diff
@@ -1,102 +1,193 @@
...
-   * @param doReturn Indicates if the party member on the field should be
-   * recalled to ball or has already left the field. Passed to {@linkcode SwitchSummonPhase}.
+   * If `true`, this phase schedules a {@linkcode SummonPhase}
+   * to run immediately after it ends
    */
-  constructor(switchType: SwitchType, fieldIndex: number, isModal: boolean, doReturn: boolean) {
-    super();
+  private readonly withSummon: boolean;
+
+  constructor(battlerIndex: number, switchType: SwitchType, switchInIndex: number = -1, withSummon: boolean = true) {
```
MokaStitcher: doc would be nice

src/phases/switch-phase.ts
```diff
@@ -1,102 +1,193 @@
...
+      switchedInPokemon.x += switchedInPokemon.getSubstituteOffset()[0];
+      switchedInPokemon.y += switchedInPokemon.getSubstituteOffset()[1];
+      switchedInPokemon.setAlpha(0.5);
+    }
+
+    // Swap the party positions of the switching Pokemon
+    party[this.switchInIndex] = activePokemon;
+    party[this.fieldIndex] = switchedInPokemon;
+
+    // Reset the switched out Pokemon's summon and turn data
```
MokaStitcher: is it maybe where the switched out mon's form should be reset?
Not sure this is a good place for the field scale though 🤔

MokaStitcher (commented):
first look at the code, no  playtesting yet

src/phases/summon-phase.ts
innerthunder: Not always. Consider this case where the player loads into a double battle:

1. The player's first Pokemon gets summoned without issue.
2. The Pokemon in the player's second party slot isn't allowed in battle, and the player doesn't have another inactive Pokemon that can take its place.
3. The game should progress with only the first player Pokemon on the field.

src/phases/return-phase.ts
innerthunder: The form change occurs in `Pokemon.leaveField`. The field scale does need to be updated, though. I figure at the end of `RecallPhase` and the start of `SummonPhase` would make the most sense?

src/phases/summon-missing-phase.ts
innerthunder: This is the same line as the `"battle:playerGo"` line in `SummonPhase` (looking at the English locales, at least). It seems like there are a few keys in `battle.json` with the same text...

src/phases/summon-phase.ts
innerthunder: Yes, this `resetSummonData` call is unnecessary

src/phases/return-phase.ts
MokaStitcher: recallPhase is only used for mons with trainers, right?
In theory there's no gmax wild Pokemon, but there's the garbodor ME, so not sure what the best approach is in this case 🤔

src/phases/summon-missing-phase.ts
MokaStitcher: 😭

src/phases/summon-phase.ts
MokaStitcher: ah I see, that makes sense. thanks

src/phases/summon-phase.ts
innerthunder: This pattern was something @SirzBenjie recommended iirc. It's slightly more flexible than returning the promise in that we could add code for after the promise resolves without a `then` if needed.

src/phases/summon-phase.ts
innerthunder: What part is confusing, the doc or the name? `loaded` was pretty much just ported over from the previous implementation.

src/phases/return-phase.ts
MokaStitcher: Maybe an event when a mon leaves the field would make sense

src/phases/summon-phase.ts
MokaStitcher: The "as if loading from a save" I think. Because it's actually being loaded from a save, right? Or are there cases where it's not but `loaded` is still true?
If not, maybe something like "`true` is the Pokemon was loaded from a save", or something like that?

src/phases/return-phase.ts
innerthunder: I changed it so that field scale should be updated at the end of `RecallPhase` and the start of `SummonPhase`. The Garbodor ME should be covered by the `SummonPhase` update.

In theory, if the player or trainer switches between two upscaled Pokemon, it should roughly look like
```
recall anim > scale down > show trainer > scale up > summon anim
```
<sub>I may have to update the "scale up" timing, though</sub>

src/phases/summon-phase.ts
innerthunder: This is also used when the player chooses to retry a wave after losing, so _technically_ it isn't only used when loading up a save. Maybe "as if the Player has just loaded into a wave" would be more accurate?

DayKev: That's supposed to act the same as loading a save, I think.

src/data/abilities/ab-attrs/post-damage-force-switch-ab-attr.ts
```diff
@@ -1,3 +1,7 @@
+/** biome-ignore-start lint/correctness/noUnusedImports: TSDoc imports */
+import type BattleScene from "#app/battle-scene";
```
MokaStitcher:
```suggestion
import type { BattleScene } from "#app/battle-scene";
```

MokaStitcher (dismissed):
Sorry it took so long to get to playtesting this

Encountered a crash when moving from a (wild) double to single battle when trying to recall the 2nd mon.

<details><summary>Steps to reproduce</summary>

- Override `BATTLE_TYPE_OVERRIDE: "odd-doubles",`
- Start a run with at least 2 mons
- Get through the first wave

=> crash
Seems like it's not finding the Pokemon, even its name is "Missingno"
image

</details>

MokaStitcher:
Once #1298 is merged you should be able to add the following test in the `double-battle` test file to check that the single <-> double battle transitions work (at least for wild battles). I confirmed the test passes in #1298 and fails here with the same error I encountered when playtesting.
<sub>Considering the number of times those single/double transitions have caused issues in the past I'm really surprised there was no such test already

```ts
it("should transition to and from double battles without crashing", async () => {
    game.override.battleType("even-doubles");
    await game.classicMode.startBattle(SpeciesId.BULBASAUR, SpeciesId.CHARMANDER);

    // Run 2 single -> double transitions and 2 double -> single transitions
    for (let waveNumber = 1; waveNumber < 5; waveNumber++) {
      const isDouble = waveNumber % 2 === 0;
      expect(game.scene.currentBattle.double).toBe(isDouble);
      expect(game.scene.currentBattle.waveIndex).toBe(waveNumber);

      game.move.select(MoveId.SPLASH);
      if (isDouble) {
        game.move.select(MoveId.SPLASH, 1);
      }
      await game.faintOpponents();
      await game.toNextWave();

      expect(game.scene.currentBattle.double).toBe(!isDouble);
    }
  });
```

failing on this branch
image
passing on the `test/fix-faint-opponents` branch
image

</details>

src/phases/summon-phase.ts
```diff
@@ -1,120 +1,272 @@
...
+    return false;
+  }
 
+  /**
+   * Plays animations for the Trainer summoning the Pokemon, then plays
+   * summon animations for the Pokemon.
+   */
+  private async playSummonSequence(): Promise<void> {
+    const { currentBattle, pbTray, pbTrayEnemy, trainer, ui } = globalScene;
+    // Update field scale in case a G-Max Pokemon or Starmobile is being summoned
```
DayKev:
```suggestion
```
Leftover comment.

src/phases/common-anim-phase.ts
```diff
@@ -29,7 +29,6 @@ export class CommonAnimPhase extends PokemonPhase {
     }
 
     this.anim = anim;
-    this.battlerIndex = battlerIndex!;
```
DayKev, line 32: This needs to still be here (or the `super()` call moved to its place is probably better).
```ts
  constructor(anim: CommonAnim, battlerIndex?: FieldBattlerIndex, targetIndex?: FieldBattlerIndex) {
    battlerIndex =
      battlerIndex
      ?? globalScene
        .getField()
        .find((p) => p?.isActive())
        ?.getBattlerIndex();
    if (battlerIndex === undefined) {
      console.warn("There are no Pokemon on the field!"); // TODO: figure out a suitable fallback behavior
    }

    // TODO: refactor `PokemonPhase` and/or this phase
    super(battlerIndex!);

    this.anim = anim;
    this.targetIndex = targetIndex;
  }
```

DayKev:
When switching due to a move like U-Turn, the switch animation happens while the party UI is open, and then it closes after the animation completes (so you don't see it lol).
Other than that everything seems to be working properly.

DayKev:
Actually the same happens with EE/Wimp Out as with U-Turn.

innerthunder:
> When switching due to a move like U-Turn, the switch animation happens while the party UI is open, and then it closes after the animation completes (so you don't see it lol). Other than that everything seems to be working properly.

Should be fixed now. `SwitchPhase` wasn't returning to message mode after going into party mode...

DayKev (dismissed):

DayKev (dismissed):

src/data/moves/move-attrs/force-switch-out-attr.ts
```diff
@@ -75,27 +76,19 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
...
-          globalScene.phaseManager.createAndPrependPhase(
-            "PostActionPhase",
-            "SwitchSummonPhase",
-            this.switchType,
-            switchOutTarget.getFieldIndex(),
-            slotIndex,
-            false,
-            true,
-          );
+          const switchInIndex = eligibleNewIndices[user.randSeedInt(eligibleNewIndices.length)];
```
MokaStitcher, line 81: would making a pokemon.randItem() function be useful in other places than here? 🤔

src/field/player-pokemon.ts
```diff
@@ -173,25 +173,23 @@ export class PlayerPokemon extends Pokemon {
    * @param switchType the {@linkcode SwitchType} for this switch-out. If this is
    * `BATON_PASS` or `SHED_TAIL`, this Pokemon's effects are not cleared upon leaving
    * the field.
+   * @deprecated
    */
   switchOut(switchType: SwitchType = SwitchType.SWITCH): Promise<void> {
```
MokaStitcher: is there a reason to keep it rather than remove it entirely?

src/phases/base/battle-phase.ts
```diff
@@ -1,12 +1,13 @@
...
 import { Phase } from "#app/phase";
 import { TrainerSlot } from "#enums/trainer-slot";
+import { playTween } from "#utils/anim-utils";
 
 /**
  * Adds functions to display and hide the enemy trainer
  */
 export abstract class BattlePhase extends Phase {
-  public showEnemyTrainer(trainerSlot: TrainerSlot = TrainerSlot.NONE): void {
+  public async showEnemyTrainer(trainerSlot: TrainerSlot = TrainerSlot.NONE): Promise<void> {
```
MokaStitcher, line 11:
```suggestion
  /** Slides the enemy trainer into view */
  public async showEnemyTrainer(trainerSlot: TrainerSlot = TrainerSlot.NONE): Promise<void> {
```
(if that's indeed what it does)

src/phases/base/battle-phase.ts
```diff
@@ -38,14 +40,22 @@ export abstract class BattlePhase extends Phase {
     });
   }
 
-  public hideEnemyTrainer(): void {
-    globalScene.tweens.add({
-      targets: globalScene.currentBattle.trainer,
+  public async hideEnemyTrainer(): Promise<void> {
```
MokaStitcher, line 49:
```suggestion
  /** Slide the enemy trainer out of view */
  public async hideEnemyTrainer(): Promise<void> {
```

src/phases/base/pokemon-phase.ts
```diff
@@ -8,9 +9,9 @@ import type { nil } from "#types/utility-types";
  * Provides helper functions to get the pokemon involved in the phase
  */
 export abstract class PokemonPhase extends FieldPhase {
-  protected battlerIndex: FieldBattlerIndex | number;
-  public isPlayer: boolean;
-  public fieldIndex: number;
+  protected readonly battlerIndex: FieldBattlerIndex | number;
```
MokaStitcher, line 13:
```suggestion
  /** FieldBattlerIndex of the target Pokemon, or its ID */
  protected readonly battlerIndex: FieldBattlerIndex | number;
```
Not sure "target" is the correct term
The field should probably be renamed to clarify too. <sub> even if it has little to do with this PR

src/phases/base/pokemon-phase.ts
```diff
@@ -32,6 +33,34 @@ export abstract class PokemonPhase extends FieldPhase {
...
+   * `undefined` if no such Pokemon exists. Unlike {@linkcode getPokemon}, this
+   * doesn't require the Pokemon to be active or on the field for it
+   * to be returned.
+   *
+   * @todo This is a bandaid fix for an issue where {@linkcode RecallPhase}
+   * would crash the game during a double -> single battle transition. If this phase is
+   * refactored, this method should be removed in favor of a more flexible {@linkcode getPokemon}
+   * @see {@link https://github.com/Despair-Games/poketernity/pull/1236#pullrequestreview-3046453380}
+   */
+  public getPokemonAtFieldIndex(): Pokemon | undefined {
```
MokaStitcher:
```suggestion
  protected getPokemonAtFieldIndex(): Pokemon | undefined {
```

could getPokemon be modified so that if the mon it finds is undefined it falls back to `getPokemonAtFieldIndex`? This would allow to make it private and have RecallPhase call getPokemon
But since it's a bandaid fix I'm fine with not wanting to modify getPokemon too

<sub>But this all means that the bang in `PokemonPhase.getPokemon` is not correct 😔

src/phases/common-anim-phase.ts
```diff
@@ -28,8 +26,9 @@ export class CommonAnimPhase extends PokemonPhase {
       console.warn("There are no Pokemon on the field!"); // TODO: figure out a suitable fallback behavior
     }
 
+    super(battlerIndex!);
```
MokaStitcher, line 29: <sub>i love that we know that it can be undefined but still get on with it xd

MokaStitcher (commented):
Playtesting:
<details><summary>Trainer sprite bug</summary>

After the battle the trainer sprite is not shown again for its defeat dialogue, unlike before
image

On beta:
image

</details> 

Other than that everything I tested seemed fine

src/data/moves/move-attrs/force-switch-out-attr.ts
DayKev: We probably shouldn't waste time modifying item stuff until we ~~steal~~ copy the item refactor from PKR.

MokaStitcher: I mean an equivalent to the randItem rng util that picks a random element in an array
Sorry for the confusion lol

DayKev: Oh I see. Well, the entire RNG system needs to be redone somewhat.

MokaStitcher: <sub> what are items anyway, I only know of modifiers :3

innerthunder:
Re: the Trainer sprite bug: the `showEnemyTrainer` and `hideEnemyTrainer` animation methods in `BattlePhase` are a bit flimsy because their tweens use additive notation, e.g.
```ts
await playTween({
  targets: trainer,
  x: "-=16",
  y: "+=16",
  alpha: 1,
  ease: "Sine.easeInOut",
  duration: 750,
});
```
I suspect the issue came from `hideEnemyTrainer` being called an extra time in `EncounterPhase`, which would offset the sprite off screen.

src/data/moves/move-attrs/force-switch-out-attr.ts
innerthunder: I figure this should be saved for a follow-up. There are a lot of problems with the force-switch-out code beyond this...

MokaStitcher:
> I suspect the issue came from `hideEnemyTrainer` being called an extra time in `EncounterPhase`, which would offset the sprite off screen.

the issue isn't fixed, right? Or is it not happening for you?
I think things were fine between mons, it's just at the end of the battle that the trainer should reappear, not sure where that was being handled before

innerthunder:
> > I suspect the issue came from `hideEnemyTrainer` being called an extra time in `EncounterPhase`, which would offset the sprite off screen.
> 
> the issue isn't fixed, right? Or is it not happening for you? I think things were fine between mons, it's just at the end of the battle that the trainer should reappear, not sure where that was being handled before

The trainer is also meant to show itself in between switches, which isn't happening here.

https://github.com/user-attachments/assets/f0d0e1b2-4c8c-4715-b098-914b8d5c249a

src/phases/base/pokemon-phase.ts
innerthunder: Yeah, I'm not sure if I want to change the existing methods in `PokemonPhase` much here because any change would propagate to other phases.

DayKev (dismissed):

DayKev (dismissed):

src/phases/base/pokemon-phase.ts
MokaStitcher: think this can still be protected rather than public

MokaStitcher (commented):
More opponent animation trouble
When starting a double battle against a single trainer (like twins), the opponent's trainer and pokeball tray appear before calling the second mon. They don't in beta, and it looks buggy that they do because the trainer sprite overlaps with the mon already on the field.

image

There may be a similar thing going on when the opponent has more than 2 mons and sends a new one during a double battle. Wasn't able to test that scenario though and I'm not sure what happens on beta in that case either

We probably also need to look at how things look in double battles with 2 independent trainer sprites as well, like evil grunts (they can be forced to be double battles line 569 of battle.ts)

DayKev:
The twins like to break things huh.

DayKev (dismissed):

DayKev (approved):

src/utils/anim-utils.ts
```diff
@@ -66,3 +68,17 @@ export function loadCommonAnimAssets(startLoad?: boolean): Promise<void> {
 export async function loadEncounterAnimAssets(startLoad?: boolean): Promise<void> {
   await loadAnimAssets(Array.from(encounterAnims.values()), startLoad);
 }
+
+/**
+ * Plays a Tween animation, resolving once the animation completes.
+ * @param config - The config for a single Tween
+ * @param scene - The {@linkcode SceneBase} on which the Tween plays (Default {@linkcode globalScene})
+ */
```
DayKev, line 82:
```suggestion
 */
// Note: The type of `config` cannot be changed to something like `Omit<TweenBuilderConfig, "onComplete">`
// due to whatever magic is happening behind the scenes with that type in Phaser.
```

DayKev (dismissed):

src/utils/anim-utils.ts
innerthunder: Added under a `privateRemarks` tag in the docs

PR #1237, opened 2025-06-03
[Test] Disable level ups in tests automatically

DayKev:
N/A

Ran into a flaky test potentially caused by inconsistent level up behavior.
image

The level cap will automatically be set to `1` (effectively preventing all Pokémon from levelling up) at the start of each test unless `OverridesHelper#disableLevelUp` is set to `false` via `game.override.disableLevelUp = false;`.

## How to test the changes?
`npm run test:silent`

## Checklist


test/test-utils/helpers/overrides-helper.ts
```diff
@@ -54,6 +54,11 @@ export class OverridesHelper extends GameManagerHelper {
    * @defaultValue `true`
    */
   public normalizeNatures: boolean = true;
+  /**
+   * If `true`, will automatically set the level cap to `1` at the start of each test (effectively disabling level ups).
+   * @defaultValue `true`
+   */
+  public disableLevelUp: boolean = true;
```
MokaStitcher: shouldn't it be named disableExpGain instead?
```suggestion
   * If `true`, will automatically set the level cap to `1` at the start of each test
   * (effectively disabling experience gain, and thus, level ups).
   * @defaultValue `true`
   */
  public disableExpGain: boolean = true;
```

innerthunder (dismissed):

MokaStitcher (commented):
The error that prompted this is looking bit weird and I'm not sure it is related to leveling up.

I ran the test 100 times on current beta and had no issue.
I then greatly increased the exp gain to force a level up and the test timed out because of the game waiting forever for input to learn a new move, not at all with the error from the screenshot in the PR description

The message from the screenshot happens when `ui.setMode` is called while being in a phase that isn't part of the phaseInterceptor. But the ShowExpBarPhase doesn't seem to call ui.setMode at any point so I'm very confused and the issue that made this test fail could be an indicator of something else going wrong. Maybe due to the test framework, the things ME tests do to the phase queue to skip fights, or due to an issue with the game itself. It's hard to tell but something seems to have gone wrong in the phases and/or UI modes somewhere, and that's a bit worrying.

Can you share which branch you had this issue on? (and the GH action link if it was a GH run)

tldr: I don't mind keeping this PR but I don't think it'll solve whatever caused this error

DayKev:
> Can you share which branch you had this issue on? (and the GH action link if it was a GH run)

https://github.com/Despair-Games/poketernity/actions/runs/15407365527/job/43352523952?pr=1233

test/test-utils/helpers/overrides-helper.ts
DayKev: Yeah that seems more precise.

MokaStitcher (approved):

innerthunder (approved):

PR #1239, opened 2025-06-05
[Addendum] Add support for multi-Phase input in Phase scheduling methods

innerthunder:

N/A

#1236 currently makes it so multiple phases need to be scheduled at once to carry out a switch (or "Pokemon") turn action. However, the turn manager appends phases from commands after the first queued `POST_ACTION` phase, which can cause problems with phase order when scheduling multiple phases. These changes should help make phase ordering more intuitive when using `appendToPhase` or `prependToPhase` and make some Phase scheduling calls more readable.

- In `PhaseManager`:
  - `pushPhase`, `unshiftPhase`, `prependToPhase`, and `appendToPhase` now accept multiple Phases as input. For example, `pushPhase`'s signature has been changed to
    ```ts
    public pushPhase(phase: Phase, ...otherPhases: Phase[]): void;
    ```
  - Updated docs
- Updated some instances of multiple consecutive `unshiftPhase` or `pushPhase` calls to make use of the new parameters

## How to test the changes?

Code review only (the parent PR runs into errors atm)

## Checklist

<sub>Other parts of the usual checklist aren't relevant for this PR...</sub>

### DIFF
--- a/src/phase-manager.ts
+++ b/src/phase-manager.ts
@@ -73,15 +73,26 @@ interface PokemonFaintInit {
  * This is responsible for managing the game's {@linkcode Phase | phases}.
  */
 export class PhaseManager {
-  /** dequeue/remove the first element to get the next phase */
+  /**
+   * The main queue where {@linkcode Phase | Phases} are scheduled.
+   * The first Phase in this queue is {@linkcode Phase.start | run} whenever
+   * {@linkcode shiftPhase} is called.
+   */
   private phaseQueue: Phase[] = [];
-  /** A temporary storage of what will be added to the front of {@linkcode phaseQueue} */
+  /**
+   * When {@linkcode shiftPhase} is called, the Phases in this queue are inserted
+   * in queue order to the front of {@linkcode phaseQueue}.
+   */
   private phaseQueuePrepend: Phase[] = [];
   /** overrides default of inserting phases to end of phaseQueuePrepend array, useful for inserting Phases "out of order" */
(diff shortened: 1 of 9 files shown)


DayKev (approved):
Seems good. Eventually we should probably remove the second phase queue that `unshift` pushes to and instead have it just work like the "actual" [`unshift`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/unshift) method as well.

src/phase-manager.ts
```diff
@@ -108,25 +119,28 @@ export class PhaseManager {
...
+   * Queues one or more phases to be run at a future point in time.
+   * @param phase - The first {@linkcode Phase} to push to the {@link phaseQueue | main queue}.
+   * @param otherPhases - Additional (optional) phases to queue. These phases are scheduled after {@linkcode phase} in array order.
    * @todo replace with a factory function for Phases based on `PhaseId`, ex: `public pushPhase<P extends Phase>(phase: PhaseId, ...params: ConstructorParameters<P>)`
    */
-  public pushPhase(phase: Phase): void {
-    this.phaseQueue.push(phase);
+  public pushPhase(phase: Phase, ...otherPhases: Phase[]): void {
+    const phases = [phase, ...otherPhases];
+    this.phaseQueue.push(...phases);
```
flx-sta:
```suggestion
    this.phaseQueue.push(phase, ...otherPhases);
```

flx-sta (approved):
lgtm

PR #1240, opened 2025-06-05
[Dev] Enforce type-only imports in `@types/` and `enums/`

DayKev:
N/A

Anything in `@types/` and `enums/` should not be importing other files.

New rule added to dependency cruiser to prevent non-type-only imports in the `@types/` and `enums/` directories.

## Checklist

### DIFF
--- a/.dependency-cruiser.cjs
+++ b/.dependency-cruiser.cjs
@@ -2,11 +2,23 @@
 module.exports = {
   forbidden: [
     {
+      name: "only-type-imports",
+      severity: "error",
+      comment: "Files in 'enums/' and '@types/' must only use type imports.",
+      from: {
+        path: ["(^|/)src/@types", "(^|/)src/enums"],
+        pathNot: ["(^|/)src/@types/phaser.d.ts"],
+      },
+      to: {
+        dependencyTypesNot: ["type-only"],
+      },
+    },
+    {
       name: "no-circular-at-runtime",
       severity: "error",
       comment:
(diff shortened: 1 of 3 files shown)


.dependency-cruiser.cjs
```diff
@@ -1,12 +1,24 @@
 /** @type {import('dependency-cruiser').IConfiguration} */
 module.exports = {
   forbidden: [
+    {
+      name: "only-type-imports",
+      severity: "error",
+      comment: "Files in 'enums/' and '@types/' must only use type imports.",
+      from: {
+        path: ["(^|/)src/@types", "(^|/)src/enums"],
+        pathNot: ["(^|/)src/@types/phaser.d.ts"],
```
DayKev, line 10: Had to add an exception for this file, seems like it's probably necessary for it to import Phaser in order to modify it.

.dependency-cruiser.cjs
```diff
@@ -18,7 +30,7 @@ module.exports = {
     {
       name: "no-orphans",
       comment:
-        "This is an orphan module - it's likely not used (anymore?). Either use it or "
+        "This is an orphan module - it's likely not used [any more]. Either use it or "
```
innerthunder:
```suggestion
        "This is an orphan module - it's likely not used [anymore]. Either use it or "
```

innerthunder (dismissed):

flx-sta (approved):

PR #1243, opened 2025-06-05
[Save Data][P3] Genderless mons no longer get male dex attribute

MokaStitcher:

- Genderless mons no longer show with a tinted ball when using the "everything" save file
- (dev) The gender of Pokemon with overridden species should now always be valid

When making the new "everything" save file in #1177 I naively assumed that genderless mons did not get a gender flag in their dex attributes. But it turns out they do get the male flag by default, causing the pokeball to appear tinted when using that save file and encountering a genderless mon.
Rather than update the save file though, this PR changes the game's behavior so that genderless mons do not get the MALE flag registered.

The override thing is an old issue, and happens if you load an existing save with the Enemy Species override. Since it's an old save the original opponent's information will be used, including its gender. So if when the file was saved the opponent was genderless but then an override to Rattata was added, you would encounter a genderless Rattata, which is not ideal and could lead to confusion when playtesting.
<sub>there may be an argument to make about simply not using the `dataSource` with the enemy species override, but that will have to be for another PR

- `Pokemon.getDexAttr()` no longer set the MALE flag by default
- Replace `DexAttrProp`'s `female` boolean in game data with actual gender
- This allows to move the `malePercent` checks away from the new game logic, and into starter select and game data logic.

The only situation where this change could cause issues is when catching a genderless mon whose pre-evolution is gendered, like Shedinja. In this case we still need to unlock a gender for Nincada. This is handled in `GameData.setPokemonCaught`, in a similar fashion to how the formIndex gets validated when getting up the evolution chain: When catching a Shedinja, if Nincada already was unlocked, leave it at that. If not, it will get the male gender unlocked by default.

## Screenshots/Videos

Before (using the everything save file):
image

After (using the everything save file):
image

## How to test the changes?

- On incomplete or fresh file:
   - Check that both gendered and genderless Pokemon show as new or not properly after getting caught.
   - With Nincada not unlocked: Catch a Shedinja, then a male Nincada. When catching the Nincada its gender should not be shown as new, proving that catching Shedinja had unlocked it.
   - With only female Nincada unlocked: Catch a Shedinja, then a male Nincada. In this case the gender should be marked as new, since there was no need to unlock the male gender when catching Shedinja.

- On everything file: override to a genderless opponent species and make sure the pokeball indicator doesn't appear tinted

## Checklist

- [] ⚠️ If this is a PR for `main` (such as a hotfix), has the game version been updated (`pnpm update-version:patch` / `pnpm update-version:minor`?

  - [X] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes?


src/ui/components/starter-container.ts
```diff
@@ -30,18 +33,12 @@ export class StarterContainer extends Phaser.GameObjects.Container {
...
+    this.icon = globalScene.add.sprite(-2, 2, species.getIconAtlasKey(formIndex, shiny, variant));
     this.icon.setScale(0.5);
     this.icon.setOrigin(0, 0);
-    this.icon.setFrame(
-      species.getIconId(defaultProps.female, defaultProps.formIndex, defaultProps.shiny, defaultProps.variant),
-    );
-    this.checkIconId(defaultProps.female, defaultProps.formIndex, defaultProps.shiny, defaultProps.variant);
+    this.icon.setFrame(species.getIconId(isFemale, formIndex, shiny, variant));
+    this.checkIconId(isFemale, formIndex, shiny, variant);
     this.icon.setTint(0);
```
DayKev:
```suggestion
    this.icon = globalScene.add.sprite(-2, 2, species.getIconAtlasKey(formIndex, shiny, variant))
      .setScale(0.5)
      .setOrigin(0, 0)
      .setFrame(species.getIconId(isFemale, formIndex, shiny, variant))
      .setTint(0);
    this.checkIconId(isFemale, formIndex, shiny, variant);
```

src/ui/handlers/starter-select-ui-handler.ts
```diff
@@ -2341,20 +2342,13 @@ export class StarterSelectUiHandler extends MessageUiHandler {
...
-      props.formIndex,
-      props.shiny,
-      props.variant,
-    );
+    const { formIndex, shiny, variant } = props;
+    const female = props.gender === Gender.FEMALE;
+
+    const icon = this.starterIcons[this.starterSpecies.length];
+    icon.setTexture(species.getIconAtlasKey(formIndex, shiny, variant));
+    icon.setFrame(species.getIconId(female, formIndex, shiny, variant));
```
DayKev:
```suggestion
    const icon = this.starterIcons[this.starterSpecies.length]
      .setTexture(species.getIconAtlasKey(formIndex, shiny, variant))
      .setFrame(species.getIconId(female, formIndex, shiny, variant));
```

src/ui/handlers/starter-select-ui-handler.ts
```diff
@@ -2364,16 +2358,20 @@ export class StarterSelectUiHandler extends MessageUiHandler {
...
     const props = globalScene.gameData.getSpeciesDexAttrProps(species, this.getCurrentDexProps(species.speciesId));
-    this.starterIcons[index].setTexture(species.getIconAtlasKey(props.formIndex, props.shiny, props.variant));
-    this.starterIcons[index].setFrame(species.getIconId(props.female, props.formIndex, props.shiny, props.variant));
-    this.checkIconId(this.starterIcons[index], species, props.female, props.formIndex, props.shiny, props.variant);
+    const { formIndex, shiny, variant } = props;
+    const female = props.gender === Gender.FEMALE;
+
+    const icon = this.starterIcons[index];
+    icon.setTexture(species.getIconAtlasKey(formIndex, shiny, variant));
+    icon.setFrame(species.getIconId(female, formIndex, shiny, variant));
```
DayKev:
```suggestion
    const icon = this.starterIcons[index]
      .setTexture(species.getIconAtlasKey(formIndex, shiny, variant))
      .setFrame(species.getIconId(female, formIndex, shiny, variant));
```

src/ui/components/starter-container.ts
MokaStitcher: Didn't know this was possible, so that's good to know, but absolutely none of the UI is built that way, including the rest of this file, so putting it in will just make it stand out from the rest.
If we want to rewrite the UI with this it should probably be its own refactor PR

And considering starter select stuff will be gone hopefully soon I don't feel it's worth it to update those 2 files in particular with this
<sub> still did it

DayKev (dismissed):

src/ui/components/starter-container.ts
DayKev: True. Figured we might as well start somewhere, but it doesn't really matter.

src/system/game-data.ts
```diff
@@ -1508,7 +1510,20 @@ export class GameData {
...
-      const dexAttr = pokemon.getDexAttr(); // Get the dex attr with the valid and starter selectable form index
+      /*
+       * Ensure that the gender is valid for this species. An invalid gender can happen when an evolved Species
+       * and its pre evolution Species don't have the same possible genders (gendered VS genderless).
+       * In this case it will default to using an already unlocked gender, or male if there is none.
+       */
+      const genderlessSpecies = isNil(species.malePercent);
+      const genderlessMon = currentGender === Gender.GENDERLESS;
+      if (genderlessMon && !genderlessSpecies) {
+        pokemon.gender = caughtAttr & DexAttr.FEMALE || species.malePercent === 0 ? Gender.FEMALE : Gender.MALE;
```
flx-sta: Is this single `&` intentional? 

I know there is a lot of bitwise stuff going on here but this one feels a bit misplaced

test/system/set-pokemon-caught.test.ts
flx-sta, line 1: I assume the usage of `toBeTruthy()` and `toBeFalsy()` here are intentional.
Usually people avoid them and use `toBe(true)` or `toBe(false)` instead as that checks the exact boolean value

MokaStitcher: I'm pretty sure in the past I used toBe(true/false) and it was suggested to use those instead
So no, there's no specific intention behind it except to keep a consistent code base I suppose

src/system/game-data.ts
MokaStitcher: It is. Dexattr are all about bitwise operations 
Could be rewritten more clearly though

flx-sta (dismissed):

PR #1245, opened 2025-06-07
[Dev] Fix circular imports in `test/`

DayKev:
N/A

Circular imports are bad.

- Moved definition of `mockCanvas` to `mock-context.ts` and deleted `mock-canvas.ts` since the only import of that object was in `mock-context.ts`.
- Added missing `type` from import in `overrides-helper.ts`

## Checklist


innerthunder (approved):
The matchers we made trigger `no-non-package-json` errors when running depcruise:

image

If we resolve this (or have the severity reduced), we should be able to update the `pnpm depcruise` script to
```
depcruise src test
```

DayKev:
Those errors will go away with the package update PR (it's because we don't have `@vitest/expect` installed as a dev dependency yet, but I addressed that in the package update PR).

MokaStitcher (approved):

PR #1253, opened 2025-06-11
[Dev] Add and implement `lint/style/useFilenamingConvention` Biome rule

DayKev:
N/A

A few files were missed when standardizing to "kebab-case" filenames previously.

The Biome rule `lint/style/useFilenamingConvention` will now enforce "kebab-case" and ASCII-only filenames.

A few files that were missed previously were renamed to be "kebab-case".

## Checklist


flx-sta (approved):
What the heck is "manip"

DayKev:
> What the heck is "manip"

If you mean "what does it stand for" then "manipulation". If you mean "what do those files do" then :shrug:.

innerthunder (approved):

PR #1258, opened 2025-06-14
[AI-Rework] Add caching for invariant scoring calculations

innerthunder:

Very minor improvements (for now) in performance within `EnemyCommandPhase`

Some implemented and/or planned Effect Scores for move attributes refer to a Pokemon's Matchup Score (MUS) or Expected Attack Score (EAS) to determine whether a bonus or penalty should apply. Before this change, these values would have to be recalculated even though they are invariant for the turn. The added runtime from this is especially noticeable in newly added unit tests which statistically analyze move selection over 300 trials. 

A new field has been added to all Pokemon's `turnData`:

```ts
scoreData: Map<number, PokemonScoreData>;
```

`scoreData` is mapped by other active Pokemon's `id`s. The data in each `PokemonScoreData` entry may contain the Pokemon's MUS as well as EAS of each of the Pokemon's moves against the target Pokemon:

```ts
export interface PokemonScoreData {
  matchupScore?: number;
  expectedAttackScores: Map<MoveId, number>;
}
```

This cached data is populated by `Pokemon.getExpectedAttackScore` and `Pokemon.getMatchupScore`, which also return the appropriate cached data if it's already populated for the turn.

## Checklist

  - [n/a] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes?

### DIFF
--- a/src/enums/move-id.ts
+++ b/src/enums/move-id.ts
@@ -1,4 +1,20 @@
+// -- start tsdoc imports --
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import type { Pokemon } from "#field/pokemon";
+/* eslint-enable @typescript-eslint/no-unused-vars */
+// -- end tsdoc imports --
+
 export enum MoveId {
+  /**
+   * The {@linkcode Pokemon.getSimulatedMoves | simulated move}
+   * of a Pokemon's primary type.
+   */
+  SIMULATED_MOVE_1 = -2,
+  /**
+   * The {@linkcode Pokemon.getSimulatedMoves | simulated move}
+   * of a Pokemon's secondary type.
+   */
+  SIMULATED_MOVE_2,
   /**{@link https://bulbapedia.bulbagarden.net/wiki/None_(move) | Source} */
   NONE,
   /**{@link https://bulbapedia.bulbagarden.net/wiki/Pound_(move) | Source} */
--- a/src/enums/move-id.ts
+++ b/src/enums/move-id.ts
@@ -8,6 +8,11 @@ export enum MoveId {
   /**
    * The {@linkcode Pokemon.getSimulatedMoves | simulated move}
    * of a Pokemon's primary type.
+   *
+   * @privateremarks
+   * This and {@linkcode SIMULATED_MOVE_2} are negative values
+   * to maintain parity between the IDs of real moves
+   * and their IDs in the mainline games.
    */
   SIMULATED_MOVE_1 = -2,
   /**
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -1431,7 +1431,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
   }
 
   /**
-   * Generates placeholder moves for this Pokemon with properties based
+   * Generates placeholder attacks for this Pokemon with properties based
    * on the Pokemon's type(s), attacking stats, and the current wave index.
    * This generates up to 2 moves:
    * - One move for each of the Pokemon's base type(s).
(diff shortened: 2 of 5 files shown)


src/field/pokemon.ts
```diff
@@ -2226,16 +2228,44 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   * Caches a {@linkcode PokemonScoreData} entry in this Pokemon's
+   * {@linkcode turnData} with the given data.
+   * @param opponent - The {@linkcode Pokemon} that {@linkcode move} is evaluated against
+   * @param move - The {@linkcode Move} being evaluated
+   * @param score - The calculated score for the move action
+   * @see {@linkcode getExpectedAttackScore}
+   */
+  private cacheEas(opponent: Pokemon, move: Move, score: number): void {
+    const oppScoreData = this.turnData.scoreData.get(opponent.id);
+    if (!oppScoreData) {
```
DayKev: Shouldn't this be `if (isNil(oppScoreData))`?

src/field/pokemon.ts
```diff
@@ -2275,16 +2310,40 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+  /**
+   * Caches a {@linkcode PokemonScoreData} entry in this Pokemon's
+   * {@linkcode turnData} with the given Matchup Score data.
+   * @param opponent - The opposing Pokemon that this Pokemon is evaluated against
+   * @param matchupScore - This Pokemon's Matchup Score against {@linkcode opponent}
+   * @see {@linkcode getMatchupScore}
+   */
+  private cacheMatchupScore(opponent: Pokemon, matchupScore: number) {
+    const oppScoreData = this.turnData.scoreData.get(opponent.id);
+    if (!oppScoreData) {
```
DayKev: Same question here wrt `isNil`.

DayKev (approved):

src/enums/move-id.ts
```diff
@@ -1,4 +1,20 @@
...
+import type { Pokemon } from "#field/pokemon";
+/* eslint-enable @typescript-eslint/no-unused-vars */
+// -- end tsdoc imports --
+
 export enum MoveId {
+  /**
+   * The {@linkcode Pokemon.getSimulatedMoves | simulated move}
+   * of a Pokemon's primary type.
+   */
+  SIMULATED_MOVE_1 = -2,
```
Tempo-anon, line 17: why is the default -2?

src/field/pokemon.ts
```diff
@@ -1451,13 +1452,14 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
    */
-  protected getSimulatedMoves(): Move[] {
+  private getSimulatedMoves(): Move[] {
     const types = this.getTypes(false, false, true);
     const category = this.getStat(Stat.ATK) >= this.getStat(Stat.SPATK) ? MoveCategory.PHYSICAL : MoveCategory.SPECIAL;
     const ret: Move[] = [];
 
     for (let i = 0; i < Math.min(types.length, 4 - this.waveData.revealedMoves.size); i++) {
-      ret.push(new AttackMove(MoveId.NONE, types[i], category, this.getSimulatedMovePower(), 100, 10, -1, 0, 0));
+      const moveId = i % 2 === 0 ? MoveId.SIMULATED_MOVE_1 : MoveId.SIMULATED_MOVE_2;
```
Tempo-anon, line 1468: nit: a comment above this line would be good

src/field/pokemon.ts
```diff
@@ -2226,16 +2228,44 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
 
+    let score: number = 0;
     if (damage >= opponent.hp) {
-      return this.getAttackScoreOnKnockOut(move);
+      score = this.getAttackScoreOnKnockOut(move);
+    } else if (damage <= 0) {
+      score = -1;
+    } else {
+      const damagePct = Math.floor((damage / opponent.getMaxHp()) * 100);
+      score = damagePct / 40;
```
Tempo-anon: 40 here is a magic number

src/enums/move-id.ts
innerthunder: `MoveIds` are currently based on mainline move IDs (except for G-Max moves), and `all-moves.test.ts` verifies move data based on this coupling. I wanted to indicate that the simulated move IDs aren't associated with a "real" move without disturbing the `all-moves` test, so setting them to negative values made the most sense to me.

DayKev (approved):

src/enums/move-id.ts
Tempo-anon: Is -1 already used? Your explanation makes sense, it just feels like it should be in the comment here

PR #1260, opened 2025-06-15
refactor(move): Remove "G-Max" descriptor from moves if the source is not Gigantamaxed

innerthunder:
> [!IMPORTANT]
> This PR requires major changes to locales! 

- If a non-Max-form Pokemon learns a G-Max move (e.g. G-Max Vine Lash), the move will be displayed in the Pokemon's summary and "Fight" menu without the "G-Max" prefix (e.g. "Vine Lash")
- Other cases where the "G-Max" prefix is removed:
  - In the Starter Select UI (e.g. when set as an Egg Move)
  - When any Pokemon is in the process of replacing a move with the G-Max move (e.g. in the Summary menu)
  - Within a Pokemon's Move Relearn options while using a Memory Mushroom
  - Wherever the move is displayed outside of the context of a specific source Pokemon. For example, if any G-Max move were available as a TM, the TM's name would exclude the prefix.

I suggested this a while ago in the Discord, but it was never acted on until now. This should more clearly indicate when G-Max moves are "empowered" and make their inclusion as potential Egg Moves less confusing.

- `Move` no longer implements `Localizable`. A move's name and description are now localized within `name` and `effect` getter methods.
  - Another getter, `i18nKey`, obtains the `i18next` key used to localize the move's name and effect. This is basically a camel-case representation of the move's `MoveId` key, except that `G_MAX_` is trimmed from the front of the identifier if applicable. For example. G-Max Vine Lash (id `G_MAX_VINE_LASH`) now has the locale key `vineLash`.
- `PokemonMove` now includes a reference to the move's source `Pokemon`

## Screenshots/Videos

<details><summary>Pre-locale changes</summary>
<p>
In G-Max Venusaur's Summary:

image

In Charmander's Summary:

image

</p>
</details>

<details><summary>Overrides Used</summary>
<p>

```ts
const overrides = {
  STARTER_SPECIES_OVERRIDE: SpeciesId.VENUSAUR,
  STARTER_FORM_OVERRIDES: {
    [SpeciesId.VENUSAUR]: 2,
  },
  MOVESET_OVERRIDE: [MoveId.G_MAX_VINE_LASH, MoveId.G_MAX_STONESURGE]
```

</p>
</details> 

## Checklist

  - [n/a] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes?

  - [x] If so, please leave a link to it here: https://github.com/Despair-Games/poketernity-locales/pull/50
- [x] Have I added the `Localization` tag to this PR?


Tempo-anon (commented):
I like this idea in theory but here are some thoughts.

1. Is the G-Max string always a prefix in all languages?
2. Just for the sake of breaking up this PR, making the change to `move` and then renaming the g-max moves feels like it can be 2 separate PRs

src/data/abilities/ab-attrs/post-defend-apply-battler-tag-ab-attr.ts
```diff
@@ -21,11 +21,14 @@ export class PostDefendApplyBattlerTagAbAttr extends PostDefendAbAttr {
     if (this.condition(pokemon, attacker, move)) {
       if (!pokemon.hasTag(this.tagType) && !simulated) {
         pokemon.addTag(this.tagType, undefined, undefined, pokemon.id);
+
+        const pokemonName = getPokemonNameWithAffix(pokemon);
+        const moveName = pokemon.getMove(move.id)?.name ?? "";
```
DayKev, line 26: Is `move` not always defined or something?

src/data/moves/move.ts
```diff
@@ -126,7 +126,12 @@ export abstract class Move {
    * Used to localize the move's {@linkcode name} and {@linkcode effect}.
    */
   private get i18nKey(): string {
-    return toCamelCaseString(MoveId[this.id]);
+    const moveKey = MoveId[this.id];
+    const gMaxPrefix = "G_MAX_";
+    if (moveKey.startsWith(gMaxPrefix)) {
+      return toCamelCaseString(moveKey.substring(gMaxPrefix.length));
+    }
+    return toCamelCaseString(moveKey);
```
DayKev: Can't you do this?
```suggestion
    return toCamelCaseString(moveKey.replace(gMaxPrefix, ""));
```

src/field/pokemon-move.ts
```diff
@@ -30,14 +32,35 @@ export class PokemonMove {
...
   public maxPpOverride?: number;
 
-  constructor(moveId: MoveId, ppUsed: number = 0, ppUp: number = 0, virtual: boolean = false, maxPpOverride?: number) {
+  constructor(
+    pokemon: Pokemon | null,
+    moveId: MoveId,
+    ppUsed: number = 0,
+    ppUp: number = 0,
+    virtual: boolean = false,
+    maxPpOverride?: number,
```
DayKev: I don't think we should be adding new pseudo-optional params. Since you have to update every call site already, it's better to just do it "properly" instead.
```suggestion
  constructor(
    moveId: MoveId,
    { pokemon, ppUsed = 0, ppUp = 0, virtual = false, maxPpOverride }: {
      pokemon?: Pokemon;
      ppUsed?: number;
      ppUp?: number;
      virtual?: boolean;
      maxPpOverride?: number;
    } = {}
```

src/field/pokemon-move.ts
```diff
@@ -17,8 +19,8 @@ import { toDmgValue } from "#utils/common-utils";
  * @see {@linkcode getPpRatio} - returns the current PP amount / max PP amount.
  * @see {@linkcode getName} - returns name of the {@linkcode Move}.
  **/
-
 export class PokemonMove {
+  public pokemon: Pokemon | null;
```
DayKev:
```suggestion
  /** The {@linkcode Pokemon} whose moveset this move is part of. Used for localization of the move name. */
  public pokemon?: Pokemon;
```

src/data/battler-tags/taunt-tag.ts
```diff
@@ -37,16 +37,18 @@ export class TauntTag extends MoveRestrictionBattlerTag {
   }
 
   override getSelectionDeniedText(pokemon: Pokemon, moveId: MoveId): string {
+    const moveName = pokemon.getMove(moveId)?.name ?? "";
```
DayKev: Shouldn't the fallback for this (and every other time this pattern appears) be "`?? allMoves.get(moveId).name`"?

test/mystery-encounter/encounters/clowning-around-encounter.test.ts
```diff
@@ -350,12 +352,14 @@ describe("Clowning Around - Mystery Encounter", () => {
...
+
       // Same type moves on lead
-      scene.getPlayerParty()[0].moveset = [new PokemonMove(MoveId.ICE_BEAM), new PokemonMove(MoveId.SURF)];
+      player1.moveset = [new PokemonMove(player1, MoveId.ICE_BEAM), new PokemonMove(player1, MoveId.SURF)];
       // Different type moves on second
-      scene.getPlayerParty()[1].moveset = [new PokemonMove(MoveId.GRASS_KNOT), new PokemonMove(MoveId.ELECTRO_BALL)];
+      player2.moveset = [new PokemonMove(player2, MoveId.GRASS_KNOT), new PokemonMove(player2, MoveId.ELECTRO_BALL)];
       // No moves on third
-      scene.getPlayerParty()[2].moveset = [];
+      player3.moveset = [];
```
DayKev: ...I'm not sure this should even be here, the player should never have a pokemon with 0 moves.

test/ui/type-hints.test.ts
```diff
@@ -30,7 +31,7 @@ describe("UI - Type Hints", () => {
     game.override.battleType("single").startingLevel(100).startingWave(1).enemyMoveset(MoveId.SPLASH);
   });
 
-  it("check immunity color", async () => {
+  it.skip("check immunity color", async () => {
```
DayKev: Is there an issue with these tests?

DayKev (commented):
#1277

test/ui/type-hints.test.ts
innerthunder: Last I checked, the added `Pokemon` reference to `PokemonMoves` broke the mock text conversion. Type hints worked normally the last time I manually tested the changes.

DayKev: Oh so the mock needs to be updated.

innerthunder: Yeah... The tests in the UI folder are generally pretty old, and I'm not sure how useful the current mocks are at this point. I think @MokaStitcher said they would look into it?

DayKev (dismissed):

src/data/arena-tags/arena-tag.ts
```diff
@@ -97,6 +97,7 @@ export abstract class ArenaTag {
   /**
    * @returns The name of the move corresponding with the tag's {@linkcode sourceMoveId},
    * or `null` if `sourceMoveId` isn't defined.
+   * @todo Should this use `PokemonMove.name`?
```
MokaStitcher, line 100: maybe. As long as there is a safeguard for the mon no longer existing it shouldn't cause any issue.
Based on the tags that use the GMax descriptor I don't think it's needed for now though, but I'm not really familiar with G-Max moves

src/field/pokemon-move.ts
```diff
@@ -17,8 +19,9 @@ import { toDmgValue } from "#utils/common-utils";
  * @see {@linkcode getPpRatio} - returns the current PP amount / max PP amount.
  * @see {@linkcode getName} - returns name of the {@linkcode Move}.
  **/
-
 export class PokemonMove {
+  /** The {@linkcode Pokemon} whose moveset this move is part of. Used for localization of the move name. */
+  public pokemon?: Pokemon;
```
MokaStitcher: Potential issues with this

- First, PokemonMove is used PokemonData, which means right now that pokemon field gets added to save data. We don't want that because it inflates it. Daily run saved on this branch: 10.2KB. Same daily run saved on beta: 9.5KB
The easiest way to solve this would probably be to replace this
https://github.com/Despair-Games/poketernity/blob/96b41f8ea9314f849b7ab7978d760606fac95dcf/src/system/pokemon-data.ts#L114
with something that maps the moveset to a new one with PokemonMoves that don't have the pokemon field
Then, in the Pokemon class, when loading a Pokemon from PokemonData, we need to rebuild the moveset to have the pokemon field.
Don't know if that's the best solution but

- Can you think of any other places where PokemonMove(s) are used which could end up staying in memory after a mon is gone, potentially preventing a destroyed Pokemon object from being garbage collected?

MokaStitcher (changes requested):
Stopped early in the review because of an issue with save data.
The comment is on the pokemon-move class

(hadn't noticed there are conflicts to solve as well)

src/field/pokemon-move.ts
DayKev: It could store the ID of the Pokémon instead of a reference to the Pokémon object. This would also mean that [de]serialization doesn't need any special handling for the field.

MokaStitcher: true

innerthunder: Ended up switching to IDs, so this shouldn't be an issue anymore 😄

src/data/arena-tags/arena-tag.ts
innerthunder: Yeah, this doesn't seem to be relevant for the existing G-Max effects.

src/data/mystery-encounters/encounters/dancing-lessons-encounter.ts
```diff
@@ -271,9 +271,9 @@ export const DancingLessonsEncounter: MysteryEncounter = MysteryEncounterBuilder
         const move = encounter.misc.selectedMove?.getMove().id;
         if (!oricorio.moveset.some((m) => m.getMove().id === move)) {
           if (oricorio.moveset.length < 4) {
-            oricorio.moveset.push(new PokemonMove(move));
+            oricorio.moveset.push(new PokemonMove(move, { pokemonId: oricorio }));
```
DayKev: Ah... somehow everything is typed as `any` here... thanks MEs.
```suggestion
            oricorio.moveset.push(new PokemonMove(move, { pokemonId: oricorio.id }));
```

src/data/mystery-encounters/encounters/dancing-lessons-encounter.ts
```diff
@@ -271,9 +271,9 @@ export const DancingLessonsEncounter: MysteryEncounter = MysteryEncounterBuilder
...
           } else {
-            oricorio.moveset[3] = new PokemonMove(move);
+            oricorio.moveset[3] = new PokemonMove(move, { pokemonId: oricorio });
```
DayKev:
```suggestion
            oricorio.moveset[3] = new PokemonMove(move, { pokemonId: oricorio.id });
```

DayKev (dismissed):

src/data/battler-tags/disabled-tag.ts
```diff
@@ -51,12 +51,13 @@ export class DisabledTag extends MoveRestrictionBattlerTag {
     }
 
     this.moveId = lastValidMove.id;
+    const moveName = pokemon.getMove(this.moveId)?.name ?? allMoves.get(this.moveId).name;
```
MokaStitcher: this is being done so many times in tags, I wonder if making some util function would be good 🤔

src/data/mystery-encounters/encounters/dancing-lessons-encounter.ts
MokaStitcher: Ah, so this is still accessing the moveset directly as well...
Fixing that: [001-dancing-lesson-me-moveset.diff.txt](https://github.com/user-attachments/files/21663823/001-dancing-lesson-me-moveset.diff.txt)

src/field/player-pokemon.ts
```diff
@@ -468,13 +466,12 @@ export class PlayerPokemon extends Pokemon {
...
-    const newMoveset: PokemonMove[] = [];
-    this.moveset.forEach((move) => {
-      newMoveset.push(new PokemonMove(move.moveId, 0, move.ppUp, move.virtual, move.maxPpOverride));
-    });
-
-    return newMoveset;
+  /**
+   * Copies this Pokemon's moveset array onto the given Pokemon.
+   * @param pokemon - The {@linkcode Pokemon} receiving the copied moveset
+   * @todo This doesn't deeply copy each move (e.g. copied moves always have full PP)
```
MokaStitcher, line 472: doesn't really matter since this is only for Shedinja, but since this takes the target Pokemon as parameter it could be done without too much trouble, right?

src/field/pokemon.ts
```diff
@@ -1458,6 +1461,10 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     }
   }
 
+  public getMove(moveId: MoveId, bypassSummonData: boolean = false): PokemonMove | undefined {
```
MokaStitcher: I think I'd like this to be renamed getPokemonMove() to avoid too much confusion

src/phases/move-phase.ts
```diff
@@ -860,7 +860,7 @@ export class MovePhase extends BattlePhase {
       variableMessage
       ?? i18next.t("battle:useMove", {
         pokemonNameWithAffix: getPokemonNameWithAffix(this.pokemon),
-        moveName: this.move.getName(),
+        moveName: this.move.name,
```
MokaStitcher: why isn't `this.move` named `pokemonMove` :'(

src/ui/handlers/run-info-ui-handler.ts
```diff
@@ -819,10 +819,11 @@ export class RunInfoUiHandler extends UiHandler {
         moveContainer.add(moveLabel);
         movesetContainer.add(moveContainer);
         const move = pokemonMoveset[m]?.getMove();
+        const moveName = move ? pokemon.getMove(move.id)!.name : "-";
         pokemonMoveBgs[m].setFrame(
           enumValueToKey(ElementalType, move ? move.type : ElementalType.UNKNOWN).toLowerCase(),
         );
```
MokaStitcher:
```suggestion
        const pokemonMove = pokemonMoveset[m];
        const moveName = pokemonMove?.name ?? "-";
        const moveType = pokemonMove?.getMove()?.type ?? ElementalType.UNKNOWN;
        pokemonMoveBgs[m].setFrame(enumValueToKey(ElementalType, moveType).toLowerCase());
```
Trying to make more sense of this

src/field/pokemon-move.ts
```diff
@@ -30,14 +34,47 @@ export class PokemonMove {
...
+    if (isNil(this.pokemon)) {
+      return this.getMove().name;
+    }
+    const gMaxDescriptor =
+      this.getMove().checkFlag(MoveFlags.G_MAX_MOVE, this.pokemon) && this.pokemon.isMax()
+        ? i18next.t("move:gMaxDescriptor")
+        : "";
+
+    const moveName = this.getMove().name;
+    return i18next.t("move:moveFormatWithGMax", { gMaxDescriptor, moveName });
```
MokaStitcher: This does not have the intended result. `moveFormatWithGMax` puts the space in even if there is no gMaxDescriptor. Maybe there's a way to make a conditional with i18next, dunno.
Otherwise we either need to trim the string before returning it, or:

```suggestion
    const moveName = this.getMove().name;
    const isMax = this.getMove().checkFlag(MoveFlags.G_MAX_MOVE, this.pokemon) && this.pokemon.isMax();
    if (isMax) {
      const gMaxDescriptor = i18next.t("move:gMaxDescriptor");
      return i18next.t("move:moveFormatWithGMax", { gMaxDescriptor, moveName });
    }
    return moveName;
```

This also fixes the UI tests you disabled. They were failing because they look for a TextObject containing "Dragon Claw", but it was " Dragon Claw" because of the space issue (the UI test would still fail on this branch until the locale PR is merged in)

test/ui/type-hints.test.ts
MokaStitcher: Had missed that, sorry. I am not working on UI tests, nor planning to for the moment 😅 
Here's the tests working with the fix suggested in PokemonMove's name getter, they can be re-enabled
image

MokaStitcher:
```suggestion
  it("check immunity color", async () => {
```

test/ui/type-hints.test.ts
```diff
@@ -62,14 +63,14 @@ describe("UI - Type Hints", () => {
     await game.phaseInterceptor.to("CommandPhase");
   });
 
-  it("check status move color", async () => {
+  it.skip("check status move color", async () => {
```
MokaStitcher:
```suggestion
  it("check status move color", async () => {
```

MokaStitcher (commented):
Renamed some `move` variables to `pokemonMove`, feel free to disregard it
[002-pokemon-move-variable-renames.diff.txt](https://github.com/user-attachments/files/21665628/002-pokemon-move-variable-renames.diff.txt)

I think the pokemonId solution is great, unfortunately it doesn't seem to work for the run summary, `globalScene.getPokemonById()` doesn't work in this situation because it looks in player and enemy parties, and there are no parties in this situation. (in the screenshot I'm looking at an ongoing run from the load screen. But looking at a past run from the run history menu should have the same problem)
Unsure what the best course of action would be here
image

src/field/pokemon-move.ts
MokaStitcher: Thinking a bit more about this. If we go this route, we could also simply merge `gMaxDescriptor` and `moveFormatWithGMax` in a single key

DayKev:
> I think the pokemonId solution is great, unfortunately it doesn't seem to work for the run summary, `globalScene.getPokemonById()` doesn't work in this situation because it looks in player and enemy parties, and there are no parties in this situation.

So the Pokémon just kind of "exist", detached from anything? Hmm... could do more or less the same thing I did in my Pokemon ID PR, have each `Pokemon` instance add itself to some sort of "`allPokemon`" `BattleScene` field in the `Pokemon` constructor and then remove itself in `Pokemon#destroy`. Then have `getPokemonById` check that instead of the party fields.

MokaStitcher:
> > I think the pokemonId solution is great, unfortunately it doesn't seem to work for the run summary, `globalScene.getPokemonById()` doesn't work in this situation because it looks in player and enemy parties, and there are no parties in this situation.
> 
> So the Pokémon just kind of "exist", detached from anything? Hmm... could do more or less the same thing I did in my Pokemon ID PR, have each `Pokemon` instance add itself to some sort of "`allPokemon`" `BattleScene` field in the `Pokemon` constructor and then remove itself in `Pokemon#destroy`. Then have `getPokemonById` check that instead of the party fields.

Kind of. The RunInfoUiHandler gets the save data from local Storage, so it has PokemonData objects, and calls PokemonData.toPokemon(), which calls globalScene.addPlayer/EnemyPokemon. But these functions don't put them in a party, and they shouldn't. If we're loading a run it makes no sense to have parties with mons. And if we're looking at run history during a run, it shouldn't add random mons from an old run to the party lol
Ideally the handler wouldn't need to create a Pokemon object, but... 😔 
<sub>this makes me wonder if these Pokemon ever get destroyed tho

DayKev:
> this makes me wonder if these Pokemon ever get destroyed tho

Apparently they do, I added logging in the constructor and destroy method of the pokemon class (ignore the "unknown"s, that's from my saves getting messed up from switching branches lol):
<details><summary>Image</summary>
<p>

image

</p>
</details>

DayKev:
Actually hm, the pokemon get created twice for some reason when I move right into the run info screen, but only one copy gets destroyed... I think?
<details><summary>Image</summary>
<p>

image

</p>
</details>

MokaStitcher:
> Actually hm, the pokemon get created twice for some reason when I move right into the run info screen, but only one copy gets destroyed... I think?
> Image

... I think it's this
https://github.com/Despair-Games/poketernity/blob/c10eb8475751f9d762bafb2c01ddf6066d3b8bca/src/ui/handlers/run-info-ui-handler.ts#L597-L603

Other instances of toPokemon get destroyed, but not these ones 💀

DayKev:
> ... I think it's this
> 
> https://github.com/Despair-Games/poketernity/blob/c10eb8475751f9d762bafb2c01ddf6066d3b8bca/src/ui/handlers/run-info-ui-handler.ts#L597-L603
> 
> Other instances of toPokemon get destroyed, but not these ones 💀

That could be changed to this to fix the issue:
```ts
const luckValue = clamp( 
  this.runInfo.party 
    .map((p) => {
      const pokemon = p.toPokemon();
      const luck = pokemon.getLuck();
      pokemon.destroy();
      return luck;
    })
    .reduce((total: number, value: number) => (total += value), 0), 
  0, 
  14, 
); 
```
Though it's a bit silly to create a `Pokemon` object and then immediately destroy it, just to get a luck value lmao. It could probably be done the first time the `Pokemon` object is created, but I didn't look through the rest of the file.

src/field/pokemon-move.ts
innerthunder: I figure the easiest fix would be to update the keys themselves so that `gMaxDescriptor` includes the space (in English at least)?

MokaStitcher: Yes but that would mean it that key can only ever be used in this very context, and is likely to confuse translators as well. And if it's limited to a single context it might as well be all in a single key I think

src/field/player-pokemon.ts
innerthunder: It's awkward because `setMove` and `setMoveset` only generate "new" moves (i.e. all of the optional parameters are default), and `pokemon.moveset` can't be accessed directly from here.

DayKev: I think Shedinja is the only time where you want to do more than create a new `PokemonMove`, right? So we could just not worry about it and let Shedinja be slightly incorrect (actually, is it incorrect? What actually happens in mainline?).

DayKev:
For the title, I'd go with `refactor` rather than `feat` since you're changing something rather than adding a new feature. You could do `i18n:` as well, though the PR does include changes that aren't directly related to localization.

src/utils/pokemon-utils.ts
```diff
@@ -93,3 +95,7 @@ export function getPokerusStarters(): PokemonSpecies[] {
 export function getRandomElementalType(): ElementalType {
   return randSeedIntRange(1, 18) as ElementalType;
 }
+
+export function getPokemonMoveName(pokemon: Pokemon, moveId: MoveId, bypassSummonData: boolean = false): string {
```
DayKev, line 106: +TSDocs

DayKev (dismissed):

innerthunder:
> Renamed some `move` variables to `pokemonMove`, feel free to disregard it [002-pokemon-move-variable-renames.diff.txt](https://github.com/user-attachments/files/21665628/002-pokemon-move-variable-renames.diff.txt)
> 
> I think the pokemonId solution is great, unfortunately it doesn't seem to work for the run summary, `globalScene.getPokemonById()` doesn't work in this situation because it looks in player and enemy parties, and there are no parties in this situation. (in the screenshot I'm looking at an ongoing run from the load screen. But looking at a past run from the run history menu should have the same problem)

Hmm... I'm not sure what the ideal solution is here, assuming we want to keep storing IDs. I could maybe add an alternate `getNameOutsideOfGame` method with a Pokemon as input, since I think both menus generate Pokemon from session data before parsing move names, etc.

<sub>It'd be nice to separate Pokemon data from sprites for this, but that's obviously out of scope</sub>

DayKev:
Tbh there shouldn't be any `Pokemon` objects created outside of the player or enemy parties (at least unless/until all the Phaser/sprite/etc stuff is removed from the `Pokemon` class; Benjie said it shouldn't be too difficult apparently). I think we just leave the "outside of battle" move names (so run history/summary/info/whatever) slightly inaccurate for now and come back to them later.

DayKev (dismissed):

src/data/battler-tags/imprisoning-tag.ts
```diff
@@ -46,7 +46,7 @@ export class ImprisoningTag extends BattlerTag implements RestrictingBattlerTag
   }
 
   public getInterruptedText(pokemon: Pokemon, moveId: MoveId): string {
-    const moveName = pokemon.getMove(moveId)?.name ?? allMoves.get(moveId).name;
+    const moveName = pokemon.getPokemonMove(moveId)?.name ?? allMoves.get(moveId).name;
```
MokaStitcher: Is there a reason this is not using `getPokemonMoveName`? (I updated it, but just in case it was intended, leaving this comment)

src/data/battler-tags/move-restriction-battler-tag.ts
```diff
@@ -30,11 +30,14 @@ export abstract class MoveRestrictionBattlerTag extends BattlerTag implements Re
...
 
-      if (this.isMoveRestricted(move.moveId, pokemon)) {
-        if (this.getInterruptedText(pokemon, move.moveId)) {
-          globalScene.phaseManager.createAndUnshiftPhase("MessagePhase", this.getInterruptedText(pokemon, move.moveId));
+      if (this.isMoveRestricted(pokemonMove.moveId, pokemon)) {
+        if (this.getInterruptedText(pokemon, pokemonMove.moveId)) {
+          globalScene.phaseManager.createAndUnshiftPhase(
+            "MessagePhase",
+            this.getInterruptedText(pokemon, pokemonMove.moveId),
+          );
```
MokaStitcher:
```suggestion
        const interruptedText = this.getInterruptedText(pokemon, pokemonMove.moveId);
        if (interruptedText) {
          globalScene.phaseManager.createAndUnshiftPhase("MessagePhase", interruptedText);
```

src/phases/select-target-phase.ts
```diff
@@ -33,7 +33,7 @@ export class SelectTargetPhase extends PokemonPhase {
           .getRestrictingTag(moveId, user, firstTarget)
           ?.getSelectionDeniedText(user, moveObject.id);
 
-        const moveName = user.getMove(moveObject.id)?.name ?? allMoves.get(moveId).name;
+        const moveName = user.getPokemonMove(moveObject.id)?.name ?? allMoves.get(moveId).name;
```
MokaStitcher: same here
<sub> love that it's using both `moveObject.id` and `moveId` when they are the same since `moveObject` is `allMoves.get(moveId)` 🍝

src/ui/handlers/run-info-ui-handler.ts
```diff
@@ -596,8 +596,13 @@ export class RunInfoUiHandler extends UiHandler {
...
       this.runInfo.party
-        .map((p) => p.toPokemon().getLuck())
-        .reduce((total: number, value: number) => (total += value), 0),
+        .map((p) => {
+          const pokemon = p.toPokemon();
+          const luck = pokemon.getLuck();
+          pokemon.destroy();
+          return luck;
+        })
+        .reduce((total: number, value: number) => total + value, 0),
```
MokaStitcher: ah, I'm making a separate PR to fix that issue since it's completely unrelated. It should be ready later today so
you can revert the change if you want (sorry for highjacking that PR with this lol)
```suggestion
        .map((p) => p.toPokemon().getLuck())
        .reduce((total: number, value: number) => (total += value), 0),
```

src/ui/handlers/run-info-ui-handler.ts
```diff
@@ -818,11 +823,10 @@ export class RunInfoUiHandler extends UiHandler {
         moveContainer.add(moveBg);
         moveContainer.add(moveLabel);
         movesetContainer.add(moveContainer);
-        const move = pokemonMoveset[m]?.getMove();
-        const moveName = move ? pokemon.getMove(move.id)!.name : "-";
-        pokemonMoveBgs[m].setFrame(
-          enumValueToKey(ElementalType, move ? move.type : ElementalType.UNKNOWN).toLowerCase(),
-        );
+        const pokemonMove = pokemonMoveset[m];
+        const moveName = pokemonMove?.name ?? "-";
```
MokaStitcher, line 823:
```suggestion
        // TODO: the name doesn't get the GMax prefix if the Pokemon is Gigantamaxed
        const moveName = pokemonMove?.name ?? "-";
```

MokaStitcher (dismissed):
Merged the locale PR and updated it on this branch. Also added a couple missing calls to getPokemonMoveName
Just a few nits and we should be good 🚀

src/data/battler-tags/imprisoning-tag.ts
innerthunder: Nope, just missed it 😅

PR #1267, opened 2025-06-22
[Refactor] Put optional params of `Pokemon#getEffectiveStat` in object

### LINKED ISSUE #1041: [Ability] Implement Protosynthesis / Quark Drive
https://bulbapedia.bulbagarden.net/wiki/Protosynthesis_(Ability)

https://bulbapedia.bulbagarden.net/wiki/Quark_Drive_(Ability)

DayKev:
Quark Drive and Protosynthesis now ignore abilities when calculating the highest stat and are no longer marked as "partial".

No more `.getEffectiveStat(stat, undefined, undefined, ...)`.

The optional parameters of `Pokemon#getEffectiveStat` are now passed via object.

## Checklist

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -216,7 +216,7 @@ interface DamageFunctionOptions {
 }
 
 interface EffectiveStatOptions {
-  /** The target {@linkcode Pokemon} */
+  /** The opposing {@linkcode Pokemon}, usually involved in an incoming or outgoing attack */
   opponent?: Pokemon;
   /** The {@linkcode Move} being used */
   move?: Move;
@@ -1135,11 +1135,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
    * Calculates and retrieves the final value of a stat considering any held items,
    * move effects, opponent abilities, and whether there was a critical hit.
    * @param stat - The desired {@linkcode EffectiveStat}
-   * @param opponent - The target {@linkcode Pokemon}
-   * @param move - The {@linkcode Move} being used
-   * @param abilityApplyMode - (Default {@linkcode AbilityApplyMode.DEFAULT}) The {@linkcode AbilityApplyMode} determining how abilities are applied
-   * @param isCritical - (Default `false`) Whether a critical hit has occurred or not
-   * @param simulated - (Default `true`) If `true`, nullifies any effects that produce any changes to game state from triggering
+   * @see {@linkcode EffectiveStatOptions} for optional params
    * @returns The final in-battle value of a stat
    */
   getEffectiveStat(
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -1156,6 +1156,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
      * Variable Attack attributes are applied only to the raw stat
      * value and associated stat stage multiplier. Other stat modifiers,
      * e.g. items and abilities, apply based on the original stat.
+     * See https://bulbapedia.bulbagarden.net/wiki/Body_Press_(move)#Effect for more info
      */
     if (move && opponent && [Stat.ATK, Stat.SPATK].includes(stat)) {
       applyMoveAttrs(VariableAtkAttr, this, opponent, move, statValue, isCritical);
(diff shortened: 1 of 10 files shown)


src/field/pokemon.ts
```diff
@@ -1112,33 +1134,34 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
     const applyAbFunc = getAbApplyFunc(abilityApplyMode);
 
     const statValue = new NumberHolder(-1);
 
-    /**
+    /*
      * Variable Attack attributes are applied only to the raw stat
      * value and associated stat stage multiplier. Other stat modifiers,
-     * e.g. items and abilities, apply based on the original {@linkcode stat}.
+     * e.g. items and abilities, apply based on the original stat.
```
DayKev, line 1164: This comment feels off but I'm not sure what to do about it.

DayKev, line 1164: The end of this comment feels off but I'm not sure what to do about it.

src/field/pokemon.ts
```diff
@@ -1112,33 +1134,34 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
      */
     if (move && opponent && [Stat.ATK, Stat.SPATK].includes(stat)) {
```
DayKev, line 1166: The end of this comment feels off but I'm not sure what to do about it.

src/field/pokemon.ts
```diff
@@ -1112,33 +1134,34 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
       applyMoveAttrs(VariableAtkAttr, this, opponent, move, statValue, isCritical);
```
DayKev: The end of this comment feels off but I'm not sure what to do about it.

Tempo-anon: Maybe something like `apply based on the original numeric value of the stat` to be more explicit that it's the numeric value of the original stat?

Tempo-anon (dismissed):

src/field/pokemon.ts
innerthunder: "Items and Abilities that modify the original stat will still apply"?

To clarify, this refers to effects like the user's Huge Power boosting the damage of Body Press/Foul Play despite them applying different stat values/stages

src/field/pokemon.ts
```diff
@@ -217,6 +217,28 @@ interface AbilityData {
   passive: boolean;
 }
 
+interface EffectiveStatOptions {
+  /** The target {@linkcode Pokemon} */
```
```suggestion
  /**
   * The opposing {@linkcode Pokemon}, usually
   * involved in an incoming or outgoing attack
   */
```
The current doc isn't accurate because, depending on the Stat being calculated, `opponent` can be the attacker or defender.

test/test-utils/matchers/to-have-effective-stat-matcher.ts
```diff
@@ -46,7 +46,12 @@ export function toHaveEffectiveStatMatcher(
     };
   }
 
-  const actualValue = received.getEffectiveStat(stat, enemy, move, AbilityApplyMode.DEFAULT, isCritical);
+  const actualValue = received.getEffectiveStat(stat, {
+    opponent: enemy,
+    move,
+    abilityApplyMode: AbilityApplyMode.DEFAULT,
```
innerthunder:
```suggestion
```

innerthunder (dismissed):
If you want to fully(?) implement Protosynthesis/Quark Drive while you're at it, you can change this `getEffectiveStat` call to ignore abilities:

https://github.com/Despair-Games/poketernity/blob/b5c3b3461e13b0ec766f290f4e36fbae360df2aa/src/data/battler-tags/highest-stat-boost-tag.ts#L50

<sub>May not be considered a "full" implementation because of post-summon jank, but that's more of a Neutralizing Gas issue...</sub>

src/field/pokemon.ts
DayKev: So this is just changing which stat is being checked?

innerthunder: This changes which stat value and stages are used in the calculation. Body Press, for example, uses the user's Defense stat value and stages, but is still affected by Attack modifiers e.g. Huge Power and Choice Band

PR #1270, opened 2025-06-25
[Dev] Use Biome for unused imports rule

DayKev:
N/A

Reducing reliance on ESLint.

When adding imports for TSDocs, we now use the following:
```ts
/* biome-ignore-start lint/correctness/noUnusedImports: tsdoc imports */
import ...;
/* biome-ignore-end lint/correctness/noUnusedImports: tsdoc imports */
```
instead of:
```ts
/* eslint-disable @typescript-eslint/no-unused-vars */
import ...;
```

## Checklist

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -49,15 +49,14 @@
   "linter": {
     "enabled": true,
     "rules": {
-      "recommended": true, // see https://biomejs.dev/linter/rules/#recommended-rules
+      "recommended": true, // see https://biomejs.dev/linter/javascript/rules/#recommended-rules and https://biomejs.dev/linter/css/rules/#recommended-rules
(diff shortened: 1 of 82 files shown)


Tempo-anon (approved):

flx-sta (approved):
Love the fact that you can actually add the note in the same comment

DayKev:
> Love the fact that you can actually add the note in the same comment

I think you can with ESLint as well, but when I first started adding the TSDoc imports I was using the single-line suppression comments instead of the range suppression comments and also didn't think to check if you could add additional text to the suppression comments (Biome requires you to add an explanation to suppression comments so I immediately realized I could put that text there).

PR #1273, opened 2025-06-27
[Dev] Generate moveset when using player Species override

DayKev:
N/A

Improved testing workflow.

Player Pokémon will now generate an NPC moveset when the player species override is used (except for Keldeo, to prevent crashes) instead of having no moves. The moveset override will still apply normally if used.

## How to test the changes?
```ts
const overrides = {
  STARTER_SPECIES_OVERRIDE: SpeciesId.VENUSAUR,
} satisfies Partial<InstanceType<OverridesType>>;
```

## Checklist

### DIFF
--- a/src/field/player-pokemon.ts
+++ b/src/field/player-pokemon.ts
@@ -67,7 +67,10 @@ export class PlayerPokemon extends Pokemon {
     }
 
     if (!dataSource) {
-      if (globalScene.gameMode.isDaily) {
+      if (
+        globalScene.gameMode.isDaily
+        || (Overrides.STARTER_SPECIES_OVERRIDE && Overrides.STARTER_SPECIES_OVERRIDE !== SpeciesId.KELDEO)
+      ) {
         this.generateAndPopulateMoveset();
       } else {
         this.moveset = [];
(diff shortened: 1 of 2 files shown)


torranx (approved):

src/field/player-pokemon.ts
```diff
@@ -67,7 +67,10 @@ export class PlayerPokemon extends Pokemon {
     }
 
     if (!dataSource) {
-      if (globalScene.gameMode.isDaily) {
+      if (
+        globalScene.gameMode.isDaily
+        || (Overrides.STARTER_SPECIES_OVERRIDE && Overrides.STARTER_SPECIES_OVERRIDE !== SpeciesId.KELDEO)
```
Tempo-anon, line 72: Ugh Keldeo is so dumb. What is the issue with it exactly though? Is it the alternate form doesn't have moves or something? Wonder if we can just copy over its moveset from its default form to its cosmetic form

Tempo-anon (approved):

src/field/player-pokemon.ts
DayKev: The issue is the form change that occurs if you have its signature move in your moveset. It triggers the form change to occur when you start the run, which is outside the time the form change [phase] expects to happen.

PR #1279, opened 2025-07-03
[Dev] Fix issues with `pnpm` and `lefthook`

DayKev:
N/A

Improved developer experience.

- Add `skip_lfs: true` to `lefthook.yml` to prevent requiring the user to hit enter after a lefthook command runs (cf https://github.com/evilmartians/lefthook/discussions/1060)

- Remove package `@hpcc-js/wasm` and command + script to create dependency cruiser graph (will remove the errors that occur when running `pnpm install` as well)

- Replace `lefthook run post-merge` with directly running the command it invoked (`git submodule update --init --recursive`) in the `postinstall` command in `package.json`

- Add `pnpm-workspace.yaml` to allow packages to be built now that the lefthook hang preventing it has been fixed
  - Also add `shellEmulator: true` to the config to allow `package.json` scripts to work more reliably cross-platform

## How to test the changes?
Run `pnpm install`, it should "Just Work :tm:".

## Checklist


Tempo-anon (dismissed):

torranx (approved):
lgtm

PR #1280, opened 2025-07-04
Review suggestions for PR1236

DayKev:


src/data/arena-tags/tailwind-tag.ts
```diff
@@ -17,7 +17,7 @@ import i18next from "i18next";
  */
 export class TailwindTag extends ArenaTag {
   // This asserts to the compiler that `sourceId` is defined at all times for this tag
-  public override sourceId: number;
+  declare public sourceId: number;
```
DayKev, line 20: `declare` doesn't emit any JS while still telling TS that we know the field is defined.

src/phases/summon-phase.ts
```diff
@@ -77,15 +73,15 @@ export class SummonPhase extends PokemonPhase {
...
-     * @todo This only uses `resetSummonData` to push data from a
-     * {@linkcode Pokemon.summonDataPrimer | primer}. This should use a
-     * separate dedicated method instead to avoid the risk of side effects
+    /*
+     * If this summon is from loading into a wave, load the Pokemon's saved summon data.
+     *
+     * TODO: This only uses `resetSummonData` to push data from `Pokemon.summonDataPrimer`.
+     * This should use a separate dedicated method instead to avoid the risk of side effects.
      */
     if (this.loaded) {
```
DayKev, line 86: TSDocs are only valid when placed on definitions (of variables/etc).

innerthunder (approved):

PR #1282, opened 2025-07-08
[Refactor] Replace `.fill().map` "loops" with real `for` loops

DayKev:
N/A

`new Array(n).fill(...).map(...);` is dumb.

Replaced most instances of `new Array(n).fill(...).map(...)` with `for` loops.

## Checklist

### DIFF
--- a/src/animations.ts
+++ b/src/animations.ts
@@ -205,7 +205,7 @@ export class Animation {
   }
 
   public addPokeballCaptureStars(pokeball: Phaser.GameObjects.Sprite): void {
-    const addParticle = () => {
+    const addParticle = (): void => {
       const particle = this.scene.add.sprite(pokeball.x, pokeball.y, "pb_particles", "4.png");
       particle.setOrigin(pokeball.originX, pokeball.originY);
       particle.setAlpha(0.5);
@@ -243,7 +243,9 @@ export class Animation {
       });
     };
 
-    new Array(3).fill(null).map(() => addParticle());
+    for (let i = 0; i < 3; i++) {
+      addParticle();
+    }
   }
 
   /**
--- a/src/animations.ts
+++ b/src/animations.ts
@@ -243,7 +243,8 @@ export class Animation {
       });
     };
 
-    for (let i = 0; i < 3; i++) {
+    const numStars = 3;
+    for (let i = 0; i < numStars; i++) {
       addParticle();
     }
   }
--- a/src/data/mystery-encounters/encounters/training-session-encounter.ts
+++ b/src/data/mystery-encounters/encounters/training-session-encounter.ts
@@ -7,7 +7,7 @@ import { BattlerTagType } from "#enums/battler-tag-type";
 import { MysteryEncounterOptionMode } from "#enums/mystery-encounter-option-mode";
 import { MysteryEncounterTier } from "#enums/mystery-encounter-tier";
 import { MysteryEncounterType } from "#enums/mystery-encounter-type";
-import type { Nature } from "#enums/nature";
+import { Nature } from "#enums/nature";
 import { getStatKey } from "#enums/stat";
 import type { PlayerPokemon } from "#field/player-pokemon";
 import type { Pokemon } from "#field/pokemon";
@@ -27,6 +27,7 @@ import { MysteryEncounterOptionBuilder } from "#mystery-encounters/mystery-encou
 import PokemonData from "#system/pokemon-data";
 import type { HeldModifierConfig } from "#types/modifiers-types";
 import type { OptionSelectItem } from "#ui/option-select-config";
+import { getTSEnumValues } from "#utils/common-utils";
 import { randSeedShuffle } from "#utils/random-utils";
 import i18next from "i18next";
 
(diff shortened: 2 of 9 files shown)


src/animations.ts
```diff
@@ -243,7 +243,9 @@ export class Animation {
       });
     };
 
-    new Array(3).fill(null).map(() => addParticle());
+    for (let i = 0; i < 3; i++) {
+      addParticle();
+    }
```
innerthunder: Two things:
1. The number of particles (3) should probably be put into a const.
2. Not sure if this fits within the PR's scope, but the tweens in `addParticle` should be reorganized into a tween chain.

src/data/mystery-encounters/encounters/training-session-encounter.ts
```diff
@@ -181,7 +182,10 @@ export const TrainingSessionEncounter: MysteryEncounter = MysteryEncounterBuilde
       .withPreOptionPhase(async (): Promise<boolean> => {
         // Open menu for selecting pokemon and Nature
         const encounter = globalScene.currentBattle.mysteryEncounter!;
-        const natures = new Array(25).fill(null).map((_val, i) => i as Nature);
+        const natures: Nature[] = [];
+        for (const nature of getTSEnumValues(Nature)) {
+          natures.push(nature);
+        }
```
innerthunder: Wouldn't either
```ts
const natures: Nature[] = getTSEnumValues(Nature);
```
or
```ts
const natures: Nature[] = [...getTSEnumValues(Nature)];
```
be more efficient?

DayKev: Oh yeah, good point.

src/animations.ts
DayKev: A global constant? It's not like we're planning on changing this ever.

DayKev: Or do you just want `const numStars = 3;` right above the loop?

src/modifier/modifier-type.ts
```diff
@@ -1767,16 +1777,10 @@ export function getPlayerModifierTypeOptions(
...
-        getModifierTypeOptionWithRetry(
-          options,
-          retryCount,
-          party,
-          modifierTiers && modifierTiers.length > i ? modifierTiers[i] : undefined,
-        ),
-      );
-    });
+    for (let i = 0; i < count; i++) {
+      const tier = modifierTiers && modifierTiers.length > i ? modifierTiers[i] : undefined;
```
MokaStitcher:
```suggestion
      const tier = modifierTiers?.length > i ? modifierTiers[i] : undefined;
```

although doing 
```suggestion
      const tier = modifierTiers?[i]; 
```
would also do the same I think?

src/ui/components/iv-graph.ts
```diff
@@ -20,7 +20,7 @@ const ivLabelOffset = [0, sideLabelOffset, -sideLabelOffset, sideLabelOffset, -s
 const ivChartLabelyOffset = [0, 5, 0, 5, 0, 0]; // doing this so attack does not overlap with (+N)
 const ivChartStatIndexes = [0, 1, 2, 5, 4, 3]; // swap special attack and speed
 
-const defaultIvChartData = new Array(12).fill(null).map(() => 0);
+const defaultIvChartData: number[] = new Array(12).fill(0);
```
MokaStitcher, line 23: <sub>lol

src/ui/handlers/egg-gacha-ui-handler.ts
```diff
@@ -274,8 +274,8 @@ export class EggGachaUiHandler extends MessageUiHandler {
 
     this.eggGachaContainer.add(this.eggGachaOptionsContainer);
 
-    new Array(getTSEnumKeys(VoucherType).length).fill(null).map((_, i) => {
-      const container = globalScene.add.container(GAME_WIDTH - 56 * i, 0);
+    for (const voucher of getTSEnumValues(VoucherType)) {
+      const container = globalScene.add.container(GAME_WIDTH - 56 * voucher, 0);
```
MokaStitcher: If we don't want to make the voucherType require to start at 0 we could also increment a variable each loop,
or use a `foreach` on getTSEnumValues to get the index as we go

src/ui/handlers/starter-select-ui-handler.ts
```diff
@@ -742,15 +744,16 @@ export class StarterSelectUiHandler extends MessageUiHandler {
 
     this.starterSelectContainer.add(starterBoxContainer);
 
-    this.starterIcons = new Array(6).fill(null).map((_, i) => {
+    this.starterIcons = [];
+    for (let i = 0; i < 6; i++) {
```
MokaStitcher, line 748: I though there was a NUM_STARTER constant somewhere lol
<sub> maybe in the in progress ssui rewrite. Don't worry about it :')

MokaStitcher (dismissed):
minor questions, otherwise lgtm

src/modifier/modifier-type.ts
DayKev: Can't do the first one:
image
The second should work I believe, based on a quick test.

src/ui/handlers/egg-gacha-ui-handler.ts
DayKev: Like this?
```ts
    getTSEnumValues(VoucherType).forEach((voucher, index) => {
      const container = globalScene.add.container(GAME_WIDTH - 56 * index, 0);
```

src/modifier/modifier-type.ts
MokaStitcher: I right, it would need tiers.length ?? 0 to work

src/ui/handlers/egg-gacha-ui-handler.ts
DayKev: Seems to be working when I tested it, so I guess I'll go with that.

PR #1290, opened 2025-07-18
[Refactor] Rename `Arena#applyTagsForSide` to `Arena#applyTags`

DayKev:
N/A

Missed it in the Arena Tag methods refactor.
Reduces the number of redundant methods.

`Arena#applyTagsForSide` is renamed to `Arena#applyTags` and the old `Arena#applyTags` is removed.

## Checklist

### DIFF
--- a/src/field/arena.ts
+++ b/src/field/arena.ts
@@ -765,35 +765,19 @@ export class Arena {
   }
 
   /**
-   * Applies each `ArenaTag` in this Arena, based on which side (self, enemy, or both) is passed in as a parameter
+   * Applies each `ArenaTag` in this Arena, based on which side (player, enemy, or both) is passed in as a parameter
    * @param tagTypes Either an {@linkcode ArenaTagType} string, or an actual {@linkcode ArenaTag} class to filter which ones to apply
    * @param side {@linkcode ArenaTagSide} which side's arena tags to apply
    * @param simulated if `true`, this applies arena tags without changing game state
    * @param args array of parameters that the called upon tags may need
    */
-  applyTagsForSide(
-    tagTypes: ArenaTagType | ArenaTagType[],
-    side: ArenaTagSide,
-    simulated: boolean,
-    ...args: unknown[]
-  ): void {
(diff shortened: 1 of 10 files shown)


MokaStitcher (dismissed):

MokaStitcher (approved):

src/field/arena.ts
```diff
@@ -765,18 +765,13 @@ export class Arena {
...
    * @param simulated if `true`, this applies arena tags without changing game state
    * @param args array of parameters that the called upon tags may need
    */
-  applyTagsForSide(
-    tagTypes: ArenaTagType | ArenaTagType[],
-    side: ArenaTagSide,
-    simulated: boolean,
-    ...args: unknown[]
-  ): void {
+  applyTags(tagTypes: ArenaTagType | ArenaTagType[], side: ArenaTagSide, simulated: boolean, ...args: unknown[]): void {
```
innerthunder, line 774: `args` needs stricter typing. For now you might want to just follow what `applyBattlerTags` is doing:

```suggestion
  applyTags<T extends ArenaTag = ArenaTag>(
    tagTypes: ArenaTagType | ArenaTagType[],
    side: ArenaTagSide,
    ...params: Parameters<T["apply"]>
  ): void {
```
<sub>`simulated` is included in apply params</sub>

DayKev: That doesn't work because you can't do `t.apply(...params)` on L780.
I was planning on doing this originally but ran into issues like that (it's also why I made the other PR that removed the `arena` param from all the `ArenaTag` methods.

innerthunder: I'm not sure why, but you have to destructure the `params` tuple before you can plug it into `apply`. Call it a TS quirk 😅

This is already done in `applyBattlerTags`:

```ts
export function applyBattlerTags<T extends BattlerTag = BattlerTag>(
  tagTypes: BattlerTagType | readonly BattlerTagType[],
  ...params: Parameters<T["apply"]>
): boolean {
  const [pokemon, simulated, ...args] = params;
  // ...
  return tags.some((tag) => tag.apply(pokemon, simulated, ...args));
}
```

DayKev: Huh, I guess that works. Well I wanted to do this anyway but I wasn't sure how, now I know lol. Will make that my next PR.

PR #1291, opened 2025-07-18
[Refactor] Remove unnecessary `arena` params

DayKev:
N/A

Basically just cleaning up leftovers from the global scene refactor.

All params used to pass around the arena were removed from various `ArenaTag` methods (and `ProtectConditionFunc` and `GameMode#isWaveTrainer`).

## Checklist

### DIFF
--- a/src/data/arena-tags/arena-room-tag.ts
+++ b/src/data/arena-tags/arena-room-tag.ts
@@ -1,11 +1,11 @@
+import { globalScene } from "#app/global-scene";
 import { ArenaTag } from "#arena-tags/arena-tag";
-import type { Arena } from "#field/arena";
 
 /**
  * Base class for moves like Trick Room which should negate their effect when used a second time.
  */
 export abstract class ArenaRoomTag extends ArenaTag {
-  override onOverlap(arena: Arena): void {
-    arena.removeTag(this.tagType);
+  override onOverlap(): void {
+    globalScene.arena.removeTag(this.tagType);
   }
 }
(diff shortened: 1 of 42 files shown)


src/data/mystery-encounters/utils/encounter-phase-utils.ts
```diff
@@ -1066,7 +1065,7 @@ export function calculateMEAggregateStats(baseSpawnWeight: number) {
           }
         }
 
-        currentArena = globalScene.newArena(currentBiome);
+        globalScene.newArena(currentBiome);
```
Tempo-anon, line 1068: Minor, but it feels a bit weird to use `newArena` as a function here now since that returns the new `Arena` object and nothing is being done with it

Tempo-anon (dismissed):

src/data/mystery-encounters/utils/encounter-phase-utils.ts
DayKev: There was only one other place using the return value of `newArena` so I changed that callsite and removed the return value from the function.

src/data/arena-tags/arena-room-tag.ts
```diff
@@ -1,11 +1,11 @@
...
-import type { Arena } from "#field/arena";
 
 /**
  * Base class for moves like Trick Room which should negate their effect when used a second time.
  */
 export abstract class ArenaRoomTag extends ArenaTag {
-  override onOverlap(arena: Arena): void {
-    arena.removeTag(this.tagType);
+  override onOverlap(): void {
+    globalScene.arena.removeTag(this.tagType);
```
MokaStitcher, line 9: Is arena always defined or could this actually cause issues in weird cases?

MokaStitcher (approved):
lgtm

src/data/arena-tags/arena-room-tag.ts
DayKev: It should always be defined when something that sets an `ArenaTag` is used. That would be a pretty big bug if it could be `undefined` here.

DayKev: If this causes a crash at some point in the future that's fine, since that'll tell us that such a bug does exist and needs to be fixed.

Though I don't see how it's possible for there to be an active `ArenaTag` and not an active `Arena`, since this is the `onOverlap` function meaning it won't run unless there's already an active Trick Room / etc tag that needs to be removed.

PR #1242, opened 2025-06-05
[Refactor][UI] Refactor Game Stats Ui Handler

MokaStitcher:
## What are the changes the user will see?

- Faster loading time for Games Stats UI
> [!NOTE]
> The loading time issue isn't noticeable on beta since all UIs are initialized during loading.
So this targets #1230 because on that branch the UI gets initialized only when it's opened from the menu

- It is now possible to wrap around to the bottom/top of stats list
- Playtime statistic no longer updates while in the UI itself
<sub>but honestly is it important? And before it would only update when scrolling down then back up

## Why am I making these changes?

When working on #1230 I noticed that the Game Stats handlers took a couple seconds to initialize on my machine.
This is because the UI creates 2 `TextObjects` per game stats on screen (one for the label and one for the value) resulting in 36 `TextObject` being created and initialized at once. While creating dozens of sprites for other UIs do not cause such issues it seems like creating text objects comes with a significant overhead
This PR replaces all those individual TextObjects by a single one per column of text, changing those 36 objects into only 4, making the UI loading instantaneous and making the UI use less resource intensive in general.

## What are the changes from a developer perspective?

- Create a new UI component `TextListContainer` that can handle creating and updating a `TextObject` based on an array of strings. This will be useful to do similar optimizations in other UIs
- Add an optional `lineSpacing` to font styles so that the line spacing works with language specific text size
- Replace the individual textObjects for each statistics by creating 4 arrays of texts:
   - labels of the first column of stats
   - values of the first column of stats
   - labels of the second column of stats
   - values of the second column of stats

<sub>it's actually generic and should all work with any number of columns even though we'll probably never want more

## Screenshots/Videos

Before:
image
german_before

After: 
image
image

## How to test the changes?

- Open game stats page
- Scroll up/down
- Close in various states of scrolling / reopen

## Checklist

- [] ⚠️ If this is a PR for `main` (such as a hotfix), has the game version been updated (`pnpm update-version:patch` / `pnpm update-version:minor`?
- [ ] Otherwise: **I'm using `beta` as my base branch**
- [X] There is no overlap with another PR?
- [X] The PR is self-contained and cannot be split into smaller PRs?
- [X] Have I provided a clear explanation of the changes?
- [X] Have I tested the changes manually?

- [X] Are all unit tests still passing? (`pnpm test:silent`)
  - [ ] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes?
[...]

PR #1295, opened 2025-07-19
refactor(ui): Refactor Settings UI Handlers

### LINKED ISSUE #1115: [Bug][UI] Cancelling out of control remapping plays the error sound
### Describe the bug

When remapping keyboard or gamepad controls the error sound is played instead of the success sound when the player chooses to cancel a remapping

### Reproduction

- Open Settings
- Keyboard
- Select a button to remap
- Press any button
- Press enter to cancel the remapping

The error sound plays

### Expected behavior

The success sound should play in this case, because cancelling was chosen by the player.
The error sound should only be played if there was an error, like trying to remap to an invalid button

### Screenshots / Videos

Image

Pressing "ENTER" to cancel here will play an error sound

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

### LINKED ISSUE #1116: [Bug][UI] Shortcut buttons not updated immediately when remapping controller buttons
### Describe the bug

When remapping the cancel or action button on a gamepad, the contextual buttons at the bottom of the screen do not get updated immediately. They only get refreshed once another button is pressed

### Reproduction

- Settings -> Gamepad
- Change the mapping to either the action or cancel button
- Notice that the button icons at the bottom of the screen do not change
- Press any button
- The icons now get updated

### Expected behavior

The icons should update immediately after the remapping takes effect

### Screenshots / Videos

Before remapping
Image
After remapping (swapped action and cancel button)
Image
After remapping + another button press
Image

</details> 

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

MokaStitcher:
Locale PR: https://github.com/Despair-Games/poketernity-locales/pull/52

- Settings ui handlers are less resource intensive and faster to load
- It is now possible to switch keyboard layout in the settings (for now, qwerty, qwertz and azerty)
- Minor QoL updates for settings remapping UIs:
   - Fixed sound effect when updating bindings
   - The "delete binding" instruction text is grayed out when the button cannot be remapped
   - All settings related text should now be localizable
   - It's now possible to cancel out of a remapping with the "cancel" button/key.
- Minor visual changes to the settings UIs:
  - The navigation tabs in the settings now take all available space
  - The various option labels get aligned with each other when possible
  - The placement of instructions buttons and labels at the bottom of the UIs is improved
  - The "reload required" indicator at the bottom is only shown when relevant
  - The text in the rebinding UI is centered properly in all languages
- Fixes in Gamepad Settings UI when using multiple gamepads:
   - If a gamepad of index >0 is selected, the related settings now highlights the gamepad name, rather than the "Change" option
   - When pressing a button on a second gamepad while in the gamepad settings, update the UI like before but ignore the actual input. Before, the UI would update but still process the pressed button, which could result in weird or unexpected behavior

When working on #1230 we realized that the setting ui handlers are way too slow to load to be able to be built and destroyed dynamically. As a result they are currently kept in memory at all times.
This happens because the current code for these handlers is extremely inefficient:
- It creates a text object for ALL possible settings and ALL possible values of these settings, rather than creating the amount of text objects that can fit on screen and updating them as the player scrolls through the settings
- The keyboard and gamepad handlers do the same but worse: they create text objects for ALL buttons, for EACH possible gamepad and keyboard type. (which results in 50+ text objects and 30+ sprites for the keyboard ui, and a total of 100+and 50+ sprites for 3 different gamepad types), which are all in memory at all times, even when playing on devices that will never even need them, like phones

This also means that the more settings and gamepad types get supported, the more these UIs are using memory

This PR aims to address the bigger performance issues with the settings UIs, so that it loads faster and doesn't need to be initialized on startup anymore

As a side effect, these issues are also fixed: #1116 and #1115

<details><summary>Settings Ui Handlers optimization</summary>

- Use the TextListContainer created in #1242 to replace all previous text objects for the settings labels with a single one
- For the options, use a rotating array of text objects, that gets updated dynamically as the player scrolls through the UI. On initialization (`setup`) only 2 text objects are created for each option that can fit on screen. When loading and scrolling through the UI, update the existing text objects with the proper label, and if there is not enough, create more as needed.
- Controls Ui Handlers:
   - updated to extend the base SettingsUiHandler, to remove a lot of duplicate code
   - On top of the normal SettingsUiHandler behavior, add a single TextListContainer for the "press action to bind" text to show for rebind options
   - For the mapping icons, use a rotating array of sprites that gets updated as the player scrolls down.

As a result the Control Ui Handlers now only use 2 Text Objects each instead of the 50+ / 100+ from before, and only 8 sprite objects.

</details> 

<details><summary>Setting ui handlers refactor</summary>

- Move the code related to specific settings like language and touch controls to their own handler rather than the parent handler
- Create a helper function `addInstructionText` to add an icon and text at the bottom of the settings ui for the instructions, and aligns them automatically. This allows the gamepad and keyboard uis to place extra instructions for deleting/resetting bindings.
- The instruction text can be modified at any point, allowing the keyboard settings ui handlers to gray out the text for non relevent instruction based on context

</details>

<details><summary>Navigation Menu / Navigation Manager refactor</summary>

- The `NavigationMenu` and `NavigationManager` classes were split in their own files and moved to the `components` and `helpers` subfolders respectively
- Both are now independent of settings
   - The `NavigationMenu` can display any menu based on given labels, and place its elements automatically based on their width image
  - The `NavigationManager` can create and update any number of `NavigationMenu`s, changing UIMode based on player input and updating its "children" menus automatically
  - A `SettingsNavigationManager` singleton that extend it was created to replace the old `NavigationManager`
</details>

<details><summary>UI update</summary>

- The temporary code that was initializing all Settings handlers during UI the game loading was removed
- This means the settings UI handlers now get initialized only when needed, like the others handlers
- When the UI detects that a settings mode is exited and a non-setting mode is entered, all initialized settings handers now get destroyed, freeing them from memory

</details> 

#### Notes for potential further optimization

For now, the structure where each tab in the settings corresponds to a single UI Mode was preserved. This means each tab has its own handler, with its own number of Text Objects which stay in memory as long as the settings are open.
To save even more on performance it would be possible to merge all of these handlers/modes in a single one that swaps its settings category on the fly.
This would also remove the need for each of them having their own NavigationMenu, and for the singleton SettingsNavigationManager.

Another avenue for NavigationMenu optimization without the need to merge the handlers would be to have settings moved to their own Scene. This would allow the scene itself (or its UI) to handle the NavigationMenu, removing the need for duplicating it, as well as the instruction labels and icons at the bottom of each tab.
<sub> The NavigationMenu could also be moved to the current UI class, similar to other elements like the achievements or bgm bar, but that would add even more bloat to it

## Screenshots/Videos

<details><summary>Memory usage</summary>

Before, heap snapshot after loading the game, on the title screen: 108 MB, 463 Phaser Text objects (subsequent snapshots were actually at 109 MB heap size)
image

After:
Snapshots 1, 3, 5: title screen (107 MB heap size, 55 text objects)
Snapshots 2, 4: in settings, after having navigated to all tabs to initialize them (108 MB heap size, 215 text objects
imageimage

</details> 

<details><summary>Screenshots, EN</summary>

Before / After comparisons:
image_en-general-after
image_en-display-after
image_en-audio-after
image_en-keyboard-after
imageimage
_en-no-gamepad-after
</details> 

<details><summary>Some other languages screenshots</summary>

de-gen-afterde-display-after
_de_gamepad_afterjp-display-after

</details> 

<details><summary>Gamepad Binding</summary>

EN:
imageimage

FR: 
Before (see the misaligned text):
imageimage

After:
imageimage

</details> 

<details><summary>Multiple Gamepad Setup</summary>

Before, when gamepad with index 1 is selected, the change option is highlighted:
(index 2 or higher would likely cause a console warning, but I don't have enough controllers to test that lol)
image
After, when gamepad with index 1 is selected:
image

(index 2 or higher should behave the same, without warning

</details> 

## How to test the changes?

- Play with various settings
- make sure they take effect and are preserved when reloading
- Ones whose code got slightly altered: language, moving touch controls (needs a touch device or to emulate mobile through the devtools)

#### Gamepad

- If you have a gamepad, try remapping some stuff
- If you have 2 gamepads, try using one then the other and switch which one to configure, see if the icons update properly
<sub>I don't think the UX is great for switching gamepads like this, and the activeConfig setting makes little sense in itself but since this is very niche, it would have to be improved in a different PR

#### Keyboard

- Go to the Keyboard tab. By default it should show the layout as "QWERTY"
- Attempt to delete the cycle form/cycle shiny mappings. It shouldn't work
- Attempt to bind R or F to any setting. It shouldn't work
- Play around with the mappings
- Switch to a different keyboard layout (AZERTY has more noticeable changes, notably to the alt mappings for the directional arrows).
- Make sure the remappings you did in QWERTY were not ported to the AZERTY layout
- Make some remappings
- Switch back to another layout then back to AZERTY/QWERTY
- Make sure your past remappings for each layout got preserved
- You can do the same after reloading the game as well

## Checklist

- ~~[ ] ⚠️ If this is a PR for `main` (such as a hotfix), has the game version been updated~~
- [X] The PR is self-contained and cannot be split into smaller PRs? :sob:

  - [ ] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes?

- [X] Has a locales PR been created on the [locales](https://github.com/despair-games/poketernity-locales) repo?
  - [X] If so, please leave a link to it here: https://github.com/Despair-Games/poketernity-locales/pull/52
- [X] Have I added the `Localization` tag to this PR?


src/ui/settings/binding-ui-handler.ts
```diff
@@ -269,4 +283,21 @@ export abstract class BindingUiHandler extends UiHandler {
     this.setCursor(0);
     this.actionsContainer.setVisible(true);
   }
+
+  protected getSelectedDevice(): string | null {
+    return globalScene.inputController?.selectedDevice[this.device];
+  }
+
+  protected swapAction() {
```
DayKev:
```suggestion
  protected swapAction(): boolean {
```

src/ui/settings/settings-ui-handler.ts
```diff
@@ -552,34 +677,40 @@ export abstract class SettingsUiHandler extends MessageUiHandler {
...
+    this.settingValuesData[settingIndex].labels[optionIndex] = newLabel;
+
+    // If the option is currently shown on screen, update the text object
+    if (!isNil(this.optionValueLabels[settingIndex - this.scrollCursor])) {
+      this.optionValueLabels[settingIndex - this.scrollCursor][optionIndex].setText(newLabel);
+    }
   }
 
-  private handleSaveSetting<V = any>(uiItem: SettingsUiItem, newValue: V) {
+  protected handleSaveSetting<V = any>(uiItem: SettingsUiItem, newValue: V) {
```
DayKev, line 694: Do you know what's going on here with `V` (which defaults to `any`?) and the `key as never` below?

src/ui/settings/controls-settings-ui-handler.ts
```diff
@@ -1,691 +1,246 @@
...
-    // Return immediately if the options container is not initialized.
-    if (!this.optionsContainer) {
-      return;
+  public override processInput(button: Button): boolean {
+    if (
+      this.noDeviceText.visible
+      && button !== Button.CYCLE_SHINY
+      && button !== Button.CYCLE_FORM
+      && button !== Button.CANCEL
+    ) {
```
DayKev:
```suggestion
    const specialButtons: Button[] = [Button.CYCLE_SHINY, Button.CYCLE_FORM, Button.CANCEL];
    if (this.noDeviceText.visible && !specialButtons.includes(button)) {
```
Not sure what to call the `const`.

src/ui/settings/gamepad-settings-ui-handler.ts
```diff
@@ -63,53 +36,94 @@ export class GamepadSettingsUiHandler extends ControlsSettingsUiHandler {
...
-          // Update the text of the first option label under the current setting to the name of the chosen gamepad,
-          // truncating the name to 30 characters if necessary.
-          this.layout[_key].optionValueLabels[index][0].setText(
-            truncateString(globalScene.inputController.selectedDevice[Device.GAMEPAD] ?? "", 20),
-          );
+  protected override handleSaveSetting<V = any>(uiItem: SettingsUiItem, newValue: V): void {
+    if (uiItem.key === SettingGamepad.Controller) {
+      if (newValue) {
+        const gp = globalScene.inputController.getGamepadsName();
+        if (globalScene.ui && gp) {
```
DayKev:
```suggestion
        const gamepad = globalScene.inputController.getGamepadsName();
        if (globalScene.ui && gamepad) {
```

src/ui/settings/gamepad-settings-ui-handler.ts
```diff
@@ -63,53 +36,94 @@ export class GamepadSettingsUiHandler extends ControlsSettingsUiHandler {
...
+            return true;
+          };
+          const changeGamepadHandler = (gamepad: string, index: number) => {
+            globalScene.inputController.setChosenGamepad(gamepad);
+            settings.update("gamepad", "activeIndex", index);
+            return cancelHandler();
+          };
+          globalScene.ui.setOverlayMode<OptionSelectUiHandler>(UiMode.OPTION_SELECT, {
+            options: [
+              ...gp.map((g: string, index) => ({
```
DayKev:
```suggestion
              ...gamepad.map((g: string, index) => ({
```

src/ui/settings/keyboard-binding-ui-handler.ts
```diff
@@ -39,57 +30,29 @@ export class KeyboardBindingUiHandler extends BindingUiHandler {
...
     }
-    const buttonIcon = activeConfig.icons[_key];
+
+    const buttonIcon = activeConfig.icons[key];
     if (!buttonIcon) {
       return;
     }
-    this.buttonPressed = key;
+    this.buttonPressed = keyCode;
     // const assignedButtonIcon = getIconWithSettingName(activeConfig, this.target);
```
DayKev:
```suggestion
```

src/ui/settings/keyboard-settings-ui-handler.ts
```diff
@@ -24,81 +21,102 @@ import i18next from "i18next";
...
+
+    // Iterate over the keys in the settingDevice enumeration.
+    for (const [index, value] of Object.values(SettingKeyboard).entries()) {
+      if (value === "layout") {
+        // Update the text of the first option label under the current setting to the name of the chosen layout
+        this.updateOptionValueLabel(
+          index,
+          0,
+          truncateString(enumValueToKey(KeyboardLayout, settings.keyboard.layout ?? KeyboardLayout.QWERTY), 25),
+        );
```
DayKev:
```suggestion
        const layoutLabel = truncateString(enumValueToKey(KeyboardLayout, settings.keyboard.layout ?? KeyboardLayout.QWERTY), 25);
        this.updateOptionValueLabel(index, 0, layoutLabel);
```

src/utils/inputs-utils.ts
```diff
@@ -3,6 +3,9 @@ import { Device } from "#enums/device";
 import type { InputInterfaceConfig, InputKeys, InputSettings } from "#types/inputs-types";
 import { isNil } from "#utils/common-utils";
 
+// TODO: Most of these are only used in the inputs controller and tests,
+// should they really be here rather than in the inputs controller?
```
DayKev, line 7: Yeah it looks like it. One of them (`getIconWithKey`) is literally only used once, in this very file, and shouldn't have even been exported in the first place lol.

src/ui/settings/settings-navigation-manager.ts
```diff
@@ -0,0 +1,32 @@
...
+      { mode: UiMode.SETTINGS_DISPLAY, label: i18next.t("settings:display") },
+      { mode: UiMode.SETTINGS_AUDIO, label: i18next.t("settings:audio") },
+      { mode: UiMode.SETTINGS_GAMEPAD, label: i18next.t("settings:gamepad") },
+      { mode: UiMode.SETTINGS_KEYBOARD, label: i18next.t("settings:keyboard") },
+    ];
+    super(options);
+  }
+
+  /**
+   * Gets the singleton instance of the NavigationManager.
```
DayKev:
```suggestion
```

src/ui/helpers/navigation-manager.ts
```diff
@@ -0,0 +1,109 @@
...
+  public getSelectedIndex(): number {
+    return this.cursor;
+  }
+
+  public addMenu(x: number, y: number): NavigationMenu {
+    const menu = new NavigationMenu(
+      x,
+      y,
+      this.tabs.map((tab) => tab.label),
+    );
```
DayKev:
```suggestion
    const tabsMap = this.tabs.map((tab) => tab.label);
    const menu = new NavigationMenu(x, y, tabsMap);
```

src/ui/settings/settings-ui-handler.ts
MokaStitcher: hmm. This was part of #50 but I don't know much more than that
I'll try and see if anything can be done about it but if I had to guess it's coming from the way the different settings categories are handled, and the fact that settings can be various types (usually a number or specific enum) 🤔 
Kind of similar to the way I had to use any to "unspecify" the gamepad configurations in #1297 maybe

src/ui/settings/keyboard-settings-ui-handler.ts
MokaStitcher: actually, updating this to 
```suggestion
        const layoutLabel = truncateString(enumValueToKey(KeyboardLayout, settings.keyboard.layout), 25);
        this.updateOptionValueLabel(index, 0, layoutLabel);
```
because the ?? shouldn't be needed

src/ui/settings/controls-settings-ui-handler.ts
MokaStitcher: I didn't realize that simply declaring it as a Button array would solve my compilation issues with this approach. thanks :')
I think `navigationButtons` should do the trick to convey why these buttons are not being blocked 🤔 
```suggestion
    const navigationButtons: Button[] = [Button.CYCLE_SHINY, Button.CYCLE_FORM, Button.CANCEL];
    if (this.noDeviceText.visible && !navigationButtons.includes(button)) {
```

src/utils/inputs-utils.ts
MokaStitcher: good catch. removed that one.
For the rest of the file I think I'll leave it for now. The way the icon stuff based on input device is handled needs a full refactoring of its own

src/ui/settings/settings-ui-handler.ts
MokaStitcher: yeaahh, looked a bit into it and it's a headache. Not sure we should open that rabbit hole

V = any could be changed to `string | number | boolean` but then the settingsManager's update function also expects any, so that would need to change, which causes issues

The "key as never" stuff is worse though
Basically the settings manager does stuff based on settings category, so you could do things like `SettingsManager.update<"general">("general", "whatever", value)` and it would scream at you because "whatever" is not a setting that is part of the general category.
That's all well and good when updating a setting "manually", but the handlers can change any setting so to make it work for the handlers we'd need to make then generic as well.
I made an attempt that seemed to work fine-ish but i'm getting weird compilation errors in Phaser stuff elsewhere, very confusing. I'll give it another go in the next few days maybe

DayKev (dismissed):

DayKev:
Btw there's a couple options that don't follow the "left = on, right = off" format (damage number display, egg skip), maybe those should have their option positions swapped?

innerthunder (approved):
Locales updated, everything looks good from playtesting 👍

DayKev (approved):

src/ui/settings/settings-ui-handler.ts
MokaStitcher: think i'll merge this PR for now. I have my attempted update on another branch and will get to it later.

MokaStitcher:
> Btw there's a couple options that don't follow the "left = on, right = off" format (damage number display, egg skip), maybe those should have their option positions swapped?

Maybe yes. That should be as simple as reordering the elements in settings ui items
There are some options I've been meaning to change as well, the iv scanner ones are very unclear
We can leave that to another PR I think

PR #1297, opened 2025-07-22
[Refactor] Refactor inputs controller and (input) config handler

### LINKED ISSUE #284: [Bug] Keyboard settings can get bricked
### Describe the bug

In the settings menu, the keybinds for switching tabs are tied to the keybinds for "Cycle Form" and "Cycle Shiny". This means that if the user somehow deletes those keybinds, the user will no longer be able to switch tabs in the settings, which prevents the user from accessing the keyboard settings to undo this mistake later on. This effectively bricks the keyboard settings (unless the user deletes settings from local storage, etc.).

### Reproduction

(See description)

### Expected behavior

There should be some sort of safeguard preventing the user from deleting these important keybinds.

### Screenshots / Videos

https://github.com/user-attachments/assets/eb32b80c-ce2c-4cb0-8ca5-a921e7583ee2

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

### LINKED ISSUE #1094: [Bug] Keyboard device mapping not always initialized properly
### Describe the bug

It seems like in some circumstances it is possible that the keyboard device mapping is not yet initialized when the game registers a keyboard key up event, resulting in this error:
Image

The error is not game breaking but the input controller file could use a cleanup, typing on function parameters and returned values, and documentation.

### Reproduction

Unfortunately haven't found a consistent way, but it seems to involve pressing keys during startup in way  such that somehow the first registered keyboard input is a key up one, rather than a key down one, where the keyboard config gets initialized
Opening dev tools during loading might make it easier to trigger

However by commenting the following line in key down, the same error can be triggered whenever a key is pressed (but then nothing will work, unlike with the actual bug)
https://github.com/Despair-Games/poketernity/blob/2933918b04861d66736f93d1e55c43d1095c9f3a/src/inputs-controller.ts#L369

### Expected behavior

No error in console when pressing keyboard during startup

### Screenshots / Videos

_No response_

### Session export file

_No response_

### User data export file

_No response_

### Additional context

The key function here seems to `ensureKeyboardIsInit` which is called whenever a key down event is received, but not key up, ultimately causing the error message is an `up` event is received because any `down` event
The same code as `ensureKeyboardIsInit` is also being executed direct in the gamepad button down event callback.

Some other places in the file likely do the same, so this should all be cleaned up and streamlined, and we probably shouldn't assume that a key up event will always be preceded by a key down event.

MokaStitcher:
Attempting to break #1295 into multiple PRs

- The default cycle shiny and cycle form keyboard mappings can no longer be changed. It is still possible to have an alternative mapping, like for all other buttons
- As a result it is no longer possible to brick the keyboard settings by deleting or overriding the mappings to switch tabs. 
- The game should no longer crash due to non initialized input device configuration
- When remapping an action/submit button and validation with said button, it should no longer be considered as pressed forever

Aside from that most functionality should be preserved. It's possible there are some minor bugs or oddities introduced in the various settings ui handlers, but these are getting refactored in #1295 right after this PR is merged, so that should be acceptable

When refactoring the settings ui handlers in #1295 it became apparent that inputs-related code needed a big overhaul: almost none of the code uses proper typing for parameters or variables, and there is little to no documentation, making everything extremely error prone. #1295 ended up devolving way too much into an inputs code refactor. This PR is attempting to isolate the inputs refactor from there

Also fixes #284 and #1094

#### InputInterfaceConfig
- The starting point is taking `InterfaceConfig` out of the inputs-controller file and moving it to an `inputs-types` file.
- `InterfaceConfig` got renamed to `InputInterfaceConfig` and the typing of its elements got reworked
- Renamed `config-handler` to `inputs-utils` and moved it to the utils folder
- Added proper parameter and return types to all functions in `inputs-utils` and `input-controller`
- Most other changes are a result of this. Since before there was absolutely no type safety surrounding that interface it had to be added everywhere

#### Device Configuration

- All gamepads and keyboard configuration files now extend InputInterfaceConfig with the proper generic types, which ensures that no button or key is forgotten in these.
- The way the bindings blacklist is handled was reworked. Before it was done partly in UI code, partly in inputs controller, partly in each individual configuration file, each with their own list. Now there are generic blacklists for all gamepads and all keyboard types:
   - `settingsBlacklist` is a list of settings that are to be grayed out in the UI
   - `keysBlacklist` is a list of the keys/button code that are not allowed to be mapped to another setting. When a device configuration initialized, this list gets populated automatically based on the defined settingsBlacklist. ie, if the setting for the "action" button is blacklisted, the button/key that is mapped by default to this setting will be locked.
- Each individual configuration can also define extra elements in these blacklists, although this is unused at the moment
- While the UI is not updated in this PR, this will allow much more consistency in remapping behavior and prevent
- To prevent the potential softlock of keyboard settings, the cycle shiny and cycle form settings (which are used to navigate the settings tabs) are now part of the blacklisted settings

#### Configuration initialization

Before, when loading the game, all custom mappings stored locally were added to newly `(Input)InterfaceConfig` objects.
This means that the inputsController `configs` list sometimes contained configuration which had not been initialized properly.
This is why in a lot of places of the code there were checks like `if(config.padID)`, to ensure the config had been properly initialized. This check wasn't present everywhere, sometimes resulting in the crash in #1094 

With proper typing added everywhere it's no longer possible to add an incomplete `(Input)InterfaceConfig` to the configs list.
Instead the custom mappings from local storage are stored in their own buffer list until they are needed and get copied when initializing a configuration.
This now means that any defined `InputInterfaceConfig` should be good to be used in any context

#### Keyboard Layout

- Since there was existing code to support multiple keyboard layouts, that code was fixed and updated
- A new category was added in settings for keyboards, with the layout being represented by a new enum
- New default configurations for AZERTY (french) and QWERTZ (german) keyboard were added as an example
- It is currently not possible to change the layout in-game, but a test was added to check the functionality.
The option will be added in #1295

#### UI folder changes

I did the bare minimum to remove compilation errors from the refactor and have the game still work.
All those UIs will be updated properly to take advantage of all the changes in #1295, so feel free to ignore those files

<sub>The Inputs Controller could still use some more work and documentation, it may happen in this PR, or in another. Unsure at this point

## Screenshots/Videos

 N/A

## How to test the changes?

`pnpm test:silent rebinding keyboard-layout`

- Run the game
- Go the gamepad and/or keyboard tab in settings
- Change some buttons / keys mappings
- Make sure they are applied properly
- Refresh the game, make sure the custom mappings got loaded properly
- Play around some more, reset the mappings, etc.

## Checklist

  - [X] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes?


src/configs/inputs/cfg-keyboard-azerty.ts
```diff
@@ -0,0 +1,279 @@
...
+    KEY_LEFT_BRACKET: "LEFT_BRACKET.png",
+    KEY_RIGHT_BRACKET: "RIGHT_BRACKET.png",
+
+    KEY_SEMICOLON: "SEMICOLON.png",
+
+    KEY_BACKSPACE: "BACK.png",
+    KEY_ALT: "ALT.png",
+  },
+  settings: {
+    [SettingKeyboard.Button_Up]: Button.UP,
```
innerthunder: Should the enum names be changed to `BUTTON_UP`, etc.? Also, can/should this enum set be changed to a const?

src/enums/keyboard-layout.ts
```diff
@@ -0,0 +1,8 @@
+import type { EnumValues } from "#types/utility-types";
+
+export const KeyboardLayout = {
```
innerthunder, line 3: QWERTZ (for German-speaking countries) is also probably worth adding

src/configs/inputs/cfg-keyboard-azerty.ts
MokaStitcher: It's very possible parts of the code rely on the name being the way it is, but I'll make a quick test to see if not

~~For changing it to a const, I think I did it in the UI PR, I'll double check as well~~ indeed. Think it'll stay in that PR otherwise the UI would have to be updated further in this one

src/enums/keyboard-layout.ts
MokaStitcher: Yup definitely. If not in this PR at least now the framework is here and working!

src/configs/inputs/cfg-keyboard-azerty.ts
MokaStitcher: Indeed some things like localization relies on the current names... I think I can change this but I'd rather do it in the UI PR because I would want to test things with the latest UI code. So this one would need to get merged first

public/locales
DayKev, line 1: Were the locales supposed to be updated in this PR? It's not synced with the locales repo.

MokaStitcher: No it will be in the UI PR linked in the description 
This one is just me breaking the input stuff out of that PR
I didn't notice, my bad. Will fix tomorrow

DayKev (commented):
[0001-Consolidate-and-deduplicate-keyboard-configs.patch](https://github.com/user-attachments/files/21499682/0001-Consolidate-and-deduplicate-keyboard-configs.patch)
Partial review

DayKev (commented):
[0001-Consolidate-and-deduplicate-keyboard-configs.patch](https://github.com/user-attachments/files/21500006/0001-Consolidate-and-deduplicate-keyboard-configs.patch)
[0002-Consolidate-gamepad-configs.patch](https://github.com/user-attachments/files/21500007/0002-Consolidate-gamepad-configs.patch)
Partial review

src/utils/inputs-utils.ts
```diff
@@ -0,0 +1,219 @@
...
+
+/**
+ * Retrieves the setting name associated with the specified key.
+ *
+ * @param config - The configuration object containing custom settings.
+ * @param key - The key to search for.
+ * @returns The setting associated with the specified key, or -1 if none.
+ */
+export function getSettingNameWithKey(config: InputInterfaceConfig, key: InputKeys): InputSettings | -1 {
+  return config.custom ? config.custom[key] : -1;
```
DayKev:
```suggestion
  return config.custom?.[key] ?? -1;
```

src/utils/inputs-utils.ts
```diff
@@ -0,0 +1,219 @@
...
+  // If the setting is mapped to a protected key, we can't change it
+  if (settingsBlacklist?.includes(settingName) || (!isNil(key) && keysBlacklist?.includes(key))) {
+    return false;
+  }
+  return true;
+}
+
+export function canIDeleteThisSetting(config: InputInterfaceConfig, settingName: InputSettings): boolean {
+  return canIOverrideThisSetting(config, settingName);
+}
```
DayKev: This function seems to be fully redundant.
```suggestion
```

src/utils/inputs-utils.ts
```diff
@@ -0,0 +1,219 @@
...
+
+export function canIAssignThisKey(config: InputInterfaceConfig, key: InputKeys): boolean {
+  const settingName = getSettingNameWithKey(config, key);
+  if (config.keysBlacklist?.includes(key) || (settingName !== -1 && config.settingsBlacklist?.includes(settingName))) {
+    return false;
+  }
+  return true;
+}
+
+export function canIOverrideThisSetting(config: InputInterfaceConfig, settingName: InputSettings): boolean {
```
DayKev:
```suggestion
export function canOverrideOrDeleteSetting(config: InputInterfaceConfig, settingName: InputSettings): boolean {
```
(Use VSCode refactor tool for this ofc)

src/utils/inputs-utils.ts
```diff
@@ -0,0 +1,219 @@
...
+export function deleteBind(config: InputInterfaceConfig, settingName: InputSettings): boolean {
+  const key = getKeyWithSettingName(config, settingName);
+  if (isNil(config.custom) || isNil(key) || !canIDeleteThisSetting(config, settingName)) {
+    return false;
+  }
+  config.custom[key] = -1;
+  return true;
+}
+
+export function canIAssignThisKey(config: InputInterfaceConfig, key: InputKeys): boolean {
```
DayKev:
```suggestion
export function canAssignKey(config: InputInterfaceConfig, key: InputKeys): boolean {
```

src/utils/inputs-utils.ts
```diff
@@ -0,0 +1,219 @@
...
+ *
+ * @param config - The configuration object containing custom settings.
+ * @param settingName - The setting name to search for.
+ * @returns The key associated with the specified setting name, or undefined if none.
+ */
+export function getKeyWithSettingName(config: InputInterfaceConfig, settingName: InputSettings): InputKeys | undefined {
+  if (isNil(config.custom)) {
+    return undefined;
+  }
+  return Object.keys(config.custom).find((key) => config.custom![key] === settingName) as InputKeys;
```
DayKev:
```suggestion
  const { custom } = config;
  if (isNil(custom)) {
    return undefined;
  }
  return Object.keys(custom).find((key) => custom[key] === settingName) as InputKeys;
```

src/utils/inputs-utils.ts
```diff
@@ -0,0 +1,219 @@
...
+/**
+ * Retrieves the icon associated with the specified setting name.
+ *
+ * @param config - The configuration object containing icons.
+ * @param settingName - The setting name to search for.
+ * @returns The icon associated with the specified setting name.
+ */
+export function getIconWithSettingName(config: InputInterfaceConfig, settingName: InputSettings): string | undefined {
+  const key = getKeyWithSettingName(config, settingName);
+  return key ? getIconWithKey(config, key) : key;
```
DayKev:
```suggestion
  return key ? getIconWithKey(config, key) : undefined;
```

src/inputs-controller.ts
```diff
@@ -220,11 +246,18 @@ export class InputsController {
...
    */
-  setChosenKeyboardLayout(layoutKeyboard: string): void {
+  private setChosenKeyboardLayout(layout: KeyboardLayout): void {
     this.deactivatePressedKey();
-    this.initChosenLayoutKeyboard(layoutKeyboard);
+
+    const layoutKey = enumValueToKey(KeyboardLayout, layout).toLowerCase();
+    this.selectedDevice[Device.KEYBOARD] = layoutKey;
+    if (!this.configs[layoutKey]) {
+      this.setupKeyboard(layout); // Setup the keyboard for this layout
```
DayKev:
```suggestion
      this.setupKeyboard(layout);
```

src/inputs-controller.ts
```diff
@@ -238,10 +271,10 @@ export class InputsController {
...
    * If a gamepad name is provided, it uses that as the chosen gamepad; otherwise, it defaults to the currently chosen gamepad.
-   * @param gamepadName - Optional parameter to specify the name of the gamepad to initialize as chosen.
+   * @param gamepadName - Name of the gamepad to initialize as chosen.
    * @param emitInitEvent - Whether to send a gamepad initialization event. Default: `true`.
    */
-  initChosenGamepad(gamepadName?: string, emitInitEvent: boolean = true): void {
+  initChosenGamepad(gamepadName: string, emitInitEvent: boolean = true): void {
     if (gamepadName) {
       this.selectedDevice[Device.GAMEPAD] = gamepadName.toLowerCase();
     }
```
DayKev:
```suggestion
    this.selectedDevice[Device.GAMEPAD] = gamepadName.toLowerCase();
```
This `if` isn't necessary anymore since `gamepadName` is required, right?

src/inputs-controller.ts
```diff
@@ -314,14 +338,68 @@ export class InputsController {
...
+    baseConfig: InputInterfaceConfig,
+    lockedSettings: readonly InputSettings[] = [],
+    lockedBindings: readonly InputKeys[] = [],
+  ) {
+    const config = deepCopy(baseConfig);
+
+    // Copy existing custom bindings if any, otherwise use the default bindings
+    if (this.customMappings[id]) {
+      config.custom = this.customMappings[id];
+      delete this.customMappings[id];
```
DayKev, line 378: Why `delete`?

src/@types/inputs-types.ts
```diff
@@ -0,0 +1,64 @@
...
+  settings: Partial<Record<S, Button>>;
+  /** Default mappings of each key to a setting. -1 means no mapping. */
+  default: InputMappings<K, S>;
+  /** Keys of binding settings that cannot be changed. */
+  settingsBlacklist?: S[];
+  /**
+   * Keys of reserved buttons which cannot be (re)mapped.
+   * Will be filled automatically with the keys mapped by default to the settingsBlacklist.
+   */
+  keysBlacklist?: K[];
```
DayKev, line 57: Won't these always be defined/filled with at least some values?

src/inputs-controller.ts
```diff
@@ -314,14 +338,68 @@ export class InputsController {
...
+    config.settingsBlacklist = [...(config.settingsBlacklist ?? []), ...lockedSettings];
+    config.keysBlacklist = [...(config.keysBlacklist ?? []), ...lockedBindings];
+    for (const key of Object.keys(config.default)) {
+      if (config.settingsBlacklist.includes(config.default[key])) {
+        config.keysBlacklist.push(key as InputKeys);
+      }
+    }
+
+    this.configs[id] = config;
+    globalScene.gameData?.saveMappingConfigs(id, this.configs[id]);
```
DayKev, line 393: `gameData` can be `undefined` at this point? That seems like a problem.

src/inputs-controller.ts
```diff
@@ -314,14 +338,68 @@ export class InputsController {
   /**
    * Initializes or updates configurations for connected keyboards.
    */
-  setupKeyboard(): void {
-    for (const layout of ["default"]) {
-      const config = deepCopy(this.getConfigKeyboard(layout)) as InterfaceConfig;
-      config.custom = this.configs[layout]?.custom || { ...config.default };
-      this.configs[layout] = config;
-      globalScene.gameData?.saveMappingConfigs(this.selectedDevice[Device.KEYBOARD], this.configs[layout]);
+  setupKeyboard(layout: KeyboardLayout = settings.keyboard.layout): void {
```
DayKev:
```suggestion
  public setupKeyboard(layout: KeyboardLayout = settings.keyboard.layout): void {
```

src/inputs-controller.ts
```diff
@@ -359,8 +437,8 @@ export class InputsController {
   keyboardKeyDown(event): void {
```
DayKev: Do you know what type `event` is supposed to be?

src/inputs-controller.ts
```diff
@@ -408,26 +490,25 @@ export class InputsController {
...
-      !this.selectedDevice[Device.GAMEPAD]
-      || (globalScene.ui.getMode() !== UiMode.GAMEPAD_BINDING
-        && this.selectedDevice[Device.GAMEPAD] !== pad.id.toLowerCase())
-    ) {
+    let gamepadID = this.selectedDevice[Device.GAMEPAD];
+    if (!gamepadID || (globalScene.ui.getMode() !== UiMode.GAMEPAD_BINDING && gamepadID !== pad.id.toLowerCase())) {
       this.setChosenGamepad(pad.id);
     }
-    if (!this.gamepadSupport || pad.id.toLowerCase() !== this.selectedDevice[Device.GAMEPAD].toLowerCase()) {
+    gamepadID = this.selectedDevice[Device.GAMEPAD];
```
DayKev: This assignment should probably go inside the `if`.

src/inputs-controller.ts
```diff
@@ -487,7 +570,7 @@ export class InputsController {
    * @param id The identifier string of the gamepad.
    * @returns InterfaceConfig The configuration object corresponding to the identified gamepad type.
    */
-  getConfig(id: string): InterfaceConfig {
+  private getGamepadConfig(id: string): GamepadInterfaceConfig<any> {
```
DayKev, line 582: `<any>`?

src/inputs-controller.ts
```diff
@@ -507,17 +590,19 @@ export class InputsController {
   }
 
   /**
-   * Retrieves the configuration object for a keyboard layout based on its identifier.
-   *
-   * @param id The identifier string of the keyboard layout.
-   * @returns InterfaceConfig The configuration object corresponding to the identified keyboard layout.
+   * Retrieves the configuration object for a given keyboard layout.
+   * @param layout The {@linkcode KeyboardLayout} identifier string of the keyboard layout.
+   * @returns {@linkcode KeyboardInterfaceConfig} for this layout.
```
DayKev:
```suggestion
   * @returns The {@linkcode KeyboardInterfaceConfig} for this layout.
```
VSCode doesn't like when you start the `@returns` line with `{@linkcode`:
image

src/inputs-controller.ts
```diff
@@ -534,20 +619,25 @@ export class InputsController {
    * Retrieves the active configuration for the currently chosen device.
    * It checks if a specific device ID is stored in configurations and returns it.
    *
-   * @returns InterfaceConfig The configuration object for the active gamepad, or null if not set.
+   * @returns The {@linkcode InputInterfaceConfig} for the active gamepad, or null if not set.
```
DayKev:
```suggestion
   * @returns The {@linkcode InputInterfaceConfig} for the active gamepad, or `null` if not set.
```

src/inputs-controller.ts
MokaStitcher: so that we don't need to keep 2 instances of them in memory, although I guess it's a reference so probably doesn't change much? 🤔 
Ideally we wouldn't be loading these at startup in the first place, and only retrieve them as needed. But I didn't want to go through another rabbit hole with this in that PR so this was an attempt at mitigating the issue

src/inputs-controller.ts
MokaStitcher: Not sure it can, but since the configs can get initialized at startup by pressing a button or something during loading, I imagine it's possible in some weird situation, so I left that `?` from the original code for safety I guess.
The thing is, this code doesn't really need gameData. `saveMappingConfigs` doesn't need to be in GameData, it just writes in local storage but doesn't use anything from the class itself. I guess just like the way the custom configs are loaded this should get reworked. Not sure if this should be in this PR or not, ~~I'll try to make~~ I made an issue for it: #1305

src/inputs-controller.ts
MokaStitcher: nah, it's probably a JS or Phaser thing, but I haven't looked at it closer

src/inputs-controller.ts
MokaStitcher: <sub>i removed some of these redundant comments, but missed this one lol

src/inputs-controller.ts
MokaStitcher: hopefully it shouldn't cause issues to remove it. Those names are provided by Phaser/the browser

src/@types/inputs-types.ts
MokaStitcher: At the moment yes, because they are getting filled when loading the configs. But I didn't want all configuration files to have to specify an empty array even if they don't specify anything, so I marked it as optional. Not sure if there's a better approach

src/inputs-controller.ts
MokaStitcher: wasn't sure how to solve this. I made `GamepadInterfaceConfig` generic so that when writting the config it would ensure that the it had everything for all buttons of this gamepad, and not more. But when we retrieve the gamepadConfig here we don't care about the specifics.
This was the easiest way I found to not make it complain about the return types. Happy to be suggested something better cause I'm not super happy with how the config types were handled

src/inputs-controller.ts
MokaStitcher: I think I also had the concern that after that initial loading of the mappings, they can get modified as part of an actual config object. So those customMappings would get outdated, or they would need to be updated as well. Deleting them solves that issue.

MokaStitcher (commented):
<sub>my comments are all marked as pending somehow, so auto review i guess

Since the configs/inputs folder got heavily cleaned up thanks to DayKey, I just renamed it to `inputs` and moved the other inputs files there

DayKev:
Uh, hm. I assigned `Y` to "Submit (Alt)", then hit "assign button" (using `Z`). It worked fine. Then I hit `Y` to assign a different key to "Submit (Alt)", assigned `T` to it, and hit `Y` on "assign button". The game then thought I had the "submit" key continuously held down until I hit `T` again.

MokaStitcher:
> Uh, hm. I assigned `Y` to "Submit (Alt)", then hit "assign button" (using `Z`). It worked fine. Then I hit `Y` to assign a different key to "Submit (Alt)", assigned `T` to it, and hit `Y` on "assign button". The game then thought I had the "submit" key continuously held down until I hit `T` again.

Hm, I noticed a few oddities with assigning stuff on this PR, which I assumed were caused by not porting the UI changes over. Hence the description saying things should be working for the most part 
I'll see if I can reproduce it both in this PR and #1295. If it's only on this branch I'll ask to let it slide until we can get #1295 in as well :pray:

DayKev (dismissed):

src/inputs/inputs-controller.ts
```diff
@@ -75,12 +89,16 @@ export class InputsController {
...
 
-  // TODO interactions and configs are defined as maps but used as objects
+  // TODO: interactions is defined as a map but used as an object
   private interactions: Map<Button, Map<string, boolean>> = new Map();
-  private configs: Map<string, InterfaceConfig> = new Map();
+
+  /** Configurations for device that have been used within the current session. */
+  private configs: { [key: string]: InputInterfaceConfig } = {};
+  /** Used to store custom mappings from local storage until the proper configuration gets initialized. */
+  private customMappings: { [key: string]: InputMappings } = {};
```
innerthunder, line 98: Can these be made into Records?

src/inputs/inputs-controller.ts
```diff
@@ -356,11 +424,11 @@ export class InputsController {
    *
    * @param event The keyboard event.
    */
-  keyboardKeyDown(event): void {
+  private keyboardKeyDown(event): void {
     this.lastSource = "keyboard";
     this.ensureKeyboardIsInit();
-    const buttonDown = getButtonWithKeycode(this.getActiveConfig(Device.KEYBOARD), event.keyCode);
-    if (buttonDown !== undefined) {
+    const buttonDown = getButtonWithKeycode(this.getActiveConfig(Device.KEYBOARD)!, event.keyCode);
```
innerthunder, line 433:
```suggestion
    const config = this.getActiveConfig(Device.KEYBOARD);
    if (!config) {
      return; // The keyboard isn't setup, ignore the input
    }
    const buttonDown = getButtonWithKeycode(config, event.keyCode);
```
This was added to `keyboardKeyUp` but not here.

src/inputs/inputs-controller.ts
```diff
@@ -534,71 +617,79 @@ export class InputsController {
    * Retrieves the active configuration for the currently chosen device.
    * It checks if a specific device ID is stored in configurations and returns it.
    *
-   * @returns InterfaceConfig The configuration object for the active gamepad, or null if not set.
+   * @returns The {@linkcode InputInterfaceConfig} for the active gamepad, or `null` if not set.
    */
-  getActiveConfig(device: Device) {
-    if (this.configs[this.selectedDevice[device]]?.padID) {
-      return this.configs[this.selectedDevice[device]];
+  public getActiveConfig(device: Device): InputInterfaceConfig | null {
```
innerthunder: Should this be changed to return `InputInterfaceConfig | undefined`?

src/inputs/inputs-controller.ts
```diff
@@ -534,71 +617,79 @@ export class InputsController {
...
-  getLastSourceDevice(): Device {
+  private getLastSourceDevice(): Device {
     if (this.lastSource === "gamepad") {
       return Device.GAMEPAD;
     }
     return Device.KEYBOARD;
   }
 
-  getLastSourceConfig() {
+  private getLastSourceConfig(): InputInterfaceConfig | null {
```
innerthunder: Same here

src/inputs/inputs-controller.ts
MokaStitcher: Record expect the keys to be part of a predefined enum or type. So it would have to be a partial record because only configs for any device that has been used during the session would be stored in there.
I guess we could create a type for it if we think there's a benefit

src/inputs/inputs-controller.ts
MokaStitcher: ensureKeyboardIsInit initializes the config in onKeyboardKeyDown, so there shouldn't be a need to double check the config there
In onKeyboardKeyUp I chose to check the config rather than initialize the keyboard because I don't want the input to be interpreted if no down input was received before.
An 'up' event being received before any 'down' event is what caused the crash in #1094

src/inputs/inputs-controller.ts
MokaStitcher: It can be 🤷

src/inputs/inputs-controller.ts
DayKev: If this is the only place the type would be used it's fine to just keep it like this.

src/inputs/inputs-controller.ts
innerthunder: it's just a nit, but I'd rather only use `null` for when things intentionally have no value. That didn't seem to be the case here from what I could tell.

src/inputs/inputs-controller.ts
MokaStitcher: Thinking more about it, I'd have to double check but I don't think that wouldn't be possible anyway because for gamepad the key is something that depends on the gamepad type, browser, etc. We can't make a definite list of that

src/inputs/inputs-controller.ts
MokaStitcher: Will do

src/utils/inputs-utils.ts
MokaStitcher: whoops, had missed that one

MokaStitcher:
> Uh, hm. I assigned `Y` to "Submit (Alt)", then hit "assign button" (using `Z`). It worked fine. Then I hit `Y` to assign a different key to "Submit (Alt)", assigned `T` to it, and hit `Y` on "assign button". The game then thought I had the "submit" key continuously held down until I hit `T` again.

Bad news, it also happens in #1295. Good (?) news: it also happens in beta, so it's not a regression :plains:
I think the issue is that it switches the mapping before recording the "up" input so it treats the button as being continually pressed.
I'll see what I can do, but possibly will be fixed in 1295 since I'd rather work with the updated UI code

Edit: the issue should be fixed. It was a timing issue. When assigning a new key all previous keys considered as being pressed should've been cleared through `deactivatePressedKey()`. The issue was that the interval for the key being pressed down was being created *after* `deactivatePressedKey` was called, so it never got cleared. I reordered things in onKeyboardButtonDown and onGamepadButtonDown to solve the issue.
Note that when doing said remapping now the "error" sound effect plays, despite it working. That is part of the oddities I mentioned due to the UI code not being updated properly. The same thing happens when remapping things on controller, and it doesn't in #1295, so this should be fine once everything gets merged. I'll make sure to double check this scenario when the time comes though

src/inputs/inputs-controller.ts
```diff
@@ -314,14 +329,68 @@ export class InputsController {
...
+  private initGamepadConfig(gamepadID: string): void {
+    this.initConfig(gamepadID, this.getGamepadConfig(gamepadID), GAMEPAD_LOCKED_BINDINGS);
+  }
+
+  /**
+   * Initializes the config object for the given keyboard layout.
+   * @param - The {@linkcode KeyboardLayout}
+   * @param - Unique string corresponding to this layout
+   */
+  private initKeyboardConfig(layout: KeyboardLayout, layoutKey: string) {
```
innerthunder:
```suggestion
  private initKeyboardConfig(layout: KeyboardLayout, layoutKey: string): void {
```

src/inputs/inputs-controller.ts
```diff
@@ -314,14 +329,68 @@ export class InputsController {
...
+   * @param baseConfig - the default {@linkcode InputInterfaceConfig} for this device type
+   * @param lockedSettings - Optional. Array of binding settings that are not allowed to be remapped for this device
+   * @param lockedBindings - Optional. Array of keys/buttons that are not allowed to be mapped for this device
+   */
+  private initConfig(
+    id: string,
+    baseConfig: InputInterfaceConfig,
+    lockedSettings: readonly InputSettings[] = [],
+    lockedBindings: readonly InputKeys[] = [],
+  ) {
```
innerthunder:
```suggestion
  private initConfig(
    id: string,
    baseConfig: InputInterfaceConfig,
    lockedSettings: readonly InputSettings[] = [],
    lockedBindings: readonly InputKeys[] = [],
  ): void {
```

src/inputs/inputs-controller.ts
```diff
@@ -398,6 +470,25 @@ export class InputsController {
     }
   }
 
+  private createInputDownInterval(buttonDown: Button, controllerType: string) {
```
innerthunder:
```suggestion
  private createInputDownInterval(buttonDown: Button, controllerType: string): void {
```

src/utils/inputs-utils.ts
```diff
@@ -0,0 +1,214 @@
...
+}
+
+/**
+ * Retrieves the setting name associated with the specified keycode.
+ *
+ * @param config - The configuration object containing custom settings.
+ * @param keycode - The keycode to search for.
+ * @returns The setting name associated with the specified keycode, or -1 if there is no mapping
+ */
+export function getSettingNameWithKeycode(config: InputInterfaceConfig, keycode: number): InputSettings | -1 {
```
innerthunder, line 24: Why `-1` instead of `undefined`?

src/utils/inputs-utils.ts
```diff
@@ -0,0 +1,214 @@
...
+}
+
+/**
+ * Retrieves the setting name associated with the specified key.
+ *
+ * @param config - The configuration object containing custom settings.
+ * @param key - The key to search for.
+ * @returns The setting associated with the specified key, or -1 if none.
+ */
+export function getSettingNameWithKey(config: InputInterfaceConfig, key: InputKeys): InputSettings | -1 {
```
innerthunder, line 75: Same here

innerthunder (dismissed):

src/utils/inputs-utils.ts
MokaStitcher: That's how it was defined in the configs before, see here <https://github.com/Despair-Games/poketernity/blob/988f5df4221fee9f73c340548b442d5c54c6c17b/src/configs/inputs/cfg-keyboard-qwerty.ts>
I didn't change that, just added type information

MokaStitcher: Do you think we should change it?
Considering such mappings are stored as is in local storage, I wonder how it would look with undefined 🤔 
I guess another option is to convert the -1s to undefined here if we wanted to

innerthunder: Apparently they would be stored in `localStorage` as a string (`"undefined"`), so `-1` seems a bit more efficient? Probably not worth changing here, then.

innerthunder (approved):

DayKev (approved):

PR #1298, opened 2025-07-23
[Tests] Fix GameManager.faintOpponents for double battles

MokaStitcher:

2 less skipped tests

After finding a crash in #1236 I wanted to add a test to ensure transitions between single and double battles worked. I realized there was already a double battle test file with a similar test, which was disabled.
This PR fixes the issue that caused this test to get disabled.

<sub>~~the test now passes also in #1236 so maybe it doesn't fully fit the testing scenario I needed, but that's another issue~~

`GameManager.faintOpponents` now runs the `PostKnockoutPhase` queued by the first opponent's `FaintPhase` before faiting the second opponent. Without it, both mon's PostKnockoutPhases would run after both are fainted, causing them both to queue a VictoryPhase, skipping an extra wave.

<sub>This was only an issue with the test framework, the game itself waits for the first PostKnockoutPhase to run before starting the second mon's FaintPhase.

## Screenshots/Videos

running the 2 previously disabled/failing tests 100 times each 
image

## How to test the changes?

`pnpm test:silent double-battle`

## Checklist

- [] Have I tested the changes manually?

  - [] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes?


f-fsantos (dismissed):

test/test-utils/game-manager.ts
```diff
@@ -282,12 +282,12 @@ export class GameManager {
...
    * Faint all opponents currently on the field
-   * @todo This is currently bugged for double battles and can cause multiple waves
-   * to be cleared at once, especially if called before a turn's {@linkcode TurnStartPhase}
    */
   async faintOpponents() {
     await this.faintPokemon(this.scene.currentBattle.enemyParty[0]);
-    if (this.scene.currentBattle.double) {
+    if (this.scene.currentBattle.double && this.scene.currentBattle.enemyParty[1]?.isOnField()) {
+      // run the first PostKnockoutPhase now, otherwise both PostKnockoutPhases will queue a VictoryPhase
+      await this.phaseInterceptor.to("PostKnockoutPhase", true);
```
innerthunder:
```suggestion
      await this.phaseInterceptor.to("PostKnockoutPhase");
```
`true` is default

innerthunder (dismissed):

PR #1302, opened 2025-07-28
[Refactor] Convert `enum`s to `const` objects part 3

DayKev:
Less bloated javascript being downloaded to their device.
> [!NOTE]
> Some settings will be set to different values due to the enum values being changed, re-setting them will fix the issue. 
> Saves may change game modes due to the enum value changes. Classic runs will be marked "unknown" but when loaded and then re-saved the mode will update correctly to Classic.
> (edit by Moka: Pokemon Gender in existing runs will also be wrong. I recommend just wiping your save files to avoid issues)

TypeScript enums are bad.

See #1194

## Checklist


MokaStitcher (commented):
<https://github.com/Despair-Games/poketernity/blob/8c08f58beb86d057273f73e869d61b670e4bb115/src/ui/handlers/starter-select-ui-handler.ts#L3154>

DayKev:
> https://github.com/Despair-Games/poketernity/blob/8c08f58beb86d057273f73e869d61b670e4bb115/src/ui/handlers/starter-select-ui-handler.ts#L3154

I thought I checked for that one, huh.

src/enums/exp-gain-speed.ts
```diff
@@ -0,0 +1,22 @@
...
+  FAST: 2,
+  /** 4x speed */
+  FASTER: 3,
+  /** Skip gaining exp animation */
+  SKIP: 4,
+} as const;
+
+export type ExpGainSpeed = EnumValues<typeof ExpGainSpeed>;
+
+export const expGainSpeedMap = {
```
MokaStitcher: should this be in one of the constants file instead of here? 🤔

src/ui/components/battle-info.ts
```diff
@@ -655,9 +656,14 @@ export class BattleInfo extends Phaser.GameObjects.Container {
       const updatePokemonHp = () => {
         let duration = !instant ? clamp(Math.abs(this.lastHp - pokemon.hp) * 5, 250, 5000) : 0;
         const speed = settings.general.hpBarSpeed;
-        if (speed) {
-          duration = speed >= 3 ? 0 : duration / Math.pow(2, speed);
-        }
+        const speedMap = {
```
MokaStitcher, line 659: Don't have strong feeling about what to do about these, but I'd like xp speed and hp speed mapping to be handled the same way. Either both like here, both maps in their respective enum file, or both maps in a constants file

MokaStitcher: Is this one here only because it's needed only here, and the xp map is used in several places?

src/ui/handlers/egg-gacha-ui-handler.ts
```diff
@@ -340,7 +340,7 @@ export class EggGachaUiHandler extends MessageUiHandler {
     this.setGachaCursor(1);
 
     for (let g = 0; g < this.gachaContainers.length; g++) {
-      this.updateGachaInfo(g);
+      this.updateGachaInfo((g + 1) as GachaType);
```
MokaStitcher: alternative, make the loop start a 1 and end at `<= this.gachaContainers.length`

test/ui/battle-info.test.ts
```diff
@@ -13,7 +13,8 @@ vi.mock("../data/exp", ({}) => {
   };
 });
 
-describe("UI - Battle Info", () => {
+// TODO: figure out how to test this now
```
MokaStitcher, line 16: what got broken?

MokaStitcher (commented):
Thanks for the warning in the description 🙏 
Wouldn't the Gender indexing change also have messed up mon's genders in existing save files? Previously female mons would now be male and males... I don't know, hopefully the game doesn't crash? :')

src/ui/components/battle-info.ts
DayKev: Yeah it was only used here so I put it here, whereas the other one was used in multiple files.

src/enums/exp-gain-speed.ts
DayKev: I think it makes sense to be here since it's intrinsicly tied to the enum.

src/ui/handlers/egg-gacha-ui-handler.ts
DayKev: Good idea, that looks better.

test/ui/battle-info.test.ts
DayKev: The test spied on `Math.pow` but that was replaced with the animation speed map.

DayKev: Tbh the test seems like it was a bit fragile, since at any point a `Math.pow` could be put somewhere else and break the test.

src/ui/components/battle-info.ts
DayKev: I could move it anyway, since it's kinda tied to the enum.

DayKev:
>  Wouldn't the Gender indexing change also have messed up mon's genders in existing save files? Previously female mons would now be male and males... I don't know, hopefully the game doesn't crash?

Looks like some/most of the code already accidentally accounted for invalid values:
image
https://github.com/Despair-Games/poketernity/blob/ff3cccb92895bb1f1ea6f5b3ffcf1f02c5d50f2b/src/data/gender.ts#L9-L17

Using attract just failed without causing problems as well.

test/ui/battle-info.test.ts
innerthunder: UI tests strike again 🙄

innerthunder (approved):

MokaStitcher (approved):

PR #1303, opened 2025-07-29
[AI-Rework] Add Effect Scores for Arena/Battler-tag-modifying effects

innerthunder:

The AI should now properly evaluate the tag-modifying effects of moves, including:
- Defog and G-Max Wind Rage
- Brick Break and Psychic Fangs
- Rapid Spin and Mortal Spin
- Tidy Up
- Court Change
- Ingrain (removing the effects of Magnet Rise)
- Spit Up and Swallow

Part of the AI Rework

Added Effect Score functions for the following move attributes:
- RemoveArenaTagsAttr (Defog, Brick Break, etc.)
- SwapArenaTagsAttr (Court Change)
- RemoveAllSubstitutesAttr (Tidy Up)
- RemoveBattlerTagAttr (Rapid Spin)

## How to test the changes?

- [x] `pnpm typecheck`
- [x] `pnpm depcruise`
- [x] `pnpm test:silent`

New unit tests (run w/ `pnpm test[:silent] move-effect-scores/testName`):
- `defog`
- `rapid-spin`
- `tidy-up`
- `court-change`

## Checklist


src/constants/arena-tag-constants.ts
```diff
@@ -56,3 +67,17 @@ export const COURT_CHANGE_ARENA_TAG_TYPES = Object.freeze<ArenaTagType[]>([
   ArenaTagType.G_MAX_CANNONADE,
   ArenaTagType.G_MAX_VOLCALITH,
 ]);
+
+/**
+ * All {@linkcode ArenaTagType | ArenaTagTypes} that can be swapped by {@linkcode MoveId.COURT_CHANGE}
+ * and are perceived by the Enemy AI to have a harmful effect.
+ */
+export const HARMFUL_COURT_CHANGE_ARENA_TAG_TYPES = Object.freeze<ArenaTagType[]>([
```
MokaStitcher, line 75: Not for these PR necessarily, but I wonder if these types of const only used for AI should be placed in their own separate ai constants file 🤔 
That file would probably end up really big so maybe not

src/data/moves/move-attrs/swap-arena-tags-attr.ts
```diff
@@ -55,4 +58,36 @@ export class SwapArenaTagsAttr extends MoveEffectAttr {
...
+            + (HARMFUL_COURT_CHANGE_ARENA_TAG_TYPES.includes(tag.tagType)
+              ? MAJOR_EFFECT_SCORE_PENALTY
+              : MAJOR_EFFECT_SCORE_BONUS),
+          0,
+        ) ?? 0
+      );
+    };
+
+    const totalScore = getScore(playerTags) - getScore(enemyTags);
+    return Math.min(totalScore, SOFT_EFFECT_SCORE_LIMIT);
```
MokaStitcher, line 98: So, positive score have a limit to them, but not negative scores, is that intended?

src/data/moves/move-attrs/swap-arena-tags-attr.ts
```diff
@@ -55,4 +58,36 @@ export class SwapArenaTagsAttr extends MoveEffectAttr {
...
+
+  /**
+   * @returns The combined score from swapping all active Arena Tags of relevant {@linkcode COURT_CHANGE_ARENA_TAG_TYPES | type}.
+   * Each relevant tag yields a {@linkcode MAJOR_EFFECT_SCORE_BONUS} or {@linkcode MAJOR_EFFECT_SCORE_PENALTY},
+   * depending on which side the tag is located and whether it is {@link HARMFUL_COURT_CHANGE_ARENA_TAG_TYPES | harmful}.
+   * The total bonus from this attribute cannot exceed the {@linkcode SOFT_EFFECT_SCORE_LIMIT}.
+   */
+  public override getEffectScore(_user: EnemyPokemon, _target: Pokemon, _move: Move): number {
+    const { arena } = globalScene;
+    const playerTags = arena.getTags((t) => COURT_CHANGE_ARENA_TAG_TYPES.includes(t.tagType), ArenaTagSide.PLAYER);
```
MokaStitcher: wonder if we should take into account to remaining turns for the tags in some way. If a tag expires at the end of the turn is there really an incentive to swap? 🤔 (the answer is probably yes, sometimes, and sometimes not lol)

MokaStitcher (approved):
lgtm

src/constants/arena-tag-constants.ts
innerthunder: There's `ai-constants.ts` which has the scoring constants already, but so far I've organized all the AI-related ability/tag lists in their respective constant files

src/data/moves/move-attrs/swap-arena-tags-attr.ts
innerthunder: Yes. Positive scores have a limit mainly so that the AI still prioritizes KOs.

I may have to tweak the limits to account for moves with multiple attributes, but that's out of scope for this PR

src/data/moves/move-attrs/swap-arena-tags-attr.ts
innerthunder: Good point. The simplest approach would probably be just to exclude tags with a `turnCount` of 1 here.

src/constants/arena-tag-constants.ts
DayKev: Some of these could conceivably be used for reasons other than AI use (like the defog tag list, at the very least it's probably useful for tests), so having them here is fine.

src/data/moves/move-attrs/remove-arena-tags-attr.ts
```diff
@@ -47,4 +50,60 @@ export class RemoveArenaTagsAttr extends MoveEffectAttr {
...
+      (t) => this.tagTypes.includes(t.tagType) && [ArenaTagSide.BOTH, ...sides].includes(t.side),
+    );
+
+    /**
+     * Each {@link affectedTags | tag} contributes its {@link getTagRemovalBonus | base score},
+     * multiplied by -1 if the tag only exists on the enemy's (i.e. user's) side
+     */
+    const totalScore =
+      affectedTags?.reduce(
+        (score, tag) => score + this.getTagRemovalBonus(tag.tagType) * (tag.side === ArenaTagSide.ENEMY ? -1 : 1),
```
DayKev: Doesn't this mean if the enemy has screens up, Defog will get an effect score penalty (`getTagRemovalBonus` returns "major bonus", then this multiplies that by `-1`)?

test/ai/move-effect-scores/defog.test.ts
DayKev, line 1: There's no tests for the non-entry hazard arena tags that can be removed.

src/data/moves/move-attrs/remove-entry-hazard-attr.ts
DayKev, line 1: It seems like this attr could be removed, it's not really doing much.

src/constants/arena-tag-constants.ts
MokaStitcher: i was specifically talking about those like these ones that indicate some sort of "judgement" on the tag quality. But not a big deal

src/data/moves/move-attrs/swap-arena-tags-attr.ts
MokaStitcher: an alternative would be to apply a multiplier to the effect score based on the time left, but we probably shouldn't over-engineer this (or make the AI too smart), so that works 👍

src/data/moves/move-attrs/remove-arena-tags-attr.ts
MokaStitcher: It's a bit confusing but this is specifically the enemy, aka not player side, not "opponent" side. The AI code does everything from the perspective of an Enemy Pokemon, it couldn't be reused as is for a friendly AI for instance.
So in this case ENEMY side = AI side. So if the AI side has a screen that defog would remove, it should indeed get a negative score so that it avoids using the move.

MokaStitcher (approved):

src/data/moves/move-attrs/remove-arena-tags-attr.ts
DayKev: Ah, right. I was confusing "enemy" and "opponent".

test/ai/move-effect-scores/defog.test.ts
```diff
@@ -63,6 +63,32 @@ describe("Move Effect Scores - Defog", () => {
...
+  ])("Screen Removal", ({ tagName, tagType }) => {
+    it(`should be preferred when ${tagName} is on the opponent's side of the field`, async () => {
+      await game.classicMode.startBattle(SpeciesId.MAGIKARP);
+
+      game.scene.arena.addTag(tagType, 0, 0, MoveId.NONE, ArenaTagSide.PLAYER, true);
+      const enemy = game.field.getEnemyPokemon();
+      expect(enemy).toPreferSelectingMove(MoveId.DEFOG);
+    });
+
+    // This may need to be removed or adjusted when scoring is added for Defog's accuracy drop
```
DayKev:
```suggestion
    // This may need to be removed or adjusted when scoring is added for Defog's evasion drop
```

DayKev (approved):

PR #1306, opened 2025-08-01
[Refactor] Change `Pokemon#moveset` to `protected`

### LINKED ISSUE #1231: [Bug] Learning moves for transformed Pokemon does not check correct moveset length
### Describe the bug

Context: Pokemon with 4 moves and Imposter ability transformed into a Pokemon with 3 moves or less

When that Pokemon wants to learn a move the player will not be offered a move to forget, resulting in the new move being learnt automatically and overriding the last move

I have not tested the opposite case, a mon with 3 moves or less transforming into a mon with 4 moves, but it will likely have the opposite issue of being offered to replace a move when it's not needed

I have also not tested a mon with 4 moves being transformed to a mon with 4 moves but suspect it to work properly

### Reproduction

- overrides
```
  ABILITY_OVERRIDE: AbilityId.IMPOSTER,
  ITEM_REWARD_OVERRIDE: [{name: "RARE_CANDY"}, {name: "MEMORY_MUSHROOM"}],
```
- Start a run with a Pokemon with 4 move slots (Squirtle is good since it learns a move at level 6)
- Any Pokemon you transform into here should have less than 4 moves
- Go through the wave and level up to learn a move, or use a memory mushroom or TM
- The move will be learnt automatically and replace the move in the 4th slot

### Expected behavior

The player should be offered the option to forget a move, and pick which one to forget

### Screenshots / Videos

Moveset of transformed Bulbasaur (the summary screen shows the mon's base moveset, not the transformed one):

Image

Moveset of transformed Bulbasaur after learning Growth at level 6:

Image

### Session export file

_No response_

### User data export file

_No response_

### Additional context

likely just need to ensure the moveset checked for length when learning move is the base one, not the transformed one

DayKev:
Transformed pokemon will be able to learn moves correctly again.

Improved code safety.
- Fixes #1231

- `Pokemon#moveset` changed to `protected` and all uses updated to use the appropriate helper methods.
- New helper method `Pokemon#swapMoves` added to swap two moves in a Pokémon's moveset.
- New helper method `Pokemon#restoreMovePP` added to reset PP usage of all moves in a Pokémon's moveset to `0`.
- New helper method `Pokemon#setMoveset` added to overwrite the Pokémon's current moveset with a new moveset.

## Checklist

### DIFF
--- a/src/data/abilities/ab-attrs/forewarn-ab-attr.ts
+++ b/src/data/abilities/ab-attrs/forewarn-ab-attr.ts
@@ -4,6 +4,7 @@ import { getPokemonNameWithAffix } from "#app/messages";
 import { MoveId } from "#enums/move-id";
 import type { Pokemon } from "#field/pokemon";
 import type { PokemonMove } from "#field/pokemon-move";
+import type { Move } from "#moves/move";
 import { OneHitKOAttr } from "#moves/one-hit-ko-attr";
 import i18next from "i18next";
 
@@ -18,27 +19,25 @@ export class ForewarnAbAttr extends PostSummonAbAttr {
       movesets.push(...opponent.getMoveset());
     }
 
-    for (const move of movesets) {
-      if (move.getMove().isStatusMove()) {
+    for (const pokemonMove of movesets) {
+      const move = pokemonMove.getMove();
+      // TS incorrectly narrows `move` from `Move` to `never` if `move.isStatusMove()` is used here
+      if ((move as Move).isStatusMove()) {
         movePower = 1;
-      } else if (move.getMove().hasAttr(OneHitKOAttr)) {
+      } else if (move.hasAttr(OneHitKOAttr)) {
         movePower = 150;
-      } else if (
-        move.getMove().id === MoveId.COUNTER
-        || move.getMove().id === MoveId.MIRROR_COAT
-        || move.getMove().id === MoveId.METAL_BURST
-      ) {
+      } else if (move.id === MoveId.COUNTER || move.id === MoveId.MIRROR_COAT || move.id === MoveId.METAL_BURST) {
         movePower = 120;
-      } else if (move.getMove().power === -1) {
+      } else if (move.power === -1) {
         movePower = 80;
       } else {
-        movePower = move.getMove().power;
+        movePower = move.power;
       }
 
       // TODO: if multiple moves have the same effective power, one should be randomly chosen from them
       if (movePower > maxPowerSeen) {
         maxPowerSeen = movePower;
-        moveName = move.getName();
+        moveName = move.name;
       }
     }
 
--- a/src/data/abilities/ab-attrs/forewarn-ab-attr.ts
+++ b/src/data/abilities/ab-attrs/forewarn-ab-attr.ts
@@ -15,8 +15,7 @@ export class ForewarnAbAttr extends PostSummonAbAttr {
     const movesets: PokemonMove[] = [];
 
     for (const opponent of pokemon.getOpponents()) {
-      // TODO: should this consider summon data movesets (such as from Transform)?
-      movesets.push(...opponent.getMoveset(true));
+      movesets.push(...opponent.getMoveset());
     }
 
     for (const move of movesets) {
--- a/src/data/abilities/ab-attrs/forewarn-ab-attr.ts
+++ b/src/data/abilities/ab-attrs/forewarn-ab-attr.ts
@@ -3,47 +3,56 @@ import { globalScene } from "#app/global-scene";
 import { getPokemonNameWithAffix } from "#app/messages";
 import { MoveId } from "#enums/move-id";
 import type { Pokemon } from "#field/pokemon";
+import type { PokemonMove } from "#field/pokemon-move";
 import { OneHitKOAttr } from "#moves/one-hit-ko-attr";
 import i18next from "i18next";
 
 export class ForewarnAbAttr extends PostSummonAbAttr {
   public override apply(pokemon: Pokemon, simulated: boolean): boolean {
     let maxPowerSeen = 0;
-    let maxMove = "";
+    let moveName = "";
     let movePower = 0;
-    for (const opponent of pokemon.getOpponents()) {
-      for (const move of opponent.moveset) {
-        if (move.getMove().isStatusMove()) {
-          movePower = 1;
-        } else if (move.getMove().hasAttr(OneHitKOAttr)) {
-          movePower = 150;
-        } else if (
-          move.getMove().id === MoveId.COUNTER
-          || move.getMove().id === MoveId.MIRROR_COAT
-          || move.getMove().id === MoveId.METAL_BURST
-        ) {
-          movePower = 120;
-        } else if (move.getMove().power === -1) {
-          movePower = 80;
-        } else {
-          movePower = move.getMove().power;
-        }
+    const movesets: PokemonMove[] = [];
 
-        if (movePower > maxPowerSeen) {
-          maxPowerSeen = movePower;
-          maxMove = move.getName();
-        }
+    for (const opponent of pokemon.getOpponents()) {
+      // TODO: should this consider summon data movesets (such as from Transform)?
+      movesets.push(...opponent.getMoveset(true));
+    }
+
+    for (const move of movesets) {
+      if (move.getMove().isStatusMove()) {
+        movePower = 1;
+      } else if (move.getMove().hasAttr(OneHitKOAttr)) {
+        movePower = 150;
+      } else if (
+        move.getMove().id === MoveId.COUNTER
+        || move.getMove().id === MoveId.MIRROR_COAT
+        || move.getMove().id === MoveId.METAL_BURST
+      ) {
+        movePower = 120;
+      } else if (move.getMove().power === -1) {
+        movePower = 80;
+      } else {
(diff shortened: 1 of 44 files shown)


src/data/mystery-encounters/encounters/absolute-avarice-encounter.ts
```diff
@@ -377,12 +377,10 @@ export const AbsoluteAvariceEncounter: MysteryEncounter = MysteryEncounterBuilde
...
-        greedent.moveset = [
-          new PokemonMove(MoveId.THRASH),
-          new PokemonMove(MoveId.BODY_PRESS),
-          new PokemonMove(MoveId.STUFF_CHEEKS),
-          new PokemonMove(MoveId.SLACK_OFF),
-        ];
+        greedent.setMove(0, MoveId.THRASH);
+        greedent.setMove(1, MoveId.BODY_PRESS);
+        greedent.setMove(2, MoveId.STUFF_CHEEKS);
+        greedent.setMove(3, MoveId.SLACK_OFF);
```
innerthunder: I feel like a `setMoveset` method would be useful here (and some other places), e.g.

```ts
greedent.setMoveset([
  MoveId.THRASH,
  MoveId.BODY_PRESS,
  MoveId.STUFF_CHEEKS,
  MoveId.SLACK_OFF,
]);
```

DayKev: Yeah I realized as I was "self-reviewing" the changes that I would need to add one anyway due to my changes to `challenge.ts` not working properly as-is.

src/data/abilities/ab-attrs/forewarn-ab-attr.ts
```diff
@@ -3,47 +3,56 @@ import { globalScene } from "#app/global-scene";
...
-          || move.getMove().id === MoveId.METAL_BURST
-        ) {
-          movePower = 120;
-        } else if (move.getMove().power === -1) {
-          movePower = 80;
-        } else {
-          movePower = move.getMove().power;
-        }
+      // TODO: should this consider summon data movesets (such as from Transform)?
+      movesets.push(...opponent.getMoveset(true));
```
innerthunder:
```suggestion
      movesets.push(...opponent.getMoveset());
```
It should :pikasmile:

innerthunder (dismissed):

src/data/abilities/ab-attrs/forewarn-ab-attr.ts
```diff
@@ -3,47 +3,55 @@ import { globalScene } from "#app/global-scene";
...
-        } else if (move.getMove().power === -1) {
-          movePower = 80;
-        } else {
-          movePower = move.getMove().power;
-        }
+      movesets.push(...opponent.getMoveset());
+    }
+
+    for (const move of movesets) {
+      if (move.getMove().isStatusMove()) {
```
MokaStitcher: should `move.getMove()` be placed in a const here, rather than call it in every if and else if?

src/field/pokemon.ts
```diff
@@ -1390,7 +1391,11 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
 
   abstract getBossSegmentIndex(): number;
 
-  getMoveset(bypassSummonData: boolean = false): PokemonMove[] {
+  /**
+   * @param bypassSummonData - Whether to get the Pokemon's actual moveset
```
```suggestion
   * @param bypassSummonData - Whether to get the unmodified Pokemon's moveset. Default: `false`
```
not sure what the best way to express this is. But I feel that "actual" could be interpreted either way

src/data/moves/move-attrs/less-pp-more-power-attr.ts
```diff
@@ -10,10 +10,10 @@ import type { NumberHolder } from "#utils/common-utils";
 export class LessPPMorePowerAttr extends VariablePowerAttr {
   override apply(user: Pokemon, _target: Pokemon, move: Move, power: NumberHolder): boolean {
     const ppMax = move.pp;
-    const ppUsed = user.moveset.find((m) => m.moveId === move.id)?.ppUsed ?? 0;
+    const ppUsed = user.getMoveset(true).find((m) => m.moveId === move.id)?.ppUsed ?? 0;
```
MokaStitcher: should we really bypass summonData here? I'm assuming if a move with that attribute would ever get used that would be because it's part of the effective, current moveset?

src/data/moves/move-attrs/random-moveset-move-attr.ts
```diff
@@ -48,7 +48,7 @@ export class RandomMovesetMoveAttr extends CallMoveAttr {
         allies = [user];
       }
 
-      const partyMoveset = allies.flatMap((p) => p.moveset);
+      const partyMoveset = allies.flatMap((p) => p.getMoveset(true));
```
MokaStitcher, line 51: <sub>I guess I have a similar question here, but that would have to be answered from mainline in this case probably.

src/data/mystery-encounters/encounters/uncommon-breed-encounter.ts
```diff
@@ -73,23 +73,23 @@ export const UncommonBreedEncounter: MysteryEncounter = MysteryEncounterBuilder.
...
         pokemon: pokemon,
       };
-      if (pokemon.moveset.length < 4) {
-        pokemon.moveset.push(new PokemonMove(randomEggMoveId));
+      const moveset = pokemon.getMoveset(true);
+      if (moveset.length < 4) {
+        pokemon.setMove(moveset.length, randomEggMoveId);
       } else {
-        pokemon.moveset[0] = new PokemonMove(randomEggMoveId);
+        pokemon.setMove(0, randomEggMoveId);
```
MokaStitcher, line 85: how many times do we have these lines :despair:

src/data/mystery-encounters/utils/encounter-phase-utils.ts
```diff
@@ -366,9 +365,8 @@ export async function initBattleWithEnemyConfig(partyConfig: EnemyPartyConfig):
 
       // Set moves
       if (config?.moveSet && config.moveSet.length > 0) {
-        const moves = config.moveSet.map((m) => new PokemonMove(m));
-        enemyPokemon.moveset = moves;
-        enemyPokemon.summonData.moveset = moves;
```
MokaStitcher, line 371: uuuuh, why was it doing that? Could it be because just changing the moveset had a timing issue and got overridden after this code ran so they had to rely on summonData moveset?

src/ui/handlers/summary-ui-handler.ts
```diff
@@ -1053,7 +1051,7 @@ export class SummaryUiHandler extends UiHandler {
 
         for (let m = 0; m < 4; m++) {
           const move: PokemonMove | null =
-            this.pokemon && this.pokemon.moveset.length > m ? this.pokemon?.moveset[m] : null;
+            this.pokemon && this.pokemon.getMoveset(true).length > m ? this.pokemon.getMoveset(true)[m] : null;
```
MokaStitcher: could we put the moveset in a const above the `for`?

MokaStitcher (commented):
partial

src/field/pokemon.ts
DayKev: Good point, I'll add some clarification to the doc.

DayKev: What do you think of this?

> (Default `true`) Whether to get the Pokemon's actual/unmodified moveset (`true`) or the Pokemon's temporary/modified moveset (such as due to Transform) (`false`).

src/data/moves/move-attrs/random-moveset-move-attr.ts
DayKev: Hm, yeah, dunno if these moves look at the real or temporary moveset.

src/data/mystery-encounters/encounters/uncommon-breed-encounter.ts
DayKev: A few times, but once or twice it sets move index 1 instead of 0 lol

src/data/mystery-encounters/utils/encounter-phase-utils.ts
DayKev: Looking at the rest of the code, it shouldn't have been necessary... it would be nice if the comment wasn't useless, then maybe we'd know the reason.

src/data/abilities/ab-attrs/forewarn-ab-attr.ts
```diff
@@ -3,47 +3,54 @@ import { globalScene } from "#app/global-scene";
...
-          movePower = 80;
-        } else {
-          movePower = move.getMove().power;
-        }
+      movesets.push(...opponent.getMoveset());
+    }
+
+    for (const pokemonMove of movesets) {
+      const move = pokemonMove.getMove();
+      // TS incorrectly narrows `move` from `Move` to `never` if `move.isStatusMove()` is used here
```
MokaStitcher, line 24: :plains:

src/data/mystery-encounters/encounters/uncommon-breed-encounter.ts
MokaStitcher: I wanted to make a helper function but some other cases then roll a random replacement index or do other things, so probably not worth it :')

MokaStitcher (approved):

PR #1307, opened 2025-08-03
refactor: prevent serialization of full species in `PokemonSummonData`

DayKev:
Smaller save data (though the most impact will be to the server not having to store as much data).

### Text from PKR 6145
Writing the entire structure of `pokemonSpecies` when serializing data is completely unnecessary, as the data for a species is static. All we need are enough fields to reconstruct.
Also, this simplifies the zod schemas for the save migrator rewrite.

### Text from PKR 6145
Added a `toJSON` method to `PokemonSummonData` that does a few things
- `null` fields are replaced with `undefined`
- Fields that used `PokemonSpeciesForm` instead serialize an object containing the species ID and the particular form index.
- A new type helper, `CoerceNullProperitesToUndefined` was added.

## How to test the changes?
Save/load a session, including one with a transformed Pokémon (such as Ditto).

## Checklist


src/@types/pokemon-types.ts
```diff
@@ -11,10 +11,38 @@ import type { AbilityId } from "#enums/ability-id";
...
+  speciesForm?: SerializedSpeciesForm;
+  ability: AbilityId;
+  passiveAbility: AbilityId;
+  gender?: Gender;
+  stats: number[];
+  moveset: PokemonMove[];
+  types: ElementalType[];
+  addedType?: ElementalType;
+  turnCount: number;
+  waveTurnCount: number; // temporary
```
MokaStitcher, line 42: what's the "temporary" about?

src/@types/pokemon-types.ts
```diff
@@ -11,10 +11,38 @@ import type { AbilityId } from "#enums/ability-id";
...
+import type { SpeciesId } from "#enums/species-id";
 import type { StatusEffect } from "#enums/status-effect";
 import type { PokemonMove } from "#field/pokemon-move";
 import type { AttackMoveResult, TurnMove } from "#types/move-types";
 
+/**
+ * The type that {@linkcode PokemonSpeciesForm} is converted to when an object containing it serializes it.
+ */
+export type SerializedSpeciesForm = {
+  id: SpeciesId;
```
MokaStitcher: can we rename id to `speciesId` or is that used to save on storage space?

src/@types/pokemon-types.ts
```diff
@@ -11,10 +11,38 @@ import type { AbilityId } from "#enums/ability-id";
...
+  formIdx: number;
```
MokaStitcher: same, can we use `formIndex`

MokaStitcher (commented):
does server code need to be updated to accommodate for this?
<sub>I know pkr just stores the serialized session string, but is that what we do? i think we wanted to have full data validation, but ensure if that's the case at the moment

src/@types/pokemon-types.ts
DayKev: It saves on storage space, though I dunno if it matters that much for this specific field since transformed Pokémon shouldn't be that common?

DayKev:
> does server code need to be updated to accommodate for this?

Not sure, meant to ask Walker about it.

Edit: For future reference, the answer was no.

> i think we wanted to have full data validation, but ensure if that's the case at the moment

In terms of client-side validation this is building towards that.

src/@types/pokemon-types.ts
DayKev: `waveTurnCount` needs to be removed (it only exists to arbitrarily buff Fake Out, First Impression and Mat Block) but that's something for another PR.

MokaStitcher (approved):
lgtm

for reference, size of fresh daily run save file on beta (top) and this branch (bottom)
image
(Ko = KB 🥖 )

flx-sta (approved):

PR #1308, opened 2025-08-05
refactor: convert `enum`s to `const` objects part 4

DayKev:
- See #1194 for info

## Checklist


src/battle-scene.ts
```diff
@@ -3036,8 +3028,8 @@ export class BattleScene extends SceneBase {
     // If no valid encounters exist at tier, checks next tier down, continuing until there are some encounters available
     while (availableEncounters.length === 0 && tier !== null) {
       availableEncounters = biomeMysteryEncounters
-        .filter((encounterType) => {
-          const encounterCandidate = allMysteryEncounters[encounterType];
+        .filter((encType) => {
+          const encounterCandidate = allMysteryEncounters[encType];
```
MokaStitcher, line 3054: why the rename lol

src/ui/handlers/run-info-ui-handler.ts
```diff
@@ -206,7 +206,7 @@ export class RunInfoUiHandler extends UiHandler {
    */
   private async parseRunResult() {
     const genderIndex = settings.display.playerGender ?? PlayerGender.UNSET;
-    const genderStr = PlayerGender[genderIndex];
+    const genderStr = enumValueToKey(PlayerGender, genderIndex);
```
MokaStitcher: huh, why is it not lowercased here..
2 options: i18next ignores the case of the context so it works like this and we don't need to lowercase everywhere
or it's never worked <sub> it's probably that :plains:

MokaStitcher (dismissed):
no playtesting but lgtm

src/battle-scene.ts
DayKev: Variable name shadowing iirc.

src/ui/handlers/run-info-ui-handler.ts
MokaStitcher: it's broken without the lowercase indeed. fully unrelated but. if you want to fix it feel free ^^"

DayKev: So it just needs `.toLowerCase()`?

MokaStitcher: Yup

MokaStitcher (dismissed):

flx-sta (approved):

PR #1310, opened 2025-08-05
deps: update packages to the latest major version

DayKev:
Probably N/A

Keep packages up to date.

Packages updated to their latest available version, including major updates if available.
- Only `dependency-cruiser` and `i18next` had available major version updates.
- Node was kept at `22.x` instead of updating to `24.x` for now.
- Biome was not updated yet, waiting for `2.2` when Prettier can be replaced at the same time.

## Checklist


Tempo-anon (approved):
being up to date with deps is good

innerthunder (approved):

PR #1311, opened 2025-08-05
dev: update to `ES2023` & add `Promise.withResolvers` polyfill

DayKev:
https://github.com/pagefaultgames/pokerogue/pull/6160
<details><summary>ES2023 (text from PKR 6160)</summary>
<p>

Most likely no visible changes, possibly better performance.

Users with very old browser versions (those prior to 2022, which is when most mainstream browsers started supporting ES2023 features) may have compatibility issues. This is less than 3% of all internet users, according to Google's measurements.

Ability to use features from ES2021-ES2023.

All features of ES2021-ES2023 are now available for use (such as `Object.hasOwn()`, `Array#toSpliced()`, etc).

</p>
</details> 

https://github.com/pagefaultgames/pokerogue/pull/6103

<details><summary>`Promise.withResolvers` (text from PKR 6103)</summary>
<p>

N/A

I have been wanting to use `Promise.withResolvers` for a while now to simplify some logic. However, it is a feature new to ES2024, which is not yet considered "Baseline - Widely available".

I did a lot of digging into best practices for polyfills. I considered vite plugins, `core-js`, and `es-shims`.

Ultimately, adding dependencies (which aren't even typed!) via `es-shims` did not seem to be the best option; we'd be adding a dependency for each new polyfill we want. `es-shims` also makes larger assumptions.

The polyfills in `core-js` are too heavy, adding too much code.

[Cloudflare's polyfills](https://cdnjs.cloudflare.com/polyfill/) do not yet have es2024, which is what introduced Promise.withResolvers.

I don't know much about polyfills. One of the benefits of using `es-shims` is that it would hopefully be updated if the spec or requirements of the polyfill were to change (we'd just have to bump the polyfill version).
However, if we go that route, then we'd need to add a new dependency for each polyfill into our `package.json` file. This really seems like a headache.

I'm open to a larger discussion about the best approach here; I'm not as comfortable with best practices for polyfills.

However, I would really like to have `Promise.withResolvers` in, as it would make quite a bit of code more ergonomic.

- Created a new file, `polyfills.ts`. for shimming our own polyfills
- Added `typings/polyfills.d.ts` to provide global typescript definitions for these polyfills
- Added an import to this file on the top of `main.ts`.

Note: Only `main.ts` needs to import `polyfills`. No other file should.

</p>
</details> 

## Checklist


Tempo-anon (commented):
The `name` field is removed from the `Pokemon` and `Trainer` class?

DayKev:
> The `name` field is removed from the `Pokemon` and `Trainer` class?

They're already defined (as `string`s) in the Phaser class they're extending, the override was never necessary.

innerthunder (approved):

PR #1312, opened 2025-08-05
ai-rework: Add Effect Scores for Status Effect inflicting moves

innerthunder:

The Enemy AI should now evaluate moves that inflict non-volatile status effects more precisely.

Part of the AI Rework.

The following attributes now have Effect Score methods:
- `StatusEffectAttr`
- `MultiStatusEffectAttr`

> [!NOTE]
> These attributes use *chance-based* scores. To recap, the raw effect score is multiplied by `move.chance / 100`, then "tiered" based on the result. For example, an effect scored at 1.4 after the multiplier has a 40% chance to yield (+2) and a 60% chance to yield (+1). 

## How to test the changes?

- [x] `pnpm typecheck`
- [x] `pnpm depcruise`
- [x] `pnpm test:silent`

New unit tests  – can be run via `pnpm test[:silent] move-effect-scores/testName`:
- `spore`
- `thunder-wave`
- `will-o-wisp`
- `toxic`

## Checklist

### DIFF
--- a/src/constants/ability-constants.ts
+++ b/src/constants/ability-constants.ts
@@ -43,3 +43,29 @@ export const ELECTRIC_IMMUNE_ABILITIES: Readonly<AbilityId[]> = Object.freeze([
   AbilityId.LIGHTNING_ROD,
   AbilityId.MOTOR_DRIVE,
 ]);
+
+/** Abilities that grant a benefit when the source is under a non-volatile status effect */
+export const STATUS_SYNERGY_ABILITIES: Readonly<AbilityId[]> = Object.freeze([
+  AbilityId.GUTS,
+  AbilityId.QUICK_FEET,
+  AbilityId.MARVEL_SCALE,
+  AbilityId.MAGIC_GUARD,
+]);
+
+/** Abilities that grant a benefit when the source is burned */
+export const BURN_SYNERGY_ABILITIES: Readonly<AbilityId[]> = Object.freeze([
+  ...STATUS_SYNERGY_ABILITIES,
+  AbilityId.FLARE_BOOST,
+]);
+
+/** Abilities that grant a benefit when the source is poisoned (or badly poisoned) */
+export const POISON_SYNERGY_ABILITIES: Readonly<AbilityId[]> = Object.freeze([
+  ...STATUS_SYNERGY_ABILITIES,
+  AbilityId.TOXIC_BOOST,
+  AbilityId.POISON_HEAL,
+]);
+
+export const POISONING_SYNERGY_ABILITIES: Readonly<AbilityId[]> = Object.freeze([
+  AbilityId.MERCILESS,
+  AbilityId.POISON_PUPPETEER,
+]);
--- a/src/constants/ability-constants.ts
+++ b/src/constants/ability-constants.ts
@@ -45,7 +45,7 @@ export const ELECTRIC_IMMUNE_ABILITIES: Readonly<AbilityId[]> = Object.freeze([
 ]);
 
 /** Abilities that grant a benefit when the source is under a non-volatile status effect */
-export const ANY_STATUS_SYNERGY_ABILITIES: Readonly<AbilityId[]> = Object.freeze([
+export const ANY_STATUS_SYNERGY_ABILITIES = Object.freeze<AbilityId[]>([
   AbilityId.GUTS,
   AbilityId.QUICK_FEET,
   AbilityId.MARVEL_SCALE,
@@ -53,20 +53,20 @@ export const ANY_STATUS_SYNERGY_ABILITIES: Readonly<AbilityId[]> = Object.freeze
 ]);
 
 /** Abilities that grant a benefit when the source is burned */
-export const BURN_SYNERGY_ABILITIES: Readonly<AbilityId[]> = Object.freeze([
+export const BURN_SYNERGY_ABILITIES = Object.freeze<AbilityId[]>([
   ...ANY_STATUS_SYNERGY_ABILITIES,
   AbilityId.FLARE_BOOST,
 ]);
 
 /** Abilities that grant a benefit when the source is poisoned (or badly poisoned) */
-export const POISON_SYNERGY_ABILITIES: Readonly<AbilityId[]> = Object.freeze([
+export const POISON_SYNERGY_ABILITIES = Object.freeze<AbilityId[]>([
   ...ANY_STATUS_SYNERGY_ABILITIES,
   AbilityId.TOXIC_BOOST,
   AbilityId.POISON_HEAL,
 ]);
 
 /** Abilities that grant a benefit when the source poisons another Pokemon */
-export const POISONING_SYNERGY_ABILITIES: Readonly<AbilityId[]> = Object.freeze([
+export const POISONING_SYNERGY_ABILITIES = Object.freeze<AbilityId[]>([
   AbilityId.MERCILESS,
   AbilityId.POISON_PUPPETEER,
 ]);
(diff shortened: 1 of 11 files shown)


src/constants/ability-constants.ts
Tempo-anon, line 1: Hm I wonder. Would it be better to use the flag system that moves currently use to track specific categories of abilities rather than having lists grouping them together?

src/constants/ability-constants.ts
```diff
@@ -43,3 +43,29 @@ export const ELECTRIC_IMMUNE_ABILITIES: Readonly<AbilityId[]> = Object.freeze([
   AbilityId.LIGHTNING_ROD,
   AbilityId.MOTOR_DRIVE,
 ]);
+
+/** Abilities that grant a benefit when the source is under a non-volatile status effect */
+export const STATUS_SYNERGY_ABILITIES: Readonly<AbilityId[]> = Object.freeze([
```
Tempo-anon: nit: I'd rename this to include something like `any` because my first thought was "flare boost and toxic boost are missing from here" rather than this list being the generic basecase for all abilities that synergize with nonvolatile status

src/constants/ability-constants.ts
```diff
@@ -43,3 +43,29 @@ export const ELECTRIC_IMMUNE_ABILITIES: Readonly<AbilityId[]> = Object.freeze([
...
+  AbilityId.GUTS,
+  AbilityId.QUICK_FEET,
+  AbilityId.MARVEL_SCALE,
+  AbilityId.MAGIC_GUARD,
```
Tempo-anon, line 52: If magic guard is included as an ability here for its effect in blocking the downsides of status, should other abilities like immunity, comatose, and water veil be here as well? My initial thought process is they probably shouldn't if these abilities are what makes certain combat items spawn

src/data/moves/move-attrs/status-effect-attr.ts
```diff
@@ -72,11 +88,92 @@ export class StatusEffectAttr extends ChanceBasedMoveEffectAttr {
...
+    if (this.effect === StatusEffect.BURN && BURN_SYNERGY_ABILITIES.some((abId) => ally.hasAbility(abId))) {
+      return this.getRandomScore(user, 80, 2);
+    }
 
-    return !pokemon.hasNonVolatileStatusEffect() && pokemon.canSetStatus(this.effect, true, false, user) ? score : 0;
+    if (
+      [StatusEffect.POISON, StatusEffect.TOXIC].includes(this.effect)
+      && POISON_SYNERGY_ABILITIES.some((abId) => ally.hasAbility(abId))
+    ) {
+      const bonusChance = this.effect === StatusEffect.POISON ? 70 : 60;
```
Tempo-anon: nit: some magic numbers

src/data/moves/move-attrs/status-effect-attr.ts
```diff
@@ -72,11 +88,92 @@ export class StatusEffectAttr extends ChanceBasedMoveEffectAttr {
...
+    return this.getStatusEffectScore(user, target, this.effect);
+  }
+
+  /**
+   * @returns A base Effect Score depending on the given {@linkcode StatusEffect}:
+   * - Poison and Toxic Poison grant (+1)/(+1.5) respectively. If the user has Poison Puppeteer or
+   * Merciless, this bonus is increased to (+2)/(+2.5).
+   * - Paralysis grants (+1) if the user outspeeds the target, and (+2) otherwise.
+   * - Sleep and Freeze grant (+2.5) in all cases
+   * - Burn grants (+2) if the target has a physical affinity (ATK > SPATK), and (+1) otherwise.
```
Tempo-anon, line 156: This is a pretty good starting point for scoring statuses

src/constants/ability-constants.ts
innerthunder: The flag system won't have the granularity we need in some cases (e.g. all post-summon effects fall under the `POST_SUMMON` flag). I'd rather just use ability lists to be consistent

src/constants/ability-constants.ts
innerthunder: The difference is that Magic Guard negates the damaging effects of Burn/Poison and not the statuses themselves. For the ally target scoring, including immunity, comatose, etc. would basically reward the user for doing nothing.

src/data/moves/move-attrs/status-effect-attr.ts
```diff
@@ -72,11 +88,100 @@ export class StatusEffectAttr extends ChanceBasedMoveEffectAttr {
...
 
-    return !pokemon.hasNonVolatileStatusEffect() && pokemon.canSetStatus(this.effect, true, false, user) ? score : 0;
+    if (
+      [StatusEffect.POISON, StatusEffect.TOXIC].includes(this.effect)
+      && POISON_SYNERGY_ABILITIES.some((abId) => ally.hasAbility(abId))
+    ) {
+      /**
+       * The chance to grant a bonus for poisoning the user's ally (with a status move).
+       * {@link StatusEffect.TOXIC | Badly poisoning} an ally is less likely to grant a bonus
+       * than {@link StatusEffect.POISON | poisoning} an ally.
```
DayKev, line 130: How come? I don't think there are any abilities that differentiate between the user being Posioned or Badly Poisoned.

src/field/enemy-pokemon.ts
```diff
@@ -211,6 +211,13 @@ export class EnemyPokemon extends Pokemon {
    * @returns the sum of this move's score components against the given target
    */
   public getMoveScore(target: Pokemon, move: Move): number {
+    // If the move is known to have no effect on the target, return a Bad Move Penalty.
+    // TODO: This is necessary for some status moves, but is inefficient for attacks
+    // since effectiveness is calculated again for Attack Score.
+    if (target.getMoveEffectiveness(this, move, AbilityApplyMode.REVEALED) === 0) {
+      return BAD_MOVE_PENALTY;
+    }
```
DayKev: Should it be `if (status move and ineffective) { return bad move } else { ... }` or something then?

src/data/moves/move-attrs/status-effect-attr.ts
innerthunder: There aren't, but being badly poisoned leaves less time for the ally to make use of Guts / Toxic Boost / etc. compared to being poisoned (assuming the ally stays active). There are some abilities that negate the damage either way, though...

src/data/moves/move-attrs/multi-status-effect-attr.ts
```diff
@@ -18,16 +19,20 @@ export class MultiStatusEffectAttr extends StatusEffectAttr {
...
+  /**
+   * @returns The average base Effect Score among each of this attribute's {@linkcode effects},
+   * according to {@linkcode getStatusEffectScore}
+   */
+  public override getRawEffectScore(user: EnemyPokemon, target: Pokemon, _move: Move): number {
+    const totalStatusEffectScore = this.effects.reduce(
+      (score, effect) => score + this.getStatusEffectScore(user, target, effect),
+      0,
+    );
+    return totalStatusEffectScore / this.effects.length;
```
MokaStitcher, line 36: do we want to round this?

MokaStitcher (approved):
lgtm

src/data/moves/move-attrs/multi-status-effect-attr.ts
innerthunder: No. The same chance-based logic applies as in `StatusEffectAttr`, so e.g. if the effects of Dire Claw are scored as
- Poison: (+1)
- Paralysis: (+1) [the user outspeeds the target]
- Sleep: (+2.5)

then the final score for Dire Claw's overall effect is 75%(+1) (1.5 average * 50% chance)

MokaStitcher (approved):

PR #1314, opened 2025-08-07
refactor: Pokemon IDs are now unique and can't be `0`

### LINKED ISSUE #598: [Bug] Twins Trainer Class Generates Pokémon with Identical IDs
Copied over from PKR 5067

### Describe the bug

the "Twins" Trainer Class generates Pokemon with identical IDs.
This results in both Pokemon having:
- Identical IVS
- Identical Shinyness
- A shared Moveset/PP consumption Flyout (V Menu)
- Both Icons Flash when one is Targeted
- Items are no longer consumed after one has Fainted
- Items are Shared between both Pokemon

Link to Discord's Bug Report (Trainer Doubles Battles Generate Pokemon with the same ID)
https://discord.com/channels/1125469663833370665/1303162564687167500

### Reproduction

Paste in Battle.ts in `export const classicFixedBattles: FixedBattleConfigs = { ... }`
https://github.com/pagefaultgames/pokerogue/blob/206419c9f176cf43cb3e8639dfee175664afa5f5/src/battle.ts#L545
and start a Classic run, then advance to Wave 2. Saving and reloading will allow you view an 'Object' that contains 'EnemyParty' in the Developer Console.

```
[2]: new FixedBattleConfig().setBattleType(BattleType.TRAINER)
    .setGetTrainerFunc(scene => new Trainer(scene, TrainerType.TWINS, TrainerVariant.DOUBLE)),
```

### Expected behavior

"Twins" Trainer class should not generate twin Pokémon with the same ID.

### Screenshots / Videos

Twins' Pokemon IDs match

SameID

Doubles Psychic's Pokemon do not share IDs

DoublesPsychicsdifferentID

Both Icons flash when a Pokemon is individually Targeted

https://github.com/user-attachments/assets/b533ef85-a0da-454a-8cb1-b63be611a7c2

Items aren't used if one of the Pokemon has Fainted

https://github.com/user-attachments/assets/14e338f4-5472-4029-9ef7-975a3c6f955c

Move/PP Flyout is Shared, is Inaccurate

https://github.com/user-attachments/assets/58ea16e3-d3d8-456c-86f7-74a335bd8a18

Shinyness is shared (search by twins in [#shiny-showcase](https://discord.com/channels/1125469663833370665/1230751815482736640))

ShinyShared

### Session export file

_No response_

### User data export file

_No response_

### Additional context

https://github.com/pagefaultgames/pokerogue/blob/206419c9f176cf43cb3e8639dfee175664afa5f5/src/field/trainer.ts#L376

The "Twins" Trainer class is the only one that uses `useSameSeedForAllMembers`

Twins can rarely generate with identical Pokémon, Seen with two Eevees an two Wurmples
https://github.com/pagefaultgames/pokerogue/blob/206419c9f176cf43cb3e8639dfee175664afa5f5/src/data/trainer-config.ts#L1502

[TwinsDoubleWurmple.txt](https://github.com/user-attachments/files/18287085/TwinsDoubleWurmple.txt)

TwinsDoubleWurmple

Double Eevee

DoubleEevee

DayKev:
Less random bugs due to multiple Pokémon having the same ID. This primarily affects the Twins trainer battle.

If multiple Pokémon have the same ID (or an ID less than `4`), things break.

- `BattleScene#nextPokemonID` added, which is the next ID that will be assigned to a Pokémon. This is a simple incrementing integer.
- `BattleScene#getNextPokemonID()` added to get the next ID to assign to a Pokémon and then increment `nextPokemonID`.
- `BattleScene#updateNextPokemonID()` added which sets `nextPokemonID` to the max of the passed in value and the current value of `nextPokemonID`. Used when loading from a save to prevent ID collisions from pre-existing pokemon.
- The `Pokemon` constructor calls `BattleScene#getNextPokemonID()` to assign an ID when not loading from a save.
- `Pokemon#personalityValue` added which is a non-unique random integer from `0` to `2^32 - 1` that is used to [determine shininess](https://bulbapedia.bulbagarden.net/wiki/Personality_value#Shininess) instead of the Pokémon's ID.

## Checklist


innerthunder (dismissed):

src/field/pokemon.ts
```diff
@@ -500,11 +502,22 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+  public generateId(): number {
+    // ID has already been generated before, don't allow assigning new IDs
+    if (!isNil(this.id)) {
+      return this.id;
+    }
+
+    let id = randSeedInt(4294967295, 1);
+    while (globalScene.activePokemonIDs.has(id)) {
+      id = randSeedInt(4294967295, 1);
+    }
```
Tempo-anon: Does this need to be randomly generated if IDs will be checked from a set? Theoretically you could just take the largest number from the set and just add 1 to it or something if a unique identifier is required

src/field/pokemon.ts
```diff
@@ -500,11 +502,22 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
    */
-  public generateId(): void {
-    this.id = randSeedInt(4294967295);
+  public generateId(): number {
+    // ID has already been generated before, don't allow assigning new IDs
+    if (!isNil(this.id)) {
+      return this.id;
+    }
+
+    let id = randSeedInt(4294967295, 1);
```
Tempo-anon: The collision rate of 2^32 is 50% with 2^16 trials. This seems pretty unlikely in any run

src/field/pokemon.ts
```diff
@@ -500,11 +502,22 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
   }
 
   /**
-   * Sets this Pokemon's ID to be a random integer from 0 to 2^32 - 1, inclusive.
-   * @todo This should be `protected` or `private` but MEs currently call it
+   * Generates a random integer from `1` to `2^32 - 1`, inclusive, to be assigned to as the Pokemon's ID.
```
Tempo-anon: What's the special issue with the ID being 0?

Tempo-anon (commented):
Does this fix the issue with twins having Pokemon with the same ID and causing bugs to happen? :pikasmile:

src/field/pokemon.ts
DayKev: There's a bunch of places where things like arena/battler tags have a `sourceId` field that can be `undefined`, and the checks to see if it's `undefined` are `if (this.sourceId)` which also wouldn't pass if the `sourceId` was `0`. It's simpler to just prevent IDs of `0` than change every single `if` statement lol.

src/field/pokemon.ts
DayKev: In any individual run it probably won't happen, but over the lifetime of the game and over every run played, it will probably happen.

src/field/pokemon.ts
DayKev: True, but it was simpler to do this than add handling to make sure the number doesn't go above `2 ** 32 - 1`.

Tempo-anon: If the only thing we care about is ensuring that there are no collisions then this could just be some kind of iterator especially since there's a upper limit to how many Pokemon will even exist in the course of a run. (Quick napkin math example of 6 * 6000 = 36000)

DayKev:
> Does this fix the issue with twins having Pokemon with the same ID and causing bugs to happen? :pikasmile:

The Twins' Pokémon have different IDs (dunno if it was this PR that fixed it, I didn't check if they had different IDs on beta).

src/field/pokemon.ts
DayKev: Wdym?

Tempo-anon: I'm just wondering about the usecase of ID. From my understanding it's just something to uniquely identify a Pokemon in the context of a specific run. Since this PR updates the global scene to keep track of all existing IDs I'm wondering if instead of a set of numbers that still get randomly generated the global scene is now an int that starts with 1 and every new Pokemon generated is set equal to that number and then increments it.

in battle-scene.ts
```
public nextPokemonId = 1;
```

in pokemon.ts
```
this.id = globalScene.nextPokemonId;
globalScene.nextPokemonId += 1;
```

Note that this means that when a Pokemon is destroyed the `nextPokemonId` is not decremented but that shouldn't matter as we're only keeping track of information by the int

MokaStitcher: It's a flair thing, similar to mainline. And especially if we intend to implement multiplayer stuff i feel like this solution would add even more headache than

MokaStitcher:
> > Does this fix the issue with twins having Pokemon with the same ID and causing bugs to happen? :pikasmile:
> 
> The Twins' Pokémon have different IDs (dunno if it was this PR that fixed it, I didn't check if they had different IDs on beta).

They had the same Id before, because they are generated with the same seed on purpose so that their mons match (plusle with minun for instance). Which is why I asked you to check how twins behaved with your changes. They seem to no longer have the same if buy retain the matching mon thing, which is good

MokaStitcher:
See #598

src/field/pokemon.ts
Tempo-anon: Well that's what I'm trying to clarify here. If the only use of the `id` field in the `Pokemon` class is to distinguish individual Pokemon then we can just make it an int and have the global scene keep track of the next id to give out. In terms of multiplayer I imagine there's still only going to be one global scene and the whole process of getting and incrementing the id being tracked in the global scene to be atomized.

The personality value of a Pokemon in the mainline games is also used to calculate other values like shininess and gender/ability/nature/characteristic/forms/size/evolution in addition to identifying a Pokemon.

DayKev: We use Pokémon IDs to calculate shininess matching the mainline method.
I think they might be used for one other thing like that, not sure.

If we weren't using them for anything other than identifiers then yeah I would've just made them unique incrementing numbers or something.

Tempo-anon: Hm okay.

Very very minor but this means the chances of shinies will change based on how many IDs are in the set of existing IDs at a certain time.

MokaStitcher: Doesn't seem minor to me at all, this number could be manipulated. Since eggs will contribute to it, you could get any guaranteed shiny from same species eggs if you know what you are doing
If a shiny id exists in the very low numbers it would also be noticed immediately by players
And if there is none... Well players won't get shinies unless they play for an extended period of time

MokaStitcher: Or do you mean with the current implementation from this PR? In which case I'm not sure I follow. The chance of pulling 2 exact same IDs naturally is almost non existent. If it ever happens it's because we messed something up (or the twins)
Which makes me wonder if we should print a console warning or something whenever an id gets rerolled

Tempo-anon: I was just asking if the id was being used for anything at all or if it was just an identifier. If it is just an identifier then I think my suggestion of just keeping a counter that increments would make the most sense but apparently it is being used 🤷 

Though now that I think about it I believe the shiny check is only done when a new Pokemon is created. Starting a run with 6 shinies won't affect new shiny rates at all but catching a shiny and then keeping it on the team will (probably a negligible amount)

Eggs hatching new Pokemon with their own IDs..... I'm not as familiar with the exact technicals of that but I wouldn't be surprised if it creates an actual `Pokemon` object that then needs its own ID that's also being used for shiny checking.

You can probably ignore most of my rambling. I don't think it matters too much but might be worth documenting somewhere.

MokaStitcher: > I wouldn't be surprised if it creates an actual Pokemon object that then needs its own ID that's also being used for shiny checkin

They do

DayKev: Eggs do create `PlayerPokemon` objects, though I believe they're properly deleted afterwards now.

In mainline, Pokémon PVs aren't unique afaik, so I could add a new PV field to the `Pokemon` class and then change it so the things that are using Pokémon IDs as PVs here instead use the new PV field (which no longer needs to be unique). Then make the existing ID field just an incrementing (but unique) number.

DayKev:
Updated to use basic incrementing integers for the Pokemon's ID. Shininess now uses `Pokemon#personalityValue` (which is a non-unique integer from `0` to `2^32 - 1`) instead of `Pokemon#id`.

src/field/pokemon.ts
```diff
@@ -512,9 +519,9 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
       return this.id;
     }
 
-    let id = randSeedInt(4294967295, 1);
+    let id = 1;
```
MokaStitcher: If we're going with this I think we might as well defer that logic to the scene itself
Call `globalScene.getNewPokemonID()` or something?

MokaStitcher: This way we don't make the activeIDs list public and can use helper functions to add/remove them instead

DayKev: True.

SirzBenjie (approved):
This is a solid way of handling it.

MokaStitcher:
Does this requires server side updates due to new pv field?

src/battle-scene.ts
```diff
@@ -725,6 +730,19 @@ export class BattleScene extends SceneBase {
     });
   }
 
+  /** Gets the next available Pokemon ID value and then updates the counter. */
+  public getNextPokemonID(): number {
+    return this.nextPokemonID++;
+  }
+
+  /**
+   * Sets {@linkcode nextPokemonID} to the maximum of its current value and the passed in value plus 1. \
```
MokaStitcher, line 743: for my own curiosity, what does the `\ ` do?

MokaStitcher (dismissed):
lgtm

We may want to consider resetting nextPokemonID to 1 in `scene.reset()` if we want to avoid the ids in a new run to sort of depend on the ids in previous run.
It shouldn't be an issue most of the time but if we imagine that a player loads an existing save, then finishes the run and starts a new one, the ids in that new run would depend on the IDs in the first run. Now if we assume that every time they load the game they load an existing run and later start another run, the IDs would keep snowballing up and up over time, leading to things imploding eventually. It's quite an unlikely scenario because we'd need to reach a very high number, but considering that at the moment Pokemon objects are being created quite liberally in different parts of the code, it's probably better to be safe than sorry :')
I think having a reset point for the ids should help prevent this risk

src/battle-scene.ts
DayKev: A `\` at the end of a line in Markdown creates a linebreak, and TSDocs use Markdown.

DayKev:
> We may want to consider resetting nextPokemonID to 1 in `scene.reset()` if we want to avoid the ids in a new run to sort of depend on the ids in previous run.

Oh yeah, I forgot to do that.

src/battle-scene.ts
DayKev: Otherwise you need to have a blank line in-between two lines to create a linebreak in Markdown.

MokaStitcher: Ah I see, thanks 👍

PR #1319, opened 2025-08-09
misc: remove unused code from `BattleScene#initStarterColors`

DayKev:
N/A

The code has no current use. It can be re-added at a later time as a script if necessary.

Unused, commented out code removed.

## Checklist


MokaStitcher (approved):
wonder where we could keep a trace of this, because it will be needed when new starters get introduced

DayKev:
That's why I made a PR just for this, so it's easy to find again. We can just search "[unused](https://github.com/Despair-Games/poketernity/pulls?q=is%3Apr+is%3Aopen+unused)" and/or "[script](https://github.com/Despair-Games/poketernity/pulls?q=is%3Apr+is%3Aopen+script)" to find it again.

Issue #1320, opened 2025-08-11
Optimize save data


DayKev:
More is stored in save data than is necessary. Unnecessary data should be removed from saves.

I'm currently in the process of porting over various save optimization/refactoring PRs from PKR that were made by Benjie.

Stuff to optimize:
- [ ] Pokemon Species data #1307 
- [ ] Arena Tags #1322
- [ ] Battler Tags (after arena tags)
- [ ] ...

Any other specifics to consider should be added as comments on this issue.

PR #1322, opened 2025-08-12
refactor: optimize Arena Tags save data & Wish persists in save

DayKev:
- Wish now correctly persists across save/load.
- Arena tags now display in the arena flyout again.

- Part of #1320

### Text from PKR 6110
- Added missing implementation of `loadTag` to `WishTag`.
- Added many new types and type helpers for the purposes of type-safe parsing of arena data.
- Added a lengthy description at the top of `arena-tag.ts` that discusses changes
- Added a new abstract class, `SerializableArenaTag` to mark arena tags that should persist in save data.
- Simplified the body of `ArenaData`'s constructor to be more readable.
  - When creating an Arena, tags that are not Serializable are now ignored when constructing this data.

### Additional Changes
- Max layers for entry hazards now defined as a getter on the tag class.
- Fixed missing usages of `enumValueToKey` for `ArenaTagType`.

## How to test the changes?
Use various moves that create `ArenaTag`s (Wish, Reflect, hazard moves like Stealth Rocks, etc), then save and reload the save.

## Checklist


src/data/arena-tags/delayed-attack-tag.ts
```diff
@@ -41,12 +52,12 @@ export class DelayedAttackTag extends ArenaTag {
    * @param targetIndex - The {@linkcode FieldBattlerIndex} targeted by the attack
    */
   public addAttack(source: Pokemon, moveId: MoveId, targetIndex: FieldBattlerIndex): void {
-    this.delayedAttacks.push({ sourceId: source.id, moveId: moveId, targetIndex, turnCount: 3 });
+    (this.delayedAttacks as DelayedAttack[]).push({ sourceId: source.id, moveId: moveId, targetIndex, turnCount: 3 });
```
innerthunder:
```suggestion
    (this.delayedAttacks as DelayedAttack[]).push({ sourceId: source.id, moveId, targetIndex, turnCount: 3 });
```

src/data/arena-tags/entry-hazard-tag.ts
```diff
@@ -1,33 +1,31 @@
...
  * Abstract class to implement arena entry hazards.
  */
-export abstract class EntryHazardTag extends ArenaTag {
-  public layers: number;
-  public maxLayers: number;
+export abstract class EntryHazardTag extends SerializableArenaTag {
+  public abstract override readonly tagType: EntryHazardTagType;
+  public readonly layers: number = 1;
+
+  public abstract get maxLayers(): number;
```
innerthunder, line 20: This could use a doc like `WeakenMoveScreenTag` explaining why this is a getter and not just `public readonly`

src/data/arena-tags/toxic-spikes-tag.ts
```diff
@@ -16,11 +16,17 @@ import i18next from "i18next";
  * Pokémon summoned into this trap remove it entirely.
  */
 export class ToxicSpikesTag extends EntryHazardTag {
-  private neutralized: boolean;
+  public override readonly tagType = ArenaTagType.TOXIC_SPIKES;
 
-  constructor(sourceId: number, side: ArenaTagSide) {
-    super(ArenaTagType.TOXIC_SPIKES, MoveId.TOXIC_SPIKES, sourceId, side, 2);
-    this.neutralized = false;
+  #neutralized: boolean;
```
innerthunder, line 21: Why `#` instead of `private`?

src/data/arena-tags/wish-tag.ts
```diff
@@ -1,46 +1,69 @@
...
-export class WishTag extends ArenaTag {
-  private battlerIndex: FieldBattlerIndex;
-  private triggerMessage: string;
-  private healHp: number;
+export class WishTag extends SerializableArenaTag {
+  // The following fields are meant to be inwardly mutable, but outwardly immutable.
+  public readonly battlerIndex: BattlerIndex;
+  public readonly healHp: number;
+  public readonly sourceName: string;
+  // End inwardly mutable fields
```
innerthunder, line 22: Out of this PR's scope, but this doesn't support Wish being used in both slots on the same side in a double battle. It might be worth adding an `edgeCase` for now...

src/system/game-data.ts
```diff
@@ -891,7 +892,9 @@ export class GameData {
                   new TagAddedEvent(tagType, side, turnCount, layers, maxLayers),
                 );
               } else {
-                globalScene.arena.eventTarget.dispatchEvent(new TagAddedEvent(tag.tagType, tag.side, tag.turnCount));
+                globalScene.arena.eventTarget.dispatchEvent(
+                  new TagAddedEvent(tag.tagType, tag.side, tag.turnCount, 0, 0),
+                );
```
innerthunder: We can do this to condense the `if/else` here:

```ts
for (const tag of globalScene.arena.tags) {
  const { tagType, side, turnCount } = tag;
  const { layers, maxLayers } = tag instanceof EntryHazardTag ? tag : {};
  globalScene.arena.eventTarget.dispatchEvent(
    new TagAddedEvent(tagType, side, turnCount, layers ?? 0, maxLayers ?? 0);
  ); 
}

innerthunder (dismissed):

src/data/arena-tags/toxic-spikes-tag.ts
DayKev: This field doesn't need to be serialized because it can only ever be `true` right before the tag is removed.

A field that is `private` or `protected` will still be serialized, however TypeScript won't be able to see it and so it will cause type errors, which is why fields that need to be serialized but are normally supposed to be `private` need the `readonly` stuff so they can't be modified outside the class. Basically, working around TypeScript limitations.

src/data/arena-tags/wish-tag.ts
DayKev: Added. There's a PR to rework Wish and Future Sight at PKR that fixes the issue that should probably be ported at some point. https://github.com/pagefaultgames/pokerogue/pull/5862

innerthunder (dismissed):

innerthunder (dismissed):

innerthunder (approved):

PR #1325, opened 2025-08-16
dev: add simple Biome rules

DayKev:
N/A

To keep code quality high/consistent.

Multiple Biome rules (most of which were already being followed) were enabled:
- https://biomejs.dev/linter/rules/no-namespace/
- https://biomejs.dev/linter/rules/no-void/
- https://biomejs.dev/linter/rules/use-while/
- https://biomejs.dev/linter/rules/no-barrel-file/
- https://biomejs.dev/linter/rules/no-default-export/
- https://biomejs.dev/linter/rules/no-done-callback/
- https://biomejs.dev/linter/rules/no-shouty-constants/
- https://biomejs.dev/linter/rules/no-substr/
- https://biomejs.dev/linter/rules/no-yoda-expression/
- https://biomejs.dev/linter/rules/use-single-var-declarator/
- https://biomejs.dev/linter/rules/use-throw-new-error/
- https://biomejs.dev/linter/rules/use-throw-only-error/
- https://biomejs.dev/linter/rules/use-trim-start-end/
- https://biomejs.dev/linter/rules/use-error-message/
- https://biomejs.dev/linter/rules/use-guard-for-in/
- https://biomejs.dev/linter/rules/use-number-to-fixed-digits-argument/

## Checklist

### DIFF
--- a/test/ui/transfer-item.test.ts
+++ b/test/ui/transfer-item.test.ts
@@ -50,6 +50,7 @@ describe("UI - Transfer Items", () => {
       handler.setCursor(1);
       handler.processInput(Button.ACTION);
 
+      // biome-ignore lint/complexity/noVoid: Unable to use `await` here
       void game.scene.ui.setModeWithoutClear<PartyUiHandler>(UiMode.PARTY, PartyUiMode.MODIFIER_TRANSFER);
     });
 
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -117,7 +117,8 @@
         "noNamespace": "error",
         "noDefaultExport": "error",
         "noDoneCallback": "error",
-        "noShoutyConstants": "error"
+        "noShoutyConstants": "error",
+        "noSubstr": "error"
       },
       "suspicious": {
         "noDoubleEquals": "error",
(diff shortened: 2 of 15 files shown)


test/ui/transfer-item.test.ts
```diff
@@ -50,6 +50,7 @@ describe("UI - Transfer Items", () => {
       handler.setCursor(1);
       handler.processInput(Button.ACTION);
 
+      // biome-ignore lint/complexity/noVoid: Unable to use `await` here
       void game.scene.ui.setModeWithoutClear<PartyUiHandler>(UiMode.PARTY, PartyUiMode.MODIFIER_TRANSFER);
```
MokaStitcher, line 54: What is this sorcery

MokaStitcher (approved):
lgtm

test/ui/transfer-item.test.ts
DayKev: We have an ESLint rule https://typescript-eslint.io/rules/no-floating-promises/ that complains if a Promise is not `await`ed (or otherwise "handled"), and putting `void` before a Promise says "yes this is intentionally not `await`ed don't complain about it".

Looking at these tests, I have no idea if they actually function or not... should probably confirm that some time lol.

MokaStitcher: Ah I see lol
<sub>almost forgot that party UI needed a big refactor too..

flx-sta (approved):

PR #1328, opened 2025-08-17
refactor: convert `CustomPokemonData` to an `interface`

DayKev:
N/A

There's no need for this to be a class.

`CustomPokemonData` converted from a class to an interface and `Pokemon#resetCustomPokemonData()` created to replace `new CustomPokemonData(...)` where necessary.

Also removed unnecessary `null` guards for `Pokemon#customPokemonData`.

## How to test the changes?
Do things that modify `Pokemon#customPokemonData` such as the clown ME or using a nature mint.

## Checklist


innerthunder (dismissed):

innerthunder (dismissed):

src/data/mystery-encounters/encounters/clowning-around-encounter.ts
```diff
@@ -151,10 +150,10 @@ export const ClowningAroundEncounter: MysteryEncounter = MysteryEncounterBuilder
         {
           // Blacephalon has the random ability from pool, and 2 entirely random types to fit with the theme of the encounter
```
MokaStitcher, line 151:
```suggestion
          // Blacephalon has the random ability from pool, and 2 entirely random types to fit with the theme of the encounter
          // TODO: should prevent it from rolling the same type twice
```

src/field/pokemon.ts
```diff
@@ -4378,6 +4381,11 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     };
   }
 
+  public resetCustomPokemonData(data: Partial<CustomPokemonData> = {}): void {
+    const { spriteScale = -1, ability = -1, passive = -1, nature = -1, types = [] } = data;
+    this.customPokemonData = { spriteScale, ability, passive, nature, types };
```
MokaStitcher: hm, guess this means the save files will be slightly larger overall since all customPokemonData attributes are now defined for all mons. Not sure how significant
edit: tested it with a fresh daily run and there was no difference so all good i suppose 🤷

MokaStitcher: should the types array be a copy of the source one rather than a reference to it?

src/system/pokemon-data.ts
```diff
@@ -121,7 +120,7 @@ export class PokemonData {
     this.stellarTypesBoosted = source.stellarTypesBoosted ?? [];
     this.status = source["status"];
 
-    this.customPokemonData = new CustomPokemonData(source.customPokemonData);
+    this.customPokemonData = source.customPokemonData;
```
MokaStitcher, line 123: I think this won't cause an issue because PokemonData should only be used temporarily for serialization, but i'm always a bit worried of doing things like this rather than a deep copy

MokaStitcher (dismissed):

src/field/pokemon.ts
DayKev: Probably a good idea. Changed.

src/system/pokemon-data.ts
DayKev: Possibly a future refactor to do of `PokemonData`/etc then, since this isn't the only time this is done here.

PR #1331, opened 2025-08-20
ai-rework: Rewrite Ally Target scoring logic

innerthunder:

The AI should no longer be incentivized to use attacks with secondary effects (e.g. Flamethrower) on their allies)

The `overridesAllyTargetPenalty` option for move attributes has proven to be a bit too cumbersome, with it requiring the option to be enabled and a private helper method to be defined in most cases. These changes should make it easier to distinguish ES calculations against opponents from those against allies

- All attacks (other than Pollen Puff) are now automatically assigned the Ally Target Penalty when targeting an ally.
- `MoveAttr` now has a public `getAllyTargetScore` function, which is used to compute Effect Score when an enemy Pokemon targets its ally with a move with the attribute (aka ATS). This can either return an integer score or `null`. If `null` is returned, the attribute does not contribute to ATS in the given game state, deferring to other attributes with a defined ATS. If none of the move's attributes have a defined ATS, then the move is given an Ally Target Penalty.
- The `overridesAllyTargetPenalty` option is removed. Attributes that previously used the option have been adjusted to override `getAllyTargetScore` instead.

## How to test the changes?

- [x] `pnpm typecheck`
- [x] `pnpm depcruise`
- [x] `pnpm test:silent`

## Checklist

### DIFF
--- a/src/data/moves/move-attrs/move-attr.ts
+++ b/src/data/moves/move-attrs/move-attr.ts
@@ -1,3 +1,7 @@
+/* biome-ignore-start lint/correctness/noUnusedImports: tsdoc imports */
+import type { ALLY_TARGET_PENALTY } from "#constants/ai-constants";
+/* biome-ignore-end lint/correctness/noUnusedImports: tsdoc imports */
+
 import type { EnemyPokemon } from "#field/enemy-pokemon";
 import type { Pokemon } from "#field/pokemon";
 import type { Move } from "#moves/move";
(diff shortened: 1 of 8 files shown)


DayKev (approved):

DayKev (changes requested):
Oh wait, you missed `HealStatusEffectAttr`

src/data/moves/move-attrs/move-attr.ts
```diff
@@ -117,6 +109,21 @@ export abstract class MoveAttr {
     return 0;
   }
 
+  /**
+   * Defines the integer Effect Score bonus (or penalty) granted by this attribute based on the
+   * current game state when targeting an ally.
+   * @param user - The {@linkcode EnemyPokemon} evaluating the move
+   * @param target - The {@linkcode Pokemon} the move is evaluated against. This can be assumed to be
+   * the {@linkcode _user | user}'s ally.
```
DayKev:
```suggestion
   * @param target - The {@linkcode Pokemon} the move is evaluated against. This can be assumed to be the user's ally.
```

DayKev (approved):

PR #1336, opened 2025-08-23
refactor: Pokemon gender no longer uses the old Gen3-5 method

### LINKED ISSUE #1335: Refactor random gender for enemy pokemon to no longer use ID
Currently it expects to use the ID as a PV so it can do the old Gen3-5 method of determining gender.

https://github.com/Despair-Games/poketernity/blob/8131f17ae2d1f82d082957fb4530e52ee08585d0/src/field/pokemon.ts#L1355-L1366

DayKev:
Wild Pokémon gender will be rolled correctly again.

Fixing an oversight from the Pokémon ID refactor.
- Closes #1335 

Instead of doing `Pokémon ID mod 256 * 0.39...` to generate a random number, a simple random number from `0-100` is generated and compared to the species' "male percent" value.

## Checklist


innerthunder (dismissed):

src/field/pokemon.ts
```diff
@@ -1322,48 +1322,47 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+  public getHpRatio(): number {
     return this.hp / this.getMaxHp();
   }
 
-  generateGender(): void {
+  /** Sets the pokemon's gender based on its species gender ratios */
+  protected generateGender(): void {
     if (this.species.malePercent === null) {
       this.gender = Gender.GENDERLESS;
+    } else if (Phaser.Math.RND.frac() * 100 <= this.species.malePercent) {
```
MokaStitcher: should we be using `this.randSeedInt` like for IVs, to make sure it's using the proper RNG?

MokaStitcher: also not 100% sure the `<=` is correct. Is it technically possible that `frac()` retruns 0? Because if yes, it would then set gender to male even if malePercent is 0

MokaStitcher: <sub> also could this have used the PV instead of ID if we wanted to preserve the previous method?

DayKev: > should we be using `this.randSeedInt` like for IVs, to make sure it's using the proper RNG?

The `malePercent` value is often fractional though, so I think it would be inaccurate? Tbh the RNG needs to be refactored in many ways (such as so that `randSeedInt` isn't the only function that can use the battle seed instead of just the global seed).

> also not 100% sure the `<=` is correct. Is it technically possible that `frac()` retruns 0? Because if yes, it would then set gender to male even if malePercent is 0

Oops, will fix.

> also could this have used the PV instead of ID if we wanted to preserve the previous method?

There's no reason to use the old method, we're not developing for Gameboy [Advance] and don't need to use the tricks that the old games did. The mainline Pokémon games haven't used that method since Gen 5 either.

innerthunder (approved):

Tempo-anon (approved):

PR #1337, opened 2025-08-24
ai-rework: Add scoring for `StatStageChangeAttr` + subclasses

innerthunder:

The AI should now properly evaluate moves that increase or decrease one or more Pokemon's stat stages.

Part of the AI Rework.

- Added scoring logic for `StatStageChangeAttr` and subclass `CutHpStatStageBoostAttr`.
- Added some helper methods for scoring to `ChanceBasedMoveEffectAttr` so that ally target scores can apply chance-based tiering on a per-attribute basis

## How to test the changes?

- [x] `pnpm typecheck`
- [x] `pnpm depcruise`
- [x] `pnpm test:silent`

- `swords-dance`
- `iron-defense`
- `agility`
- `charm`
- `screech`
- `scary-face`
- `strength-sap` (1 newly enabled test)
- `belly-drum`

## Checklist

### DIFF
--- a/src/data/moves/move-attrs/chance-based-move-effect-attr.ts
+++ b/src/data/moves/move-attrs/chance-based-move-effect-attr.ts
@@ -81,16 +81,21 @@ export abstract class ChanceBasedMoveEffectAttr extends MoveEffectAttr {
    * and {@linkcode getMoveChance | chance to apply}.
    */
   public override getEffectScore(user: EnemyPokemon, target: Pokemon, move: Move): number {
+    /** The attribute's effect score, assuming its effect always applies */
+    const rawScore = this.getRawEffectScore(user, target, move);
+
+    return this.getTieredScore(user, target, move, rawScore);
+  }
+
+  protected getTieredScore(user: EnemyPokemon, target: Pokemon, move: Move, score: number): number {
     /**
      * The attribute's chance to apply its effect
      * @todo this chance calculation may prematurely reveal abilities
      */
     const chance = this.getMoveChance(user, target, move);
-    /** The attribute's effect score, assuming its effect always applies */
-    const rawScore = this.getRawEffectScore(user, target, move);
 
     if (chance < 0) {
-      return rawScore;
+      return score;
     }
 
     /**
@@ -98,7 +103,7 @@ export abstract class ChanceBasedMoveEffectAttr extends MoveEffectAttr {
      * This may be a decimal number; the final output is either
      * `floor(chanceWeightedScore)` or `floor(chanceWeightedScore) + 1`
      */
-    const chanceWeightedScore = (chance * rawScore) / 100;
+    const chanceWeightedScore = (chance * score) / 100;
     /** The minimum integer score this function can return */
     const minScore = Math.floor(chanceWeightedScore);
     /**
--- a/src/data/moves/move-attrs/chance-based-move-effect-attr.ts
+++ b/src/data/moves/move-attrs/chance-based-move-effect-attr.ts
@@ -87,6 +87,22 @@ export abstract class ChanceBasedMoveEffectAttr extends MoveEffectAttr {
     return this.getTieredScore(user, target, move, rawScore);
   }
 
+  /**
+   * Translates the given "raw" decimal Effect Score into a final integer Effect Score, accounting for the
+   * effect's chance to apply. This is done by multiplying the raw score by the effect chance, then
+   * randomly "tiering" the product into either `floor(score)` or `floor(score) + 1`.
+   *
+   * **Example:**
+   *
+   * If an effect's raw score is `1.5`, and the chance to apply said effect is 50 percent, the resulting
+   * chance-adjusted score is `0.75`. In turn, the final score has a 25% chance of being (+0) and a 75%
+   * chance of being (+1)
+   * @param user - The {@linkcode EnemyPokemon} evaluating the move
+   * @param target - The {@linkcode Pokemon} the move is evaluated against
+   * @param move - The {@linkcode Move} being evaluated
+   * @param score - The previously calculated "raw" Effect Score for the move action
+   * @returns The final integer score after tiering
+   */
   protected getTieredScore(user: EnemyPokemon, target: Pokemon, move: Move, score: number): number {
     /**
      * The attribute's chance to apply its effect
--- a/src/data/moves/move-attrs/cut-hp-stat-stage-boost-attr.ts
+++ b/src/data/moves/move-attrs/cut-hp-stat-stage-boost-attr.ts
@@ -1,4 +1,4 @@
-import { ATTACK_SCORE_HP_THRESHOLD, MAJOR_EFFECT_SCORE_PENALTY } from "#constants/ai-constants";
+import { ATTACK_SCORE_HP_THRESHOLD, BAD_MOVE_PENALTY } from "#constants/ai-constants";
 import { HitResult } from "#enums/hit-result";
 import type { BattleStat } from "#enums/stat";
 import type { EnemyPokemon } from "#field/enemy-pokemon";
@@ -48,6 +48,14 @@ export class CutHpStatStageBoostAttr extends StatStageChangeAttr {
       user.getHpRatio() > 1 / this.cutRatio && this.stats.some((s) => user.getStatStage(s) < 6);
   }
 
+  /**
+   * @returns An additional penalty for reducing the user's HP to go with the score
+   * for this effect's stat stage changes. This checks each of the user's opponents'
+   * max {@link Pokemon.getExpectedAttackScore | EAS} among their estimated attacks
+   * to predict how much damage the user will receive this turn. If the user is
+   * projected to receive more damage than the effect's {@linkcode cutRatio}, this
+   * grants a {@linkcode BAD_MOVE_PENALTY}.
+   */
   public override getRawEffectScore(user: EnemyPokemon, target: Pokemon, move: Move): number {
     const oppMaxEas = user.getOpponents().map((opp) =>
       opp.estimateAttackMoves().reduce((maxEas, move) => {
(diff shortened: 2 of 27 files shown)


src/data/moves/move-attrs/chance-based-move-effect-attr.ts
```diff
@@ -81,24 +81,29 @@ export abstract class ChanceBasedMoveEffectAttr extends MoveEffectAttr {
    * and {@linkcode getMoveChance | chance to apply}.
    */
   public override getEffectScore(user: EnemyPokemon, target: Pokemon, move: Move): number {
+    /** The attribute's effect score, assuming its effect always applies */
+    const rawScore = this.getRawEffectScore(user, target, move);
+
+    return this.getTieredScore(user, target, move, rawScore);
+  }
+
+  protected getTieredScore(user: EnemyPokemon, target: Pokemon, move: Move, score: number): number {
```
DayKev, line 106: Should have docs.

src/data/moves/move-attrs/cut-hp-stat-stage-boost-attr.ts
```diff
@@ -41,8 +43,33 @@ export class CutHpStatStageBoostAttr extends StatStageChangeAttr {
...
+   * for this effect's stat stage changes. This checks each of the user's opponents'
+   * max {@link Pokemon.getExpectedAttackScore | EAS} among their estimated attacks
+   * to predict how much damage the user will receive this turn. If the user is
+   * projected to receive more damage than the effect's {@linkcode cutRatio}, this
+   * grants a {@linkcode BAD_MOVE_PENALTY}.
+   */
+  public override getRawEffectScore(user: EnemyPokemon, target: Pokemon, move: Move): number {
+    const oppMaxEas = user.getOpponents().map((opp) =>
+      opp.estimateAttackMoves().reduce((maxEas, move) => {
+        const eas = opp.getExpectedAttackScore(user, move);
```
DayKev: https://biomejs.dev/linter/rules/no-shadow/
```suggestion
      opp.estimateAttackMoves().reduce((maxEas, mv) => {
        const eas = opp.getExpectedAttackScore(user, mv);
```

src/data/moves/move-attrs/stat-stage-change-attr.ts
```diff
@@ -76,49 +99,275 @@ export class StatStageChangeAttr extends ChanceBasedMoveEffectAttr {
...
+
+  /**
+   * Calculates the Effect Score gained (before factoring in effect chance) from modifying a single
+   * stat stage with this effect on either the user or its ally. The heuristic for this scoring varies based
+   * on the stat changed:
+   * - `ATK` / `SPATK`: If the target has a move of matching {@linkcode MoveCategory}, grant (+0.5) per stat stage.
+   * - `DEF` / `SPDEF`: Check the offensive stat affinities of each opponent (i.e. which is higher between
+   * `ATK` and `SPATK`). This grants (+0.5) per stat stage, per opponent with a matching affinity. For `DEF`
+   * boosts, if the target has Body Press, this grants an additional (+1).
+   * - `SPD`: Grants (+1) per opponent that the target would outspeed as a result of this effect.
```
DayKev: Does this only grant the `+1` if the user['s ally] doesn't already outspeed both of its opponents?

src/data/moves/move-attrs/stat-stage-change-attr.ts
```diff
@@ -76,49 +99,275 @@ export class StatStageChangeAttr extends ChanceBasedMoveEffectAttr {
...
+   * - `EVA`: Grants (+0.5) per stat stage.
+   * @param user - The {@linkcode Pokemon} evaluating this effect
+   * @param target - The {@linkcode Pokemon} this effect is evaluated against. This can be assumed to be either the
+   * user or its ally.
+   * @param move - The {@linkcode Move} whose effect is evaluated.
+   * @param stat - The {@linkcode BattleStat} being evaluated
+   * @returns The "raw" Effect Score bonus (or penalty) for the given stat. This score is combined with other
+   * stats and {@link getTieredScore | tiered} to yield the final Effect Score.
+   */
+  private getAllyTargetScoreByStat(user: EnemyPokemon, target: Pokemon, move: Move, stat: BattleStat): number {
```
DayKev: Is it possible to do `target: EnemyPokemon`/`target: PlayerPokemon` when the target should always be on a specific side of the field?

test/ai/move-effect-scores/screech.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    game.override
+      .battleType("single")
+      .enemyAbility(AbilityId.BALL_FETCH)
+      .ability(AbilityId.BALL_FETCH)
+      .enemyMoveset([MoveId.SCREECH, MoveId.SPLASH, MoveId.TACKLE])
+      .startingLevel(100)
+      .enemyLevel(100);
+  });
+
+  it("should be preferred when the user is a physical attacker", async () => {
```
DayKev, line 33:
```suggestion
  it("should be preferred when the target is a physical attacker", async () => {
```

test/ai/move-effect-scores/screech.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    game.override.enemySpecies(SpeciesId.DRILBUR);
+
+    await game.classicMode.startBattle(SpeciesId.MAGIKARP);
+
+    const enemy = game.field.getEnemyPokemon();
+    expect(enemy).toPreferSelectingMove(MoveId.SCREECH);
+  });
+
+  it("should not be preferred when the user is a special attacker", async () => {
```
DayKev, line 42:
```suggestion
  it("should not be preferred when the target is a special attacker", async () => {
```

DayKev (commented):
I wonder if maybe the further the effect target is away from 0 in the same direction as the expected stat stage change (such as going from +2 Def to +4 Def with Iron Defense), the effect should be deprioritized slightly; and the opposite for when the stat stage level change is against the current stat stages (such as going from -2 Def to +0 Def with Iron Defense).

Tempo-anon:
> I wonder if maybe the further the effect target is away from 0 in the same direction as the expected stat stage change (such as going from +2 Def to +4 Def with Iron Defense), the effect should be deprioritized slightly; and the opposite for when the stat stage level change is against the current stat stages (such as going from -2 Def to +0 Def with Iron Defense).

Oh yeah there's definitely diminishing returns from extra stats after basically +2. Seems like it may be complex to fully account for all the cases but some general rules of thumb:
* Speed gains matter the most from -1 to 0 to 1. From 1->2 it still might matter a little bit but it's basically useless after +2 (except for very niche cases of slow speed boosters like shell smash Crustle)
* Boosting defense is generally not worth it unless the user has a recovery move or a move that scales with increased defense like body press or stored power
* Extra attacking stats are nice but unless the opponent has a lot of defense stat stages then there's really not much of a point in going over +2
* With regards to adding more priority when the user is at stages under 0, I think this case should actually be deprioritized since in most cases you don't want to be spending turns and moves boosting your lowered stats. At that point it's better to just switch out and get all your stats reset

innerthunder:
> I wonder if maybe the further the effect target is away from 0 in the same direction as the expected stat stage change (such as going from +2 Def to +4 Def with Iron Defense), the effect should be deprioritized slightly; and the opposite for when the stat stage level change is against the current stat stages (such as going from -2 Def to +0 Def with Iron Defense).

In some ways (and for some stat types) this is already indirectly enforced:
- Attack/Sp. Atk: Ideally the enemy would reach a stage where their attacks' scores surpass the score of the stat boosting move. E.g. if the AI has just used Swords Dance and now expects that one of their attacks will KO an opponent, it will always attack on the next turn.
- Defense/Sp. Def: I don't think there's much in the way of enforcing diminishing returns here, but everything that applies to Attack/Sp Atk also applies to Defense with Body Press.
- Speed: The AI should only focus on speed boosts/reduction if using the move once will immediately improve its turn order compared to opponents (as far as it can tell, anyway). Diminishing returns don't need to be accounted for here.
- Accuracy: Nothing enforced here, but using `calculateBattleAccuracy` instead of base move accuracy would definitely help.
- Evasion: I'm not sure that evasion boosts have the same diminishing returns as other stat types. We might want to set a hard limit on evasion boosts (and accuracy reduction) just to improve the general gameplay experience, though 😅 

<sub>This is all in reference to ally-targeted scoring</sub>

EDIT: I added some thresholds to reduce incentive for certain stat types when the target already has high (or low) stat stages:
- Defense / Sp. Def (ally): When the user or targeted ally already has 2 or more stages of the relevant stat, the effect score for additional boosts is halved. This does not effect the bonus from the ally having Body Press.
- Evasion (ally): No longer grants incentive when the user or targeted ally is already at 2 or more stages.
- Accuracy (opponent): No longer grants incentive when the targeted opponent is already at -2 or fewer stages.

The change to defensive stat scoring is meant to dissuade the AI from spamming defense-boosting moves aimlessly with diminishing returns. The changes to evasion and accuracy should prevent the AI from trying to cheese the player too much by fishing for misses. When AI Types are re-implemented, it might be worth removing these limits for certain types 🙃

src/data/moves/move-attrs/stat-stage-change-attr.ts
innerthunder: Yes. This is enforced in one of the Agility tests ("should not be preferred if using the move would not improve the user's turn order").

test/ai/move-effect-scores/screech.test.ts
innerthunder: `user` is correct here. It's checking if the user would be able to take advantage of lowering the target's Defense.

DayKev: Oh oops, for some reason I thought Screech lowered attack lmao.

DayKev (approved):

DayKev (approved):

PR #1340, opened 2025-08-28
ai-rework: Add Effect Scores for misc. stat-modifying attributes

innerthunder:

The AI should now properly evaluate the following moves:
- Spectral Thief
- Psych Up
- Order Up
- Acupressure
- Fell Stinger
- Topsy-Turvy
- Haze, Freezy Frost, and Clear Smog
- Guard Swap, Power Swap, and Heart Swap
- Speed Swap
- Power Split and Guard Split

Part of the AI Rework

Added Effect Score overrides for the following move attributes:
- `StealPositiveStatsAttr`
- `CopyStatsAttr`
- `OrderUpStatBoostAttr`
- `AcupressureStatStageChangeAttr`
- `PostVictoryStatStageChangeAttr`
- `InvertStatsAttr`
- `ResetStatsAttr`
- `SwapStatStagesAttr`
- `SwapStatAttr`
- `AverageStatsAttr`

## How to test the changes?

- [x] `pnpm typecheck`
- [x] `pnpm depcruise`
- [x] `pnpm test:silent`

New unit tests (can be run with `pnpm test move-effect-scores/testName`):
- `spectral-thief`
- `psych-up`
- `acupressure`
- `fell-stinger`
- `topsy-turvy`
- `haze`
- `clear-smog`
- `swap-stat-stages`
- `speed-swap`
- `average-stats`

## Checklist

### DIFF
--- a/src/data/moves/move-attrs/copy-stats-attr.ts
+++ b/src/data/moves/move-attrs/copy-stats-attr.ts
@@ -1,7 +1,9 @@
 import { globalScene } from "#app/global-scene";
 import { getPokemonNameWithAffix } from "#app/messages";
+import { BAD_MOVE_PENALTY, SOFT_EFFECT_SCORE_LIMIT } from "#constants/ai-constants";
 import { BattlerTagType } from "#enums/battler-tag-type";
-import { BATTLE_STATS } from "#enums/stat";
+import { BATTLE_STATS, type BattleStat } from "#enums/stat";
+import type { EnemyPokemon } from "#field/enemy-pokemon";
 import type { Pokemon } from "#field/pokemon";
 import type { Move } from "#moves/move";
 import { MoveEffectAttr } from "#moves/move-effect-attr";
@@ -34,5 +36,38 @@ export class CopyStatsAttr extends MoveEffectAttr {
     );
 
     return true;
+  }
+
+  /**
+   * @returns (+0.5) for each stat stage the user would gain from this effect against the given
+   * target. If the projected gain is 0 or less, this grants a {@linkcode BAD_MOVE_PENALTY} instead.
+   */
+  public override getEffectScore(user: EnemyPokemon, target: Pokemon, _move: Move): number {
+    const totalStatGain = BATTLE_STATS.reduce(
+      (total, stat) => total + this.getProjectedStatGain(user, target, stat),
+      0,
+    );
+
+    if (totalStatGain <= 0) {
+      return BAD_MOVE_PENALTY;
+    }
+
+    return Math.min(Math.floor(totalStatGain * 0.5), SOFT_EFFECT_SCORE_LIMIT);
+  }
+
+  /** This effect treats allies in the same way as opponents */
+  public override getAllyTargetScore(user: EnemyPokemon, target: EnemyPokemon, move: Move): number {
+    return this.getEffectScore(user, target, move);
+  }
+
+  /**
+   * @param user - The {@linkcode EnemyPokemon} evaluating this effect
+   * @param target - The {@linkcode Pokemon} this effect is evaluated against
+   * @param stat - The {@linkcode BattleStat} for which the stat stage gain is calculated
+   * @returns The number of stat stages the given user gains in the given stat when this
+   * attribute's effect is applied to the target
+   */
+  private getProjectedStatGain(user: EnemyPokemon, target: Pokemon, stat: BattleStat) {
+    return target.getStatStage(stat) - user.getStatStage(stat);
   }
 }
--- a/src/data/moves/move-attrs/copy-stats-attr.ts
+++ b/src/data/moves/move-attrs/copy-stats-attr.ts
@@ -39,7 +39,7 @@ export class CopyStatsAttr extends MoveEffectAttr {
   }
 
   /**
-   * @returns (+0.5) for each stat stage the user would gain from this effect against the given
+   * @returns (+1) for every `2` stat stages the user would gain from this effect against the given
    * target. If the projected gain is 0 or less, this grants a {@linkcode BAD_MOVE_PENALTY} instead.
    */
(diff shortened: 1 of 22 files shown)


src/data/moves/move-attrs/copy-stats-attr.ts
```diff
@@ -35,4 +37,37 @@ export class CopyStatsAttr extends MoveEffectAttr {
 
     return true;
   }
+
+  /**
+   * @returns (+0.5) for each stat stage the user would gain from this effect against the given
```
DayKev:
```suggestion
   * @returns (+1) for each `2` stat stages the user would gain from this effect against the given
```
Since you are `floor`ing the result, it won't actually return fractional values.

src/data/moves/move-attrs/invert-stats-attr.ts
```diff
@@ -26,4 +28,29 @@ export class InvertStatsAttr extends MoveEffectAttr {
...
+    return Math.min(this.getProjectedStatChange(target) * MINOR_EFFECT_SCORE_BONUS, SOFT_EFFECT_SCORE_LIMIT);
+  }
+
+  /**
+   * @param target - The {@linkcode Pokemon} on which this effect is evaluated
+   * @returns The effective change (in number of stat stages) this attribute's effect is expected
+   * to apply to the given target
+   */
+  private getProjectedStatChange(target: Pokemon): number {
+    return BATTLE_STATS.reduce((total, stat) => total + target.getStatStage(stat) * -2, 0);
```
DayKev, line 54: Why is this multiplied by `-2`?

src/data/moves/move-attrs/shift-stat-attr.ts
DayKev, line 1: There doesn't seem to be a replacement implementation for this attr?

src/data/moves/move-attrs/swap-stat-attr.ts
```diff
@@ -35,4 +39,35 @@ export class SwapStatAttr extends MoveEffectAttr {
 
     return true;
   }
+
+  /**
+   * @returns A {@link MINOR_EFFECT_SCORE_BONUS | minor bonus} for each opponent the user
+   * would outspeed as a result of applying this attribute's effect on the target. All
+   * abilities (including revealed abilities) are ignored for the stat calculations in this scoring.
```
DayKev: Why are abilities ignored?

src/data/moves/move-attrs/swap-stat-stages-attr.ts
DayKev, line 1: Should it not consider negative stat stages that are being passed to the target?

src/data/moves/move.ts
DayKev, line 1: > Attacks (other than Pollen Puff) are always penalized against allies

Should this take into consideration things like Water Gun into an ally with Water Absorb?

src/data/moves/move.ts
```diff
@@ -861,7 +863,9 @@ export abstract class Move {
 
     let attrs: MoveAttr[] = this.attrs;
     if (isKnockOut) {
-      attrs = attrs.filter((attr) => attr.selfTarget);
+      attrs = attrs.filter(
```
DayKev, line 871: Should probably add a comment explaining what this filter is doing.
```suggestion
      // <comment goes here>
      attrs = attrs.filter(
```

src/data/moves/move-attrs/invert-stats-attr.ts
innerthunder: Example: If the target's at +2 Attack, Topsy-Turvy would set its Attack to -2 for a total stat change of -4

src/data/moves/move-attrs/shift-stat-attr.ts
innerthunder: The score sheet lists this attribute as unscored (i.e. Power Shift will always be scored at 0). I think I gave up on trying to score it initially for how niche its utility is 😅 . Maybe this should gain incentive if the user has no physical attacks and its Attack is higher than its Defense (by a certain amount)?

src/data/moves/move-attrs/invert-stats-attr.ts
DayKev: Ah, gotcha.

src/data/moves/move-attrs/swap-stat-stages-attr.ts
innerthunder: I could just increase the score multiplier (from 0.5) I guess, since the user's stat gain is the same as the target's stat loss...

src/data/moves/move.ts
innerthunder: Perhaps. There are still a lot of ability interactions the AI isn't aware of...

src/data/moves/move-attrs/swap-stat-attr.ts
innerthunder: I rewrote this just now, but it's hard to separate different elements of stat calculation (raw stats, stages, abilities, and other effects) within the `getEffectiveStat` method as is. Only the raw stats transfer to the user, so I originally had this ignore abilities so that they don't interfere with stat projections.

test/ai/move-effect-scores/topsy-turvy.test.ts
```diff
@@ -0,0 +1,74 @@
...
+  });
+
+  it("should not be preferred when the target has no stat changes", async () => {
+    await game.classicMode.startBattle(SpeciesId.MAGIKARP);
+
+    const enemy = game.field.getEnemyPokemon();
+    expect(enemy).not.toPreferSelectingMove(MoveId.TOPSY_TURVY);
+  });
+
+  it("should be preferred when the target has an increased stat stage", async () => {
```
DayKev, line 43: There should probably be inverse tests for this and the ally one at the bottom as well.

DayKev (approved):

PR #1343, opened 2025-08-30
dev: remove unused packages/functions/etc

DayKev:
N/A

Removing dead code and other unused stuff.

- Removed unused package `json-stable-stringify`.
- Added `depcruiser` rule to prevent importing anything from `test/` into `src/`.
- Deleted `post-turn-heal-ab-attr.ts` (this was never used lol) and `daily-run-scoreboard.ts` (we're not going to add scoreboard functionality back).
- Gave Hydrapple the correct variable for its evo level.
- Deleted various unused functions/variables/etc.
- Removed `export` from various functions/variables/etc that were never imported.
- Moved `printContainerList()` to a new file `debug-utils.ts` and added a `depcruiser` rule to prevent importing from that file in production code.

## Checklist


innerthunder (approved):

src/data/abilities/ab-attrs/post-turn-heal-ab-attr.ts
Tempo-anon, line 1: I assume this was created for something like poison heal or dry skin but they use other attr's. The [postTurnHeal](https://github.com/search?q=repo%3ADespair-Games%2Fpoketernity-locales%20postTurnHeal&type=code) message is also being unused and can be removed

src/utils/color-utils.ts
Tempo-anon, line 1: weird to think these had to be implemented in the codebase instead of just using a dependency

Tempo-anon (approved):

Issue #1276, opened 2025-06-28
[Bug] Wimp Out/Emergency Exit does not Trigger if a Self-Damage Move is Used Preceding an Enemy's Attack, Triggers on Clangorous Soul and Fillet Away


Snailman11:
### Describe the bug

Currently, wimp out "prevents" proccing from self inflicted HP cuts by returning false if Belly Drum, Substitute, curse or Pain Split was the last move used by the ability holder.

This has a few problems:

Fillet Away and Clangorous Soul are not included, despite being self-HP cutting moves.

The use of getLastXMoves effectively assumes the prior HP cutting move is the thing triggering Wimp Out, which may not be the case! If an HP cut move like Substitute immediately precedes an instance of damage that DOES drop the user below half, wimp out will see the prior substitute and chicken out.

This also causes non-ghost type Curses to block Wimp Out from triggering in the same manner.

- It seems that Wimp Out switching also does not occur despite Substitute not being successfully executed

https://github.com/Despair-Games/poketernity/blob/1850410098bd144c704eb32fa0a002a9d9b93f41/src/data/abilities/ab-attrs/post-damage-force-switch-ab-attr.ts#L52
```
  const forbiddenAttackingMoves = [MoveId.BELLY_DRUM, MoveId.SUBSTITUTE, MoveId.CURSE, MoveId.PAIN_SPLIT];
    if (moveHistory.length > 0) {
      const lastMoveUsed = moveHistory[moveHistory.length - 1];
      if (forbiddenAttackingMoves.includes(lastMoveUsed.move.id)) {
```
Clangorous Soul and Fillet Away might have been oversights.

### Reproduction

Choose Wimpod **and Togepi/other mon**
```
	MOVESET_OVERRIDE: [MoveId.BELLY_DRUM, MoveId.SUBSTITUTE, MoveId.ENDURE, MoveId.SPLASH],

	ENEMY_LEVEL_OVERRIDE: 3,

	ENEMY_MOVESET_OVERRIDE: [MoveId.TACKLE],
```

```
	MOVESET_OVERRIDE: [MoveId.CLANGOROUS_SOUL, MoveId.FILLET_AWAY, MoveId.ENDURE, MoveId.SPLASH],

	ENEMY_LEVEL_OVERRIDE: 3,
	ENEMY_SPECIES_OVERRIDE: SpeciesId.REGIELEKI,
	ENEMY_MOVESET_OVERRIDE: [MoveId.TACKLE],

```

[...]

PR #1348, opened 2025-09-05
refactor: Rewrite force-switch and fleeing logic

innerthunder:

- Some inconsistencies in forced switches between switching moves and Wimp Out/Emergency Exit have been fixed:
  - Dondozo with a Tatsugiri in its mouth via Commander can no longer be forced to switch out or flee under *any* circumstance. If a "commanded" Pokemon with Wimp Out is damaged below half HP, Wimp Out will not activate.
  - ALL Boss Pokemon now cannot be forced to flee (not just Wave X0 bosses). Boss Pokemon can still be forced to switch out if they have a valid and inactive ally.
  - Mystery Encounters that disable fleeing no longer erroneously disable forced switching on enemies.
- Wimp Out and Emergency Exit no longer activate from the source using Clangorous Soul or Fillet Away

Switching moves and Wimp Out et al. used a lot of duplicate code that wasn't properly maintained to be consistent in their effects. These changes centralize the logic for forcing switches and fleeing so that moves, abilities, and future items that apply these effects are easier to maintain.

This is also a partial fix to #1276. However, self-damaging moves still erroneously prevent Wimp Out et al. from activating in some cases.

- Added four new public `BattleScene` methods related to forced switching and fleeing:
  - `canForceSwitchPokemon`
  - `tryForceSwitchPokemon`
  - `canForceFleePokemon`
  - `tryForceFleePokemon`
- Rewrote `ForceSwitchAttr` and `PostDamageForceSwitchAbAttr` to take advantage of the new `BattleScene` methods

## How to test the changes?

`pnpm test[:silent] wimp-out`
`pnpm test[:silent] whirlwind`
`pnpm test[:silent] u-turn` (1 new test)

There are a handful of other existing tests related to forced switching and fleeing, so running `pnpm test:silent` is fairly comprehensive.

## Checklist


src/data/moves/move-attrs/force-switch-out-attr.ts
```diff
@@ -22,239 +19,87 @@ import i18next from "i18next";
...
-          globalScene.redirectPokemonMoves(switchOutTarget, allyPokemon);
-        }
+  /**
+   * Status moves with this effect fail if the effect's target cannot be forced to
+   * switch out or flee, depending on the current {@linkcode BattleType}
+   */
+  public override getCondition(): MoveConditionFunc {
+    return (user, target, move) => {
+      if (move.isAttackMove(user, target)) {
+        return true;
```
DayKev, line 65: According to Bulbapedia, Dragon Tail should fail if it's a wild double battle and there's multiple pokemon on the field, or the user is lower level than the target.

> Dragon Tail inflicts damage and has decreased priority. In a Trainer battle, it will switch out the target to the next Pokémon in line that has not fainted (if there is one). In a wild Pokémon battle with a single wild Pokémon, this move will cause the battle to end if the user's level is equal or greater than the target's; however, if there are multiple wild Pokémon in the battle, it works the same as in a Trainer battle.

innerthunder: Added the level requirement. I don't think the double battle behavior is necessary to replicate since it's from a fairly old gen atp. I'm fine with being able to force Wild Pokemon to flee in any type of battle (aside from the "wild" ME battles).

DayKev: Wdym? It appears to have never changed since it was added in Gen 5 up through Gen 9.

innerthunder: https://bulbapedia.bulbagarden.net/wiki/Wild_Pok%C3%A9mon

Conventional wild double battles haven't appeared since Gen 5. Gen 6/7 had variations of it with horde battles and SOS battles, respectively.

DayKev: Ah, I see. Well, BDSP is technically Gen 8, but I suppose so.

src/battle-scene.ts
```diff
@@ -2842,6 +2845,170 @@ export class BattleScene extends SceneBase {
...
+    const blockedByAbility = new ValueHolder(false);
+    applyAbAttrs<ForceSwitchOutImmunityAbAttr>(AbAttrFlag.FORCE_SWITCH_OUT_IMMUNITY, pokemon, false, blockedByAbility);
+
+    return !(blockedByAbility.value || pokemon.isMax());
+  }
+
+  /**
+   * Forces the active enemy Pokemon at the given battler index to flee.
+   * @param battlerIndex - The {@linkcode FieldBattlerIndex} of the enemy Pokemon to make flee
+   * @param pokemon - The {@linkcode Pokemon} that initiated this effect
```
```suggestion
   * @param source - The {@linkcode Pokemon} that initiated this effect
```

DayKev:
There should probably be some tests added/modified/etc (can look at copying some from https://github.com/pagefaultgames/pokerogue/pull/5825 possibly). The current Wimp Out tests are a bit of a mess iirc.

Other than that lgtm.

test/abilities/wimp-out.test.ts
```diff
@@ -45,80 +46,106 @@ describe("Abilities - Wimp Out", () => {
...
+    game.override //
+      .enemyAbility(AbilityId.WIMP_OUT)
+      .enemyLevel(1)
+      .enemyHealthSegments(2);
 
-    game.move.select(MoveId.SPLASH);
+    await game.classicMode.startBattle(SpeciesId.WIMPOD);
+
+    const enemy = game.field.getEnemyPokemon();
+    game.move.use(MoveId.FALSE_SWIPE);
```
DayKev: False Swipe interacts incorrectly with boss bars atm (the "leave the target at 1 HP" damage reductions happens before boss bars are calculated instead of after) so it's not reliable to use False Swipe in a test involving boss pokemon. Sturdy should be used instead (in this case as the passive since the normal ability needs to be Wimp Out).

test/abilities/wimp-out.test.ts
```diff
@@ -558,8 +559,8 @@ describe("Abilities - Wimp Out", () => {
     await game.phaseInterceptor.to("PostActionPhase");
 
     expect(player1.getHpRatio()).toBeLessThanOrEqual(0.5);
-    expect(game.phaseInterceptor.log).not.toContain("SwitchSummonPhase");
-    expect(enemy1.isActive(true)).toBeTruthy();
-    expect(enemy1.waveData.abilitiesApplied.includes(AbilityId.WIMP_OUT)).toBeFalsy();
+    expect(player1).not.toHaveFainted();
+    expect(player1.isOnField()).toBeTruthy();
+    expect(player1.waveData.abilitiesApplied.includes(AbilityId.WIMP_OUT)).toBeFalsy();
```
DayKev: The enemy Pokémon is the one with Wimp Out in this test, the player's ability was set to No Guard.

DayKev (approved):

PR #1350, opened 2025-09-05
refactor: remove Endless Mode, move Daily Run to "New Game"

DayKev:
- Obsolete Endless Mode removed from the game.
- "Daily Run" option moved to "New Game" submenu.

Implementing the plan to remove endless mode.

All code related to endless mode is hopefully removed (there's some stuff in the force switch move/ability code that was untouched due to being removed by a different PR).

The option to start a new Daily Run was moved to the "New Game" submenu.

## Checklist


innerthunder (dismissed):

src/data/mystery-encounters/utils/encounter-phase-utils.ts
```diff
@@ -1062,9 +1062,6 @@ export function calculateMEAggregateStats(baseSpawnWeight: number) {
           currentBiome = biomeLinks[currentBiome] as BiomeId;
         } else if (i % 50) {
           currentBiome = globalScene.generateRandomBiome(i);
-        } else {
-          // Special logic for endless mode
-          currentBiome = BiomeId.END;
```
MokaStitcher, line 1075: should the `if (i%50)` just be replaced by an `else`? I think the check was only here because endless did a thing every 50 wave, and it shouldn't be needed anymore

src/data/splash-messages.ts
```diff
@@ -144,7 +144,6 @@ const commonSplashMessages = [
   "betterNerfGreninja",
   "inCaseOfUpdateClearYourCache",
   "insertTextHere",
-  "endingEndlessNotFound",
```
MokaStitcher, line 147: hm. should we make a locale repo PR as well to remove endless related keys?

src/game-mode.ts
```diff
@@ -327,17 +291,13 @@ export class GameMode implements GameModeConfig {
       case GameModes.CHALLENGE:
       case GameModes.DAILY:
         return isBoss ? 6 : 18;
-      case GameModes.ENDLESS:
-        return isBoss ? 4 : 12;
     }
   }
 
   getName(): string {
```
MokaStitcher, line 297: ugh, why is there `getName` and a static `getModeName`. This one should probably call the other to avoid having to maintain both I think? (assuming the other is needed)

src/enums/unlockables.ts
```diff
@@ -1,5 +1,5 @@
 export enum Unlockables {
-  ENDLESS_MODE,
+  CLASSIC_CLEAR,
```
MokaStitcher: This should be renamed `CHALLENGE_MODE` imo, a classic clear is not exactly an unlockable, especially considering the mini black hole is also unlocked by clearing classic

src/phases/unlock-phase.ts
```diff
@@ -1,9 +1,21 @@
...
-import type { Unlockables } from "#enums/unlockables";
-import { getUnlockableName } from "#system/unlockables";
+import { GameModes } from "#enums/game-modes";
+import { Unlockables } from "#enums/unlockables";
 import i18next from "i18next";
 
+function getUnlockableName(unlockable: Unlockables) {
+  switch (unlockable) {
+    case Unlockables.CLASSIC_CLEAR:
+      return `${GameMode.getModeName(GameModes.CHALLENGE)} Mode`;
```
MokaStitcher: Dunno where this is used in game at the moment, but this is not proper localizing. This needs a dedicated key if there is none (good opportunity to also remove the endless localization keys).

src/data/splash-messages.ts
DayKev: We need to figure out what we want to do with the locales, we don't have any of the last like 9 months worth of work that's been done to improve them.

src/enums/unlockables.ts
DayKev: Ah, hm. It's weird that they're separate then since they're unlocked by the same thing (and previously this was used for both challenge and endless mode...).
It should probably not be a list of unlockable things, but rather a list of events that unlock things.

src/game-mode.ts
DayKev: I think I'll move the static method to a helper function, and then this method can call the helper function with `this.modeId` as the parameter.

src/data/mystery-encounters/utils/encounter-phase-utils.ts
DayKev: Hmm, probably.

src/phases/unlock-phase.ts
DayKev: That's how it was originally lol.

innerthunder: I do think this should be localized, but to keep things organized I'd say keep it separate from whatever we decide to do with unused/outdated locale keys

PR #1351, opened 2025-09-06
ai-rework: Add Effect Scores for Terrain- and Weather-changing effects

innerthunder:

The AI should now properly evaluate moves that can change the terrain or weather on the field, including:
- Rain Dance et al
- Defog
- Electric Surge et al
- Ice Spinner and Steel Roller

Part of the AI Rework.

- Added Effect Score overrides for the following attributes:
  - `ChangeWeatherAttr` (renamed from `WeatherChangeAttr`)
  - `ClearWeatherAttr`
  - `ChangeTerrainAttr` (renamed from `TerrainChangeAttr`)

  Effect Scores are calculated by computing each non-fainted Pokemon's "Terrain/Weather Benefit Score", which checks the Pokemon's type, ability, and moves for synergies to factor into the final score.
- Refactored `ClearTerrainAttr` to extend `ChangeWeatherAttr` (and in turn reuse its Effect Score methods).
- Reorganized some existing localization functions in `terrain.ts` and `weather.ts` to `utils/terrain-utils.ts` and `utils/weather-utils.ts`, respectively. These new files also include some helper functions for AI move scoring.

## How to test the changes?

- [x] `pnpm typecheck`
- [x] `pnpm depcruise`
- [x] `pnpm test:silent`
  - Some unrelated tests seem to be flaky; this will be resolved within the `ai-rework` branch

- `sunny-day`
- `rain-dance`
- `sandstorm`
- `snowscape`
- `misty-terrain`
- `electric-terrain`
- `grassy-terrain`
- `psychic-terrain`
- `defog` (added tests for terrain removal)

## Checklist

### DIFF
--- a/src/constants/ability-constants.ts
+++ b/src/constants/ability-constants.ts
@@ -1,6 +1,7 @@
 /* biome-ignore-start lint/correctness/noUnusedImports: tsdoc imports */
 import type { Ability } from "#abilities/ability";
 import type { SuppressWeatherEffectAbAttr } from "#abilities/suppress-weather-effect-ab-attr";
+import type { WeatherType } from "#enums/weather-type";
 /* biome-ignore-end lint/correctness/noUnusedImports: tsdoc imports */
 
 import { AbilityId } from "#enums/ability-id";
@@ -81,4 +82,52 @@ export const POST_STAT_STAGE_REDUCTION_ABILITIES = Object.freeze<AbilityId[]>([
 export const RECOIL_DAMAGE_PREVENTION_ABILITIES = Object.freeze<AbilityId[]>([
   AbilityId.ROCK_HEAD,
   AbilityId.MAGIC_GUARD,
+]);
+
+/**
+ * Abilities that grant a benefit when {@link WeatherType.SUNNY | harsh sunlight}
+ * or {@link WeatherType.HARSH_SUN | extremely harsh sunlight} is in effect
+ */
+export const SUN_SYNERGY_ABILITIES = Object.freeze<AbilityId[]>([
+  AbilityId.CHLOROPHYLL,
+  AbilityId.SOLAR_POWER,
+  AbilityId.FLOWER_GIFT,
+  AbilityId.LEAF_GUARD,
+  AbilityId.PROTOSYNTHESIS,
+  AbilityId.FORECAST,
+]);
+
+/**
+ * Abilities that grant a benefit when {@link WeatherType.RAIN | rain}
+ * or {@link WeatherType.HEAVY_RAIN | heavy rain} is in effect
+ */
+export const RAIN_SYNERGY_ABILITIES = Object.freeze<AbilityId[]>([
+  AbilityId.SWIFT_SWIM,
+  AbilityId.RAIN_DISH,
+  AbilityId.DRY_SKIN,
+  AbilityId.HYDRATION,
+  AbilityId.FORECAST,
+]);
+
+/**
+ * Abilities that grant a benefit when a {@link WeatherType.SANDSTORM | sandstorm}
+ * is in effect
+ */
+export const SAND_SYNERGY_ABILITIES = Object.freeze<AbilityId[]>([
+  AbilityId.SAND_FORCE,
+  AbilityId.SAND_RUSH,
+  AbilityId.SAND_VEIL,
+  AbilityId.MAGIC_GUARD,
+  AbilityId.OVERCOAT,
+]);
+
+/**
+ * Abilities that grant a benefit when {@link WeatherType.HAIL | hail}
+ * or {@link WeatherType.SNOW | snow} is in effect
+ */
+export const SNOW_SYNERGY_ABILITIES = Object.freeze<AbilityId[]>([
+  AbilityId.ICE_BODY,
+  AbilityId.SNOW_CLOAK,
+  AbilityId.SLUSH_RUSH,
+  AbilityId.ICE_FACE,
 ]);
--- a/src/constants/ability-constants.ts
+++ b/src/constants/ability-constants.ts
@@ -88,46 +88,37 @@ export const RECOIL_DAMAGE_PREVENTION_ABILITIES = Object.freeze<AbilityId[]>([
  * Abilities that grant a benefit when {@link WeatherType.SUNNY | harsh sunlight}
  * or {@link WeatherType.HARSH_SUN | extremely harsh sunlight} is in effect
  */
-export const SUN_SYNERGY_ABILITIES = Object.freeze<AbilityId[]>([
-  AbilityId.CHLOROPHYLL,
-  AbilityId.SOLAR_POWER,
-  AbilityId.FLOWER_GIFT,
-  AbilityId.LEAF_GUARD,
-  AbilityId.PROTOSYNTHESIS,
-  AbilityId.FORECAST,
-]);
+export const SUN_SYNERGY_ABILITIES = Object.freeze<Set<AbilityId>>(
+  new Set([
+    AbilityId.CHLOROPHYLL,
+    AbilityId.SOLAR_POWER,
+    AbilityId.FLOWER_GIFT,
+    AbilityId.LEAF_GUARD,
+    AbilityId.PROTOSYNTHESIS,
+    AbilityId.FORECAST,
+  ]),
+);
 
 /**
  * Abilities that grant a benefit when {@link WeatherType.RAIN | rain}
  * or {@link WeatherType.HEAVY_RAIN | heavy rain} is in effect
  */
-export const RAIN_SYNERGY_ABILITIES = Object.freeze<AbilityId[]>([
-  AbilityId.SWIFT_SWIM,
-  AbilityId.RAIN_DISH,
-  AbilityId.DRY_SKIN,
-  AbilityId.HYDRATION,
-  AbilityId.FORECAST,
-]);
+export const RAIN_SYNERGY_ABILITIES = Object.freeze<Set<AbilityId>>(
+  new Set([AbilityId.SWIFT_SWIM, AbilityId.RAIN_DISH, AbilityId.DRY_SKIN, AbilityId.HYDRATION, AbilityId.FORECAST]),
+);
 
 /**
  * Abilities that grant a benefit when a {@link WeatherType.SANDSTORM | sandstorm}
  * is in effect
  */
-export const SAND_SYNERGY_ABILITIES = Object.freeze<AbilityId[]>([
-  AbilityId.SAND_FORCE,
-  AbilityId.SAND_RUSH,
-  AbilityId.SAND_VEIL,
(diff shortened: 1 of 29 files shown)


src/utils/terrain-utils.ts
```diff
@@ -0,0 +1,165 @@
...
+    }
+  }
+}
+
+/**
+ * @param terrainType - The {@linkcode TerrainType} to check
+ * @returns A set of {@link AbilityId | IDs} for abilities that benefit from
+ * the given terrain type
+ */
+export function getTerrainSynergyAbilities(terrainType: TerrainType): Readonly<Set<AbilityId>> {
```
DayKev:
```suggestion
export function getTerrainSynergyAbilities(terrainType: TerrainType): ReadonlySet<AbilityId> {
```

src/utils/weather-utils.ts
```diff
@@ -0,0 +1,249 @@
...
+function getStrongWindsSynergyScore(elementalType: ElementalType): number {
+  return elementalType === ElementalType.FLYING ? 1 : 0;
+}
+
+/**
+ * @param weatherType - The {@linkcode WeatherType} to check
+ * @returns A set of {@link AbilityId | IDs} for abilities that benefit from
+ * the given weather type.
+ */
+export function getWeatherSynergyAbilities(weatherType: WeatherType): Readonly<Set<AbilityId>> {
```
DayKev:
```suggestion
export function getWeatherSynergyAbilities(weatherType: WeatherType): ReadonlySet<AbilityId> {
```

src/constants/ability-constants.ts
```diff
@@ -82,3 +83,42 @@ export const RECOIL_DAMAGE_PREVENTION_ABILITIES = Object.freeze<AbilityId[]>([
   AbilityId.ROCK_HEAD,
   AbilityId.MAGIC_GUARD,
 ]);
+
+/**
+ * Abilities that grant a benefit when {@link WeatherType.SUNNY | harsh sunlight}
+ * or {@link WeatherType.HARSH_SUN | extremely harsh sunlight} is in effect
+ */
+export const SUN_SYNERGY_ABILITIES = Object.freeze<Set<AbilityId>>(
```
DayKev, line 91: Why use a set for these instead of an array?

src/field/pokemon.ts
```diff
@@ -2479,6 +2503,164 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   * @returns A decimal score approximating the Pokemon's potential synergy with the given weather
+   */
+  public getWeatherBenefitScore(weatherType: WeatherType, estimate: boolean = false): number {
+    const typeScore = this.getTypes(true, true).reduce(
+      (score, elemType) => score + getWeatherTypeSynergyScore(weatherType, elemType),
+      0,
+    );
+
+    // TODO: This ability check doesn't account for ability suppression
+    const abilities = this.getAbilities({ revealedOnly: estimate });
```
MokaStitcher, line 2523: is something preventing us from adding `canApplyOnly: true` to the options here to exclude suppressed abilities?

src/field/pokemon.ts
```diff
@@ -2479,6 +2503,164 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+      return 0;
+    }
+
+    const typeScore = this.getTypes(true, true).reduce(
+      (score, elemType) => score + getTerrainTypeSynergyScore(terrainType, elemType),
+      0,
+    );
+
+    // TODO: This ability check doesn't account for ability suppression
+    const abilities = this.getAbilities({ revealedOnly: estimate });
```
MokaStitcher, line 2561: same

src/field/pokemon.ts
```diff
@@ -2479,6 +2503,164 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   * Calculates a score to measure how much this Pokemon benefits from the
+   * given type of weather. The final score consists of three components:
+   * - An assigned score for the Pokemon's type(s) according to {@linkcode getWeatherTypeSynergyScore}.
+   * - (+1) for each Ability the Pokemon has (and can apply) that benefits from the weather.
+   * - (+0.5) for each Move the Pokemon knows that benefits from the weather.
+   * @param weatherType - The {@link WeatherType | type} of weather to evaluate this Pokemon against
+   * @param estimate - If `true`, limits checked abilities and moves to those revealed in battle
+   * @returns A decimal score approximating the Pokemon's potential synergy with the given weather
+   */
+  public getWeatherBenefitScore(weatherType: WeatherType, estimate: boolean = false): number {
```
MokaStitcher, line 2516: are there other AI specific functions in the Pokemon class? I wonder if it should be here or in utils along with `getWeatherTypeSynergyScore` (same with getTerrainBenefitScore)

test/ai/move-effect-scores/misty-terrain.test.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    const enemy = game.field.getEnemyPokemon();
+    expect(enemy).toPreferSelectingMove(MoveId.MISTY_TERRAIN);
+  });
+
+  it("should be strongly preferred when the player has multiple Dragon-type Pokemon", async () => {
+    await game.classicMode.startBattle(SpeciesId.AXEW, SpeciesId.GIBLE, SpeciesId.BAGON);
+
+    const enemy = game.field.getEnemyPokemon();
+    expect(enemy).toNeverSelectMove((move) => move.id !== MoveId.MISTY_TERRAIN);
```
MokaStitcher, line 55: would it make sense to have `toAlwaySelectMove` util, or am I misunderstanding something?

~~I wonder if we should have at least one of these terrain/weather tests do stuff with double battles to test that it is indeed looking at allies and not just the user~~ edit: no clue how this remark ended up here

test/ai/move-effect-scores/psychic-terrain.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  it("should not be preferred when no synergies are on the field", async () => {
+    await game.classicMode.startBattle(SpeciesId.FEEBAS);
+
+    equalizeSpd();
+    const enemy = game.field.getEnemyPokemon();
+    expect(enemy).not.toPreferSelectingMove(MoveId.PSYCHIC_TERRAIN);
+  });
+
+  it("should be preferred when the user is Psychic-type", async () => {
```
MokaStitcher, line 48: just to be sure, when those tests say "user" it actually is "user or its ally"?

test/ai/move-effect-scores/sandstorm.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    await game.classicMode.startBattle(SpeciesId.AVALUGG);
+
+    const enemy = game.field.getEnemyPokemon();
+    expect(enemy).toPreferSelectingMove(MoveId.SANDSTORM);
+  });
+
+  // Ground and Steel are penalized less, so Sandstorm is discouraged, but not avoided.
+  // However, testing for "not preferring" Sandstorm may produce false positives (see "no synergy" test)
+  it.skip.each(typeTestCases)("should be avoided when the opponent is $typeName-type", async ({ testSpecies }) => {
```
MokaStitcher, line 59: what would we want to do in the future to no longer skip this test? Is the behavior as we want it, thus the test could be changed, or would the AI need to be updated?

src/constants/ability-constants.ts
innerthunder: These are only used for lookups, and `Set.has(X)` performs better than `Array.includes(X)`.

src/field/pokemon.ts
innerthunder: `canApplyOnly` would also check whether the ability's conditions are met, and a few of the weather-based abilities use conditions to enforce them only applying in certain weather types. If the option were enabled, those abilities wouldn't be accounted for when evaluating a change to their respective weather type.

src/field/pokemon.ts
innerthunder: The Pokemon class also includes functions related to Attack Score and Matchup Score. I'm also planning on using this score in MUS calculations in a future PR.

test/ai/move-effect-scores/misty-terrain.test.ts
innerthunder: I could add an alias, I suppose. That would also apply to tests outside of this PR, though.

test/ai/move-effect-scores/sandstorm.test.ts
innerthunder: The Rock-type test suggests that the penalty is applying correctly. The problem is that the penalty for an opposing Ground/Steel-type is 50%(-1), which (A) isn't enough of a penalty for the AI to never use the move, and (B) doesn't change the AI's "preferences" compared to a no-synergy battle state (it'd rather use Tackle regardless).

We'd either have to adjust the penalty to (-1) or more or change the test so that it somehow compares results to the no-synergy state

src/field/pokemon.ts
MokaStitcher: ugh

test/ai/move-effect-scores/psychic-terrain.test.ts
MokaStitcher: "I wonder if we should have at least one of these terrain/weather tests do stuff with double battles to test that it is indeed looking at allies and not just the user"

This ended up on another comment for some reason, but is meant to be here I think

MokaStitcher (approved):

test/ai/move-effect-scores/psychic-terrain.test.ts
innerthunder: I added tests to `sunny-day` and `misty-terrain` for double battles and challenges. The double battle tests only cover type synergies, but it should be enough to show that scoring applies to all allies and enemies

src/data/moves/move-attrs/change-terrain-attr.ts
```diff
@@ -0,0 +1,61 @@
...
+  public override getCondition(): MoveConditionFunc | null {
+    return (_user, _target, _move) => !globalScene.arena.hasTerrain(this.terrainType);
+  }
+
+  /**
+   * @returns An Effect Score based on how much the user's non-fainted party benefits
+   * from the Terrain to set compared to the opponents' benefit.
+   */
+  public override getEffectScore(user: EnemyPokemon, _target: Pokemon, _move: Move): number {
+    const currentTerrain = globalScene.arena.terrain?.terrainType ?? TerrainType.NONE;
```
DayKev, line 36: Oh I missed this the first time, this should be `const currentTerrain = globalScene.arena.getTerrainType();`

DayKev: Actually I'm gonna make a few changes to `Arena` wrt this (the matching weather one is missing too) so don't worry about it.

PR #898, opened 2025-04-01
[Refactor] Biome random weather + trainers added

Tempo-anon:
## What are the changes the user will see?

Volcano is 90% sun instead of 100% sun during dawn/day, otherwise biome weathers are the same as before the refactor

Also #902 

## Why am I making these changes?

Remake of #880, part of the overall biome refactor

## What are the changes from a developer perspective?
* Changed `weatherPool` into a `Partial<Record<WeatherType, number>>`
* Deleted `weather.getRandomWeatherType` and moved biome weather data to each biome
* Created `utils.weightedPick` as a helper function for picking items with associated weights
* Use that helper function in creating `arena.setRandomWeather` and call that in `encounter-phase.ts` and `new-biome-encounter-phase.ts`

## Screenshots/Videos

image

## How to test the changes?

Override to a biome with weather

## Checklist

- [ ] ⚠️ If this is a PR for `main` (such as a hotfix), has the game version been updated (`npm run update-version:patch` / `npm run update version:minor`?
- [x] Otherwise: **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?

- [x] Are all unit tests still passing? (`npm run test:silent`)
  - [ ] Have I created new automated tests (`npm run test:create`) or updated existing tests related to the PR's changes?
- [ ] Have I provided screenshots/videos of the changes (if applicable)?
  - [ ] Have I made sure that any UI change works for both UI themes (dark and light)?

#### Are there any localization additions or changes? If so:

- [ ] Has a locales PR been created on the [locales](https://github.com/despair-games/poketernity-locales) repo?
  - [ ] If so, please leave a link to it here:
- [ ] Have I added the `Localization` tag to this PR?

#### If there are no locale changes:
- [ ] Have I made sure **not** to commit any changes to the locale repo on this branch?

PR #1352, opened 2025-09-06
refactor: remove `randSeedWeightedItem`

DayKev:
Slightly more random trainer parties?

We added a replacement for `randSeedWeightedItem` but didn't remove it. Now it's gone.

The single use of `randSeedWeightedItem` was replaced with `randSeedItem`. A very minor cleanup of the `Trainer` class constructor was also done.

## Checklist

### DIFF
--- a/src/field/trainer.ts
+++ b/src/field/trainer.ts
@@ -22,7 +22,7 @@ import type { EnemyPokemon } from "#field/enemy-pokemon";
 import type { PersistentModifier } from "#modifier/modifier";
 import { allTrainerConfigs } from "#trainer-configs/all-trainer-configs";
 import { getPokemonSpecies } from "#utils/pokemon-utils";
-import { randSeedInt, randSeedItem, randSeedWeightedItem } from "#utils/random-utils";
+import { randSeedInt, randSeedItem } from "#utils/random-utils";
 import i18next from "i18next";
 
 export class Trainer extends Phaser.GameObjects.Container {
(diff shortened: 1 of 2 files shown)


innerthunder (approved):

MokaStitcher (commented):
> We added a replacement for randSeedWeightedItem but didn't remove it. Now it's gone.

What's the replacement?

RNG.weightedPick favors the elements at the beginning of the array [as per Phaser documentation](https://docs.phaser.io/api-documentation/class/math-randomdatagenerator), so even if it wasn't documented by us we know what it did and it's not the same as randSeedItem.

If everyone is fine with the change though 🤷

DayKev:
> > We added a replacement for randSeedWeightedItem but didn't remove it. Now it's gone.
> 
> What's the replacement?

https://github.com/Despair-Games/poketernity/blob/82ebca528e009d12a9ebed9a8b46dfa8f08a5b6c/src/utils/random-utils.ts#L93-L110

Added in #898

> RNG.weightedPick favors the elements at the beginning of the array [as per Phaser documentation](https://docs.phaser.io/api-documentation/class/math-randomdatagenerator), so even if it wasn't documented by us we know what it did and it's not the same as randSeedItem.

Well it just says "favoring the earlier entries" but that doesn't tell us *how* it does that. The new function lets us explicitly set the weights we want.

MokaStitcher (approved):
> Well it just says "favoring the earlier entries" but that doesn't tell us how it does that. The new function lets us explicitly set the weights we want.

Ok but this PR doesn't use that new function in the spot where it removes the old one. Whether we want to keep the same or similar behavior is a game design decision. I don't know how much it impacts matters so :shrug:

DayKev:
> > Well it just says "favoring the earlier entries" but that doesn't tell us how it does that. The new function lets us explicitly set the weights we want.
> 
> Ok but this PR doesn't use that new function in the spot where it removes the old one. Whether we want to keep the same or similar behavior is a game design decision. I don't know how much it impacts matters so 🤷

We might want to use weighted party templates, but the current ones weren't designed with the understanding that the weighted randomness was even being used here so maybe not.

PR #1356, opened 2025-09-11
dev: upgrade `lint/nursery/noShadow` rule from "warn" to "error"

DayKev:
N/A

Applying more lint rules.

Upgraded the `lint/nursery/noShadow` rule from "warn" to "error" and applied it to the codebase.

## Checklist


src/data/animations/battle-anims.ts
```diff
@@ -403,47 +402,47 @@ export abstract class BattleAnim {
...
+                 * Move the position that the moveSprite is rendered in based on the priority.
+                 * Priority values:
                  * - 0: Draw the sprite in front of the pokemon on the field.
                  * - 1: Draw the sprite in front of the user pokemon.
                  * - 2: Draw the sprite in front of its `bgSprite` (if it has one), or its
                  * `AnimFocus` (if that is user/target), otherwise behind everything.
                  * - 3: Draw the sprite behind its `AnimFocus` (if that is user/target), otherwise in front of everything.
                  */
-                const setSpritePriority = (priority: number) => {
+                {
```
MokaStitcher: is there a re reason to keep the brackets even though we're removing `setSpritePriority`?

src/field/arena.ts
```diff
@@ -810,8 +810,15 @@ export class Arena {
...
       if (ENTRY_HAZARD_ARENA_TAG_TYPES.includes(existingTag.tagType)) {
-        const { tagType, side, turnCount, layers, maxLayers } = existingTag as EntryHazardTag;
-        this.eventTarget.dispatchEvent(new TagAddedEvent(tagType, side, turnCount, layers, maxLayers));
+        const { layers, maxLayers } = existingTag as EntryHazardTag;
+        const tagEvent = new TagAddedEvent(
+          existingTag.tagType,
+          existingTag.side,
+          existingTag.turnCount,
+          layers,
+          maxLayers,
```
MokaStitcher: ngl it feels weird to deconstruct half of it, I'd prefer to keep things consistent because at a glance no one would have any clue why it is done that way
```suggestion
        const tagEvent = new TagAddedEvent(
          existingTag.tagType,
          existingTag.side,
          existingTag.turnCount,
          existingTag.layers,
          existingTag.maxLayers,
```

src/field/arena.ts
```diff
@@ -827,9 +834,8 @@ export class Arena {
         ? (newTag as EntryHazardTag)
         : {};
 
-      this.eventTarget.dispatchEvent(
-        new TagAddedEvent(newTag.tagType, newTag.side, newTag.turnCount, layers, maxLayers),
-      );
+      const tagEvent = new TagAddedEvent(newTag.tagType, newTag.side, newTag.turnCount, layers, maxLayers);
```
MokaStitcher: <sub> i guess we already did half and half here 😔

src/field/trainer.ts
```diff
@@ -380,23 +380,23 @@ export class Trainer extends Phaser.GameObjects.Container {
 
         // Filter out the species that are already in the enemy party from the main trainer species pool
         const speciesPoolFiltered = speciesPool
-          .filter((species) => {
+          .filter((speciesId) => {
```
MokaStitcher: humm, maybe `pool` would be a more fitting name? Since it's possible that it's an array (see 2 lines below),  speciesId feels a bit misleading

src/system/achievements.ts
```diff
@@ -1,3 +1,5 @@
+/** biome-ignore-all lint/nursery/noShadow: changed in another PR */
```
MokaStitcher, line 1: which one?

src/ui/handlers/starter-select-ui-handler.ts
```diff
@@ -2840,28 +2839,27 @@ export class StarterSelectUiHandler extends MessageUiHandler {
...
     const sort = this.filterBar.getVals(DropDownColumn.SORT)[0];
-    this.filteredStarterContainers.sort((a, b) => {
+    this.filteredStarterContainers.sort((contA, contB) => {
+      const { starterData } = globalScene.gameData;
+      const speciesIdA = contA.species.speciesId;
+      const speciesIdB = contB.species.speciesId;
       switch (sort.val) {
         case SortCriteria.NUMBER:
-          return (a.species.speciesId - b.species.speciesId) * -sort.dir;
+          return (speciesIdA - speciesIdB) * -sort.dir;
```
MokaStitcher, line 2843: <sub> why are be doing `(a - b) * -dir` rather than `(b - a) * dir` everywhere :plains:

MokaStitcher (commented):
not super fond of the abbreviated names in some places, but mostly lgtm

src/field/arena.ts
innerthunder: I made this suggestion on another file in #1322, maybe something similar can be done here?
```ts
for (const tag of globalScene.arena.tags) {
  const { tagType, side, turnCount } = tag;
  const { layers, maxLayers } = tag instanceof EntryHazardTag ? tag : {};
  globalScene.arena.eventTarget.dispatchEvent(
    new TagAddedEvent(tagType, side, turnCount, layers ?? 0, maxLayers ?? 0);
  ); 
}
```

src/system/achievements.ts
DayKev: The achievement scene PR.

src/field/trainer.ts
DayKev: Good idea.

src/data/animations/battle-anims.ts
DayKev: Oops, I must have missed those.

src/field/arena.ts
DayKev: > ngl it feels weird to deconstruct half of it, I'd prefer to keep things consistent because at a glance no one would have any clue why it is done that way

If it's not done that way then you have to do `(existingTag as EntryHazardTag).layers, (existingTag as EntryHazardTag).maxLayers`. I can convert it to use inner's suggestion though.

DayKev: Actually in the first place I changed this due to `noShadow` so I can't do that.

DayKev: Okay I renamed the method params instead of removing the destructuring, and simplified the second one as well.
Also fixed the TSDocs while I was at it.

innerthunder (dismissed):

PR #1358, opened 2025-09-13
ai-rework: Update Effect Score calculation for charge moves

innerthunder:

The Enemy AI should now properly evaluate two-turn moves (AKA "charge moves"). In general, they should be less keen on using them unless they are able to be used in one turn (e.g. Solar Beam in sun). Moves that grant semi-invulnerability (including Sky Drop) are also now more accurately scored.

Part of the AI Rework.

- The `ChargeMove` mixin now overrides `getCombinedAttributeScore` to include charge attributes in scoring and grant a (-2) penalty if instant-charge conditions are not met.
- Added Effect Score overrides for `SemiInvulnerabilityAttr` and `SkyDropAttr`

## How to test the changes?

- [x] `pnpm typecheck`
- [x] `pnpm depcruise`
- [x] `pnpm test:silent`

- `solar-beam`
- `sky-drop`

## Checklist


src/data/moves/move-attrs/semi-invulnerable-attr.ts
```diff
@@ -18,7 +21,28 @@ export class SemiInvulnerableAttr extends MoveEffectAttr {
...
+   * @returns A {@linkcode MINOR_EFFECT_SCORE_BONUS} if the following conditions apply:
+   * - The user is faster than all of its active opponents
+   * - All opponents cannot bypass the effects of semi-invulnerability (e.g. via No Guard)
+   *
+   * This effectively halves the penalty applied to charge moves under favorable conditions.
+   */
+  public override getEffectScore(user: EnemyPokemon, _target: Pokemon, _move: Move): number {
+    const opponents = user.getOpponents();
+    /**
+     * @todo
```
MokaStitcher, line 38: Could it be useful to have an `aiEdgeCase` or something somewhere to keep track of all the loose ends in ai more easily? :thinking:

MokaStitcher (approved):
lgtm

src/data/moves/move-attrs/semi-invulnerable-attr.ts
innerthunder: The edge cases tend to be tied to attributes rather than moves, so I'm not sure if that would be the best way to track them

MokaStitcher: Makes sense

DayKev (approved):

PR #1360, opened 2025-09-14
refactor: add getter for weather type to `Arena`

DayKev:
N/A

Simplify code.

- Added missing getter for the current weather to `Arena`.
- Replaced `Arena#getTerrainType` method with getter `Arena#terrainType`.
- Updated a few instances where `Arena#hasWeather`, `Arena#hasTerrain` and `Arena#terrainType` should have been used.

## Checklist


innerthunder (dismissed):
lgtm

We should also make `Weather.weatherType` and `Terrain.terrainType` `readonly`.

PR #1371, opened 2025-09-24
feat: Add dynamic Phase scheduling

innerthunder:
> [!NOTE]
> Much of this PR is adapted from https://github.com/pagefaultgames/pokerogue/pull/6036

## What are the changes the user will see?

In-battle effects (e.g. post-summon abilities) are now executed in order of their source Pokemon's Speed. As in Gen 8+, the order of these effects is dynamic, changing to reflect any modifications to Speed during the turn.

## Why am I making these changes?

#262 and #885 introduced dynamic ordering for turn commands. This does the same for non-command-based effects.

## What are the changes from a developer perspective?

There's a lot to cover, but the most important change is the overhaul to `PhaseManager`'s internal queue structure. Instead of the current phase queue and prepend, it now uses two key components, explained further below.

### `PhaseTree` - for Static Phase Scheduling

The main phase queue is now implemented as a multi-leveled "tree" of phases. From PKR#6036:

> The main queue is a 2D array of "levels". Phases are popped from the "highest" level first, with each level running in FIFO order. "unshifted" phases are pushed to the level above the currently running phase. Pushed phases are instead pushed to the lowest level. See phase-tree.ts.

Our version of `PhaseTree` has _three_ explicit methods to add phases to the tree:
- `push`: Appends phases to the lowest (or "root") level
- `unshift`: Appends phases to the highest (or "top") level. If unshifting for the first time in a cycle, a new "unshifted" level is added to the top of the tree before the phase is added.
- `defer`: Creates a level below the cycle's unshifted level and appends phases to that level.

A _cycle_ may be defined as the interval between fetch (`getNextPhase`) calls, which is analogous to the runtime of the Phase Manager's current phase.

<details><summary><b>Example: Phase Tree Scheduling</b></summary>
<p>

Suppose the Phase Tree is in the following state as a phase is running:

```
[
  [APhase, BPhase, CPhase], // root level
  [DPhase, EPhase], // top level
]
```

During the current phase's runtime, the following calls are made to schedule phases:
1. `unshift(FPhase)`
2. `push(GPhase)`
3. `defer(HPhase)`
4. `unshift(JPhase)`
5. `defer(KPhase)`

As the phases are scheduled, the Phase Tree is updated as follows.

First, a level is created as the top level, and `FPhase` is added to the new top level.

```
[
  [APhase, BPhase, CPhase],
  [DPhase, EPhase],
  [FPhase], // new top level
]
```

`GPhase` is then appended to the root level.

```
[
[...]

PR #1372, opened 2025-09-28
misc: Change ability flyout timing to match mainline behavior

### LINKED ISSUE #1262: [Refactor] Ability Flyout usage and timing
PokeRogue fixed flyout timing a few months ago with https://github.com/pagefaultgames/pokerogue/pull/5267. The PR highlights several implementation issues that are worth investigating:
- The ability flyout should be displayed before the ability's effects take place
- When multiple abilities apply at the same time, the flyout may overlap "show" and "hide" animations.
- Some abilities may trigger the flyout animation multiple times in one activation (e.g. Storm Drain)
- The ability flyout animation is async and managed across multiple Phases, which can run the risk of desync.

That said, we've made a lot of changes to ability application logic, so it'd be better to do our own investigation into the issue.

Open for discussion:
- How strictly should we follow mainline's "rules" on ability flyouts? In mainline, the ability flyout is generally restricted to additional effects (e.g. Levitate granting a Ground-immunity) and not amplifiers to existing effects (e.g. Huge Power boosting the source's Attack).

innerthunder:

- Ability Flyout displays are now synchronous with visual effects from ability application. In other words, the sequence of a displayed ability application is now well-defined as

```
Show flyout > Show trigger message > Apply ability > Hide flyout
```

- Async "instant" flyout displays have been fully removed. With this being removed, several abilities are no longer displayed through the ability flyout (most, if not all, of which are also not displayed in mainline)

To make ability displays more stable and standardized. This also fixes ability flyout timing issues introduced in #1371.

- Refactored `AbilityBar` so that the player and enemy flyout are implemented by separate `AbilityBars`. Constant values are also extracted to make ability bar positioning easier to tweak.
- Cleaned up `ShowAbilityPhase`
- Added `HideAbilityPhase` to synchronously hide either the player's or enemy's ability bar.
- Split the `apply` methods of most ability attributes into `canApply` (which checks if the effect can apply in current conditions) and `apply` (which applies the effect unconditionally)
- Some minor attribute code cleanup (e.g. replacing `Boolean/NumberHolder` instances with `ValueHolder`)

## Screenshots/Videos

https://github.com/user-attachments/assets/609c3d39-19f3-41b4-8729-bd3a97d4817d

<details><summary>Overrides</summary>
<p>

```ts
const overrides = {
  ENEMY_ABILITY_OVERRIDE: AbilityId.LEVITATE,
  ENEMY_MOVESET_OVERRIDE: MoveId.EARTHQUAKE,
  PASSIVE_ABILITY_OVERRIDE: AbilityId.LEVITATE,
  MOVESET_OVERRIDE: MoveId.EARTHQUAKE,
```

</p>
</details> 

## How to test the changes?

The ability flyout needs to be tested manually. Using overrides to force a post-summon ability (e.g. Intimidate) is probably the fastest and easiest way to test.

## Checklist


src/data/abilities/ab-attrs/pre-apply-battler-tag-immunity-ab-attr.ts
DayKev, line 1: The chain of `AbAttr`s involving this one are a bit... lol

src/data/abilities/ab-attrs/force-switch-out-immunity-ab-attr.ts
```diff
@@ -1,16 +1,15 @@
...
 import { AbAttrFlag } from "#enums/ab-attr-flag";
 import type { Pokemon } from "#field/pokemon";
-import type { BooleanHolder } from "#utils/common-utils";
+import type { ValueHolder } from "#utils/common-utils";
 
 export class ForceSwitchOutImmunityAbAttr extends AbAttr {
-  constructor(showAbility: boolean = true, showAbilityInstant: boolean = false) {
-    super(showAbility, showAbilityInstant);
+  constructor() {
+    super();
```
DayKev, line 15: Suction Cups should show a flyout: https://www.youtube.com/watch?v=lW6p9GCrRS4
and Guard Dog shouldn't (for some reason...): https://www.youtube.com/watch?v=rggthHKGgTM
```suggestion
  constructor(showAbility: boolean) {
    super(showAbility);
```

src/data/abilities/ab-attrs/full-hp-resist-type-ab-attr.ts
```diff
@@ -24,27 +24,22 @@ export class FullHpResistTypeAbAttr extends PreDefendAbAttr {
...
-    if (pokemon.isFullHp() && typeMultiplier.value > 0.5) {
-      typeMultiplier.value = 0.5;
-      if (!simulated) {
-        pokemon.turnData.moveEffectiveness = 0.5;
-      }
-      return true;
-    }
-    return false;
+  public override canApply(...[pokemon, , , move, typeMultiplier]: Parameters<this["apply"]>): boolean {
+    return pokemon.isFullHp() && !move.hasAttr(FixedDamageAttr) && typeMultiplier.value > 0.5;
```
DayKev: Just a small micro-optimization
```suggestion
    return pokemon.isFullHp() && typeMultiplier.value > 0.5 && !move.hasAttr(FixedDamageAttr);
```

src/data/abilities/ab-attrs/mock-status-effect-ab-attr.ts
```diff
@@ -11,14 +11,13 @@ import type { NumberHolder } from "#utils/common-utils";
 export class MockStatusEffectAbAttr extends AbAttr {
   public mockedStatus: StatusEffect;
   constructor(mockedStatus: StatusEffect) {
-    super(false);
+    super();
```
DayKev, line 14: Hm, Comatose is supposed to have an ability flyout (once when it comes out, and whenever it blocks a status move https://www.youtube.com/watch?v=I1Eoh6OFq38) but the current setup means you can't just pass `true` here...

src/data/abilities/ab-attrs/move-effect-chance-multiplier-ab-attr.ts
```diff
@@ -3,41 +3,36 @@ import { AbAttrFlag } from "#enums/ab-attr-flag";
...
  * Modifies moves additional effects with multipliers, ie. Sheer Force, Serene Grace.
  * @see {@linkcode apply}
  */
 export class MoveEffectChanceMultiplierAbAttr extends AbAttr {
   private readonly chanceMultiplier: number;
+  /**
+   * Secondary effect chance multipliers do not apply to these moves
+   * even though they are implemented with 100 base chance.
+   */
+  private static exceptMoves = [MoveId.ORDER_UP, MoveId.ELECTRO_SHOT];
```
DayKev: Instead of a `private` `static` class member, you could just make it a non-exported `const` outside of the class. It would minify better that way too.
```suggestion
  private static readonly exceptMoves: readonly MoveId[] = [MoveId.ORDER_UP, MoveId.ELECTRO_SHOT];
```

src/data/abilities/ab-attrs/move-effect-chance-multiplier-ab-attr.ts
```diff
@@ -3,41 +3,36 @@ import { AbAttrFlag } from "#enums/ab-attr-flag";
...
-    this.showAbility = showAbility;
-
-    const exceptMoves = [MoveId.ORDER_UP, MoveId.ELECTRO_SHOT];
-    if (moveChance.value <= 0 || exceptMoves.includes(move.id)) {
-      return false;
-    }
-
+  public override apply(_pokemon: Pokemon, _simulated: boolean, moveChance: ValueHolder<number>, _move: Move): void {
     moveChance.value *= this.chanceMultiplier;
     moveChance.value = Math.min(moveChance.value, 100);
```
DayKev:
```suggestion
    moveChance.value = Math.min(moveChance.value * this.chanceMultiplier, 100);
```

src/data/abilities/ab-attrs/non-super-effective-immunity-ab-attr.ts
```diff
@@ -17,20 +17,19 @@ export class NonSuperEffectiveImmunityAbAttr extends TypeImmunityAbAttr {
...
-  ): boolean {
-    if (move.isAttackMove() && typeMultiplier.value < 2) {
-      cancelled.value = true; // Suppresses "No Effect" message
-      typeMultiplier.value = 0;
-      return true;
-    }
+    cancelled: ValueHolder<boolean>,
+    typeMultiplier: ValueHolder<number>,
+  ): void {
+    cancelled.value = true; // Suppresses "No Effect" message
```
DayKev: This isn't necessary anymore since you're making it call `super.appy`, which you changed to set `cancelled.value = true`.
```suggestion
```

src/data/abilities/ab-attrs/type-immunity-ab-attr.ts
```diff
@@ -14,50 +15,54 @@ import type { BooleanHolder, NumberHolder } from "#utils/common-utils";
...
+   * flyout to display correctly. By default, this is set to the baseline no-effect message
+   * ("It doesn't affect {Pokemon}!").
+   */
+  public override getTriggerMessage(pokemon: Pokemon, _abilityName: string): string | null {
+    return i18next.t("battle:hitResultNoEffect", { pokemonName: getPokemonNameWithAffix(pokemon) });
+  }
+
+  public override getCondition(): AbAttrCondition {
     return this.condition;
   }
```
DayKev, line 67: Hmm... there's three layers of conditions: `AbAttr#extraCondition` which is used by `.conditionalAttr`, the various `[ability attr]#condition` fields that are added on which are usually (but apparently not always) evaluated within `canApply`, and `AbAttr#canApply` itself. There doesn't appear to be a difference between whether `extraCondition` evaluates to false or `canApply` returns false either, from looking at `apply-ab-attrs.ts`. It should be possible to consolidate all of this somehow (in the future).

src/data/abilities/ab-attrs/post-battle-loot-ab-attr.ts
```diff
@@ -2,28 +2,42 @@ import { PostBattleAbAttr } from "#abilities/post-battle-ab-attr";
...
+  public override apply(pokemon: Pokemon, simulated: boolean, _isVictory: boolean): void {
     const postBattleLoot = globalScene.currentBattle.postBattleLoot;
+    if (this.randItem == null) {
+      this.randItem = randSeedItem(postBattleLoot);
+    }
+
+    if (
+      !simulated
+      && globalScene.tryTransferHeldItemModifier(this.randItem, pokemon, true, 1, true, undefined, false)
+    ) {
```
DayKev:
```suggestion
    if (globalScene.tryTransferHeldItemModifier(this.randItem, pokemon, true, 1, true, undefined, false)) {
```

src/data/abilities/ab-attrs/post-battle-loot-ab-attr.ts
```diff
@@ -2,28 +2,42 @@ import { PostBattleAbAttr } from "#abilities/post-battle-ab-attr";
...
+import type { PokemonHeldItemModifier } from "#modifier/modifier";
 import { randSeedItem } from "#utils/random-utils";
 import i18next from "i18next";
 
 export class PostBattleLootAbAttr extends PostBattleAbAttr {
-  public override apply(pokemon: Pokemon, simulated: boolean, isVictory: boolean): boolean {
+  private randItem?: PokemonHeldItemModifier;
+
+  public override apply(pokemon: Pokemon, simulated: boolean, _isVictory: boolean): void {
     const postBattleLoot = globalScene.currentBattle.postBattleLoot;
```
DayKev, line 17: Technically this `AbAttr` doesn't need to check `simulated` because it should never be `true` for this.
```suggestion
    if (simulated) {
      return;
    }
    const postBattleLoot = globalScene.currentBattle.postBattleLoot;
```

src/data/abilities/ab-attrs/post-biome-change-ab-attr.ts
```diff
@@ -1,9 +1,12 @@
 import { AbAttr } from "#abilities/ab-attr";
 import { AbAttrFlag } from "#enums/ab-attr-flag";
+import type { Pokemon } from "#field/pokemon";
 
 export abstract class PostBiomeChangeAbAttr extends AbAttr {
```
DayKev: Huh... this `AbAttr` and its subclasses are entirely reundant, `NewBiomeEncounterPhase` could just do `applyAbAttrs<PostSummonTerrainChangeAbAttr>(...)` and `applyAbAttrs<PostSummonWeatherChangeAbAttr>(...)` since they have the exact same code as these `AbAttr`s.

Though it probably doesn't even need to do that anymore, I think that was solely for Endless (since the X5 biome changes didn't cause an arena reset) which was removed. So just delete this `AbAttr` and it's subclasses, and remove the application from `NewBiomeEncounterPhase`.

src/data/abilities/ab-attrs/post-defend-steal-held-item-ab-attr.ts
```diff
@@ -1,48 +1,53 @@
...
   }
 
-  getTargetHeldItems(target: Pokemon): PokemonHeldItemModifier[] {
+  public override canApply(...[pokemon, , attacker, move]: Parameters<this["apply"]>): boolean {
+    if (!move.isAttackMove(attacker, pokemon) || !this.condition(pokemon, attacker, move)) {
+      return false;
+    }
+
+    const heldItems = this.getTargetHeldItems(attacker).filter((i) => i.isTransferable);
+    return heldItems.length > 0;
```
DayKev:
```suggestion
    return this.getTargetHeldItems(attacker).some((i) => i.isTransferable);
```

src/data/abilities/ab-attrs/post-summon-copy-ability-ab-attr.ts
```diff
@@ -14,37 +13,24 @@ export class PostSummonCopyAbilityAbAttr extends PostSummonAbAttr {
...
-      target = target!;
-    } else {
-      target = targets[0];
+  public override apply(pokemon: Pokemon, simulated: boolean): void {
+    if (!simulated) {
+      this.targetAbilityName = allAbilities[this.target.getAbility().id].name;
+      pokemon.summonData.ability = this.target.getAbility().id;
+      this.target.waveData.abilitiesRevealed.push(this.target.getAbility().id);
+      pokemon.updateInfo();
     }
```
DayKev: Reduces `.getAbility()` calls from 3 to 1, and removes the unnecessary `allAbilities[abilityId]`.
```suggestion
    if (simulated) {
      return;
    }

    const targetAbility = this.target.getAbility();
    this.targetAbilityName = targetAbility.name;
    pokemon.summonData.ability = targetAbility.id;
    this.target.waveData.abilitiesRevealed.push(targetAbility.id);
    pokemon.updateInfo();
```

src/data/abilities/ab-attrs/post-summon-user-field-remove-status-effect-ab-attr.ts
```diff
@@ -11,34 +11,32 @@ import { getStatusEffectHealText } from "#utils/status-effect-utils";
...
+        p.updateInfo();
       }
-    }
-    return true;
+    });
+  }
+
+  public override canApply(...[pokemon]: Parameters<this["apply"]>): boolean {
+    const affectedPokemon = pokemon.getField().filter((p) => p.isActive(true));
+    return affectedPokemon.some((p) => p.hasStatusEffect(this.statusEffects, false, true));
```
DayKev:
```suggestion
    return pokemon.getField().some((p) => p.isActive(true) && p.hasStatusEffect(this.statusEffects, false, true));
```

src/data/abilities/ab-attrs/anticipation-ab-attr.ts
```diff
@@ -9,23 +10,38 @@ import { OneHitKOAttr } from "#moves/one-hit-ko-attr";
...
-          }
-          if (move.hasAttr(OneHitKOAttr)) {
-            return super.apply(pokemon, simulated);
-          }
-        }
+  /** This is just to remove the `args` parameter */
+  public override apply(pokemon: Pokemon, simulated: boolean): void {
+    super.apply(pokemon, simulated);
+  }
+
```
DayKev: This doesn't appear to be necessary (anymore?).
```suggestion
```

src/data/abilities/ab-attrs/prevent-bypass-speed-chance-ab-attr.ts
```diff
@@ -15,19 +15,20 @@ export class PreventBypassSpeedChanceAbAttr extends AbAttr {
...
     const turnCommand = globalScene.currentBattle.turnManager.findCommandFromPokemon(pokemon);
     const isCommandFight = turnCommand?.command === BattleCommand.FIGHT;
     const move = turnCommand?.turnMove?.move;
-    if (move && this.condition(pokemon, move) && isCommandFight) {
-      cancelled.value = true;
-      return true;
-    }
-    return false;
+
+    return move != null && this.condition(pokemon, move) && isCommandFight;
```
DayKev: Another micro-optimization.
```suggestion
    return move != null && isCommandFight && this.condition(pokemon, move);
```

src/data/abilities/ab-attrs/reduce-berry-use-threshold-ab-attr.ts
```diff
@@ -1,22 +1,19 @@
...
-      return threshold.value >= hpRatio;
-    }
+  public override apply(_pokemon: Pokemon, _simulated: boolean, threshold: ValueHolder<number>): void {
+    threshold.value *= 2;
+  }
 
-    return false;
+  public override canApply(...[pokemon, , threshold]: Parameters<this["apply"]>): boolean {
+    return isBetween(pokemon.getHpRatio(), threshold.value, threshold.value * 2);
   }
```
DayKev: The Ability should always apply, since it's not what determines whether the berry activates or not, it's just changing it from `<= 25%` to `<= 50%`.
```suggestion
```

src/data/abilities/ab-attrs/suppress-weather-effect-ab-attr.ts
```diff
@@ -2,9 +2,13 @@ import { PreWeatherEffectAbAttr } from "#abilities/pre-weather-effect-ab-attr";
...
 import type { Pokemon } from "#field/pokemon";
-import type { BooleanHolder } from "#utils/common-utils";
+import type { ValueHolder } from "#utils/common-utils";
 
 export class SuppressWeatherEffectAbAttr extends PreWeatherEffectAbAttr {
+  /**
+   * @todo Should this be removed?
+   * Every instance of this attr sets this to `true`.
+   */
   public readonly affectsPrimal: boolean;
```
DayKev: I think it's safe to remove it. If necessary it can be re-added later, though it probably won't need to be.

src/field/arena.ts
```diff
@@ -376,6 +376,14 @@ export class Arena {
     return true;
   }
 
+  /**
+   * @param terrainType - The {@link TerrainType | type} of terrain to check
+   * @returns `true` if Terrain of the given type can be set on the field.
+   */
+  public canSetTerrain(terrain: TerrainType): boolean {
+    return this.terrain?.terrainType !== terrain;
```
```suggestion
    return (this.terrain?.terrainType ?? TerrainType.NONE) !== terrain;
```

src/utils/anim-utils.ts
```diff
@@ -73,14 +73,8 @@ export async function loadEncounterAnimAssets(startLoad?: boolean): Promise<void
  * Plays a Tween animation, resolving once the animation completes.
  * @param config - The config for a single Tween
  * @param scene - The {@linkcode SceneBase} on which the Tween plays (Default {@linkcode globalScene})
- *
- * @privateRemarks
- * The `config` input should not include an `onComplete` field as that callback is
- * used to resolve the Promise containing the Tween animation. However, `config`'s type
- * cannot be changed to something like `Omit<TweenBuilderConfig, "onComplete">` due to
- * how Phaser interprets `TweenBuilderConfig` inputs.
```
DayKev, line 93: Why remove the comment?

src/battle-scene.ts
```diff
@@ -2146,6 +2152,51 @@ export class BattleScene extends SceneBase {
     }
   }
 
+  private hasModifier(modifier: PersistentModifier, enemy = false): boolean {
+    const modifiers = enemy ? this.enemyModifiers : this.modifiers;
+    return modifiers.indexOf(modifier) > -1;
```
```suggestion
    return modifiers.includes(modifier);
```

src/battle-scene.ts
```diff
@@ -2146,6 +2152,51 @@ export class BattleScene extends SceneBase {
...
+    itemModifier: PokemonHeldItemModifier,
+    target: Pokemon,
+    simulated: boolean = true,
+    transferQuantity: number = 1,
+  ): boolean {
+    const mod = itemModifier.clone() as PokemonHeldItemModifier;
+    const source = mod.pokemonId ? mod.getPokemon() : null;
+    const cancelled = new ValueHolder(false);
+
+    if (source && source.isPlayer() !== target.isPlayer()) {
```
```suggestion
    if (source?.isOpponent(target)) {
```

src/data/abilities/ab-attrs/stab-boost-ab-attr.ts
```diff
@@ -51,7 +44,11 @@ export class StabBoostAbAttr extends AbAttr {
       // If the move type is one of the pokemon's original types then the STAB multiplier is increased by 0.5
       stabMultiplier.value += 0.5;
     }
+  }
 
-    return initialStabMultiplier !== stabMultiplier.value;
+  public override canApply(...[pokemon, , move]: Parameters<this["apply"]>): boolean {
+    return (
+      pokemon.getMoveType(move) !== ElementalType.STELLAR && pokemon.getTypes(true).includes(pokemon.getMoveType(move))
+    );
```
DayKev: 1 less function call.
```suggestion
    const moveType = pokemon.getMoveType(move);
    return moveType !== ElementalType.STELLAR && pokemon.getTypes(true).includes(moveType);
```

src/data/abilities/ab-attrs/synchronize-status-ab-attr.ts
```diff
@@ -10,37 +10,36 @@ import type { Pokemon } from "#field/pokemon";
  * Used for {@linkcode https://bulbapedia.bulbagarden.net/wiki/Synchronize_(Ability) | Synchronize}.
  */
 export class SynchronizeStatusAbAttr extends AbAttr {
-  constructor(showAbility: boolean = true, showAbilityInstant: boolean = false) {
-    super(showAbility, showAbilityInstant);
+  /** All status effects that can be transferred by Synchronize */
+  private static syncStatuses = new Set<StatusEffect>([
```
DayKev: Same comment as in `MoveEffectChanceMultiplierAbAttr` re: private static fields -> un-exported `const` outside the class.
```suggestion
  private static readonly syncStatuses = new ReadonlySet<StatusEffect>([
```

src/data/abilities/ab-attrs/force-switch-out-immunity-ab-attr.ts
innerthunder: This would require an almost full rewrite of the attribute because the activation message is currently handled externally wherever this is applied. It used to use an async flyout alongside the external messages to show the ability.

Also, I'd rather just display Guard Dog anyway. Both abilities currently use a generic "can't be switched out" message when activated. Should they use different messages?

src/data/abilities/ab-attrs/mock-status-effect-ab-attr.ts
innerthunder: Made an issue #1376

src/data/abilities/ab-attrs/post-biome-change-ab-attr.ts
innerthunder: `NewBiomeEncounterPhase` is entirely unused since Endless was removed. Should it be removed as well?

src/data/abilities/ab-attrs/pre-apply-battler-tag-immunity-ab-attr.ts
innerthunder: yep... something to untangle later 😆

src/data/abilities/ab-attrs/force-switch-out-immunity-ab-attr.ts
DayKev: Having both display the flyout is fine, it makes more sense that way (Gamefreak please...).
Suction Cups has its own message, but Guard Dog can continue to use a generic one.

src/data/abilities/ab-attrs/post-biome-change-ab-attr.ts
DayKev: Ah, yeah.

src/data/abilities/ab-attrs/force-switch-out-immunity-ab-attr.ts
Tempo-anon: Just slap a comment and call it a day. It's not a major thing and a lot of Pokemon fangames have qol ability flyouts even when it's not technically canon 🤷

src/data/abilities/ab-attrs/post-biome-change-ab-attr.ts
innerthunder: I marked it as deprecated. Removing it here seems kind of out of scope.

src/data/abilities/ab-attrs/force-switch-out-immunity-ab-attr.ts
innerthunder: Added a TODO to the attribute

src/data/abilities/ab-attrs/type-immunity-ab-attr.ts
innerthunder: Made an issue #1378

src/utils/ability-utils.ts
```diff
@@ -34,7 +34,20 @@ export function getWeatherCondition(...weatherTypes: WeatherType[]): AbAttrCondi
...
+export const anyTypeMoveConversionCondition: PokemonAttackCondition = (user, _target, move) =>
+  !move?.hasAttr(VariableMoveTypeAttr) || (TERA_MOVES.includes(move.id) && !user?.isTerastallized);
+
+/**
+ * Similar to {@linkcode anyTypeMoveConversionCondition}, except that the given move must also be Normal-type.
+ *
+ * Used for {@link https://bulbapedia.bulbagarden.net/wiki/Pixilate_(Ability) | Pixilate} et al.
+ */
+export const normalTypeMoveConversionCondition: PokemonAttackCondition = (user, target, move) =>
+  anyTypeMoveConversionCondition(user, target, move) && move?.type === ElementalType.NORMAL;
```
DayKev:
```suggestion
  move?.type === ElementalType.NORMAL && anyTypeMoveConversionCondition(user, target, move);
```

src/utils/ability-utils.ts
```diff
@@ -34,7 +34,20 @@ export function getWeatherCondition(...weatherTypes: WeatherType[]): AbAttrCondi
...
 }
 
-/** Used for Aerialate, Refrigerate, Pixilate, Galvanize */
-export const normalTypeMoveConversionCondition: PokemonAttackCondition = (user, _target, move) =>
-  move?.type === ElementalType.NORMAL
-  && (!move.hasAttr(VariableMoveTypeAttr) || (TERA_MOVES.includes(move.id) && !user?.isTerastallized));
+/**
+ * @returns `true` if the move isn't a variable-type move. Tera-based variable-type moves may
+ * still have their type changed by abilities with this condition if the user is not Terastallized.
+ *
```
DayKev:
```suggestion
 * @remarks
```

DayKev (dismissed):

DayKev (dismissed):

DayKev (approved):

PR #1377, opened 2025-10-07
feat(move): Fully implement Pursuit

### LINKED ISSUE #1032: [Move] Implement Pursuit
> [!CAUTION]
> BLOCKED by #940  

https://bulbapedia.bulbagarden.net/wiki/Pursuit_(move)

innerthunder:

Pursuit now allows the user to attack an opponent before it retreats.

closes #1032 📈 

`RecallPhase` now prompts the `TurnCommandManager` to schedule the first command in turn order of an opponent using the move Pursuit against the recalling Pokemon. If a command is scheduled from this prompt, the `RecallPhase` is postponed until the command fully resolves.

> [!NOTE]
> There is a known edge case where if a Pokemon using Pursuit also intends to Terastallize, it will always act before any of its allies that are also using Pursuit, regardless of speed order. 

## Screenshots/Videos

https://github.com/user-attachments/assets/7d845959-4183-4e73-a08c-0786f655d4a2

<details><summary>Overrides</summary>
<p>

```ts
const overrides = {
  ENEMY_SPECIES_OVERRIDE: SpeciesId.MAGIKARP,
  ENEMY_MOVESET_OVERRIDE: MoveId.PURSUIT,
  MOVESET_OVERRIDE: MoveId.SPLASH,
```

</p>
</details> 

## How to test the changes?

`pnpm test[:silent] pursuit`

## Checklist

### DIFF
--- a/src/data/battler-tags/pursuing-battler-tag.ts
+++ b/src/data/battler-tags/pursuing-battler-tag.ts
@@ -1,0 +1,21 @@
+import { BattlerTag } from "#battler-tags/battler-tag";
+import { BattlerTagLapseType } from "#enums/battler-tag-lapse-type";
+import { BattlerTagType } from "#enums/battler-tag-type";
+import type { Pokemon } from "#field/pokemon";
+import type { ValueHolder } from "#utils/common-utils";
+
+/**
+ * Tag to double the power of {@link https://bulbapedia.bulbagarden.net/wiki/Pursuit_(move) | Pursuit}
+ * when used against a retreating Pokemon.
+ * @todo Can this be consolidated with `MeFirstPowerBoostTag`?
+ */
+export class PursuingTag extends BattlerTag {
+  constructor() {
+    super(BattlerTagType.PURSUING, BattlerTagLapseType.AFTER_MOVE, 1);
+  }
+
+  public override apply(_pokemon: Pokemon, _simulated: boolean, power: ValueHolder<number>): boolean {
+    power.value *= 2;
+    return true;
+  }
+}
--- a/src/data/moves/move-attrs/pursuit-attr.ts
+++ b/src/data/moves/move-attrs/pursuit-attr.ts
@@ -1,0 +1,3 @@
+import { MoveAttr } from "#moves/move-attr";
+
+export class PursuitAttr extends MoveAttr {}
--- a/src/data/moves/move-attrs/pursuit-attr.ts
+++ b/src/data/moves/move-attrs/pursuit-attr.ts
@@ -7,7 +7,7 @@ import { MoveAttr } from "#moves/move-attr";
 
 /**
  * Attribute for moves that allow the user to attack an opposing Pokemon before it
- * switches out, i.e. {@link https://bulbapedia.bulbagarden.net/wiki/Pursuit_(move) | Pursuit}.
+ * switches out, i.e. {@link https://bulbapedia.bulbagarden.net/wiki/Pursuit_(move) | Pursuit}. \
  * When attacking a retreating target, moves with this attribute double in power.
  * @see {@linkcode TurnCommandManager.tryPursueTarget}
  * @see {@linkcode PursuingTag}
--- a/src/data/moves/move-attrs/pursuit-attr.ts
+++ b/src/data/moves/move-attrs/pursuit-attr.ts
@@ -1,3 +1,15 @@
+/* biome-ignore-start lint/correctness/noUnusedImports: tsdoc imports */
+import type { TurnCommandManager } from "#app/turn-command-manager";
+import type { PursuingTag } from "#battler-tags/pursuing-battler-tag";
+/* biome-ignore-end lint/correctness/noUnusedImports: tsdoc imports */
+
 import { MoveAttr } from "#moves/move-attr";
 
+/**
+ * Attribute for moves that allow the user to attack an opposing Pokemon before it
+ * switches out, i.e. {@link https://bulbapedia.bulbagarden.net/wiki/Pursuit_(move) | Pursuit}.
+ * When attacking a retreating target, moves with this attribute double in power.
+ * @see {@linkcode TurnCommandManager.tryPursueTarget}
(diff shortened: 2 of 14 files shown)


src/data/battler-tags/pursuing-battler-tag.ts
```diff
@@ -0,0 +1,21 @@
+import { BattlerTag } from "#battler-tags/battler-tag";
+import { BattlerTagLapseType } from "#enums/battler-tag-lapse-type";
+import { BattlerTagType } from "#enums/battler-tag-type";
+import type { Pokemon } from "#field/pokemon";
+import type { ValueHolder } from "#utils/common-utils";
+
+/**
+ * Tag to double the power of {@link https://bulbapedia.bulbagarden.net/wiki/Pursuit_(move) | Pursuit}
+ * when used against a retreating Pokemon.
+ * @todo Can this be consolidated with `MeFirstPowerBoostTag`?
```
flx-sta, line 10: This is your chance if you want to unify them or create a new general class for those two

src/data/moves/move-attrs/pursuit-attr.ts
```diff
@@ -0,0 +1,3 @@
+import { MoveAttr } from "#moves/move-attr";
+
+export class PursuitAttr extends MoveAttr {}
```
flx-sta, line 15: NIT: doc

src/phases/recall-phase.ts
```diff
@@ -31,13 +31,27 @@ export class RecallPhase extends PokemonPhase {
...
+      // TODO: this might cause issues if Pursuit KOs the target
+      phaseManager.unshiftPhase(this);
+      super.end();
+      return;
+    }
+
+    this.recall()
+      .then(() => this.end())
+      .catch(() => {
+        throw new Error("RecallPhase: Recall process failed unexpectedly");
```
flx-sta: Would this error even be catchable, since it's async

innerthunder: Not atm, but this was a dangling Promise before and Biome issued a warning for it.

src/data/battler-tags/pursuing-battler-tag.ts
innerthunder: It's awkward because of how `getBattlerTag` assembles tags and how `applyBattlerTags` groups tags by `BattlerTagType`. I think the best I could do to consolidate them now is an abstract `MovePowerMultiplierTag` with these as subclasses, but it doesn't seem like the added hierarchy would be all that useful.

test/moves/pursuit.test.ts
```diff
@@ -0,0 +1,248 @@
...
+
+      game.switchPokemon(1);
+      await game.toEndOfTurn();
+
+      expect(magikarp).not.toHaveFullHp();
+      expect(enemy).toHaveMoveResult(MoveResult.SUCCESS);
+    });
+  });
+
+  describe("in Double Battles", () => {
```
DayKev, line 143: Needs a test for bypassing/ignoring Follow Me (and one to confirm it doesn't do that when there's no switching happening as well).

src/phases/recall-phase.ts
```diff
@@ -31,13 +31,27 @@ export class RecallPhase extends PokemonPhase {
...
+      });
```
DayKev: Can you make `RecallPhase#start` async and then do this?
```ts
await this.recall();
this.end();
```

If not, then just add a todo instead of useless code.
```suggestion
    // biome-ignore lint/nursery/noFloatingPromises: TODO: refactor async code
    this.recall().then(() => this.end());
```

src/phases/recall-phase.ts
```diff
@@ -29,15 +29,25 @@ export class RecallPhase extends PokemonPhase {
 
   // #region Public methods
 
-  public override start(): void {
+  public override async start(): Promise<void> {
```
flx-sta, line 32: Q: Wouldn't `Phase.start()` also have to be updated to support `: void | Promise<void>`?

test/moves/pursuit.test.ts
```diff
@@ -0,0 +1,285 @@
...
+      .enemySpecies(SpeciesId.MAGIKARP)
+      .enemyAbility(AbilityId.BALL_FETCH)
+      .enemyMoveset(MoveId.PURSUIT)
+      .startingLevel(100)
+      .enemyLevel(100);
+  });
+
+  describe("in Single Battles", () => {
+    beforeEach(() => {
+      game.override //
```
flx-sta, line 42: Does the `//` serve a purpose here?

test/moves/pursuit.test.ts
```diff
@@ -0,0 +1,285 @@
...
+      vi.spyOn(pursuit, "calculateBattlePower");
+
+      await game.classicMode.startBattle(SpeciesId.MAGIKARP, SpeciesId.FEEBAS);
+
+      game.setTurnOrder([BattlerIndex.PLAYER, BattlerIndex.ENEMY]);
+      game.move.use(MoveId.SPLASH);
+      await game.toEndOfTurn();
+
+      expect(pursuit.calculateBattlePower).toHaveLastReturnedWith(40);
+    });
```
flx-sta, line 58: Q: Missing confirmation/check that no enemy was switched out?

flx-sta (approved):

src/phases/recall-phase.ts
innerthunder: I was thinking that too, but for some reason only changing the signature here just worked? No complaints from TS...

test/moves/pursuit.test.ts
innerthunder: It prevents biome from condensing all the overrides into one line (which is a bit harder to read imo)

test/moves/pursuit.test.ts
innerthunder: The enemy is the one using Pursuit, and we aren't inputting a switch command or move for the player in this test 🤷

DayKev (approved):

PR #1373, opened 2025-10-01
refactor: clean up some weather & terrain stuff

DayKev:
## What are the changes the user will see?
- Flower Gift will work in the primal sun weather.
- The arena flyout will no longer display "none" after weather/terrain ends.

## Why am I making these changes?
Code cleanup, etc.

## What are the changes from a developer perspective?
- `Arena#hasWeather` and `Arena#hasTerrain` now use a spread parameter.
- Added `readonly` to a bunch of stuff.
- Added `Arena#canSetTerrain`.
- Added helper function `getTerrainAnim`.
- Converted `WeatherType` from `enum` to `const` object due to some overlapping changes I had already made.
- Incidentally fixed Flower Gift so that it works in the primal sun weather too.
- Fixed the arena flyout so it no longer displays "none" after weather/terrain ends.

## Checklist
- [x] Otherwise: **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?

PR #1380, opened 2025-10-15
dev: enable `strictFunctionTypes` in `tsconfig.json`

DayKev:
N/A

Another step towards enabling all the `strict` options in `tsconfig.json`.

Enabled [`strictFunctionTypes`](https://www.typescriptlang.org/tsconfig/#strictFunctionTypes) in `tsconfig.json` and updated code as needed.
Consolidated the tests in `status-effect.test.ts`.

## Checklist

### DIFF
--- a/src/@types/utility-types.ts
+++ b/src/@types/utility-types.ts
@@ -1,21 +1,21 @@
 /**
  * Alias for the constructor of a class.
  * Can be used to build an object of templated type.
- *
+ * @remarks
  * Use {@linkcode AbstractConstructor} instead if comparing types
  */
-export type Constructor<T> = new (...args: unknown[]) => T;
+export type Constructor<T> = new (...args: any[]) => T;
 
 /**
  * Alias for an abstract constructor of a class.
  * Should be used when comparing types, e.g. with `instanceof`.
  */
-export type AbstractConstructor<T> = abstract new (...args: unknown[]) => T;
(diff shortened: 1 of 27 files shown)


innerthunder (commented):
I figure this is a good opportunity to use the non-empty array util from #1373 more widely

DayKev:
> I figure this is a good opportunity to use the non-empty array util from #1373 more widely

Added it to the `AbAttr` that got changed in the PR.

src/@types/utility-types.ts
```diff
@@ -4,18 +4,18 @@
  *
```
DayKev:
```suggestion
 * @remarks
```

src/phases/select-modifier-phase.ts
```diff
@@ -163,46 +168,52 @@ export class SelectModifierPhase extends BattlePhase {
               }
               globalScene.audioManager.playSound("se/buy");
               break;
-            case 1:
+            case 1: {
+              const partyModifierTransferSelectCallback: PartyModifierTransferSelectCallback = (
```
innerthunder, line 172: This should probably be a private method instead.

innerthunder (approved):

src/phases/select-modifier-phase.ts
DayKev: It can't, because it's using other variables from the outer scopes (`party` and `modifierSelectCallback`).

PR #1250, opened 2025-06-10
[Refactor] Add Phase factory methods to `PhaseManager`

innerthunder:
## What are the changes the user will see?

Weather-based form changes from Forecast and Flower Gift should now properly trigger the ability flyout (they previously just constructed `ShowAbilityPhases` without running them).

## Why am I making these changes?

#426 resolved circular imports in phase scheduling in two ways:
- `PhaseId` enum fields were added for each distinct and non-abstract Phase to resolve their types without the use of `instanceof`.
- Bespoke Phase scheduling methods were added to `BattleScene` (now `PhaseManager`) to pull the responsibility of Phase construction from other Phases to the phase manager.

These methods, however, aren't as robust as I and some others would like. Because the basic phase scheduling methods still directly take a phase as input, there's still a risk of circular dependencies popping up as new phases get introduced. The tests in #1236, for example, are currently bricked (likely) because of circular imports between `test` and `src` files.

This PR, modeled after https://github.com/pagefaultgames/pokerogue/pull/5868, improves upon the above methods by:
- Providing alternative Phase scheduling methods that do not require imports
- Adopting Phase identifiers that are more in line with other Phase-related API

## What are the changes from a developer perspective?

- `PhaseManager` (in `src/phase-manager.ts`):
  - Now uses the `PHASES` phase map to construct phases internally
  - Added `createPhase` generic method to construct a Phase from the given constructor parameters
  - Added shorthand methods combining `createPhase` with common scheduling methods:
    - `createAndPushPhase`
    - `createAndUnshiftPhase`
    - `createAndPrependPhase`
    - `createAndAppendPhase`
  - Removed bespoke methods in favor of the new shorthand methods (TODO)
- `Phase` (in `src/phase.ts`):
  - Now uses `phaseName` (a key from `PHASES`) instead of `phaseId` (an enum) as an identifier
  - Rewrote `is` to evaluate phases via the new identifier type

## How to test the changes?

You can playtest this branch at https://poketernity.com/phase-factory/ or run tests locally with `pnpm test:silent`.

## Checklist

- [ ] ⚠️ If this is a PR for `main` (such as a hotfix), has the game version been updated (`pnpm update-version:patch` / `pnpm update-version:minor`?
- [x] Otherwise: **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?

[...]

PR #1383, opened 2025-10-26
refactor: Replace `AbAttrFlag` with key-based attribute lookup

innerthunder:

- Liquid Ooze's ability flyout should now display whenever its effect applies.
- Poison Puppeteer should now activate when its source transfers (Toxic) Poison to another Pokemon via Psycho Shift.

- Replaced the `AbAttrFlag` system with factory-based key lookup (similar to #1250). Unlike the flag system, ability attributes can only have one key at a time. Attributes that relied on multiple flags have been adjusted accordingly:
  - `ReverseDrainAbAttr` (Liquid Ooze) no longer extends `PostDefendAbAttr` and now only uses its own key. It previously used the post-defend trigger to implement a "trigger message". This logic has been reorganized into a `getTriggerMessage` override which now fires on all "draining" effects.
  - `ConfusionOnStatusAbAttr` (Poison Puppeteer) is now applied within `Pokemon#trySetStatus` (immediately after the status is set) instead of within `StatusEffectAttr#applyEffect`. Any instance of `trySetStatus` with a non-null source should now apply Poison Puppeteer (when relevant). This includes Psycho Shift and several ability interactions e.g. Poison Touch.

- Fewer imports
- Stronger type inference for `applyAbAttrs` and related functions (the templated type doesn't need to be specified)
- Setting up for a dedicated ability builder class (it can contain the `AbilityAttrs` list after this PR)

## How to test the changes?

`pnpm test:silent`

## Checklist

  - [?] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes?


DayKev (changes requested):
~Poison Puppeteer is only supposed to apply on moves, but the change makes it apply on abilities too (e.g. Synchronize).
cf https://bulbapedia.bulbagarden.net/wiki/Poison_Puppeteer_(Ability)
I added a test for the behavior.~
Nevermind, it makes sense to do it this way.

~Other than that~ everything looks good.

DayKev:
Then again maybe it's fine, this interaction can't happen legitimately in mainline so of course they wouldn't consider it when making the ability... and it simplifies the code to do it this way as well.

DayKev (approved):

PR #1400, opened 2025-12-11
dev: update handling of dev and beta modes

DayKev:
The title of the webpage and the version string displayed on the title screen will have " (Beta)" appended to them when running in "beta" (e.g. some sort of beta website) or "development" mode (e.g. locally).

General code simplification and cleanup.

- Replaced the `start` command in `package.json` with `start:prod` and `start:beta`, which run the game in "production" and "beta" mode respectively.
  Previously, `start` would run the game in "development" mode just like `start:dev` due to the `vite` command running in "development" mode by default if no `--mode` arg is passed.

- Added `IS_DEV` global constant that checks for "development" mode.

- Removed `Api#isLocal` and replace its uses with `BYPASS_LOGIN`.

- Added " (Beta)" to window title and game version displayed on the title screen when in "development" or "beta" mode.

- Replaced `.then()` pattern with `await` in `main.ts`.

- Removed obsolete commented code in `game-data.ts`.

- Added return types to api methods.

- Updated `deploy-beta.yml` workflow to use "beta" mode when building (though the workflow is currently disabled anyway).

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "version": "0.1.0",
   "type": "module",
   "scripts": {
-    "start": "vite",
+    "start:prod": "vite --mode production",
+    "start:beta": "vite --mode beta",
     "start:dev": "vite --mode development",
     "build": "vite build",
     "build:beta": "vite build --mode beta",
--- a/.github/workflows/deploy-beta.yml
+++ b/.github/workflows/deploy-beta.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Build
         run: pnpm build:beta
         env:
-          NODE_ENV: production
+          NODE_ENV: beta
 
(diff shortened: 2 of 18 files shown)


package.json
```diff
@@ -4,7 +4,8 @@
   "version": "0.1.0",
   "type": "module",
   "scripts": {
-    "start": "vite",
```
flx-sta, line 7: Maybe we should add a `start` that give you a selection of which mode you wanna launch via prompter.

flx-sta (approved):
LGTM.
A bit polluted with style fixes but easy

package.json
DayKev: I don't think we need an extra layer of indirection for something so simple.

PR #1410, opened 2025-12-28
test: update custom test matchers and add new ones

DayKev:
N/A

To improve test writing.

`type-fest` added as dev dep.

Added new test matchers:
- `toBeAtPhase`
- `toEqualUnsorted`
- `toHaveArenaTag`
- `toHaveKey`
- `toHaveShownMessage`
- `toHaveTypes`
- `toHaveUsedPP`

Improved the messages/output of many existing matchers, and they now utilize the `actual`/`expected` fields of the matcher return object.

Combined the functionality of `toHaveMoveResult` into `toHaveUsedMove`. You can pass a `TurnMove` object to the matcher to have it check the fields passed in (`moveId` can be used in place of `move` when passing in an object even though `TurnMove` doesn't normally have that field, the matcher will automatically convert as appropriate).

Some matchers will display type errors when trying to pass invalid values, so you don't need to run the test to find out it would fail. For example, `expect(pkmn).toHaveStatStage(Stat.ATK, 13)` would show that `13` isn't valid.

The `toHaveStatusEffect` matcher will optionally accept objects matching the `Status` interface. For example: `expect(pkmn).toHaveStatusEffect({ effect: StatusEffect.SLEEP, sleepTurnsRemaining: 2 });`

The `toHaveHp` matcher now has an optional `roundDown` param that defaults to `true`, which determines whether the `expected` value is rounded via `Math.floor` or `Math.round`.

## Checklist


flx-sta (dismissed):
Blind approve since it's only tests

PR #1416, opened 2026-01-16
dev: reconfigure Typedoc and adjust Pages workflow

DayKev:
The Typedoc site will be re-enabled.

Typedoc can be useful for examining project structure.

Replaced `typedoc.json` config file with `typedoc.config.js` and filled out many settings.

Updated the workflow:
- Stale runs will be automatically cancelled.
- Permissions were minimized.
- The workflow will only run if relevant files were changed (e.g. `*.ts` files).

3 Typedoc plugins were added as dev deps:
- `typedoc-github-theme`: makes the docs site look similar to GitHub's default theme
- `typedoc-plugin-mdn-links`: replaces `{@link[code] <some built-in, like Array>}` with links to the MDN docs site (though we shouldn't be using `@link[code]` on built-ins anyway, but some do exist already)
- `typedoc-plugin-missing-exports`: "[Automatically document symbols which aren't exported but are referenced](https://www.npmjs.com/package/typedoc-plugin-missing-exports)"

The command to run Typedoc was changed from `docs` to `typedoc` in `package.json` to fix the issue where `pnpm docs` wouldn't run the command, which required using `pnpm run docs` instead.

## Screenshots/Videos

image

</details> 

## How to test the changes?
`pnpm typedoc`

## Checklist

### DIFF
--- a/.github/workflows/github-pages.yml
+++ b/.github/workflows/github-pages.yml
@@ -34,7 +34,6 @@ jobs:
         with:
           sparse-checkout: |
             .nvmrc
-          sparse-checkout-cone-mode: false
 
       - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
         id: filter
--- a/.github/workflows/github-pages.yml
+++ b/.github/workflows/github-pages.yml
@@ -3,7 +3,8 @@ name: "Github Pages"
 on:
   push:
     branches:
-      - main
+      # - main
+      - beta
   pull_request:
     branches:
       - main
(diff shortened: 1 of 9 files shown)


.github/workflows/github-pages.yml
```diff
@@ -3,70 +3,104 @@ name: "Github Pages"
...
+    outputs:
+      all: ${{ steps.filter.outputs.all }}
+    steps:
+      # ensure the correct commit is detected on push workflow
+      - name: Checkout GitHub repository
+        uses: actions/checkout@v6
+        with:
+          sparse-checkout: |
+            .nvmrc
+          sparse-checkout-cone-mode: false
```
DayKev:
```suggestion
```

PR #1417, opened 2026-01-16
github: optimize lint and test workflows

DayKev:
N/A

Reduce unnecessary workflow runs.

- Test and lint workflows will only run if relevant files are changed.
- Workflow permissions are minimized.
- Stale workflow runs will automatically be cancelled.
- All test shards will run to completion even if one shard fails.
- Lint workflows will now show errors without needing an output collection step.
- JS files are now typechecked in the lint workflow.

## Checklist


.github/workflows/lint.yml
```diff
@@ -11,81 +11,79 @@ on:
...
+      all: ${{ steps.filter.outputs.all }}
+      ts-files: ${{ steps.filter.outputs.ts-files }}
+      js-files: ${{ steps.filter.outputs.js-files }}
+    steps:
+      - name: Checkout GitHub repository
+        uses: actions/checkout@v6
+        with:
+          sparse-checkout: |
+            .github/lint-filters.yml
+          sparse-checkout-cone-mode: false
```
DayKev:
```suggestion
```

.github/workflows/tests.yml
```diff
@@ -1,23 +1,51 @@
...
+      pull-requests: read
+    outputs:
+      all: ${{ steps.filter.outputs.all }}
+    steps:
+      - name: Checkout GitHub repository
+        uses: actions/checkout@v6
+        with:
+          sparse-checkout: |
+            .github/test-filters.yml
+          sparse-checkout-cone-mode: false
```
DayKev:
```suggestion
```

